PR #12, opened 2020-12-17
Fix compose CI

vdo:


.github/workflows/main.yml
```diff
@@ -46,7 +46,7 @@ jobs:
           # export COMPOSE_DVOTE_PORT_MAPPING="9090" # Will use a random available port mapping
           cd dockerfiles/testsuite
           docker-compose build
-          sh start_test.sh
+          ./start_test.sh
```
p4u, line 49: what about using `bash start_test.sh` ?

mvdan (approved):

PR #18, opened 2020-12-18
types: use HexBytes instead of hex strings in a few API fields

mvdan:
This means the json marshal/unmarshal steps will automatically do the
hex encoding/decoding, and we can simply use the []byte data directly.

Removes over 100 lines of boilerplate thanks to the above. More "hex
string" fields remain in the API structs, but they are left alone for
now to keep this refactor smaller.

Note that this should be an entirely backwards compatible change, since
the encoded JSON messages should look exactly the same. Replacing REST
with gRPC is still planned for the future, but this is a clear win we
can get today. Plus, our actual business logic gets clearer.

### DIFF
--- a/client/util.go
+++ b/client/util.go
@@ -43,9 +43,9 @@ func CreateEthRandomKeysBatch(n int) []*ethereum.SignKeys {
 }
 
 type keysBatch struct {
-	Keys      []signKey `json:"keys"`
-	CensusID  string    `json:"censusId"`
-	CensusURI string    `json:"censusUri"`
+	Keys      []signKey      `json:"keys"`
+	CensusID  types.HexBytes `json:"censusId"`
+	CensusURI string         `json:"censusUri"`
 }
 type signKey struct {
 	PrivKey string `json:"privKey"`
@@ -53,7 +53,7 @@ type signKey struct {
 	Proof   []byte `json:"proof"`
 }
 
-func SaveKeysBatch(filepath string, censusID, censusURI string, keys []*ethereum.SignKeys, proofs [][]byte) error {
+func SaveKeysBatch(filepath string, censusID []byte, censusURI string, keys []*ethereum.SignKeys, proofs [][]byte) error {
 	if proofs != nil && (len(proofs) != len(keys)) {
 		return fmt.Errorf("lenght of Proof is different from lenght of Signers")
 	}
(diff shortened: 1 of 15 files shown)


client/util.go
```diff
@@ -43,17 +45,17 @@ func CreateEthRandomKeysBatch(n int) []*ethereum.SignKeys {
 }
 
 type keysBatch struct {
-	Keys      []signKey `json:"keys"`
-	CensusID  string    `json:"censusId"`
-	CensusURI string    `json:"censusUri"`
+	Keys      []signKey        `json:"keys"`
+	CensusID  tmbytes.HexBytes `json:"censusId"`
```
p4u: Is this a reasonable trade-off? I mean, we are not depending on a tendermint library for a package that has nothing to do with tendermint. Shouldn't make more sense to have our own "HexBytes" type instead?

test/census_test.go
```diff
@@ -213,18 +215,18 @@ func TestCensus(t *testing.T) {
...
 	resp = doRequest("getRoot", nil)
-	root = resp.Root
+	root = testutil.Hex2byte(t, resp.Root)
 	if len(root) < 1 {
 		t.Fatalf("got invalid root")
 	}
 
 	// getRoot from published census and check censusID=root
-	req.CensusID = root
+	req.CensusID = hex.EncodeToString(root)
```
p4u, line 224: Is not better to use `fmt.Sprintf("%x", root)` instead? We use fmt in general for hex encoding.

types/api.go
```diff
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"reflect"
 	"strings"
+
+	tmbytes "github.com/tendermint/tendermint/libs/bytes"
```
p4u: Same here... I really don't like to depend on tendermint for something as basic as our API types.

p4u:
For the rest LGTM

test/census_test.go
mvdan: I intend to move CensusID and the other fields in part 2, so I don't think it matters :)

mvdan:
copied HexBytes; our version is smaller, more efficient (less allocs), and has a test.

p4u:
Excelent job! that was very necessary :)

PR #23, opened 2020-12-25
types: turn signatures to HexBytes

mvdan:
Makes code more intuitive, and removes over fifty lines of boilerplate.

Also changed the code that directly interacted with signatures. For
example, crypto/ethereum's Go APIs no longer use hex-encoded string
parameters for signatures.


p4u:
Looks good! Thank you @mvdan 

TODO: apply the same changes to manager-backend repository

PR #26, opened 2020-12-29
types: allow 0x prefixes when decoding HexBytes

mvdan:
This is for backwards compatibility with some fields which regularly
used 0x prefixes, like EntityId. Should fix breakages encountered by
frontends.


mvdan:
I think @brickpop wants to do some testing before this is merged, FYI

p4u:
Should be merged ASAP in my opinion, but I prefer to let the merge decision for you.

p4u (approved):

p4u:
After reading the discord discussion, I understand this should be merged

PR #30, opened 2021-01-04
Integrate Scrutinizer into Oracle

jordipainan:
When the process results are computed, the Oracle will send a Vochain transaction that will add the results of the process on the Results fields of the process data and will change the process status to Results.

### DIFF
--- a/chain/ethevents/events.go
+++ b/chain/ethevents/events.go
@@ -145,96 +145,93 @@ func NewEthEvents(contractsAddresses []common.Address, signer *ethereum.SignKeys
 }
 
 // OnComputeResults is called once a process result is computed by the scrutinizer.
-func (ev *EthereumEvents) OnComputeResults(pid []byte, results *models.ProcessResult) {
+func (ev *EthereumEvents) OnComputeResults(results *models.ProcessResult) {
 	tctx, cancel := context.WithTimeout(context.Background(), time.Minute*5)
 	defer cancel()
 	// check ethereum
 	// check ethereum process status
 	var fixedSizePID [types.ProcessIDsize]byte
-	copy(fixedSizePID[:], pid)
+	copy(fixedSizePID[:], results.ProcessId)
 	ethProcessData, err := ev.VotingHandle.VotingProcess.Get(&ethbind.CallOpts{Context: tctx}, fixedSizePID)
 	if err != nil {
-		log.Errorf("error fetching process %x from Ethereum: %s", pid, err)
+		log.Errorf("error fetching process %x from Ethereum: %s", results.ProcessId, err)
 		return
 	}
 	if models.ProcessStatus(ethProcessData.Status) == models.ProcessStatus_RESULTS {
 		// check ethereum process results
 		ethResultsData, err := ev.VotingHandle.VotingProcess.GetResults(&ethbind.CallOpts{Context: tctx}, fixedSizePID)
 		if err != nil {
-			log.Errorf("error fetching process %x from Ethereum: %s", pid, err)
+			log.Errorf("error fetching process %x from Ethereum: %s", results.ProcessId, err)
 			return
 		}
 		if ethResultsData.Height > 0 {
-			log.Errorf("process %x results already added to ethereum", pid)
+			log.Errorf("process %x results already added to ethereum", results.ProcessId)
 			return
 		}
 	} else if models.ProcessStatus(ethProcessData.Status) != models.ProcessStatus_ENDED {
-		log.Errorf("invalid process %x status for setting the results", pid)
+		log.Errorf("invalid process %x status for setting the results", results.ProcessId)
 		return
-	} else {
-		// valid eth
-		// check vochain
-		// check vochain process status
-		vocProcessData, err := ev.VochainApp.State.Process(pid, true)
-		if err != nil {
-			log.Errorf("error fetching process %x from the Vochain: %s", pid, err)
-			return
-		}
-		if models.ProcessStatus(vocProcessData.Status) == models.ProcessStatus_RESULTS {
-			// check vochain process results
-			if len(vocProcessData.Results.Votes) > 0 {
-				log.Errorf("process %x results already added on the Vochain", pid)
-				return
-			}
-		} else if models.ProcessStatus(vocProcessData.Status) != models.ProcessStatus_ENDED {
-			log.Errorf("invalid process %x status %s for setting the results", pid, vocProcessData.Status)
-			return
-		} else {
-			// check scrutinizer
-			_, err = ev.Scrutinizer.Storage.Get(ev.Scrutinizer.Encode("results", pid))
-			if err == nil {
-				log.Errorf("process %x already computed", pid)
-				return
-			}
-			// not results found on ethereum, vochain or scrutinizer
-			// create setProcessTx
-			setprocessTxArgs := new(models.SetProcessTx)
-			// process id
-			setprocessTxArgs.ProcessId = pid
-			// process results
-			setprocessTxArgs.Results = results
-			// set also process status
-			setprocessTxArgs.Status = models.ProcessStatus_RESULTS.Enum()
-			setprocessTxArgs.Txtype = models.TxType_SET_PROCESS_RESULTS
-
-			vtx := models.Tx{}
-			resultsTxBytes, err := proto.Marshal(setprocessTxArgs)
-			if err != nil {
-				log.Errorf("cannot marshal set process results tx: %s", err)
-				return
-			}
-			vtx.Signature, err = ev.Signer.Sign(resultsTxBytes)
-			if err != nil {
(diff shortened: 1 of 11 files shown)


chain/ethevents/events.go
```diff
@@ -127,7 +133,108 @@ func NewEthEvents(contractsAddresses []common.Address, signer *ethereum.SignKeys
...
+		ethev.Scrutinizer = scrutinizer
+		ethev.Scrutinizer.AddEventListener(ethev)
+	}
+
+	return ethev, nil
+}
+
+// OnComputeResults is called once a process result is computed by the scrutinizer.
+func (ev *EthereumEvents) OnComputeResults(results *models.ProcessResult) {
+	tctx, cancel := context.WithTimeout(context.Background(), time.Minute*5)
```
mvdan: it would be more idiomatic if this func received a parent ctx too, then the parent can set the timeout to whatever it wants in the call site

chain/ethevents/events.go
```diff
@@ -127,7 +133,108 @@ func NewEthEvents(contractsAddresses []common.Address, signer *ethereum.SignKeys
...
+			return
+		}
+		if ethResultsData.Height > 0 {
+			log.Errorf("process %x results already added to ethereum", results.ProcessId)
+			return
+		}
+	} else if models.ProcessStatus(ethProcessData.Status) != models.ProcessStatus_ENDED {
+		log.Errorf("invalid process %x status for setting the results", results.ProcessId)
+		return
+	} else {
```
mvdan: if the other two branches always return, you can drop this else and unindent the rest of the code, which will help readability

chain/ethevents/events.go
```diff
@@ -127,7 +133,108 @@ func NewEthEvents(contractsAddresses []common.Address, signer *ethereum.SignKeys
...
+			}
+		} else if models.ProcessStatus(vocProcessData.Status) != models.ProcessStatus_ENDED {
+			log.Errorf("invalid process %x status %s for setting the results", results.ProcessId, vocProcessData.Status)
+			return
+		} else {
+			// check scrutinizer
+			_, err = ev.Scrutinizer.Storage.Get(ev.Scrutinizer.Encode("results", results.ProcessId))
+			if err == nil {
+				log.Errorf("process %x already computed", results.ProcessId)
+				return
```
mvdan: this func has a dozen returns preceded by log.Errorf - why not just return an error, then the caller can do the Errorf a single time?

chain/ethevents/events.go
```diff
@@ -127,7 +133,108 @@ func NewEthEvents(contractsAddresses []common.Address, signer *ethereum.SignKeys
...
+			// not results found on ethereum, vochain or scrutinizer
+			// create setProcessTx
+			setprocessTxArgs := new(models.SetProcessTx)
+			// process id
+			setprocessTxArgs.ProcessId = results.ProcessId
+			// process results
+			setprocessTxArgs.Results = results
+			// set also process status
+			setprocessTxArgs.Status = models.ProcessStatus_RESULTS.Enum()
+			setprocessTxArgs.Txtype = models.TxType_SET_PROCESS_RESULTS
```
mvdan: use `:= &models.SetProcessTx{ProcessId: results.ProcessId, etc etc`

vochain/process.go
```diff
@@ -176,6 +177,47 @@ func (v *State) SetProcessStatus(pid []byte, newstatus models.ProcessStatus, com
...
+// allows and the format of the results allows to do so
+func (v *State) SetProcessResults(pid []byte, result *models.ProcessResult, commit bool) error {
+	process, err := v.Process(pid, false)
+	if err != nil {
+		return err
+	}
+	// Check if the state transition is valid
+	switch process.Status {
+	case models.ProcessStatus_ENDED:
+		break
```
mvdan: breaks are implicit in go

vochain/process.go
```diff
@@ -176,6 +177,47 @@ func (v *State) SetProcessStatus(pid []byte, newstatus models.ProcessStatus, com
...
+	case models.ProcessStatus_RESULTS:
+		return fmt.Errorf("results already added")
+	default:
+		return fmt.Errorf("cannot set results, invalid status: %s", process.Status)
+	}
+
+	if process.Results != nil {
+		if len(process.Results.Votes) > 0 {
```
mvdan: join the two ifs?

vochain/scrutinizer/process.go
```diff
@@ -184,6 +184,10 @@ func (s *Scrutinizer) encode(t string, data []byte) []byte {
 	panic("scrutinizer encode type not known")
 }
 
+func (s *Scrutinizer) Encode(t string, data []byte) []byte {
+	return s.encode(t, data)
```
mvdan: just export the existing encode method?

chain/ethevents/events.go
```diff
@@ -127,7 +133,108 @@ func NewEthEvents(contractsAddresses []common.Address, signer *ethereum.SignKeys
...
+				log.Errorf("error marshaling set process results tx: %s", err)
+				return
+			}
+			log.Debugf("broadcasting Vochain Tx: %s", setprocessTxArgs.String())
+
+			res, err := ev.VochainApp.SendTX(txb)
+			if err != nil || res == nil {
+				log.Errorf("cannot broadcast tx: %s, res: %+v", err, res)
+				return
+			}
```
mvdan: it almost seems to me like this "marshal inner message, sign inner message, marshal outer message, then send the entire thing" should be a helper method somewhere, like `EncodeAndSendTX` or something like that. I don't think it's the first time I see this code.

chain/ethevents/events.go
jordipainan: This is a bit tricky. `OnComputeResults` is called by `checkFinishedProcesses` at a given height, the height is determined by the Vochain itself and is autoincremental on each commit which happens on every consensus round (atm 10s). I think in this case and given your comment is it better to remove this context and figure out how can we tie the commit time with the results processing timeout.
Another reason is that the defined interface on the Vochain does not accept context and adding the mentioned context will require to make more changes on all the other event listener methods. We can discuss more in depth :)

chain/ethevents/events.go
jordipainan: More or less what I commented before, the event listeners do not accept returning an error at this moment and adding so has some implications on the Vochain side. Happy to add it but needs some more analysis.

chain/ethevents/events.go
jordipainan: In progress :)

chain/ethevents/events.go
mvdan: gotcha, then maybe add TODOs

chain/ethevents/events.go
mvdan: TODO as before? :)

vochain/process.go
```diff
@@ -176,6 +177,45 @@ func (v *State) SetProcessStatus(pid []byte, newstatus models.ProcessStatus, com
...
+// allows and the format of the results allows to do so
+func (v *State) SetProcessResults(pid []byte, result *models.ProcessResult, commit bool) error {
+	process, err := v.Process(pid, false)
+	if err != nil {
+		return err
+	}
+	// Check if the state transition is valid
+	switch process.Status {
+	case models.ProcessStatus_ENDED:
+		// process must be ended for setting the results
```
p4u, line 195: Should return an error

vochain/process.go
```diff
@@ -176,6 +177,45 @@ func (v *State) SetProcessStatus(pid []byte, newstatus models.ProcessStatus, com
...
+	case models.ProcessStatus_RESULTS:
+		return fmt.Errorf("results already added")
```
p4u: Check if added results are the same as ours, else return an error

vochain/process.go
```diff
@@ -176,6 +177,45 @@ func (v *State) SetProcessStatus(pid []byte, newstatus models.ProcessStatus, com
...
+	default:
+		return fmt.Errorf("cannot set results, invalid status: %s", process.Status)
+	}
+
+	if process.Results != nil && len(process.Results.Votes) > 0 {
+		return fmt.Errorf("results already added: %+v", process.Results)
```
p4u: That cannot happend, if so it should be a panic

p4u (changes requested):

vochain/process.go
p4u: Add a TO-DO: allow result confirm with another Tx (in order to have N/M oracle signatures)

p4u:
LGTM

p4u (approved):

chain/ethevents/events.go
jordipainan: @mvdan will add in another PR

p4u:
feel free to rebase and merge yourself @jordipainan 
nice work btw :+1:

PR #31, opened 2021-01-04
all: remove the last uses of go-amino

mvdan:
We now use either plain json or protobuf for all of our internal
encodings.

The only places that still used go-amino were mainly for interacting
with Tendermint, which still makes use of amino-flavored JSON for some
bits for backwards-compatibility.

Those places can use tendermint's libs/json, which implements that JSON
flavor without requiring all of go-amino. Plus, we already depend on the
tendermint module anyway.

The gateway seems to work just like before, all tests pass, and the
"amino key" log lines still print useful information:

	amino private key: {"type":"tendermint/PrivKeyEd25519","value":"Pl9T4AJTYENY+rlFlw8z9/5iiwUFlhkSvkphkwIqD/cZUBQt0obuUPBOYEX7/EKBdCa90zKCPHi/tLRuLNfeJg=="}
	amino public key: {"type":"tendermint/PubKeyEd25519","value":"GVAULdKG7lDwTmBF+/xCgXQmvdMygjx4v7S0bizX3iY="}

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -55,7 +55,6 @@ require (
 	github.com/spf13/jwalterweatherman v1.1.0 // indirect
 	github.com/spf13/pflag v1.0.5
 	github.com/spf13/viper v1.7.1
-	github.com/tendermint/go-amino v0.16.0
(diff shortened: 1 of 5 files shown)


p4u:
Looks good.

PR #35, opened 2021-01-06
Avoid ethereum lookup on results processing

jordipainan:
Avoid sending results tx's on each commit when live processes.

Also fixed some prints.


p4u:
thanks!

PR #38, opened 2021-01-07
mod: update to the final tendermint 0.34.1 version

mvdan:


mvdan:
This fails because we were on master, so the "upgrade" to 0.34.1 is actually a downgrade to a release branch which does not have a lot of the changes in master we had.

@p4u do we want to stay in master, or move back to stable releases? I think we prefer stable releases, but I'm not sure if you'll want to go back to master soon again.

mvdan:
Spoke to the Tendermint people a bit:

* 0.35 isn't coming for months, at least. staying on master is not a good idea for stability.
* when they do important bug fixes, they get backported to the 0.34 release branch immediately, so we could always rely on that to get the latest fixes even if we go back to 0.34.1.

So I think we should do this (by amending the code to be compatible with 0.34 again).

p4u:
Reverting some changes should not be much painful. I think its worth it to go back to 0.34.1

mvdan:
@p4u should be done now. I didn't straight up revert all the API changes you did in December, because I think a few were good. I changed the minimum necessary to work with 0.34 again.

PR #39, opened 2021-01-08
census: allow using values (in addition to keys)

p4u:
Until now only keys were allowed to be used as part of the census merkle
tree. This commit introduces also values, which are stored as a leaf of
the census tree. The values do not change the position of the claim into
the merkle tree, its aim is to be used as a numeric weight.

This is the first step to support weighted elections.

WARNING: API breaking change


censustree/gravitontree/trie.go
```diff
@@ -182,18 +192,23 @@ func (t *Tree) treeWithRoot(root []byte) statedb.StateTree {
...
 	tree := t.treeWithRoot(root)
 	if tree == nil {
 		return nil, fmt.Errorf("dump: root not found %x", root)
 	}
+	dump := exportData{}
 	tree.Iterate(nil, func(k, v []byte) bool {
-		claims = append(claims, k)
+		ee := exportElement{Key: make([]byte, len(k)), Value: make([]byte, len(v))}
+		copy(ee.Key, k[:])
+		copy(ee.Value, v[:])
```
mvdan, line 205: document this copy? something like `it's not safe to hold on to the []byte values from Iterate`.

censustree/iden3tree/tree.go
```diff
@@ -207,23 +217,27 @@ func (t *Tree) Root() []byte {
...
 	}
+	dump := exportData{}
 	for _, c := range claims {
 		cb, err := hex.DecodeString(c)
 		if err != nil {
 			return nil, err
 		}
-		keys = append(keys, cb)
+		ee := exportElement{Key: make([]byte, len(cb)), Value: []byte{}}
+		copy(ee.Key, cb[:])
```
mvdan: I think this copy does not need to happen. cb is just decoded from hex, so it's safe to just use directly.

censustree/iden3tree/tree.go
```diff
@@ -207,23 +217,27 @@ func (t *Tree) Root() []byte {
...
 	rootHash := new(merkletree.Hash)
 	copy(rootHash[:], root)
 	t.updateAccessTime()
 	claims, err := t.Tree.DumpClaims(rootHash)
 	if err != nil {
 		return nil, err
 	}
+	dump := exportData{}
 	for _, c := range claims {
 		cb, err := hex.DecodeString(c)
```
mvdan, line 232: would be nice to replace this hex at some point too, e.g. with base64 or just []byte directly. do we have a TODO or issue about it?

censustree/gravitontree/trie.go
```diff
@@ -228,11 +243,15 @@ func (t *Tree) DumpPlain(root []byte) ([][]byte, [][]byte, error) {
...
 
 // ImportDump imports a partial or whole tree previously exported with Dump()
-func (t *Tree) ImportDump(keys [][]byte) error {
+func (t *Tree) ImportDump(data []byte) error {
 	t.updateAccessTime()
 	var err error
-	for _, cb := range keys {
-		if err = t.Tree.Add(cb, []byte{}); err != nil {
+	census := new(exportData)
+	if err = bare.Unmarshal(data, census); err != nil {
```
mvdan: try to use `:=` in all these `if err := f(); err != nil {` pieces of code, then you often don't need the `var err error` at the top at all. plus, you don't have to worry about overwriting a previous error.

mvdan (approved):

censustree/iden3tree/tree.go
p4u: Thought the same but I did not want to risk :sweat_smile:  But since we both got the same feeling, it makes sense to remove the copy. Thanks.

censustree/iden3tree/tree.go
p4u: The dump returned by this function is already []bytes. I suppose you refer to the t.Tree.DumpClaims(). That should be changed upstream on the iden3 merkle tree implementation. We have to update this dependency, actually there exist a branch with this work started but it was abandoned in some point.

mvdan: Ah, gotcha. Maybe we can revive that now, if you think it's worth it.

p4u: https://gitlab.com/vocdoni/go-dvote/-/tree/feature/snark-int

censustree/iden3tree/tree.go
mvdan: Also, you can drop the []byte{} for the empty value, I think.

censustree/gravitontree/trie.go
p4u: Mmm, I'd like to understand what would be, in general, the best approach. 

To me declaring a single `err` variable for a whole function makes sense since it reuses a variable which is constantly used for the same objective (check an error immediately after receiving it). There is no possible overwriting because the check is always performed on the next line. 

What's the point for declaring new N error variables in a single function?

mvdan: From a human's point of view, variables should have the smallest scope possible. If a function shares an err variable for its entire body, there's a chance there will be a mistake somewhere, like:

```
var err error
// many lines here...
if err2 := f(); err2 != nil {
    return err // wrong one!
}
```

It's true that more variables are generally worse, but think of it from the point of view of variables in the *function body scope*. With your method, there's one more variable in that scope, not fewer.

mvdan: Also, more short-lived variables are not a problem for the compiler at all. It can be clever enough to just reuse registers or stack space for them.

p4u: > From a human's point of view, variables should have the smallest scope possible. If a function shares an err variable for its entire body, there's a chance there will be a mistake somewhere, like:

Sure, it makes sense for most of the variables. But the err variable is somehow special in GoLang. Everyone is using just `err` most of the times. The example you posted is quite "weird" IMO, most of the users (even not experienced) would do:
```
var err error
// many lines here...
if err := f(); err != nil {
    return err // wrong one!
}
```

> Also, more short-lived variables are not a problem for the compiler at all. It can be clever enough to just reuse registers or stack space for them.

I'm pretty sure the GoLang compiler is quite smart and there is also the Garbage collector. However from the developer point of view, making things easier for the compiler/collector looks like a good approach.

In any case I don't have any very strong point for declaring a single `err` variable on functions. It's just I don't see any strong point either for not doing it... But if declaring a new err variable on each `if` statement is more idiomatic, lets follow this approach.

Btw it would be nice to have a document on go-dvote Wiki explaining best practices for writing code in our repositories.

censustree/iden3tree/tree.go
p4u: The code is quite old and probably requires quite an effort to update it. We are not using iden3 merkle-tree anywhere in the code right now, it will be used once we integrate zk-SNARKs (few months?). So I'm not sure if its worth it. However if you like to revive it and apply the changes, please do it :+1:

censustree/iden3tree/tree.go
p4u: Mmm, I feel like `[]byte{}` is safer than nil. Don't you think so?

mvdan: Not really. The only exception is if the format/API treats both differently and you want the non-nil version, like encoding/json with null vs [].

As far as I can tell, go-bare encodes lists as length + elements, so both should encode the same. In general, I think you should try to do the easiest thing first, which is just using the zero value (nil).

censustree/gravitontree/trie.go
mvdan: To be fair, some people do prefer the whole "declare err at the beginning of the func and reuse it". I don't think it's the most common style, so I wouldn't say it's "idiomatic". I think most people who use it are either used to old C, or aren't comfortable enough with variable shadowing and prefer to have a single declaration of err.

Regarding the doc; I think https://golang.org/doc/effective_go.html covers the basics. The "control structures" part touches a bit on `if err := ...`. I don't think we need a page of our own to describe how to write Go, except perhaps linking to existing well-maintained docs like that one.

censustree/iden3tree/tree.go
mvdan: Gotcha, I'll give that a go when this is merged.

censustree/iden3tree/tree.go
p4u: ACK

p4u:
DO NOT MERGE until we adapt also the clients (should be during the next days).

p4u:
Updated documentation https://docs.vocdoni.io/#/architecture/services/census-service @brickpop @elboletaire @emmdim

The only changes you need to apply right now are:

1. Change "claimData" by "censusKey"
2. You can just ignore the new parameter "censusValue", its fine if you don't use it
3. On "dumpPlain" method now you need to get the parameter "censusKeys" (and also "censusValues" if you want to support weighted census)

p4u:
In case you want to to add support for weighted census, the "censusValue" is expected to be the Hexadecimal string representation of a big Float number.

PR #48, opened 2021-01-14
remove unnecessary type conversions

mvdan:
Found by unconvert. Also two unnecessary uses of Sprintf found by a
newer staticcheck.


p4u:
That's nice. Would it make sense to add uncovert to the CI check?

PR #50, opened 2021-01-15
start using frankban/quicktest

mvdan:
It's similar to other Go testing libraries, but much lighter. The
documentation contains under 30 APIs and takes a few pages, while most
other testing libraries like testify/assert contain many times more APIs
without being much better. quicktest also builds on top of go-cmp, which
is pretty powerful.

This results in simpler test code, and better and more consistent test
failure output. For example:

	error:
	  unexpected success
	len(got):
	  int(0)
	got:
	  types.HexBytes(nil)
	want length:
	  <same as "len(got)">
	stack:
	  /home/mvdan/src/dvote/test/census_test.go:159
	    qt.Assert(t, siblings, qt.Not(qt.HasLen), 0)

This is a first proof of concept, transitioning just one file. We can
use quicktest more and more over time if we like it.


p4u:
Looks nice.

PR #54, opened 2021-01-19
mod: update tendermint

mvdan:
Since their latest bugfix fixes a security issue.


p4u:
Probably the CI failure is because of a recent Census change. I'll take a look tomorrow.

p4u:
For some reason tendermint needs more time to bootstrap and the API is not available when the test tries to start (5 seconds). I fixed this and also updated tendermint on #55. So closing this PR.

PR #65, opened 2021-01-25
add support for updateCensus tx

jordipainan:


p4u:
Before starting the review, could you please add some more details on the commit message?

chain/contractWrappers.go
```diff
@@ -233,9 +233,36 @@ func (ph *VotingHandle) SetStatusTxArgs(ctx context.Context, pid [types.ProcessI
...
+	if err != nil {
+		return nil, fmt.Errorf("error fetching process from Ethereum: %w", err)
+	}
+	// create setProcessTx
+	setprocessTxArgs := new(models.SetProcessTx)
+	// process id
+	setprocessTxArgs.ProcessId = pid[:]
+	// process mkRoot
+	if setprocessTxArgs.CensusRoot, err = hex.DecodeString(util.TrimHex(processData.MetadataCensusRootCensusUri[1])); err != nil {
+		return nil, fmt.Errorf("invalid mkRoot: %w", err)
```
p4u: Its not a "mkRoot" anymore but a "censusRoot"

chain/contractWrappers.go
```diff
@@ -233,9 +233,36 @@ func (ph *VotingHandle) SetStatusTxArgs(ctx context.Context, pid [types.ProcessI
...
+	}
+	// process mkUri
+	// check origin
+	switch t := models.CensusOrigin(processData.ModeEnvelopeTypeCensusOrigin[2]); {
```
p4u: Looks more idiomatic something like:
```
t := models.CensusOrigin(processData.ModeEnvelopeTypeCensusOrigin[2])
switch t {
  case models.CensusOrigin_OFF_CHAIN_CA, models.CensusOrigin_OFF_CHAIN_TREE ...:
```

chain/contractWrappers.go
```diff
@@ -233,9 +233,36 @@ func (ph *VotingHandle) SetStatusTxArgs(ctx context.Context, pid [types.ProcessI
...
+	case t == models.CensusOrigin_OFF_CHAIN_CA, t == models.CensusOrigin_OFF_CHAIN_TREE, t == models.CensusOrigin_OFF_CHAIN_TREE_WEIGHTED:
+		if types.CensusOrigins[t].NeedsURI && len(processData.MetadataCensusRootCensusUri[2]) == 0 {
+			return nil, fmt.Errorf("census %s needs URI but non has been provided", t.String())
```
p4u: typo

chain/contractWrappers.go
```diff
@@ -233,9 +233,36 @@ func (ph *VotingHandle) SetStatusTxArgs(ctx context.Context, pid [types.ProcessI
...
+		}
+		setprocessTxArgs.CensusURI = &processData.MetadataCensusRootCensusUri[2]
+	default:
+		return nil, fmt.Errorf("invalid census origin, only off-chain processes can update the census")
```
p4u: I'd remove `only off-chain processes can update the census` since its true now but in might not be in the future.

chain/ethevents/handlers.go
```diff
@@ -139,6 +139,55 @@ func HandleVochainOracle(ctx context.Context, event *ethtypes.Log, e *EthereumEv
 		}
 		log.Debugf("broadcasting tx: %s", log.FormatProto(setProcessTx))
 
+		res, err := e.VochainApp.SendTX(tx)
+		if err != nil || res == nil {
+			log.Warnf("cannot broadcast tx: (%s)", err)
```
p4u: Lets not print the error if we are returning it.

chain/ethevents/handlers.go
```diff
@@ -139,6 +139,55 @@ func HandleVochainOracle(ctx context.Context, event *ethtypes.Log, e *EthereumEv
...
+			return fmt.Errorf("cannot broadcast tx: %w, res: %+v", err, res)
+		}
+		log.Infof("oracle transaction sent, hash: %s", res.Hash)
```
p4u: I'd say Hash is bytes? So `%x`

chain/ethevents/handlers.go
```diff
@@ -139,6 +139,55 @@ func HandleVochainOracle(ctx context.Context, event *ethtypes.Log, e *EthereumEv
...
+
+	case ethereumEventList["processesCensusUpdated"]:
+		tctx, cancel := context.WithTimeout(ctx, time.Minute)
+		defer cancel()
+		setProcessTx, err := processCensusUpdatedMeta(tctx, &e.ContractsABI[0], event.Data, e.VotingHandle)
+		if err != nil {
+			return err
```
p4u: Wrap the errors and give some information regarding where it comes from (same for the rest)

chain/ethevents/handlers.go
```diff
@@ -139,6 +139,55 @@ func HandleVochainOracle(ctx context.Context, event *ethtypes.Log, e *EthereumEv
...
+		// check dynamic census enabled
+		if !p.Mode.DynamicCensus {
+			return fmt.Errorf("process needs dynamic census in order to update its census")
+		}
+		// check status
+		if (p.Status != models.ProcessStatus_READY) && (p.Status != models.ProcessStatus_PAUSED) {
+			return fmt.Errorf("process status %s does not accept census updates", p.Status.String())
+		}
+		// check census origin
+		if (p.CensusOrigin != models.CensusOrigin_OFF_CHAIN_CA) && (p.CensusOrigin != models.CensusOrigin_OFF_CHAIN_TREE) && (p.CensusOrigin != models.CensusOrigin_OFF_CHAIN_TREE_WEIGHTED) {
```
p4u: Please do not add this kind of static checks here... You can extend the types.CensusOrigins struct and add a new flag for `allowCensusUpdate`

chain/ethevents/handlers.go
```diff
@@ -139,6 +139,55 @@ func HandleVochainOracle(ctx context.Context, event *ethtypes.Log, e *EthereumEv
...
+			return fmt.Errorf("process census origin %s does not accept census updates", p.CensusOrigin.String())
+		}
+
+		vtx := models.Tx{}
```
p4u, line 180: I'm missing a check to ensure the updated census is not already updated (so compare current census with new census)

vochain/process.go
```diff
@@ -213,6 +213,38 @@ func (v *State) SetProcessResults(pid []byte, result *models.ProcessResult, comm
...
+	if err != nil {
+		return err
+	}
+	// check valid state transition
+	// dynamic census
+	if !process.Mode.DynamicCensus {
+		return fmt.Errorf("cannot update census, only processes with dynamic census can update its census")
+	}
+	// census origin
+	if !(process.CensusOrigin == models.CensusOrigin_OFF_CHAIN_CA) &&
```
p4u: A baby cat is dying...

p4u (changes requested):

chain/contractWrappers.go
jordipainan: changed on all files :)

chain/contractWrappers.go
jordipainan: good catch

chain/contractWrappers.go
jordipainan: yep, it's more flexible, changed!

chain/ethevents/handlers.go
jordipainan: changed on all handlers, also added more precise info

chain/ethevents/handlers.go
jordipainan: yep, code will be less verbose and cleaner, changed!

vochain/process.go
jordipainan: solved! hope the cat is alive :cat:

chain/ethevents/handlers.go
jordipainan: added :)

chain/contractWrappers.go
```diff
@@ -233,9 +233,35 @@ func (ph *VotingHandle) SetStatusTxArgs(ctx context.Context, pid [types.ProcessI
...
+	setprocessTxArgs := new(models.SetProcessTx)
+	// process id
+	setprocessTxArgs.ProcessId = pid[:]
+	// process censusRoot
+	if setprocessTxArgs.CensusRoot, err = hex.DecodeString(util.TrimHex(processData.MetadataCensusRootCensusUri[1])); err != nil {
+		return nil, fmt.Errorf("invalid census root: %w", err)
+	}
+
+	censusOrigin := models.CensusOrigin(processData.ModeEnvelopeTypeCensusOrigin[2])
+	if types.CensusOrigins[censusOrigin].AllowCensusUpdate {
```
p4u: This is a bit confusing if chain statement, I'd try to keep a format such as `if something wrong => return, else keep runing the main func code`

```
if !types.CensusOrigins[censusOrigin].AllowCensusUdate {
    return nil, fmt.Errorf("cannot update census, invalid census origin")
 }
if types.CensusOrigins[censusOrigin].NeedsURI && len(processData.MetadataCensusRootCensusUri[2]) == 0 {
			return nil, fmt.Errorf("census %s needs URI but an empty census URI has been provided", censusOrigin.String())
}
```

chain/ethevents/handlers.go
```diff
@@ -141,10 +142,66 @@ func HandleVochainOracle(ctx context.Context, event *ethtypes.Log, e *EthereumEv
...
+			return fmt.Errorf("cannot fetch the process from the Vochain: %w", err)
+		}
+
+		// process censusURI
+		if bytes.Equal(p.CensusRoot, setProcessTx.CensusRoot) {
+			return fmt.Errorf("censusRoot cannot be the same")
+		}
+		// check origin
+		if p.CensusURI == setProcessTx.CensusURI {
+			return fmt.Errorf("censusURI cannot be the same")
```
p4u: I don't think this check should be done. The censusURI might be the same. I suppose you are thinking on a IPFS census but we would extend it to HTTP(s) or other URL typey.

types/vochain.go
```diff
@@ -132,16 +132,17 @@ type ScrutinizerOnProcessData struct {
...
+	NeedsURI          bool
+	WeightedSupport   bool
 }
 
 var CensusOrigins = map[models.CensusOrigin]CensusProperties{
-	models.CensusOrigin_OFF_CHAIN_TREE:          {Name: "offchain tree", NeedsDownload: true, NeedsURI: true},
-	models.CensusOrigin_OFF_CHAIN_TREE_WEIGHTED: {Name: "offchain weighted tree", NeedsDownload: true, NeedsURI: true, WeightedSupport: true},
-	models.CensusOrigin_ERC20:                   {Name: "erc20", NeedsDownload: true, WeightedSupport: true, NeedsIndexSlot: true},
-	models.CensusOrigin_OFF_CHAIN_CA:            {Name: "ca", WeightedSupport: true, NeedsURI: true},
+	models.CensusOrigin_OFF_CHAIN_TREE:          {Name: "offchain tree", NeedsDownload: true, NeedsURI: true, AllowCensusUpdate: true},
```
p4u, line 144: You don't need to add the false values here.

p4u (changes requested):

vochain/transaction.go
```diff
@@ -97,6 +98,14 @@ func AddTx(vtx *models.Tx, state *State, txID [32]byte, commit bool) ([]byte, er
 					return []byte{}, fmt.Errorf("set process results, results is nil")
 				}
 				return []byte{}, state.SetProcessResults(tx.ProcessId, tx.Results, true)
+			case models.TxType_SET_PROCESS_CENSUS:
+				if bytes.Equal(tx.CensusRoot, []byte{}) {
+					return []byte{}, fmt.Errorf("set process census, census root is empty")
+				}
+				if *tx.CensusURI == "" {
```
p4u: That's a PANIC if tx.CensusURI == nil since tx.CensusURI is a pointer.

When dealing with the Protobuf models that's something very critic, any pointer field must be checked before accessing to it. Please, make sure all proto accesses are safe.

p4u (approved):

jordipainan:
Thank you for the extensive review and comments, much appreciated.

PR #67, opened 2021-01-28
crypto/ethereum: small types refactor

p4u:
- use []byte instead of string for most crypto/ethereum functions and
methods.

- use only compressed keys.

- update the protobuf models and use the pubKey (instead of address)
for OFF_CHAIN_CA census root.


p4u:
This is a breaking change, the clients should adapt and use only compressed public keys: @brickpop 

Also some components using go-dvote as library would need to adapt to []bytes instead of strings: @NateWilliams2 @emmdim

PR #68, opened 2021-01-29
remove net and subpub packages, use the ones implemented in multirpc

p4u:
Subpub and the net package (including the proxy, http and websockets implementation)
are now on a separated repository: https://github.com/vocdoni/go-dvote

In addition update libp2p and ipfs modules (due a conflict of versions
created by using a new libp2p package set in the mulritpc module)


mvdan (approved):
SGTM. note that the link to multirpc is actually a link to go-dvote in the commit message

p4u:
By the way, the router is still not migrated. I did not do it yet because it requires a huge change on go-dvote, but the idea is to do it in some moment. 

Also the vocdoni-manager-backend should use the packages on multirpc instead of its own. That way we'll have a single transport layer for all vocdoni projects that require it.

p4u:
Also @mvdan I think the transport interface needs some re-thinking and refactor to be more consistent and flexible.

PR #70, opened 2021-02-01
add go linter to CI

p4u:
A lot of work to do in order to have the go-CI happy. However I do see it as a nice improvement (fix all errors reported and keep it enabled from now on).

So I let this PR open so anyone can just push fixes on the code time to time until everything is ready to merge: @vdo @jordipainan @mvdan please feel free if you found some sparse time :)

Instructions for installing the linter locally:

```
git checkout feature/add_go_linter
curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.30.0
golangci-lint run --timeout=5m -c .golangci.yml
```
You can create as many commits as needed, we'll squash it on the merge.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,17 @@
+issues:
+        max-same-issues: 0
+        exclude-use-default: false
+linters:
+        enable:
+        - whitespace
+        - gosec
+        - gci
+        - misspell
+        - gomnd
+        - gofmt
+        - goimports
+        - lll
+        - golint
+linters-settings:
+        lll:
+                line-length: 100
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,6 +17,11 @@ jobs:
     - name: Checkout code
       uses: actions/checkout@v2
 
+    - name: Lint
+      run: |
+        curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.30.0
+        $(go env GOPATH)/bin/golangci-lint run --timeout=5m -c .golangci.yml
+
     - name: Go test
       run: |
         # we run vet in another step
@@ -30,10 +35,12 @@ jobs:
         go vet ./...
         curl -L https://github.com/dominikh/go-tools/releases/download/2020.2.1/staticcheck_linux_amd64.tar.gz | tar -xzf -
         ./staticcheck/staticcheck ./...
+ 
     - name: Install goveralls
       env:
         GO111MODULE: off
       run: go get github.com/mattn/goveralls
+
     - name: Send coverage to coverall.io
       env:
         COVERALLS_TOKEN: ${{ secrets.github_token }}


p4u:
Actually I think it will take long and there will be probably merge conflicts. So its probably better to make separated PR for a bunch of fixes in order to get in merged to master ASAP.

I'll let this PR open in order to remember a continuous work on this matter.

.golangci.yml
```diff
@@ -0,0 +1,17 @@
...
+        enable:
+        - whitespace
+        - gosec
+        - gci
+        - misspell
+        - gomnd
+        - gofmt
+        - goimports
+        - lll
+        - golint
```
mvdan, line 14: golint is deprecated and archived; using it at this point is not a good idea

.golangci.yml
```diff
@@ -0,0 +1,17 @@
...
+        max-same-issues: 0
+        exclude-use-default: false
+linters:
+        enable:
+        - whitespace
+        - gosec
+        - gci
+        - misspell
+        - gomnd
+        - gofmt
```
mvdan, line 11: personally, I would replace gofmt+gci+whitespace with gofumpt, which does more than all three combined too :wink: 

it will also handle long lines soon, so that would also make lll redundant.

.golangci.yml
```diff
@@ -0,0 +1,17 @@
...
+        - goimports
```
mvdan, line 12: I don't think this is actually helpful, because both missing and unused imports are already a build error. AFAIK goimports is useful while developing code, to fix those missing/unused build errors. It doesn't really help for CI as those cases will already fail.

.golangci.yml
```diff
@@ -0,0 +1,17 @@
+issues:
+        max-same-issues: 0
+        exclude-use-default: false
+linters:
+        enable:
+        - whitespace
+        - gosec
+        - gci
+        - misspell
+        - gomnd
```
mvdan, line 10: wouldn't this require changing *tons* of lines? I'm not sure if it's worth the effort

mvdan (commented):
I also personally think that using golangci-lint is not a great idea; I recently made the same argument at work when establishing CI best practices, so you could give this a read: https://github.com/ipld/ipld/pull/92#issuecomment-759356290

That markdown file also has some general advice on how to do CI for Go :) I was going to ping you about it when it was merged.

.golangci.yml
p4u: Please, feel free to propose the golangci.yml file that you think we should have.

Regarding using a golang-ci linter I do see a lot of possible future value here. Most of the reviews we do on a PR would be automatically find by the linter so that would save a lot of time doing always the same kind of review (empty line not necessary, imports not cgi-ed, error not check, variable not needed, etc.). Currently we only have staticcheck which is nice but not enough.

That said, I am very open to suggestion on which would be the best tool or the best configuration for golang-ci.

mvdan: I made some suggestions for linters in the other comments. I agree that using more linters is a good thing. I don't like using golangci-lint because its quality is questionable as a tool (see my link from earlier), and at the end of the day I only end up running two or three linters on my projects.

p4u: Then I'll apply your proposals and lets try how it goes... we can always change.

mvdan: I'm confused. This PR is open, but golangci-lint is on master without these changes applied.

p4u: Yes. This PR will never be merged, its just a "reminder". The idea is that we'll perform commits time to time in order to apply the fixes (not a single MR but several).

I decided to merge golangci.yml so all of us applying the CI suggestions can use the same config options.

p4u: I have updated it with your comments.

p4u:
After reading the documentation of golang-ci and considering @mvdan opinion, I made this list. What do you think?
```        
        - gosec
        - misspell
        - gofumpt
        - lll
        - staticcheck
        - structcheck
        - varcheck
        - unused
        - typecheck
        - ineffassign
        - errcheck
        - gosimple
        - goconst
```
It includes staticcheck so we can remove the current CI job.

PR #73, opened 2021-02-03
Use golang container in tests

vdo:
Use golang base container as a environment for tests, instead of using the host machine

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -7,23 +7,20 @@ on: [push, pull_request]
 jobs:
   go-tests:
     runs-on: self-hosted
+    container: golang:1.15
     if: github.repository == 'vocdoni/vocdoni-node'
+    defaults:
+      run:
+        shell: bash
     steps:
-    - name: Install Go
-      uses: actions/setup-go@v2
-      with:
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -25,6 +21,7 @@ jobs:
         go test -vet=off -timeout=15m -race -covermode atomic -coverprofile=covprofile ./...
 
     - name: Go analyze
+      shell: bash
```
mvdan: can we just make bash the default shell for the entire job?

mvdan (approved):

.github/workflows/main.yml
vdo: shell is a property of run, not of the job, as per https://json.schemastore.org/github-workflow

mvdan: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_iddefaultsrun

vdo: Great, I will use this default instead then

p4u (approved):

p4u:
Merge it yourself @vdo

PR #75, opened 2021-02-03
Clean workspace before compose test

vdo:
This needs to be applied because the `go-tests` job, running in parallel, leaves behind some files owned by root, and they conflict with the actions being run afterwards

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -7,7 +7,9 @@ on: [push, pull_request]
 jobs:
   go-tests:
     runs-on: self-hosted
-    container: golang:1.15
+    container: 
+      image: golang:1.15
+      options: --user 1000
     if: github.repository == 'vocdoni/vocdoni-node'
     defaults:
       run:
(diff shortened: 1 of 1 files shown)


mvdan:
Can you clarify why this is a good thing? I guess there's an advantage to running compose-test in parallel to go-tests, since it's a bit slow.

I also don't understand why we should delete previous files. Are we leaving any behind? I kind of assumed that each job started with a clean directory.

mvdan:
Also, coveralls is failing for no reason. Honestly, I would not use it as a CI check. We could still upload the data though. fyi @p4u

vdo:
> Can you clarify why this is a good thing? I guess there's an advantage to running compose-test in parallel to go-tests, since it's a bit slow.
> I also don't understand why we should delete previous files. Are we leaving any behind? I kind of assumed that each job started with a clean directory.

Yes, we do need to split them in 2 steps because we use a docker container in the first step, and leaves a workspace with files owned by `root` behind. I know it's faster to parallelize, but I want to avoid conflicts with the two jobs running on the same workspace at the same time.

mvdan:
👍 explain that in the commit message, then LGTM

vdo:
Also, since we will have many runners, this won't impact much. The big bottleneck was the number of runners, not the time of the jobs themselves.

vdo:
@mvdan Nevermind, I found a way to run the container as UID 1000, so it doesn't conflict with later/parallel jobs.

mvdan (approved):

PR #76, opened 2021-02-04
WIP: all, add events mechanism

jordipainan:
cc: @mvdan

### DIFF
--- a/events/dispatcher.go
+++ b/events/dispatcher.go
@@ -1,0 +1,189 @@
+package events
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/nsqio/go-diskqueue"
+	"go.vocdoni.io/dvote/crypto/ethereum"
+	"go.vocdoni.io/dvote/log"
+	"go.vocdoni.io/dvote/types"
+	"go.vocdoni.io/dvote/vochain"
+	"go.vocdoni.io/proto/build/go/models"
+	"google.golang.org/protobuf/proto"
+)
+
+// SourceType represents the origin of an event (e.g. scrutinizerOnProcessResult)
+type SourceType byte
+
+const (
+	// ScrutinizerOnprocessresult represents the event created once
+	// the scrutinizer computed successfully the results of a process
+	ScrutinizerOnprocessresult = SourceType(0)
+)
+
+// eventList ties each source type to the event callbacks of the type
+var eventList = [5]map[SourceType]eventCallback{
+	// scrutinizer
+	{
+		ScrutinizerOnprocessresult: onComputeResults,
+	},
+	// keykeeper
+	// ethereum
+	// vochain
+	// census
+}
+
+// Event represents all the info required for an event to be processed
+type Event struct {
+	Data       proto.Message
+	SourceType SourceType
+}
+
+// Dispatcher responsible for pulling work requests
+// and distributing them to the next available worker
+type Dispatcher struct {
+	Signer *ethereum.SignKeys
+	// Storage stores the unprocessed events
+	Storage [len(types.EventSources)]diskqueue.Interface
+	// VochainApp is a pointer to the Vochain BaseApplication allowing to call SendTx method
+	VochainApp *vochain.BaseApplication
+}
+
+// NewDispatcher creates a new dispatcher
+func NewDispatcher(
+	signer *ethereum.SignKeys,
+	vochain *vochain.BaseApplication,
+	dbPath string,
+	maxBytesPerFile int64,
+	minMsgSize,
+	maxMsgSize int32,
+	syncEvery int64,
+	syncTimeout time.Duration,
+	logger diskqueue.AppLogFunc,
+) *Dispatcher {
+	d := &Dispatcher{Signer: signer, VochainApp: vochain}
+	for idx := range d.Storage {
+		d.Storage[idx] = diskqueue.New(
+			types.EventSources[idx],
+			dbPath,
+			maxBytesPerFile,
+			minMsgSize,
+			maxMsgSize,
+			syncEvery,
+			syncTimeout,
+			logger,
+		)
+	}
+	return d
+}
+
+// StartDispatcher creates and starts the workers
+func (d *Dispatcher) StartDispatcher() {
+	// create all of our event workers.
+	for i := 0; i < len(types.EventSources); i++ {
+		log.Debugf("starting event worker", i+1)
+		eventWorker := &eventWorker{
+			storage:    d.Storage[i],
+			signer:     d.Signer,
+			vochainApp: d.VochainApp,
+			handlers:   eventList[i],
+		}
+		go eventWorker.start()
+	}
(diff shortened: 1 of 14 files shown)


jordipainan:
@p4u added diagram and got rid of the cmd/flag syncEvery

events/dispatcher.go
```diff
@@ -0,0 +1,189 @@
...
+func NewDispatcher(
+	signer *ethereum.SignKeys,
+	vochain *vochain.BaseApplication,
+	dbPath string,
+	maxBytesPerFile int64,
+	minMsgSize,
+	maxMsgSize int32,
+	syncEvery int64,
+	syncTimeout time.Duration,
+	logger diskqueue.AppLogFunc,
```
mvdan, line 63: IMO you should not expose all these go-diskqueue options in the constructor. Whatever you can hard-code, like minMsgSize and maxMsgSize, shouldn't be a parameter. Otherwise the NewDispatcher calls have all these obscure internal options that noone will really understand :)

vochain/scrutinizer/vote.go
```diff
@@ -57,12 +58,8 @@ func (s *Scrutinizer) ComputeResult(processID []byte) error {
...
 
 	// add results if process is not live or isLive and status is ended
-	if !isLive || (isLive && p.Status == models.ProcessStatus_ENDED) {
-		for _, l := range s.eventListeners {
-			pv.EntityId = p.EntityId
-			pv.ProcessId = p.ProcessId
-			l.OnComputeResults(pv)
-		}
+	if (!isLive || (isLive && p.Status == models.ProcessStatus_ENDED)) && s.EventDispatcher != nil {
+		go s.EventDispatcher.Collect(pv, events.ScrutinizerOnprocessresult)
```
mvdan, line 62: if you call this in a goroutine, we'll lose the order of the incoming events.

events/dispatcher.go
```diff
@@ -0,0 +1,189 @@
...
+			syncEvery,
+			syncTimeout,
+			logger,
+		)
+	}
+	return d
+}
+
+// StartDispatcher creates and starts the workers
+func (d *Dispatcher) StartDispatcher() {
```
mvdan, line 82: you can just call this Start

events/dispatcher.go
```diff
@@ -0,0 +1,189 @@
...
+		}
+		go eventWorker.start()
+	}
+	log.Infof("events dispatcher started with %d workers", len(types.EventSources))
+}
+
+// Collect receives, check and push the events to the event queue
+func (d *Dispatcher) Collect(eventData proto.Message, sourceType SourceType) {
+	// Now, we take the delay, and the person's name, and make a WorkRequest out of them.
+	if sourceType < SourceType(0) || sourceType > SourceType(types.MaxUint8) {
```
mvdan, line 100: the wrap method already checks sourcetype; remove this

events/dispatcher.go
```diff
@@ -0,0 +1,189 @@
...
+	// receive the event
+	event, err := d.wrap(eventData, sourceType)
+	if err != nil {
+		log.Debugf("cannot receive event: %w", err)
+		return
+	}
+
+	// add event to the database
+	if err := d.add(event); err != nil {
+		log.Debugf("cannot add event into database: %w", err)
```
mvdan, line 112: I assume that lost events should be more than debug logs - at least warnings. Another option is to have Collect return an error, then the producer of events can decide if it's a problem or not for events to be lost.

events/events_test.go
```diff
@@ -0,0 +1,151 @@
...
+		EntityId:  Hex2byte(nil, "2b7222146a805bba0dbb61869c4b3a03209dffba"),
+	}
+	eventData.Votes[0] = &models.QuestionResult{
+		Question: [][]byte{{1}, {1}, {1}},
+	}
+	dispatcherMock.Collect(eventData, events.ScrutinizerOnprocessresult)
+	time.Sleep(time.Millisecond * 500)
+}
+
+func Hex2byte(tb testing.TB, s string) []byte {
```
mvdan, line 59: reuse testutil

events/workers.go
```diff
@@ -0,0 +1,89 @@
...
+		event, err := ew.nextEvent()
+		if err != nil {
+			log.Warnf("cannot get next event: %s", err)
+			continue
+		}
+		if event == nil {
+			time.Sleep(time.Millisecond * types.EventWorkerSleepTime)
+			continue
+			// TODO: mvdan improve this
+		}
```
mvdan, line 54: this event nil check and sleep is now unnecessary, I think, as go-diskqueue will block until there's another event to receive.

events/dispatcher.go
```diff
@@ -0,0 +1,189 @@
...
+type SourceType byte
+
+const (
+	// ScrutinizerOnprocessresult represents the event created once
+	// the scrutinizer computed successfully the results of a process
+	ScrutinizerOnprocessresult = SourceType(0)
+)
+
+// eventList ties each source type to the event callbacks of the type
+var eventList = [5]map[SourceType]eventCallback{
```
mvdan, line 26: I assume this 5 should be a constant, like len(EventSources)

events/dispatcher.go
```diff
@@ -0,0 +1,189 @@
...
+		return fmt.Errorf(
+			"cannot add event %+v to persistent storage, unsupported event origin: %d",
+			event,
+			event.SourceType,
+		)
+	}
+	queueIdx := 0
+	switch est := event.SourceType; {
+	case est < types.ScrutinizerSourceTypeLimit:
+		// nothing to do, 0 is fine
```
mvdan, line 165: I'd just assign to 0 still, for consistency. The default index above can be -1 to separate it from this one.

events/events_test.go
```diff
@@ -0,0 +1,151 @@
...
+			panic(err)
+		}
+		tb.Fatal(err)
+	}
+	return b
+}
+
+// EventWorker is responsible for performing a unit of work given an event
+type EventWorkerMock struct {
+	Storage diskqueue.Interface
```
mvdan, line 72: call this DiskQueue maybe? A few times I misunderstood this field to be a database

events/dispatcher.go
```diff
@@ -0,0 +1,189 @@
...
+	case est >= types.VochainSourceTypeLimit && est < types.CensusSourceTypeLimit:
+		queueIdx = 4
+	default:
+		return fmt.Errorf("cannot add event, invalid queue index")
+	}
+	// byte[0]  -> event type
+	// byte[1:] -> event data
+	err = d.Storage[queueIdx].Put(
+		append(
+			[]byte(fmt.Sprintf("%x", event.SourceType)),
```
mvdan, line 181: I don't understand this. When you read the []byte messages from the disk queue, you use the first byte as the SourceType directly, without any decoding. But here, you encode the source type as hex.

In other words, you don't want to hex-encode here, I think. You can just do `append([]byte{event.SourceType}, eventDataBytes...)`

PR #79, opened 2021-02-05
vochain: select tendermint's DB via build tags

mvdan:
See the comment in vochain/start.go.

I verified that "go test ./..." still succeeds with both -tags=cleveldb
and -tags=badger.

### DIFF
--- a/vochain/start.go
+++ b/vochain/start.go
@@ -109,24 +109,6 @@ func NewTenderLogger(artifact string, disabled bool) *TenderLogger {
 	return &TenderLogger{Artifact: artifact, Disabled: disabled}
 }
 
-/*func checkDBavailable(name string) bool {
-	available := true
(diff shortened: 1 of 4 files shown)


mvdan:
FYI @p4u, as discussed. We can merge this to master today, as it shouldn't break anything.

Note that I'm pretty sure Dockerfile.release *did not* use cleveldb in master before this change, because of the commented out code. It should use cleveldb after this patch.

With this change, in master we can start benchmarking badger vs cleveldb and see if we can fully switch to badger.

p4u:
> Note that I'm pretty sure Dockerfile.release did not use cleveldb in master before this change, because of the commented out code. It should use cleveldb after this patch.

I know, I commented that code when we switched to tendermint 0.34, but in previous versions it was used.

PR #85, opened 2021-02-09
update contract bindings for supporting chainID

jordipainan:
Changes on the contract does not affect the vochain logic as the processID is given by the event created on ethereum.


chain/contracts/namespaces.go
```diff
@@ -30,7 +30,7 @@ var (
...
...18301838901865b8381101561150957601f198784030185526114f7838351611425565b948601949250908501906001016114db565b5050868103606088015287518082529084019450915050818601845b8281101561154a5781516001600160a01b031685529383019390830190600101611525565b50929998505050505050505050565b60006040825261156c6040830185611425565b905061ffff831660208301529392505050565b60208082526011908201527037b7363ca1b7b73a3930b1ba27bbb732b960791b604082015260600190565b6020808252600d908201526c092dcecc2d8d2c840d2dcc8caf609b1b604082015260600190565b602080825260129082015271092dcc8caf05ad6caf240dad2e6dac2e8c6d60731b604082015260600190565b6020808252600b908201526a26bab9ba103234b33332b960a91b604082015260600190565b6020808252600f908201526e105b1c9958591e481c1c995cd95b9d608a1b604082015260600190565b61ffff91909116815260200190565b60405181810167ffffffffffffffff8111828210171561167957600080fd5b604052919050565b600067ffffffffffffffff821115611697578081fd5b5060209081020190565b60005b838110156116bc5781810151838201526020016116a4565b838111156116cb576000848401525b50505050565b6001600160a01b038116811461108757600080fd5b61ffff8116811461108757600080fdfea264697066735822122083dd59a17980180260f9a47cb0d1d0b5fd00415c3dafa67084542b2593a7a5e964736f6c634300060c0033"
```
p4u, line 33: To me it makes more sense to add it in base64 instead than hex. Is that possible?

jordipainan: this is generated code, the only thing that changes is the pkg name to be unified between all contracts.

PR #90, opened 2021-02-10
chain: add oracle secure list for event processing

jordipainan:
Add a list of whitelisted addresses on the Oracle
in order to check if the tx comming from the source of
truth is one of the whitelisted. If the address
that created the tx on the source of truth is on the
list the Oracle will process the event. If not
the event will be ignored.

- add secureTransactorList flag for setting
  the valid addresses, by default all are accepted.
- initialize the events service with the secureTransactorList
- check if the address that created the tx on the source of
  truth is on the oracle secureList before creating the vochain tx

The address of the tx creator on the source of truth is obtained
thanks to the logs. Logs include a tx hash and through
this tx hash we can obtain the tx creator.

Careful: This is not working for internal calls
e.g. 
`Call chain: Contract A -> Contract B -> Contract C`
`msg.Sender: C(from) == B != A`

e.g.
`EOA A -> Contract B`
`msg.Sender: B(from) == A`


chain/ethevents/events.go
```diff
@@ -117,6 +126,16 @@ func NewEthEvents(contractsAddresses []common.Address, signer *ethereum.SignKeys
 	if err != nil {
 		return nil, fmt.Errorf("cannot read token storage proof contract abi: %w", err)
 	}
+
+	secureAddrList := make([]common.Address, len(secureList))
```
p4u: Better use a Map.
A Map is a hash table so for querying if a value exist is much faster than a slice.

chain/ethevents/handlers.go
```diff
@@ -364,3 +379,31 @@ func namespaceOracleRemovedMeta(
...
+	tx, _, err := ethclient.TransactionByHash(ctx, txHash)
+	if err != nil {
+		return fmt.Errorf("cannot fetch tx by hash: %w", err)
+	}
+	if tx != nil {
+		sender, err := ethclient.TransactionSender(ctx, tx, blockHash, txIndex)
+		if err != nil {
+			return fmt.Errorf("cannot fetch tx sender: %w", err)
+		}
+		for _, secureAddr := range secureList {
```
p4u: Same

cmd/dvotenode/dvotenode.go
```diff
@@ -117,6 +117,7 @@ func newConfig() (*config.DvoteCfg, config.Error) {
 	globalCfg.VochainConfig.MempoolSize = *flag.Int("vochainMempoolSize", 20000, "vochain mempool size")
 	globalCfg.VochainConfig.KeyKeeperIndex = *flag.Int8("keyKeeperIndex", 0, "if this node is a key keeper, use this index slot")
 	globalCfg.VochainConfig.ImportPreviousCensus = *flag.Bool("importPreviousCensus", false, "if enabled the census downloader will import all existing census")
+	globalCfg.VochainConfig.SecureTransactorList = *flag.StringArray("secureList", []string{}, "if empty any ethereum address can create and Oracle tx, else only the addresses on the list")
```
p4u: I'd rather use this naming and description:
`ethereumWhiteListAddrs: list of allowed ethereum address for creating processes on the vochain (oracle mode only)`

jordipainan:
Threads addressed, ignore de PR message (commit description changed)

jordipainan:
btw, @p4u this applies for all oracle transactions not just for newProcessTx

chain/ethevents/handlers.go
```diff
@@ -364,3 +379,33 @@ func namespaceOracleRemovedMeta(
...
+	txIndex uint,
+	ethereumWhiteList map[common.Address]bool,
+	ethclient *ethclient.Client,
+) error {
+	// get tx from ethereum
+	tx, _, err := ethclient.TransactionByHash(ctx, txHash)
+	if err != nil {
+		return fmt.Errorf("cannot fetch tx by hash: %w", err)
+	}
+	if tx != nil {
```
mvdan: as far as I can tell, tx is never nil here.

chain/ethevents/events.go
```diff
@@ -117,6 +126,16 @@ func NewEthEvents(contractsAddresses []common.Address, signer *ethereum.SignKeys
 	if err != nil {
 		return nil, fmt.Errorf("cannot read token storage proof contract abi: %w", err)
 	}
+
+	secureAddrList := make(map[common.Address]bool, len(ethereumWhiteList))
+	for _, sAddr := range ethereumWhiteList {
+		addr := common.HexToAddress(sAddr)
+		if (addr == common.Address{}) || (len(addr) != types.EthereumAddressSize) {
```
mvdan: you can drop the parentheses on the right side

chain/ethevents/events.go
```diff
@@ -117,6 +126,16 @@ func NewEthEvents(contractsAddresses []common.Address, signer *ethereum.SignKeys
...
+			return nil, fmt.Errorf("cannot create ethereum whitelist, address %x not valid", addr)
```
mvdan: you should use sAddr in the error message, since addr here could be empty.

and when you do, use %q, since it's already hex-encoded so %x would be double hex encoding.

chain/ethevents/handlers.go
```diff
@@ -364,3 +379,33 @@ func namespaceOracleRemovedMeta(
...
+		// get from
+		sender, err := ethclient.TransactionSender(ctx, tx, blockHash, txIndex)
+		if err != nil {
+			return fmt.Errorf("cannot fetch tx sender: %w", err)
+		}
+		// check from is whitelisted
+		if val, ok := ethereumWhiteList[sender]; !ok {
```
mvdan:
```
if !ethereumWhiteList[sender] {
    return fmt.Errorf(...)
}
```

chain/ethevents/handlers.go
```diff
@@ -364,3 +379,33 @@ func namespaceOracleRemovedMeta(
...
+			return fmt.Errorf("recovered address not in ethereum whitelist")
+		} else {
+			if val {
+				return nil
+			}
+		}
+	}
+	return fmt.Errorf("obtained tx from Ethereum is nil")
```
mvdan: as said before, I don't think this can ever happen since you already checked the error.

mvdan (approved):
I think indexing by ethereum's Address is fine, since it's an array.

jordipainan:
@mvdan thanks for your suggestions, now looks nicer :)

Issue #98, opened 2021-02-13
bug: oracle panics if event log subscription is lost


p4u:
**Describe the bug**

Oracle panics if event log fails/disconnect in some moment

```
2021-02-13T21:29:33Z    FATAL   ethevents/events.go:296 EOF
go.vocdoni.io/dvote/chain/ethevents.(*EthereumEvents).SubscribeEthereumEventLogs
        /src/chain/ethevents/events.go:296
go.vocdoni.io/dvote/service.EthEvents.func1
        /src/service/ethevents.go:56
```

**To Reproduce (please complete the following information)**

Saw it under high load.

jordipainan:
Do not fatal on eth connection lost

PR #100, opened 2021-02-17
do not stop oracle if ethclient disconnects

jordipainan:


jordipainan:
closes #98

chain/chain.go
```diff
@@ -413,10 +408,10 @@ func (e *EthChainContext) SyncGuard(ctx context.Context) {
 			log.Warn("ethereum is experiencing sync problems, restarting node...")
 			e.RestartLock.Lock()
 			if err = e.Node.Close(); err != nil {
-				log.Fatal(err)
+				log.Error(err)
 			}
 			if err = e.Node.Start(); err != nil {
-				log.Fatal(err)
+				log.Error(err)
```
p4u: This is not related to an external Web3 connection, this case should be a Fatal

chain/chain.go
```diff
@@ -248,26 +248,24 @@ func (e *EthChainContext) Start() {
...
-		if err != nil || client == nil {
-			log.Fatalf("cannot create a client connection: (%s), tried %d times.", err, types.EthereumDialMaxRetry)
-		}
+
 		tctx, cancel := context.WithTimeout(context.Background(), types.EthereumReadTimeout)
 		defer cancel()
 		nid, err := client.NetworkID(tctx)
 		if err != nil || nid == nil {
-			log.Fatalf("cannot get network ID from external web3: (%s)", err)
+			log.Errorf("cannot get network ID from external web3: (%s)", err)
```
p4u: log.Error() does not stop the execution. It looks like this is not taken into account on this PR. Make the proper flow modifications to handle Errors instead of Fatals correctly.

p4u (changes requested):

chain/chain.go
jordipainan: this is the case where the node is not synced  and we try to restart it but fails on close or fails on restart. Do we want to try again the restart or the close ? Or do we want to stop all the eth node operations without trying again these operations?

jordipainan:
Behavior:
- If starting node from scratch:
    - If internal node fails: Fatal
    - If external node fails: Error and retry 10 times, if still not working, init all other components and start ethevents only once the client connection is established
- If node already running:
    - Warn user and keep trying to reconnect without affecting other components like keykeeper
    - Once the client is connected again resume the eth events subscription

chain/ethevents/events.go
```diff
@@ -235,55 +235,116 @@ func (ev *EthereumEvents) AddEventHandler(h EventHandler) {
...
+			}
+			lastBlock = blk.Number().Int64()
+			break
 		}
 		// For security, read also the new passed blocks before subscribing
-		ev.processEventLogsFromTo(ctx, startBlock, blk.Number().Int64(), client)
+		// Same as the last processEventLogsFromTo function call
+		if err := ev.processEventLogsFromTo(ctx, lastBlock, blk.Number().Int64(), client); err != nil {
+			log.Errorf("cannot process event logs from block %d to block %d with error: %s", lastBlock, blk.Number().Int64(), err)
+
```
p4u: empty line

chain/chain.go
```diff
@@ -198,7 +198,7 @@ func (e *EthChainContext) Start() {
 
 	if len(e.Keys.Accounts()) < 1 {
 		if err := e.createAccount(); err != nil {
-			log.Error(err)
+			log.Errorf("cannot create ethereum account: %s", err)
```
p4u: this looks like a Fatal

chain/chain.go
```diff
@@ -422,3 +404,24 @@ func (e *EthChainContext) SyncGuard(ctx context.Context) {
...
+func EthClientConnect(dialAddr string, attempts int) (*ethclient.Client, error) {
+	var attemptsCounter int
+	for {
+		if attempts != 0 && attemptsCounter == attempts {
+			return nil, fmt.Errorf("tryied %d to reconnect, connection cannot be established", attemptsCounter)
+		}
+		client, err := ethclient.DialContext(context.Background(), dialAddr)
+		if err != nil || client == nil {
+			log.Errorf("cannot create a client connection: (%s), trying again ...", err)
+			time.Sleep(time.Second * 30)
```
p4u: 30 seconds looks too much

chain/contractWrappers.go
```diff
@@ -59,18 +59,15 @@ func NewVotingHandle(contractsAddress []common.Address, dialEndpoint string) (*V
...
 	ph := new(VotingHandle)
 	// try connect to the w3endpoint using an RPC connection
-	for i := 0; i < types.EthereumDialMaxRetry; i++ {
+	for {
 		ph.EthereumRPC, err = ethrpc.Dial(dialEndpoint)
 		if err != nil || ph.EthereumRPC == nil {
-			log.Warnf("cannot create a ethereum rpc connection: (%s), trying again (%d of %d)", err, i+1, types.EthereumDialMaxRetry)
-			time.Sleep(time.Second * 2)
+			log.Warnf("cannot create a ethereum rpc connection: (%s), trying again ...", err)
+			time.Sleep(time.Second * 30)
```
p4u: Same...

p4u (changes requested):

chain/chain.go
jordipainan: Done, also capture the error on the account unlock. Fatal if cannot unlock account

chain/chain.go
jordipainan: Reduced to 10

chain/contractWrappers.go
jordipainan: Also reduced to 10

PR #107, opened 2021-02-24
api: add namespace arg on getProcessList

jordipainan:
- The namespace of a processes contract can be passed
  as an argument to the getProcessList request.
  This allows to fetch only the processes that are registred
  on a specific namespace and not all the processes of the vochain
  for a given entity.
- The 0 value is valid and represents the namespace 0.
- Is possible to obtain all the processes of all namespaces
  just by not passing the namespace argument on the request at all.

### DIFF
--- a/vochain/scrutinizer/process.go
+++ b/vochain/scrutinizer/process.go
@@ -21,7 +21,7 @@ func (s *Scrutinizer) ProcessInfo(pid []byte) (*models.Process, error) {
 }
 
 // ProcessList returns the list of processes (finished or not) for a specific entity.
-func (s *Scrutinizer) ProcessList(entityID []byte, fromID []byte, max int64) ([][]byte, error) {
+func (s *Scrutinizer) ProcessList(entityID []byte, fromID []byte, max int64, namespace *uint32) ([][]byte, error) {
 	plistkey := []byte{types.ScrutinizerEntityPrefix}
 	plistkey = append(plistkey, entityID...)
 	processList, err := s.Storage.Get(plistkey)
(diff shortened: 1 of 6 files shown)


vochain/scrutinizer/process.go
```diff
@@ -37,8 +37,18 @@ func (s *Scrutinizer) ProcessList(entityID []byte, fromID []byte, max int64) ([]
 		if !fromLock {
 			keyCopy := make([]byte, len(process))
 			copy(keyCopy, process)
-			processListResult = append(processListResult, keyCopy)
-			max--
+			if p, err := s.VochainState.Process(keyCopy, true); err != nil {
+				log.Debugf("error fetching process %x: %s", p.ProcessId, err)
+				continue
+			} else if namespace == nil {
```
p4u, line 43: There is no need to do an else here if there is a `continue` on the line above.

vochain/scrutinizer/process.go
```diff
@@ -37,8 +37,18 @@ func (s *Scrutinizer) ProcessList(entityID []byte, fromID []byte, max int64) ([]
 		if !fromLock {
 			keyCopy := make([]byte, len(process))
 			copy(keyCopy, process)
-			processListResult = append(processListResult, keyCopy)
-			max--
+			if p, err := s.VochainState.Process(keyCopy, true); err != nil {
```
p4u, line 40: I don't like this approach, is very expensive. For each ProcessList query we'll be fetching all processes from the state...
I'd rather try to save on the processList scrutinizer storage the namespace. So splinting by bytes (reserving N bytes for the namespace) would make the trick.

vochain/scrutinizer/process.go
```diff
@@ -37,8 +37,18 @@ func (s *Scrutinizer) ProcessList(entityID []byte, fromID []byte, max int64) ([]
...
+				log.Debugf("error fetching process %x: %s", p.ProcessId, err)
+				continue
```
p4u, line 42: If there is a continue here the else is not required....

p4u (changes requested):
I can work on this if you want. Just let me know and I'll take the work from your current branch.

jordipainan:
Yes ofc, feel free to continue the work from here.
Just to point it out, any solution will be a tmp workaround since the right solution is to not use the scrutinizer for getting the entities processes but the process tree (indexing each process by its entityid)

vochain/scrutinizer/process.go
jordipainan: the else is required for accessing p afterwards. If you do not want the else here you should define p outside the if

vochain/scrutinizer/process.go
jordipainan: agree, this is much more efficient

vochain/scrutinizer/process.go
jordipainan: same as before

p4u:
Superseded by #110

Issue #34, opened 2021-01-05
IAVL statedb version does not exist


p4u:
```
2021-01-05T11:00:00Z    FATAL    vochain/start.go:32    cannot init vochain application: cannot create vochain state: (version does not exist)
go.vocdoni.io/dvote/vochain.NewVochain
    /home/mvdan/src/dvote/vochain/start.go:32
go.vocdoni.io/dvote/service.Vochain
    /home/mvdan/src/dvote/service/vochain.go:105
main.main
    /home/mvdan/src/dvote/cmd/dvotenode/dvotenode.go:461
runtime.main
    /home/mvdan/tip/src/runtime/proc.go:225
exit status 1
```
Probably triggered by this line: https://github.com/vocdoni/go-dvote/blob/master/vochain/state.go#L99

It can be reproduced by starting a dvotenode (from scratch) and stopping it before the first block is commited (CTRL+C).

jordipainan:
```
### Trees stacktrace

## Run for first time

service/vochain.go:90	        genesis does not exist, using hardcoded genesis
--------- INIT ---------
iavlstate/iavlstate.go:44	    creating new GolevelDB database
iavlstate/iavlstate.go:58	    creating version tree (NewMutableTree)
iavlstate/iavlstate.go:62	    loading iavlState version 0
--------- LOAD_VERSION(0) ---------
iavlstate/iavlstate.go:115	    calling LoadVersion(0)
iavlstate/iavlstate.go:125	    loading latest versions tree from disk
--------- UPDATE_IMMUTABLES ---------
iavlstate/iavlstate.go:156	    no error on updateImmutables
--------- ADD_TREE ---------
iavlstate/iavlstate.go:161	    calling AddTree(app)
iavlstate/iavlstate.go:186	    added app tree with latest commited version 0
iavlstate/iavlstate.go:161	    calling AddTree(process)
iavlstate/iavlstate.go:186	    added process tree with latest commited version 0
iavlstate/iavlstate.go:161	    calling AddTree(vote)
iavlstate/iavlstate.go:186	    added vote tree with latest commited version 0
vochain/state.go:119	        application trees successfully loaded at version 0

--------- Add header key to App tree (app.go 130) ---------

--> After key added a SAVE is performed
--> SAVE calls COMMIT and Version 0 is saved

--> New version is 1

iavlstate/iavlstate.go:138  	found app i.tree
[...]

PR #108, opened 2021-02-25
fix vochain db fatal on startup:

jordipainan:
First, when trying to read priv_validator_state.json,
if empty the Tendermint lib executes an os.Exit internally
as the file does not have any bytes to read.
The os.Exit cannot be handled at a lib level.
The applied fix opens or creates the mentioned file
and adds the default values that Tendermint expects to have,
before the Tendermint lib tries to open the file.

Second, the syncronization of the node starts always at height
N-1. There is an edge case where the node is halted before
the first block is processed. Then when the node is started again
the latest version does not exist.
If the latest version fails to load, also version 0, there is
nothing that can be done rather than restarting the entire database,
thus avoiding corruption on state bytes.
A backup of all the vochain data is stored for security reasons.

### DIFF
--- a/vochain/state.go
+++ b/vochain/state.go
@@ -5,10 +5,14 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"os"
+	"path/filepath"
 	"sync"
 	"time"
 
+	"github.com/cosmos/iavl"
 	"github.com/ethereum/go-ethereum/common"
+	dcopy "github.com/otiai10/copy"
 	tmcrypto "github.com/tendermint/tendermint/crypto"
 	ed25519 "github.com/tendermint/tendermint/crypto/ed25519"
 	"go.vocdoni.io/dvote/log"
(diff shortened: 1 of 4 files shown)


jordipainan:
closes #34

jordipainan:
@p4u I have some doubts of storing a backup of the /data directory. That is because if the Load fails at version 0 the /data directory does not have any important info to backup, but long as someone wants to have the first evidences, index or whatever the tendermint daemon collected I do not dislike the idea of having this backup.

I choose this way because I consider it the most secure. Better to have a backup before deleting something.

vochain/state.go
```diff
@@ -79,33 +83,61 @@ type ImmutableState struct {
...
+		return nil, fmt.Errorf("cannot init db: %s", err)
 	}
-	if err := vs.Store.AddTree(VoteTree); err != nil {
-		return nil, err
-	}
-
 	// Must be -1 in order to get the last committed block state, if not block replay will fail
 	if err = vs.Store.LoadVersion(-1); err != nil {
-		return nil, err
+		if err == iavl.ErrVersionDoesNotExist {
```
p4u, line 89: Don't do that, its breaking the State interface. I think this error should be handled on the statedb/iavltree implementation, not here.

p4u:
Speaking in general, I think you are taking the wrong approach. Restarting the state should not be the fix, we might find a way to avoid this problem by making the disk storage safe (instead of accepting corruption).

Also that looks like a iavltree specific problem. Are you able to reproduce it by switching to graviton?

p4u:
After some private conversations I think your workaround makes sense as a temporary solution but I'd prefer to explore a bit more the internals of this issue before merging it. Mainly because some months ago I already made some initial explorations regarding this issue, I'll try to recover the branch that I started at that time and check if there is something we might reuse from that.

p4u:
@jordipainan lets merge your workaround but let's keep the issue so we take a deeper look after Open Access. However I'd prefer to remove the backup stuff.

jordipainan:
@p4u db backup deleted and rebased with master

vochain/state.go
```diff
@@ -79,33 +81,50 @@ type ImmutableState struct {
...
```
p4u, line 89: What about moving this snippet to the `statedb/iavltree` implementation? It could be added to the `LoadVersion()` method. Then we would make this only executable if IAVL is actually used.

jordipainan: Tried that, but I am getting weird errors like:
- Cannot init state, resource temporary unavailable.
- Oracle already added (On init chain), which is a fatal.
- Same as commented before but for validators.

None of this errors is raised going back to the original code, I can spend some time investigating this weird behavior or we could just merge this and we can tackle this afterwards.

PR #110, opened 2021-03-02
use badgerhold as KV database replacement on scrutinizer

p4u:
Using badgerhold instead of plain KV store (badger) brings a new set of
possibilities for querying the Vochain. We can now easily request
information with a SQL-like style of syntax while preserving the
simplicity and low resource consumption of using a KV store.

This is the first step for renaming the current scrutinizer to indexer
(on a more generic purpose) and for decopling the consensus state from
the API querying database.


vochain/scrutinizer/process.go
```diff
@@ -90,144 +142,127 @@ func (s *Scrutinizer) isLiveResultsProcess(processID []byte) (bool, error) {
...
-		return append([]byte{types.ScrutinizerEntityPrefix}, data...)
-	case "liveProcess":
-		return append([]byte{types.ScrutinizerLiveProcessPrefix}, data...)
-	case "results":
-		return append([]byte{types.ScrutinizerResultsPrefix}, data...)
-	case "processEnding":
-		return append([]byte{types.ScrutinizerProcessEndingPrefix}, data...)
-	}
-	panic("scrutinizer encode type not known")
+	/*
```
jordipainan: Should this commented code be deleted?

types/api.go
```diff
@@ -28,11 +28,11 @@ type MetaRequest struct {
 	Content      []byte     `json:"content,omitempty"`
 	Digested     bool       `json:"digested,omitempty"`
 	EntityId     HexBytes   `json:"entityId,omitempty"`
-	From         int64      `json:"from,omitempty"`
-	FromID       HexBytes   `json:"fromId,omitempty"`
-	ListSize     int64      `json:"listSize,omitempty"`
+	From         int        `json:"from,omitempty"`
+	ListSize     int        `json:"listSize,omitempty"`
```
mvdan, line 32: removing FromID is a breaking change, right?

Also, I think using `int` instead of `int64` for the public API type is a bad idea. the API will have different integer precision depending on the machine that's running the server. you want it to be fixed, so int64, int32, uint64, etc.

vochain/scrutinizer/process.go
```diff
@@ -1,87 +1,139 @@
...
-			max--
-		}
-		if fromLock && bytes.Equal(fromID, process) {
-			fromLock = false
-		}
+func (s *Scrutinizer) ProcessList(
+	entityID []byte,
+	namespace uint32,
+	from,
+	max int) ([][]byte, error) {
```
mvdan: any particular reason you write func definitions like this now? it almost seems like you're trying to stick to 40 columns or something. I personally don't find it particularly easy to read, and it takes a ton of vertical space :)

mvdan: if this is because of the line length linter, that's part of why I think it's a bad linter. it enforces strict line length limits, which are often a bad idea.

it's true that a func decl taking 140 columns is generally bad, but then you should just split that line into two of ~70 columns. A function definition that takes 60, 80, or even 100 columns in a single line is perfectly fine and should not be split.

vochain/scrutinizer/process.go
```diff
@@ -1,87 +1,139 @@
...
-	process := []string{}
-	for _, p := range s.List(max, fromID, []byte{types.ScrutinizerResultsPrefix}) {
-		process = append(process, fmt.Sprintf("%x", p))
-	}
-	return process
+// ProcessListWithResults returns the list of process ID with already computed results.
+// TODO: use [][]byte instead of []string as return value
+func (s *Scrutinizer) ProcessListWithResults(max, from int) []string {
+	procs := []string{}
+	_ = s.db.ForEach(
```
mvdan: ignoring the error altogether seems like a bad idea. at least you should log it.

vochain/scrutinizer/process.go
```diff
@@ -1,87 +1,139 @@
...
 // ProcessListWithLiveResults returns the list of process ID which have live results (not encrypted)
-func (s *Scrutinizer) ProcessListWithLiveResults(max int64, fromID []byte) []string {
-	process := []string{}
-	for _, p := range s.List(max, fromID, []byte{types.ScrutinizerLiveProcessPrefix}) {
-		process = append(process, fmt.Sprintf("%x", p))
-	}
-	return process
+func (s *Scrutinizer) ProcessListWithLiveResults(max, from int) []string {
+	procs := []string{}
+	_ = s.db.ForEach(
```
mvdan: ditto

vochain/scrutinizer/scrutinizer.go
```diff
@@ -75,22 +62,19 @@ func (s *Scrutinizer) Commit(height int64) {
...
+		if err = s.newEmptyProcess(p.ProcessID); err != nil {
+			log.Warnf("commit: cannot create new empty process: %v", err)
 		}
 	}
 
-	for i, p := range s.resultsPool {
-		s.registerPendingProcess(p.ProcessID, height+int64(i+1))
+	// Update existing processes
+	for _, p := range s.updateProcessPool {
+		if err = s.updateProcess(p); err != nil {
```
mvdan: use `:=`

vochain/scrutinizer/scrutinizer_test.go
```diff
@@ -84,20 +91,38 @@ func testProcessList(t *testing.T, procsCount int) {
...
 	// Add 10 entities and process for storing random content
 	for i := 0; i < 10; i++ {
-		sc.addEntity(util.RandomBytes(20), util.RandomBytes(32))
+		pid := util.RandomBytes(32)
+		if err := state.AddProcess(&models.Process{
+			ProcessId:   pid,
+			EntityId:    util.RandomBytes(20),
+			VoteOptions: &models.ProcessVoteOptions{MaxCount: 8, MaxValue: 3},
+		}); err != nil {
+			t.Fatal(err)
```
mvdan: might as well use qt.Assert everywhere? :)

vochain/scrutinizer/vote.go
```diff
@@ -167,78 +129,64 @@ func (s *Scrutinizer) addLiveResultsVote(envelope *models.Vote) error {
...
-	if err := s.Storage.Put(s.Encode("liveProcess", pid), processBytes); err != nil {
-		return err
+		return fmt.Errorf("too many elements on addVote")
 	}
 
-	log.Debugf("addVote %v on process %x", vote.Votes, pid)
-	return nil
+	return s.db.UpdateMatching(&Results{},
+		badgerhold.Where("ProcessID").Eq(pid),
+		func(record interface{}) error {
```
mvdan, line 151: would be nice if badgerhold allowed you to use `func(*Results) error` here, just like with ForEach. Maybe we could file an issue about that.

mvdan (approved):
I assume you've tested this very well? It's a large refactor and I bet at least a couple of things will be broken :)

vochain/scrutinizer/process.go
```diff
@@ -90,144 +142,127 @@ func (s *Scrutinizer) isLiveResultsProcess(processID []byte) (bool, error) {
...
-	}
-	if err == badger.ErrKeyNotFound {
-		log.Infof("added new entity %x to scrutinizer", eid)
-	}
-	if len(pid) != types.ProcessIDsize {
-		log.Errorf("addEntity: pid size is not correct, got %d", len(pid))
-		return
+// updateProcess synchronize those fields that can be updated on a existing process
+// with the information obtained from the Vochain state
+func (s *Scrutinizer) updateProcess(pid []byte) error {
```
jordipainan, line 242: This function is called upon some specific events happening on the Vochain state like onCancel() or OnRevealKeys().
The OnRevealKeys event changes the process keys fields but not other fields and with updateProcess some of the unchanged fields are asigned again.
Can this function behaves depending on the event that is triggered? e.g. if the only field changed is Status there is no need to asign again other fields. Does it make sense ? I see this function too generic.

vochain/scrutinizer/scrutinizer.go
```diff
@@ -43,27 +32,25 @@ func (s *Scrutinizer) AddEventListener(l EventListener) {
...
-	VochainState   *vochain.State
-	Storage        db.Database
-	votePool       []*models.Vote
-	processPool    []*types.ScrutinizerOnProcessData
-	resultsPool    []*types.ScrutinizerOnProcessData
-	entityCount    int64
+	VochainState      *vochain.State
+	votePool          []*models.Vote
+	newProcessPool    []*types.ScrutinizerOnProcessData
+	updateProcessPool [][]byte
```
jordipainan, line 41: Can you explain a bit more the introduction of the `updateProcessPool` field on the commit msg ?

vochain/scrutinizer/db.go
```diff
@@ -0,0 +1,48 @@
+package scrutinizer
+
+import (
+	"time"
+
+	"github.com/timshannon/badgerhold/v3"
+	"go.vocdoni.io/proto/build/go/models"
+)
+
+type Process struct {
```
jordipainan, line 10: Is it possible to avoid this definitions and use the protobuf models created? Badgerhold uses `encoding/gov` as a default encoding mechanism but it can be changed for your desired one on badgerhold.Options.

types/api.go
p4u: I used "int" because that's the same type used by badgerhold for the Skip() statement. I think that is the most correct approach since if the machine have int=64b then both badgerhold and the API will use the same type.

vochain/scrutinizer/process.go
p4u: Yes, the reason is because a limit of 111 characters per line is enforced by the CI and I think that's a good number to respect. Gofumt (you recommended it) also likes this way of writing function headers btw.

I can write it with less columns per function definition (splitting in two or three columns max). However that's a personal style preference, AFAIK there is nothing officially written regarding this matter.

vochain/scrutinizer/scrutinizer.go
p4u: Well, that's part of the badgerhold usage requirement, not sure adding it to the commit would be of use to anyone. I can instead add a comment in the code.

vochain/scrutinizer/db.go
p4u: What do you mean? Envelope, Mode and VoteOpts already use the protobuf types. For the rest I don't see the need since adding the Proto types means adding more complexity to the database and also size. I.e Status is a int32 type in the protobuf models but with some extra fields which are only required for Protobuf (some of them are also not exported). If you need to have the protobuf type, a simple conversion type is enough (models.Status(int32)).

vochain/scrutinizer/scrutinizer.go
p4u: That's the never ending discussion and still not clear to me whats the most correct approach... But ok, I'll declare a new var for this error.

vochain/scrutinizer/scrutinizer_test.go
p4u: Actually a full qt.Assert refactor needs to be performed on this test file. I only added qt.Assert to the new lines to not change everything on a commit which is not directly related with this refactor (note that this snippet is a copy&paste from an old snippet). But sure, I could change it anyway on this specific case.

vochain/scrutinizer/vote.go
p4u: Sure

vochain/scrutinizer/process.go
p4u: The reason i ignored it was because I'm returning "nil" on the ForEach callback. But looking deeper it could also return an error if the ForEach statement has a wrong type. Thus I'll capture the error, makes sense. Thanks.

vochain/scrutinizer/process.go
p4u: Yes. Thanks.

vochain/scrutinizer/process.go
p4u: I was hoping that badgerHold Update() function would only change those non-zero value defined fields, but that's not the case. So instead of adding more complexity (like doing specific calls for each type of event) to the code and thus more probability of adding bugs, I decided to use a generic approach since the costs of writing 1 field or 5 fields of a struct, once the badger index has been found, loaded to memory and so on, its almost negligible.

vochain/scrutinizer/db.go
p4u: Ah, now I get what you mean. Using Protobuf as serialization mechanism for the whole badgerHold database.

TBH I'm not sure that's a good idea. That means we need to define the static types on a Protobuf types thus we loose flexibility for performing changes and we are also very limited to the Protobuf v3 types (which are not as rich as all GoLang types). 

If Protobuf encoding is much better than encoding/gov for the kind of types we use in therms of performance and size, then it might be worth it. However I'd let this for a future refactor since it requires some benchmarking and research. So we can add a new issue in order to do it in the future but I don't think its a good idea to do it on this very same PR (which is already quite big and risky).

vochain/scrutinizer/process.go
jordipainan: Oki, fine with that :) Thank you

vochain/scrutinizer/scrutinizer.go
jordipainan: Oki, yes would be nice to have it commented somewhere. As it is a requirement of badgerhold better on the code than on the commit like you said. Thanks

vochain/scrutinizer/db.go
jordipainan: Yes I am referring to use protobuf for the whole db.
I am fine to tackle this on another moment and we can do a more in depth analysis.

Thank you for the explanation !

p4u:
> I assume you've tested this very well? It's a large refactor and I bet at least a couple of things will be broken :)

Well, the `scrutinizer_test` is quite complete, so I assume all previous functionalities work as expected.

PR #113, opened 2021-03-04
Add extra commands

vdo:
This PR adds a MVP version of a CLI for the commands:

census add
census size
census root
info
file fetch
file add
file list
block height
block status
process list
process keys
process results

### DIFF
--- a/cmd/dvotecli/commands/census.go
+++ b/cmd/dvotecli/commands/census.go
@@ -1,0 +1,216 @@
+package commands
+
+import (
+	"encoding/hex"
+	"fmt"
+
+	"github.com/spf13/cobra"
+	"go.vocdoni.io/dvote/client"
+	"go.vocdoni.io/dvote/types"
+	"go.vocdoni.io/dvote/util"
+)
+
+var censusCmd = &cobra.Command{
+	Use:   "census",
+	Short: "census subcommands",
+}
+
+var censusAddCmd = &cobra.Command{
+	Use:   "add",
+	Short: "add [census id] [claim1] ...",
+	RunE:  censusAdd,
+}
+
+var claimCmd = &cobra.Command{
+	Use:   "claim [id] [key]",
+	Short: "adds a single claim (hashed pubkey) to a census id",
+	RunE:  addClaim,
+}
+
+var getRootCmd = &cobra.Command{
+	Use:   "root [id]",
+	Short: "return the root hash of the census merkle tree",
+	RunE:  getRoot,
+}
+
+var getSizeCmd = &cobra.Command{
+	Use:   "size [id]",
+	Short: "return the size of the census",
+	RunE:  getSize,
+}
+
+var genProofCmd = &cobra.Command{
+	Use:   "genproof [id] [key] [value]",
+	Short: "generate a proof of a given value",
+	RunE:  genProof,
+}
+
+func init() {
+	rootCmd.AddCommand(censusCmd)
+	censusCmd.AddCommand(censusAddCmd)
+	censusCmd.AddCommand(claimCmd)
+	censusCmd.AddCommand(getRootCmd)
+	censusCmd.AddCommand(getSizeCmd)
+	censusCmd.AddCommand(genProofCmd)
+	claimCmd.Flags().BoolVarP(&opt.digested, "digested", "", false,
+		"will digest value in the gateway if false")
+}
+
+func censusAdd(cmd *cobra.Command, args []string) error {
+	if err := opt.checkSignKey(); err != nil {
+		return err
+	}
+
+	if len(args) < 1 {
+		return fmt.Errorf("you must provide a census id")
+	}
+
+	cl, err := client.New(opt.host)
+	if err != nil {
+		return err
+	}
+	defer cl.CheckClose(&err)
+
+	req := types.MetaRequest{
+		Method:   "addCensus",
+		CensusID: args[0],
+	}
+
+	if len(args) > 1 {
+		req.PubKeys = args[1:]
+	}
+
+	resp, err := cl.Request(req, opt.signKey)
+	if err != nil {
+		return err
+	}
+	if !resp.Ok {
+		return fmt.Errorf(resp.Message)
+	}
+	fmt.Printf("CensusID: %v\n", resp.CensusID)
+	fmt.Printf("URI: %v\n", resp.URI)
+
+	return err
+}
+
+func addClaim(cmd *cobra.Command, args []string) error {
+	if err := opt.checkSignKey(); err != nil {
+		return err
+	}
+
+	if len(args) < 2 {
+		return fmt.Errorf("you must provide a census id and a claim key")
+	}
+
+	cl, err := client.New(opt.host)
+	if err != nil {
+		return err
+	}
+	defer cl.CheckClose(&err)
+
+	req := types.MetaRequest{
+		Method:    "addClaim",
+		CensusID:  args[0],
+		Digested:  opt.digested,
+		CensusKey: []byte(args[1]),
+	}
+
+	resp, err := cl.Request(req, opt.signKey)
+	if err != nil {
+		return err
+	}
+
+	if !resp.Ok {
+		return fmt.Errorf(resp.Message)
+	}
+	fmt.Printf("Root: %v\n", resp.Root)
+
+	return err
+}
+
+func getRoot(cmd *cobra.Command, args []string) error {
+	if len(args) < 1 {
+		return fmt.Errorf("you must provide a census id")
+	}
+
+	cl, err := client.New(opt.host)
+	if err != nil {
+		return err
+	}
+	defer cl.CheckClose(&err)
+
+	req := types.MetaRequest{
+		Method:   "getRoot",
+		CensusID: args[0],
+	}
+	resp, err := cl.Request(req, nil)
+	if err != nil {
+		return err
+	}
+
+	if !resp.Ok {
+		return fmt.Errorf(resp.Message)
+	}
+	fmt.Printf("Root: %v", hex.EncodeToString(resp.Root))
+	return err
+}
+
+func getSize(cmd *cobra.Command, args []string) error {
+	if len(args) < 1 {
+		return fmt.Errorf("you must provide a census id")
+	}
+
+	cl, err := client.New(opt.host)
+	if err != nil {
+		return err
+	}
+	defer cl.CheckClose(&err)
+
+	req := types.MetaRequest{
+		Method:   "getSize",
+		CensusID: args[0],
+	}
+	resp, err := cl.Request(req, nil)
+	if err != nil {
+		return err
+	}
+
+	if !resp.Ok {
+		return fmt.Errorf(resp.Message)
+	}
+	fmt.Printf("Size: %v", resp.Size)
+	return err
+}
+
+func genProof(cmd *cobra.Command, args []string) error {
+	if len(args) < 3 {
+		return fmt.Errorf("you must provide a census id, a key, and a value to proof")
+	}
+
+	cl, err := client.New(opt.host)
+	if err != nil {
+		return err
+	}
+	defer cl.CheckClose(&err)
+
+	req := types.MetaRequest{
+		Method:    "genProof",
+		CensusID:  args[0],
+		Digested:  opt.digested,
+		CensusKey: []byte(args[1]),
+	}
+	req.CensusValue, err = hex.DecodeString(util.TrimHex(args[2]))
+	if err != nil {
+		return err
+	}
+	resp, err := cl.Request(req, nil)
+	if err != nil {
+		return err
+	}
+
+	if !resp.Ok {
+		return fmt.Errorf(resp.Message)
+	}
+	fmt.Printf("Siblings: %v", resp.Siblings)
+	return err
+}
--- a/cmd/dvotecli/commands/entity.go
+++ b/cmd/dvotecli/commands/entity.go
@@ -1,0 +1,65 @@
+package commands
+
+import (
+	"bytes"
+	"fmt"
+
+	"github.com/spf13/cobra"
+	"go.vocdoni.io/dvote/client"
+	"go.vocdoni.io/dvote/router"
+	"go.vocdoni.io/dvote/types"
+)
+
+var (
+	entityCmd = &cobra.Command{
+		Use:   "entity list",
+		Short: "entity subcommands",
+	}
+	entityListCmd = &cobra.Command{
+		Use:   "list [prefix]",
+		Short: "list entities matching an optional prefix",
+		RunE:  entityList,
+	}
+)
+
+func init() {
+	rootCmd.AddCommand(entityCmd)
+	entityCmd.AddCommand(entityListCmd)
+}
+
+func entityList(cmd *cobra.Command, args []string) error {
+	cl, err := client.New(opt.host)
+	if err != nil {
+		return err
+	}
+	defer cl.CheckClose(&err)
+
+	req := types.MetaRequest{Method: "getEntityList"}
+	resp, err := cl.Request(req, nil)
+	if err != nil {
+		return err
+	}
+	if !resp.Ok {
+		return fmt.Errorf(resp.Message)
+	}
+
+	entities := append([]string{}, resp.EntityIDs...)
+	if len(entities) == router.MaxListSize {
+		for i, count := 1, len(entities); count > 0 && i < MaxListIterations; i++ {
+			req.From = i * router.MaxListSize
+			resp, err := cl.Request(req, nil)
+			if err != nil {
+				return err
+			}
+			entities = append(entities, resp.EntityIDs...)
+			count = len(resp.EntityIDs)
+		}
+	}
+
+	buffer := new(bytes.Buffer)
+	for _, ent := range entities {
+		buffer.WriteString(ent + "\n")
+	}
+	fmt.Print(buffer.String())
+	return err
+}
--- a/cmd/dvotecli/commands/file.go
+++ b/cmd/dvotecli/commands/file.go
@@ -1,0 +1,144 @@
+package commands
+
+import (
+	"bytes"
+	"encoding/base64"
+	"encoding/json"
+	"fmt"
+
+	"github.com/spf13/cobra"
+	"go.vocdoni.io/dvote/client"
+	"go.vocdoni.io/dvote/types"
+)
+
+var fileCmd = &cobra.Command{
+	Use:   "file",
+	Short: "file subcommands",
+}
+
+var fileAddCmd = &cobra.Command{
+	Use:   "add [name] [base64 payload]",
+	Short: "add file from a Base64 payload",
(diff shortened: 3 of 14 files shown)


p4u:
Looks good, congratulations @vdo

As a first impression, I think it is worth to use HTTP(s) instead of Websocket since your code is not taking profit of persistent connections. That should be a quite straight forward change.

Also it would be nice if you can install golangci-lint and be sure you apply all suggestions (there is a `.golangci.yml` file already in go-dvote that you might use.) It can also be integrated into vs-code very easily.

p4u:
Regarding the vs-code configuration, that is the one I use (and works quite well).

+ Go: Build on save: package
+ Go: Format tool: gofumpt
+ Go: Lint Flags: --fast
+ Go Lint On Save: package
+ Go: Lint Tool: golangci-lint
+ Go: Use Language Server

@jordipainan @elboletaire that could be of interest for you too

vdo:
> As a first impression, I think it is worth to use HTTP(s) instead of Websocket since your code is not taking profit of persistent connections. That should be a quite straight forward change.

You can actually use any of both protocols seamlessly, but I will change the example so it uses https

PS: Ok, now I see what you mean. The default connection should be HTTP(s) only, and reserve WS only for json-client, right?

p4u:
@vdo the CLI connection seems websockets only, not HTTP+WS 

https://github.com/vocdoni/vocdoni-node/pull/113/files#diff-06f81856bc419bca73622ac9f74b7aa26321966a542a2b8d7ae7854329db6b50R40

p4u:
WS is legacy and will probably disappear in some moment

vdo:
> WS is legacy and will probably disappear in some moment


Shouldn't this be implemented in go.vocdoni.io/dvote  first ? I'm using it since it implements MetaRequests and so on.

p4u:
Yes, change the client to HTTP is a pending task. Until now the client package was only used for testing but now its giving support to a cli-tool, so that could be a moment for making the migration to HTTP.

p4u:
Lets merge this using WebSockets and let's add an issue for switching to WS to HTTP on the near future @vdo

cmd/dvotecli/commands/process.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	cl, err := client.New(host)
+	if err != nil {
+		return err
+	}
+	defer cl.Conn.Close(websocket.StatusNormalClosure, "")
+
+	var req types.MetaRequest
+	req.Method = "getProcessList"
```
p4u: Needs pagination. If the number of processIDs is equal to 64, then use the field "From=64" on a next call to fetch the remaining processes. Do this until the result is empty or the length is smaller than 64

p4u: Same for all "list" methods.

cmd/dvotecli/commands/process.go
```diff
@@ -0,0 +1,142 @@
...
+	RunE:  getResults,
+}
+
+func init() {
+	rootCmd.AddCommand(processCmd)
+	processCmd.AddCommand(processListCmd)
+	processCmd.AddCommand(processKeysCmd)
+	processCmd.AddCommand(processResultsCmd)
+}
+
```
p4u, line 44: Add "entityList" also to fetch the list of entities registered in the blockchain: https://docs.vocdoni.io/#/architecture/services/gateway?id=get-entity-list

p4u (changes requested):

cmd/dvotecli/commands/process.go
vdo: I will use `entityList` instead of `getScrutinizerEntities`  in `entity.go`, also with pagination

cmd/dvotecli/commands/census.go
```diff
@@ -0,0 +1,210 @@
...
+		return fmt.Errorf("you must provide a census id")
+	}
+
+	cl, err := client.New(host)
+	if err != nil {
+		return err
+	}
+	defer cl.Conn.Close(websocket.StatusNormalClosure, "")
+
+	var req types.MetaRequest
```
p4u: This is more idiomatic: `req := types.MetaRequest{ Method: "getSize", CensusID: args[0] }`

cmd/dvotecli/commands/census.go
```diff
@@ -0,0 +1,210 @@
...
+		return fmt.Errorf("you must provide a census id, a key, and a value to proof")
+	}
+
+	cl, err := client.New(host)
+	if err != nil {
+		return err
+	}
+	defer cl.Conn.Close(websocket.StatusNormalClosure, "")
+
+	var req types.MetaRequest
```
p4u: Same

cmd/dvotecli/commands/common.go
```diff
@@ -0,0 +1,26 @@
+package commands
+
+import (
+	"fmt"
+
+	"go.vocdoni.io/dvote/crypto/ethereum"
+)
+
+const MaxListIterations = 256
```
p4u: Shouldn't you take MaxListSize from the scrutinizer package?

cmd/dvotecli/commands/common.go
```diff
@@ -0,0 +1,26 @@
...
+
+var (
+	digested bool
+	signKey  *ethereum.SignKeys
+)
+
+func checkSignKey() error {
+	signKey = ethereum.NewSignKeys()
```
p4u: Using signKey and privKey as global vars is not a good practice.  But in case you really need to have it globally declared, this function would need a refactor:
```
if privKey == "" { return fmt.Errorf("...") }
return signKey.AddHexKey(privKey)
```

cmd/dvotecli/commands/entity.go
```diff
@@ -0,0 +1,63 @@
...
+
+func entityList(cmd *cobra.Command, args []string) error {
+
+	cl, err := client.New(host)
+	if err != nil {
+		return err
+	}
+	defer cl.Conn.Close(websocket.StatusNormalClosure, "")
+
+	var req types.MetaRequest
```
p4u: same

cmd/dvotecli/commands/entity.go
```diff
@@ -0,0 +1,63 @@
...
+			resp, err := cl.Request(req, nil)
+			if err != nil {
+				return err
+			}
+			entities = append(entities, resp.EntityIDs...)
+			count = len(resp.EntityIDs)
+		}
+	}
+	for _, ent := range entities {
+		fmt.Printf("%v\n", ent)
```
p4u: use `Println()`

cmd/dvotecli/commands/entity.go
```diff
@@ -0,0 +1,63 @@
...
+	var req types.MetaRequest
+	req.Method = "getEntities"
+	resp, err := cl.Request(req, nil)
+	if err != nil {
+		return err
+	}
+	if !resp.Ok {
+		return fmt.Errorf(resp.Message)
+	}
+	entities := resp.EntityIDs
```
p4u: Here you are just copying the pointer, this is not a deep copy. So I'm not sure what would be the behavior when the loop changes the content of resp.EntityIDs. I think the correct way to do this is:
```
entities  := []string{}
entities = append(entities, resp.EntityIDs...)
if len(entities) == .....
```

cmd/dvotecli/commands/file.go
```diff
@@ -0,0 +1,145 @@
...
+		return fmt.Errorf("you must provide a file name and payload")
+	}
+
+	cl, err := client.New(host)
+	if err != nil {
+		return err
+	}
+	defer cl.Conn.Close(websocket.StatusNormalClosure, "")
+
+	var req types.MetaRequest
```
p4u: same

cmd/dvotecli/commands/file.go
```diff
@@ -0,0 +1,145 @@
...
+
+func init() {
+	rootCmd.AddCommand(fileCmd)
+	fileCmd.AddCommand(fileAddCmd)
+	fileCmd.AddCommand(fileFetchCmd)
+	fileCmd.AddCommand(pinListCmd)
+}
+
+func fileAdd(cmd *cobra.Command, args []string) error {
+
```
p4u: Is this empty line allowed by "gofumpt" ? I think not. Are you using golanci-lint with the `.golangci.yml` file included in this repository?

cmd/dvotecli/commands/file.go
```diff
@@ -0,0 +1,145 @@
...
+	req.Method = "fetchFile"
+	req.URI = args[0]
+	resp, err := cl.Request(req, signKey)
+	if err != nil {
+		return err
+	}
+	if !resp.Ok {
+		return fmt.Errorf(resp.Message)
+	}
+	b64content := base64.StdEncoding.EncodeToString(resp.Content)
```
p4u: No need to declare a new variable here

cmd/dvotecli/commands/file.go
```diff
@@ -0,0 +1,145 @@
...
+		return fmt.Errorf("you must provide a file URI")
+	}
+
+	cl, err := client.New(host)
+	if err != nil {
+		return err
+	}
+	defer cl.Conn.Close(websocket.StatusNormalClosure, "")
+
+	var req types.MetaRequest
```
p4u: same

cmd/dvotecli/commands/file.go
```diff
@@ -0,0 +1,145 @@
...
+	cl, err := client.New(host)
+	if err != nil {
+		return err
+	}
+	defer cl.Conn.Close(websocket.StatusNormalClosure, "")
+
+	// Increase the default response size for large lists
+	cl.Conn.SetReadLimit(32768 * 16)
+
+	var req types.MetaRequest
```
p4u: same

cmd/dvotecli/commands/file.go
```diff
@@ -0,0 +1,145 @@
...
+	req.Method = "pinList"
+	resp, err := cl.Request(req, signKey)
+	if err != nil {
+		return err
+	}
+	if !resp.Ok {
+		return fmt.Errorf(resp.Message)
+	}
+	var pins *map[string]string
```
p4u: Do this really works? Usually a map is always declared using `make(map[string]string)`, else there will be a null pointer reference when trying to access it. Maybe `json.Unmarshal()` does the make but we need to be sure.

cmd/dvotecli/commands/info.go
```diff
@@ -0,0 +1,45 @@
...
+	var req types.MetaRequest
+	req.Method = "getInfo"
+	resp, err := cl.Request(req, nil)
+	if err != nil {
+		return err
+	}
+	if !resp.Ok {
+		return fmt.Errorf(resp.Message)
+	}
+	fmt.Printf("Health: %v\n", resp.Health)
```
p4u, line 36: Use a buffer here:
```
var b bytes.Buffer
b.WriteString("string here")
for .... { b.WriteString("string here") }
fmt.Println(b.String)
```

cmd/dvotecli/commands/process.go
```diff
@@ -0,0 +1,157 @@
...
+		return fmt.Errorf("you must provide a entity id")
+	}
+
+	cl, err := client.New(host)
+	if err != nil {
+		return err
+	}
+	defer cl.Conn.Close(websocket.StatusNormalClosure, "")
+
+	var req types.MetaRequest
```
p4u: more

cmd/dvotecli/commands/process.go
```diff
@@ -0,0 +1,157 @@
...
+	req.Method = "getProcessList"
+	req.EntityId, _ = hex.DecodeString(util.TrimHex(args[0]))
+	resp, err := cl.Request(req, nil)
+	if err != nil {
+		return err
+	}
+	if !resp.Ok {
+		return fmt.Errorf(resp.Message)
+	}
+	procs := resp.ProcessList
```
p4u: Same as before

cmd/dvotecli/commands/process.go
```diff
@@ -0,0 +1,157 @@
...
+			if err != nil {
+				return err
+			}
+			procs = append(procs, resp.ProcessList...)
+			count = len(resp.ProcessList)
+		}
+	}
+
+	for _, proc := range procs {
+		fmt.Printf("%v\n", proc)
```
p4u: Buffer it!

cmd/dvotecli/commands/process.go
```diff
@@ -0,0 +1,157 @@
...
+		return fmt.Errorf("you must provide a process id")
+	}
+
+	cl, err := client.New(host)
+	if err != nil {
+		return err
+	}
+	defer cl.Conn.Close(websocket.StatusNormalClosure, "")
+
+	var req types.MetaRequest
```
p4u: same

p4u (changes requested):

p4u:
Enough comments for now. 

In general I'd prefer to avoid the usage of global variables and use instead local variables passed by parameter to the functions that require them.

Also I'd move all helper functions to the `client/` package (or create a new one `client/cmd`) with a specific type in order to keep the state (which would be the signing key, api cli connection and so on). This way the code can be reused by other packages or by other repositories.

p4u:
Actually I think it would make more sense to use and extend the existing `client` type.

https://github.com/vocdoni/vocdoni-node/blob/master/client/connection.go#L19

cmd/dvotecli/commands/common.go
vdo: No, this is the number of iterations on the total, and it's an arbitrary number. For a CLI, I think it should show the total number of elements, without requiring any pagination. 

`router.MaxListSize` is already being used in the pagination loop, so the total would be 16384 lines, already higher than must terminal buffers, but it may be needed to dump all values in a file, for example.

Does it makes sense?

cmd/dvotecli/commands/file.go
vdo: I installed all the required linters configuration you provided, and I didn't get any warning, but I will review.

cmd/dvotecli/commands/info.go
p4u: I did not mean to do buffer on each Printf (wrong comment position), just those which are inside a loop

cmd/dvotecli/commands/common.go
p4u: Yes, thanks!

cmd/dvotecli/commands/common.go
vdo: I declared a `options` type, and refactored it

vdo:
> In general I'd prefer to avoid the usage of global variables and use instead local variables passed by parameter to the functions that require them.

I added an `options`  type to avoid global vars, but I kept `checkSignKey` as a helper for options.
Another helper action was added to client, to make use of deferred close properly, and I avoided any websockets to use the client in a protocol agnostic way.

The linting suggestions have been applied too.

cmd/dvotecli/commands/process.go
vdo: This one is missing

vdo:
Also, to use cobra the command functions have to follow the definition: 
`func(cmd *cobra.Command, args []string)` 
so extra arguments cannot be passed.
That's why I use global variables and structs, and all projects using cobra do it in a similar way, afaik.

p4u:
> Also, to use cobra the command functions have to follow the definition:
> `func(cmd *cobra.Command, args []string)`
> so extra arguments cannot be passed.
> That's why I use global variables and structs, and all projects using cobra do it in a similar way, afaik.

You could do something like this:

```
type Client struct { 
  Conn websockets.Conn
  ....
}

func (c *Client) blockHeight(cmd *cobra.Command, args []string) error {
  c.Conn.Request(....)
}
```

Then initialize a new Client instance
```
client := new(cmd.Client)
client.Init("wss://12.3.4.5...", whateverelse)
```

And use the method instead of a standalone function
```
blockCmd.AddCommand(client.BlockHeight)
```

p4u:
Since you are using Websockets I'd say its quite convenient to use a single shared persistent connection rather than opening and closing it for each command. Don't you think so?

vdo:
> Since you are using Websockets I'd say its quite convenient to use a single shared persistent connection rather than opening and closing it for each command. Don't you think so?

Some commands may not require opening any client (like the genesis one, and maybe others in the future)

p4u:
You can skip the Websockets Connection also. It does not need to be mandatory. I.e you can have a command to open a remote gateway connection `client.Connect(..., "wss://gw1.vocdoni.net")`. The user then will only open a connection if the command he wants to run requires it. Else an error will be returned. Those commands that don't require remote connection will just ignore if there exist an active remote connection.

vdo:
> `blockCmd.AddCommand(client.BlockHeight)`

This is not how it works, you need to pass a *cobra.Command here, not a function or method

Also, Cobra.command does not accept methods of arbitrary types, only functions.

vdo:
> You can skip the Websockets Connection also. It does not need to be mandatory. I.e you can have a command to open a remote gateway connection `client.Connect(..., "wss://gw1.vocdoni.net")`. The user then will only open a connection if the command he wants to run requires it. Else an error will be returned. Those commands that don't require remote connection will just ignore if there exist an active remote connection.

I will review if there's a way to do this selectively, but I would prefer not to create a connection by default. If you don't provide any host param, the cli tires to connect to localhost:9090/dvote (the local daemon) so this could be a problem in some scenarios, like generating a genesis in a host without any daemon or host to connect to....

p4u:
> > You can skip the Websockets Connection also. It does not need to be mandatory. I.e you can have a command to open a remote gateway connection `client.Connect(..., "wss://gw1.vocdoni.net")`. The user then will only open a connection if the command he wants to run requires it. Else an error will be returned. Those commands that don't require remote connection will just ignore if there exist an active remote connection.
> 
> I will review if there's a way to do this selectively, but I would prefer not to create a connection by default. If you don't provide any host param, the cli tires to connect to localhost:9090/dvote (the local daemon) so this could be a problem in some scenarios, like generating a genesis in a host without any daemon or host to connect to....

You can modify the cli or just write a new one. No problem.

p4u:
In any case if you want we can merge this as it is now and then make a refactor following the approach I commented (I can do that but up to you).

vdo:
> In any case if you want we can merge this as it is now and then make a refactor following the approach I commented (I can do that but up to you).

I'm not sure that's possible following your proposal, see `RunE` here:
https://pkg.go.dev/github.com/spf13/cobra?utm_source=godoc#Command

But maybe there's a way to do it I'm not aware of (my Go experience is limited as you know :P)

In any case, I think the current commit is good enough to be merged if you agree, and then we iterate from here to not extend this PR anymore.  I will review what you mention before adding more commands.

p4u:
I'm pretty sure it can be done following my previous example. You can pass to cobra a method of a struct instead of a standalone function. This method will preserve the struct state (so the same vars stored into the struct will be available inside the function called by cobra).

But yes, lets merge this and we'll make the proper improvements in some moment.

p4u:
In order to merge it you need to set the PR as ready (now WIP)

PR #115, opened 2021-03-05
Feature/signed bytes tx

p4u:
See the commit message

I'll squash the three commits in a single one.


benchmark/vochain_benchmark_test.go
```diff
@@ -115,17 +115,16 @@ func BenchmarkVochain(b *testing.B) {
 		Process: processData,
 	}
 
-	txBytes, err := proto.Marshal(process)
+	stx := &models.SignedTx{}
+	stx.Tx, err = proto.Marshal(&models.Tx{Payload: &models.Tx_NewProcess{NewProcess: process}})
```
jordipainan, line 119: Since stx is declared the line before, would not be more idiomatic to use `if stx.Tx, err = ...; err != nil {...}` ?

benchmark/vochain_benchmark_test.go
```diff
@@ -115,17 +115,16 @@ func BenchmarkVochain(b *testing.B) {
...
 	if err != nil {
 		b.Fatal("cannot marshal process")
 	}
-
-	vtx := models.Tx{}
-	vtx.Signature, err = dvoteServer.Signer.Sign(txBytes)
+	stx.Signature, err = dvoteServer.Signer.Sign(stx.Tx)
```
jordipainan, line 123: Same as before

router/voteCallbacks.go
```diff
@@ -31,7 +31,9 @@ func (r *Router) submitRawTx(request routerRequest) {
 	log.Infof("broadcasting tx hash:%s", res.Hash)
 	var response types.MetaResponse
 	response.Payload = fmt.Sprintf("%x", res.Data) // return nullifier or other info
-	request.Send(r.buildReply(request, &response))
+	if err = request.Send(r.buildReply(request, &response)); err != nil {
```
jordipainan, line 34: `err :=`

router/voteCallbacks.go
```diff
@@ -74,7 +69,9 @@ func (r *Router) submitEnvelope(request routerRequest) {
 	log.Infof("broadcasting vochain tx hash:%s code:%d", res.Hash, res.Code)
 	var response types.MetaResponse
 	response.Nullifier = fmt.Sprintf("%x", res.Data)
-	request.Send(r.buildReply(request, &response))
+	if err = request.Send(r.buildReply(request, &response)); err != nil {
```
jordipainan, line 72: `err :=`

benchmark/vochain_benchmark_test.go
mvdan: in this case you can't, because stx.Tx is not a variable name, and `:=` can only be used to define, not assign

jordipainan:
LGTM, just adding few things I noticed: 

- @mvdan  we need to agree on what is the maximum line length, in some places we have all splited, on others we have all in one line and in others seems to be a mix. We already had this conversation but we need to establish an agreement.
- Use `if x, err = ... ; err != nil { }` when possible, is more idiomatic.
- If just the error is expected, is recommended to use `err :=` and not `err =`

mvdan:
I think a hard line length limit, like what `lll` does, doesn't make sense. A soft limit of 100-120 when a line can easily be split into multiple lines could make sense, and it's what I'm trying to implement in gofumpt.

test/vochain_test.go
```diff
@@ -1,77 +0,0 @@
...
-import (
-	"testing"
-
-	"go.vocdoni.io/dvote/test/testcommon"
-	"go.vocdoni.io/dvote/util"
-	"go.vocdoni.io/dvote/vochain"
-	models "go.vocdoni.io/proto/build/go/models"
-)
-
-func TestCreateProcess(t *testing.T) {
```
mvdan, line 12: don't we want to keep this test?

mvdan:
I keep seeing the same 20 lines that do Marshal+Sign+Marshal. I wonder if we could deduplicate all of them, now that the crypto packages mostly use interfaces.

mvdan (approved):
The new code looks much nicer. The old code even had to do two marshals anyway, because of the sign step; if anything, we're saving work now by not marshalling the inner message twice.

benchmark/vochain_benchmark_test.go
jordipainan: cannot do 
```	stx := &models.SignedTx{}
	if stx.Tx, err = proto.Marshal(&models.Tx{Payload: &models.Tx_NewProcess{NewProcess: process}}); err != nil {
		b.Fatal("cannot marshal process")
	}
	if stx.Signature, err = dvoteServer.Signer.Sign(stx.Tx); err != nil {
		b.Fatalf("cannot sign oracle tx: %s", err)
	}
```
??

mvdan: sure, you can, I'd probably do it that way. I was just replying to your earlier use of `:=` :)

jordipainan: ahh oki, yeah thats why I edited the comment (2 seconds after) :)

p4u:
@jordipainan this PR did not need to be merged before finishing the reviews... You could just push this into dev if you wanted to have it deployed. Dev is a dirty branch that can be reset in any moment.

p4u:
> I keep seeing the same 20 lines that do Marshal+Sign+Marshal. I wonder if we could deduplicate all of them, now that the crypto packages mostly use interfaces.

Agree on that, these lines can be simplified by using a function. But the refactor is big and I did not had time to look deeper into it

router/voteCallbacks.go
p4u: These `=` were introduced in purpose, I don't really see the need for declaring a new variable on this specific scope. I do see it makes sense in some others however.

PR #118, opened 2021-03-08
ethevents: subscription on reconect fails

jordipainan:
When the ethevents service is running and the
web3 conection is lost, the ethevents service
tries to reconect to the web3 endpoint.
For some reason (unknown atm) the ethereum
logs subscription fails on reconect.

As a temporary workaround let's fatal on
subscription failure and rely on docker restart
always policy. This will restart the vocdoni-node
from scratch and lost events will be also processed.

This PR just affects the nodes running the
ethevents service (Oracles).

Future work solution: Restart all the subscription
service freeing all the resources and recreating from
scratch without halting the node. This would be done
using an error channel on SubscribeEthereumEventLogs(...)
method.

(cherry picked from commit f0435c86169c2488297ea3d096a299116db6cef2)

### DIFF
--- a/chain/ethevents/events.go
+++ b/chain/ethevents/events.go
@@ -353,7 +353,8 @@ func (ev *EthereumEvents) SubscribeEthereumEventLogs(ctx context.Context, fromBl
 	for {
 		select {
 		case err := <-sub.Err():
-			log.Errorf("ethereum events subscription error on channel: %s", err)
+			// TODO: @jordipainan do not fatal here, handle error on event subscription channel
+			log.Fatalf("ethereum events subscription error on channel: %s", err)
 		case event := <-logs:
 			ev.EventProcessor.Events <- event
 		}


jordipainan:
@p4u I did not want to delete the previous work on reconnect as I think is useful. Let's fatal on the only point where the new subscription fails. I tried to implement a lighting fix using channels but the logic is not that simple ... I do not want to spend more time on this. If you agree, this fix should work (tested on dev env).

p4u:
I agree with this workaound, however the commit should:

- have a big TBD on the code
- make the execution consistent, so if `Fatal()` is executed, remove the code behind it since it is never gonna be executed

Don't worry, the code will be always available on the Git history. You can link this PR from the TBD comment so it can be easily found

jordipainan:
> I agree with this workaound, however the commit should:
> 
>     * have a big TBD on the code
> 
>     * make the execution consistent, so if `Fatal()` is executed, remove the code behind it since it is never gonna be executed
> 
> 
> Don't worry, the code will be always available on the Git history. You can link this PR from the TBD comment so it can be easily found

The TODO on the code is not enough ? What kind of TBD are you suggesting?
Also regarding the fatal, what is the code behind it that makes the execution not consistent ?

p4u:
I think the TODO should be bigger and have a better explanation. But that works also, lets add an issue to not forget about this.

Regarding the code, you're right. I was talking about the reconnect code but its not directly related.

PR #119, opened 2021-03-09
mod: udpate go-libp2p and ipfs/go-bitswap

mvdan:
(see commit message)


mvdan:
@p4u we probably want to update to the latest releases of direct deps like ipfs, too, but I left that out here to just get the race fixes.

PR #120, opened 2021-03-09
Use MaxListSize as paging limit

vdo:


router/voteCallbacks.go
```diff
@@ -11,8 +11,10 @@ import (
 	"google.golang.org/protobuf/proto"
 )
 
-const MaxListSize = 256
-const MaxListIterations = 64
+const (
+	MaxListSize       = 256
+	MaxListIterations = 64
```
p4u: Remove `MaxListIterations` since we don't need it anymore (AFAIK)

vdo: It's useful in the CLI as a paging loop limit, but I can move it there, maybe makes more sense.

p4u: But there is something I don't get... Why MaxListSize is 256? The maximum we were using for pagination was 64, not 256. The point behind this commit is to get rid of MaxListIterations since it is not used anywhere in the code, right? Can you clarify please?

vdo: By what we talked in the chat, I thought you wanted to increase this limit to 256, but we can keep the variable name and keep it as `MaxListSize == 64`.

If we do that,  the limit in `getProcessList()` will be too low though?

p4u: Well, if the client want more, they can just the pagination `from` field on the next call.

I think 64 is a good number since printing more than 64 proceses on the client side looks a bit too much for a friendly UX.

vdo: Ok, I'm doing a new  PR to decrease it then...

PR #135, opened 2021-03-20
support HTTP for API client and make block time configurable

p4u:
1. api/client and cmd/dvotecli now support HTTP (in addition to WS)
transparently for client connections. Use HTTP by default everywhere.

2. introduce a new flag `vochainBlockTime` to set the vochain consensus
block time (in seconds). Defaults to 10s

3. use a blockTime of 6s for compose testsuit (will make tests faster)

4. finally, minor improvements on vochain benchmark


p4u:
@vdo this commit brings HTTP support for dvotecli (transparently in addition to WS)

PR #143, opened 2021-03-31
Merge explorer apis into indexer

NateWilliams2:


p4u:
In general I think you are taking the good approach.

My only comment is on the `searchTerm` type. For consistency, and since we are searching on byte slices, I would use the same type `[]byte` for the methods and `HexBytes`for the API. For searching you can use `bytes.Contains()` the same way you use `string.Contains`.

By the way, this is a garbage `Where("ID").Ne(&[]byte{})` (introduced by myself) that you could remove.

NateWilliams2:
@p4u 
> My only comment is on the `searchTerm` type. For consistency, and since we are searching on byte slices, I would use the same type `[]byte` for the methods and `HexBytes`for the API. For searching you can use `bytes.Contains()` the same way you use `string.Contains`.

This makes sense to me, the only issue (which is something I ran into implementing the search on the original explorer) is that, if a user searches an odd-length partial id string (ie `4011d`), the search term would have to be truncated in the frontend in order to convert into bytes. Obviously this is pretty minor, the weird use behavior would be this search term `4011d` returning the process `4011a....`. 

> By the way, this is a garbage `Where("ID").Ne(&[]byte{})` (introduced by myself) that you could remove.

Will do

p4u:
> @p4u
> 
> > My only comment is on the `searchTerm` type. For consistency, and since we are searching on byte slices, I would use the same type `[]byte` for the methods and `HexBytes`for the API. For searching you can use `bytes.Contains()` the same way you use `string.Contains`.
> 
> This makes sense to me, the only issue (which is something I ran into implementing the search on the original explorer) is that, if a user searches an odd-length partial id string (ie `4011d`), the search term would have to be truncated in the frontend in order to convert into bytes. Obviously this is pretty minor, the weird use behavior would be this search term `4011d` returning the process `4011a....`.
> 
> > By the way, this is a garbage `Where("ID").Ne(&[]byte{})` (introduced by myself) that you could remove.
> 
> Will do

You are right. Using HexBytes on the API field means that the user cannot enter an odd length string for searching. And obviously if we complete the missing 4 bits with some padding the search won't match. So either we use String as your first proposal or either we do the search at `bit`level (not bytes). @mvdan any thought on this?

mvdan:
Right now, the JSON API is a `[]byte`, so that's base64. Presumably, the input must be whole bytes, because otherwise the base64 decoding that json does would fail.

So, the same happens with hexadecimal. The backend would reject invalid hex, just like it would reject invalid base64.

So I don't get the problem here, to be honest. I think the search should always be on whole bytes. I don't think base64 or hex would make a big difference for the encoding, but HexBytes would be more consistent with the other API fields at least.

In the UI, if the user enters invalid hex like `4011d`, we should error, just like if they entered `zz`.

NateWilliams2:
Well the problem here is that we're allowing users to search for partial IDs, and users might not know that these ids are really bytes. So it could be a weird user experience if they are required to enter an even-length search term without knowing why. 

I first used String for the json api SearchTerm type, so we could return to that. If that's bad practice the usability can come second.

mvdan:
OK, I think I see what you mean now. To the end users, the IDs are just strings, and they might not realise that they're interpreted as hex, which must be of even length to be valid. Searching for the substring `abc` doesn't seem wrong to the end user, and there's no good reason to break it other than "we'd like to interpret the search term as hexadecimal", which is an implementation detail.

If that's the case, then the search term must be string, I think. Using HexBytes won't work, because it forcefully interprets the input as hexadecimal. Using `[]byte` wouldn't work either, because json will attempt base64-decoding that.

mvdan:
I should clarify that the substring search should then end up being like: `strings.Contains(hex.EncodeToString(field), searchTerm)`. That does encode every field when searching, but it's not terribly expensive.

If it turns out to be expensive, you could be extra clever and only do the string-based substring search if the length of searchTerm is odd. If it's even, then decode it as hex and do a `bytes.Contains` check without having to `hex.EncodeToString` every field. But I'd personally leave this as a TODO as it feels like premature optimization until we benchmark/profile and we find a performance problem.

NateWilliams2:
Yes, exactly- it's about not exposing the user experience to implementation details they don't need to know about. 

This sounds good, I'll revert the commit and use Strings instead. And when I use the search feature I'll take note if it seems slow.

mvdan:
I doubt it would seem slow right now, but if later on we find relatively high CPU usage when searching through tens of thousands of IDs, then I wouldn't be surprised if the hex encoding stuff was to blame :)

cmd/vochaintest/cache-keys.json
```diff
@@ -0,0 +1 @@
...29fdb0b","proof":"AQEJ/4AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAABFsVt7EcVCs3fi9IJUCGlGhXV3pZw5ffcvC1DZEFQjyM5Zbf++l4HS8tbXtDOh8VK+uWiPJusnGklOdEihfLlgmv1DDBH+ZT1Xwx/TqsAASElCtX5m8X+Nn4fkZo3M7guz8hdmWaYFx3Fd2DBr3fM0wScTutgJLJJTGnRvLf+2nA6hq6FA/1gBf38CQQrqRCfBUiswIe7w1iYfEGew0KmNRIfATDKPsyVuZYwQbo0gfVciEgH7iKtr4IukkEe3BSw9H3Ox3V6u2TdEYqlM77dNRdtUjfNiqgr7m8m9VtcK3rwk435T6n7/r6Nw+8tt2jK2kM2RTM3piDI6CwUvIvU1GCRE4f9Ec7WH1YZqQz67h6JAmfMng4rnTnYddIE3qPgA"},{"privKey":"955dc72a51f24afb90dad4bbf408a875dcfaa486a4a0ba47525b6a2efb06e02f","pubKey":"02465c36547c0861201e0f1f07a4e323452e09cc2291e4ef08d3e67fb78d5f169a","proof":"AQEO/oQAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAABFsVt7EcVCs3fi9IJUCGlGhXV3pZw5ffcvC1DZEFQjyGfJRDN8bbxZvbkEvCfNPEH+OEjVxh1NWJ0rhlabox+z9nnu2DQ5RG6FUncqoZjqBprRI5xNkxxC0ut731Sumhl300+kjNJoUramGuUdCYB7G+LKswFhah449YVC5JOTROhMvp8nyaDmugNZA7vZ6dWuWHXp2UyG774XM49VapjHMynzsP4TUx8su/9f1DMhYkk1szWYNOWq8uOV407jF/fVe8yRhAC9uZwK5hu30HsYoc9q4hj1RCEFlxNaINycwEdohnnjJrAnR0fpYEKBvDL1e0dirA73Mn5bsa/CMUJcN14t8/fRs8Z45YO1bWOwsTKuQUnBpxKuW0+rwa0ljSwA"}],"censusId":"179f160d1f0b3cd8ba396fcc2c0d6a6e00b69d4b9428e0fd0d9588410fc47b17","censusUri":"ipfs://QmXqFXR61DGnz85ET7GuuVpFjSCxC4HsF5B3CbNHxHyoP6"}
```
p4u: This file should not be here

router/voteCallbacks.go
```diff
@@ -73,6 +74,37 @@ func (r *Router) submitEnvelope(request routerRequest) {
 	}
 }
 
+func (r *Router) getStats(request routerRequest) {
+	stats := new(models.VochainStats)
+
+	stats.BlockHeight = uint32(r.vocapp.State.Header(true).Height)
```
p4u: Use the new Height() function instead

router/voteCallbacks.go
```diff
@@ -73,6 +74,37 @@ func (r *Router) submitEnvelope(request routerRequest) {
...
+	stats.BlockTimeStamp = int32(r.vocapp.State.Header(true).Timestamp)
+
```
p4u: no need for so much empty lines

router/voteCallbacks.go
```diff
@@ -73,6 +74,37 @@ func (r *Router) submitEnvelope(request routerRequest) {
...
+	stats.EntityCount = r.Scrutinizer.EntityCount()
```
p4u: We should cache the entity count as we do with `countTotalEnvelopes` (see scrutinizer.go line 65)

router/voteCallbacks.go
```diff
@@ -73,6 +74,37 @@ func (r *Router) submitEnvelope(request routerRequest) {
...
+
+	votes, err := r.Scrutinizer.GetEnvelopeHeight([]byte{})
```
p4u: better do something like: 
```
if stats.EnvelopeCount, err = r.Scrutinizer.GetEnvelopeHeight(nil); err != nil {
...
}
```

router/voteCallbacks.go
```diff
@@ -73,6 +74,37 @@ func (r *Router) submitEnvelope(request routerRequest) {
...
+	if err != nil {
+		log.Warnf("could not count vote envelopes: %s", err)
+	}
+	stats.EnvelopeCount = votes
+
+	stats.ProcessCount = r.Scrutinizer.ProcessCount([]byte{})
```
p4u: We should cache the process count the same way we cache the envelope count, else it might become very costly.

router/voteCallbacks.go
```diff
@@ -73,6 +74,37 @@ func (r *Router) submitEnvelope(request routerRequest) {
...
+	vals, _ := r.vocapp.State.Validators(true)
+	stats.ValidatorCount = int32(len(vals))
```
p4u: We should have  a scrutinizer function for ValidatorCount()

router/voteCallbacks.go
p4u: I think you can get this from `r.vocinfo`, that would be an easy solution. Up to you.

router/voteCallbacks.go
```diff
@@ -73,6 +74,37 @@ func (r *Router) submitEnvelope(request routerRequest) {
...
+	stats.BlockTime = r.vocinfo.BlockTimes()[:]
+
+	stats.ChainID = r.vocapp.Node.GenesisDoc().ChainID
```
p4u: Lets store this in the vocapp directly on app.go NewBaseApplication constructor and make a function to export it `ChainID()`

router/voteCallbacks.go
```diff
@@ -73,6 +74,37 @@ func (r *Router) submitEnvelope(request routerRequest) {
...
+	stats.GenesisTimeStamp = r.vocapp.Node.GenesisDoc().GenesisTime.Unix()
```
p4u: Same

router/voteCallbacks.go
```diff
@@ -181,18 +219,28 @@ func (r *Router) getProcessList(request routerRequest) {
 func (r *Router) getProcessInfo(request routerRequest) {
 	var response types.MetaResponse
 	var err error
-	response.ProcessInfo, err = r.Scrutinizer.ProcessInfo(request.ProcessID)
+	processInfo, err := r.Scrutinizer.ProcessInfo(request.ProcessID)
+	if err != nil {
+		r.sendError(request, fmt.Sprintf("cannot get process info: %v", err))
+		return
+	}
+	process := processInfo.Mirror()
```
p4u: no need to declare a new var `process`

router/voteCallbacks.go
```diff
@@ -181,18 +219,28 @@ func (r *Router) getProcessList(request routerRequest) {
...
+	processBytes, err := proto.Marshal(process)
```
p4u: no need to declare processBytes either

router/voteCallbacks.go
```diff
@@ -257,13 +305,136 @@ func (r *Router) getEnvelopeList(request routerRequest) {
 	if request.ListSize == 0 {
 		request.ListSize = MaxListSize
 	}
-	nullifiers := r.vocapp.State.EnvelopeList(request.ProcessID, request.From, request.ListSize, true)
+	envelopes, err := r.Scrutinizer.GetEnvelopes(request.ProcessID)
```
p4u: This is extremly costly and does not scale, no way we can return such list. We should instead return the nullifiers (as it was before) and get the envellope content only when requested individually.

router/voteCallbacks.go
```diff
@@ -257,13 +305,136 @@ func (r *Router) getEnvelopeList(request routerRequest) {
...
+	if err != nil {
+		r.sendError(request, fmt.Sprintf("cannot get envelope list: (%s)", err))
+		return
+	}
+	listBytes, err := proto.Marshal(envelopes)
```
p4u: No way, its too costly  :-1:

router/voteCallbacks.go
```diff
@@ -257,13 +305,136 @@ func (r *Router) getEnvelopeList(request routerRequest) {
...
-	strnull := []string{}
-	for _, n := range nullifiers {
-		strnull = append(strnull, fmt.Sprintf("%x", n))
+	response.Content = listBytes
+	request.Send(r.buildReply(request, &response))
+}
+
+func (r *Router) getValidatorList(request routerRequest) {
+	var response types.MetaResponse
+	validators, err := r.vocapp.State.Validators(true)
```
p4u: Lets store the validators list in memory when the App is created and add the modificators on the scrutinizer event callbacks

router/voteCallbacks.go
```diff
@@ -257,13 +305,136 @@ func (r *Router) getEnvelopeList(request routerRequest) {
...
 	}
-	response.Nullifiers = &strnull
+	validatorList := new(models.ValidatorList)
+	validatorList.Validators = validators
+	validatorBytes, err := proto.Marshal(validatorList)
+	if err != nil {
+		r.sendError(request, fmt.Sprintf("cannot get validator list: %v", err))
+		return
+	}
+	response.ValidatorList = validatorBytes
```
p4u: Can you explain what's this list please?

router/voteCallbacks.go
```diff
@@ -257,13 +305,136 @@ func (r *Router) getEnvelopeList(request routerRequest) {
...
+	request.Send(r.buildReply(request, &response))
+}
+
+func (r *Router) getBlock(request routerRequest) {
+	var response types.MetaResponse
+	block := r.Scrutinizer.App.GetBlockByHeight(int64(request.BlockHeight))
```
p4u: Check first if the request.BlockHeight is smaller than the current height.

The height is available and cached in the App directly

router/voteCallbacks.go
```diff
@@ -257,13 +305,136 @@ func (r *Router) getEnvelopeList(request routerRequest) {
...
+	if block == nil {
+		r.sendError(request, fmt.Sprintf("cannot get block: no block with height %d", request.BlockHeight))
+		return
+	}
+	blockHeader := vochain.MirrorTendermintBlock(block)
```
p4u: No need to declare blockHeader and blockBytes

router/voteCallbacks.go
```diff
@@ -257,13 +305,136 @@ func (r *Router) getEnvelopeList(request routerRequest) {
...
+	blockBytes, err := proto.Marshal(blockHeader)
+	if err != nil {
+		r.sendError(request, fmt.Sprintf("cannot get block: %v", err))
+		return
+	}
+	response.Content = blockBytes
```
p4u: I'm not sure using Protobuf for this case (and also previous) is the best approach. What's the point for not using JSON?

router/voteCallbacks.go
```diff
@@ -257,13 +305,136 @@ func (r *Router) getEnvelopeList(request routerRequest) {
...
+	request.Send(r.buildReply(request, &response))
+}
+func (r *Router) getBlockByHash(request routerRequest) {
```
p4u: I reviewed until here.

p4u (changes requested):

router/voteCallbacks.go
NateWilliams2: Currently the scrutinizer registers entities in `newEmptyProcess`, where it uses the db `Upsert` method: 

`"Upsert inserts the record into the badgerhold if it doesn't exist. If it does already exist, then it updates the existing record"`

In order to cache the entity height, afaik, here we would need to first check if the entity exists already, which means an extra database query on every new process. Is this trade-off worth it @p4u ?

router/voteCallbacks.go
NateWilliams2: ValidatorList is a protobuf model for a list of validators. I implemented some other protobuf wrappers/list models for this api as well, such as EnvelopePackage, EnvelopePackageList, VochainStats, BlockHeader, BlockHeaderList, TxPackage, TxPackageList. These models are just to serialize content for the api. I realize it doesn't make sense to have `response.ValidatorList` if we're doing this, I'll change it to use `response.Content`

router/voteCallbacks.go
NateWilliams2: Possibly json would make more sense. My reasoning was 
a) use the same encoding for everything
b) smaller payloads
Would be easy for me to refactor as json, though.

vochain/scrutinizer/process.go
```diff
@@ -28,6 +28,9 @@ func (s *Scrutinizer) ProcessInfo(pid []byte) (*Process, error) {
 // Status is one of READY, CANCELED, ENDED, PAUSED, RESULTS
 func (s *Scrutinizer) ProcessList(entityID []byte, searchTerm string, namespace uint32,
 	status string, withResults bool, from, max int) ([][]byte, error) {
+	if from < 0 {
```
p4u, line 37: if from is smaller than zero, an error should be returned

router/voteCallbacks.go
p4u: The Upsert() of newEmptyProcess seems not correct to me, since the CreationTime is being updated on each new process. So we should change it and make two operations (see if exists and create it if not). Then we can keep an Entity count cache as we do with countTotalEnvelopes. If think is worth it, we need to minimize the DB transactions in order to scale.

router/voteCallbacks.go
p4u: I'm not sure the cost of encoding/decoding protobuf is woth it the win on size. What do you think @mvdan ?

router/voteCallbacks.go
NateWilliams2: This makes sense to me, I'll update the code

vochain/scrutinizer/process.go
```diff
@@ -255,15 +255,15 @@ func (s *Scrutinizer) newEmptyProcess(pid []byte) error {
 	// Add the entity to the indexer database
 	eid := p.GetEntityId()
 	entity := &Entity{}
-	// If entity is not registered in db, add to entity count cache
+	// If entity is not registered in db, add to entity count cache and insert to db
 	if s.db.FindOne(entity, badgerhold.Where("ID").Eq(eid)); entity == nil {
```
p4u: `s.db.FindOne` returns a "badgerhold.ErrNotFound" if the record does not exist, that should be used instead of `engity == nil`. Also if the error is another, return it.

In general no error can be unattended.

router/voteCallbacks.go
p4u: I think we should use JSON, for consistency. This API is exposed by any gateway and could be used either by the Explorer or by a third party application. Having a mixed type API (JSON/Protobuf) is weird.

The work you did with the protobuf types can still be useful in the moment that we decide to migrate the full API from JSON to Protobuf.

router/voteCallbacks.go
p4u: Let's use JSON for this list too. It's just 2 []byte fields, 1 int and 1 string. Its not worth it using protobuf here.

p4u:
@NateWilliams2 I think before continuing the review we would need to iterate the new API methods.

Lets first do it in a HackMD or Notion so afterwars we can copy&paste it directly to docs.vocdoni.io

p4u:
For future pull requests, I'd prefer to do it by smaller batches of code commits instead of 1 single pull requests.

types/api.go
```diff
@@ -18,6 +18,7 @@ type RequestMessage struct {
 // MetaRequest contains all of the possible request fields.
 // Fields must be in alphabetical order
 type MetaRequest struct {
+	BlockHeight  uint32   `json:"blockHeight,omitempty"`
```
p4u: Lets call it "Height" so we can reuse the field for other requests.

vochain/app.go
```diff
@@ -157,16 +157,16 @@ func (app *BaseApplication) GetBlockByHash(hash []byte) *tmtypes.Block {
...
-		return nil, fmt.Errorf("unable to get block by height: %d", height)
+		return nil, nil, fmt.Errorf("unable to get block by height: %d", height)
 	}
 	if int32(len(block.Txs)) <= txIndex {
-		return nil, fmt.Errorf("txIndex overflow on GetTx (height: %d, txIndex:%d)", height, txIndex)
+		return nil, nil, fmt.Errorf("txIndex overflow on GetTx (height: %d, txIndex:%d)", height, txIndex)
 	}
 	tx := &models.SignedTx{}
-	return tx, proto.Unmarshal(block.Txs[txIndex], tx)
+	return tx, tmtypes.Tx(block.Txs[txIndex]).Hash(), proto.Unmarshal(block.Txs[txIndex], tx)
```
p4u, line 198: I don't like this, Hash() computation is a costly operation. GetTx() is used by scrutinizer for computing results and that would slow it down.

Create a `GetTxHash()` function instead

vochain/scrutinizer/process.go
```diff
@@ -114,20 +134,34 @@ func (s *Scrutinizer) ProcessList(entityID []byte, namespace uint32,
...
 // ProcessCount return the number of processes indexed
-func (s *Scrutinizer) ProcessCount() int64 {
-	c, err := s.db.Count(&Process{}, nil)
+func (s *Scrutinizer) ProcessCount(entityID []byte) int64 {
+	var c int
+	var err error
+	if len(entityID) == 0 {
+		c, err = s.db.Count(&Process{}, nil)
+	} else {
+		c, err = s.db.Count(&Process{}, badgerhold.Where("EntityID").Eq(entityID))
```
p4u: Index by EntityID

vochain/scrutinizer/process.go
```diff
@@ -114,20 +134,34 @@ func (s *Scrutinizer) ProcessList(entityID []byte, namespace uint32,
 }
 
 // ProcessCount return the number of processes indexed
```
p4u, line 133: Specify in the description that entityID nil is allowed

vochain/scrutinizer/process.go
p4u: Add a TO-DO if you want for Caching also the Process count by EntityID (we already doing it with envelopes)

vochain/scrutinizer/process.go
```diff
@@ -114,20 +134,34 @@ func (s *Scrutinizer) ProcessList(entityID []byte, namespace uint32,
...
+	}
 	if err != nil {
 		log.Warnf("cannot count processes: %v", err)
 	}
 	return int64(c)
 }
 
 // EntityList returns the list of entities indexed by the scrutinizer
-func (s *Scrutinizer) EntityList(max, from int) []string {
+func (s *Scrutinizer) EntityList(searchTerm string, max, from int) []string {
```
p4u: Make searchTerm the last value since it is "optional". Also please, specify it can be empty on the method description

vochain/scrutinizer/vote.go
```diff
@@ -59,7 +59,14 @@ func (s *Scrutinizer) GetEnvelope(nullifier []byte) (*models.VoteEnvelope, []byt
 	if envelope == nil {
 		return nil, nil, fmt.Errorf("transaction is not an Envelope")
 	}
-	return envelope, stx.Signature, nil
```
p4u, line 62: I am not sure regarding returning a new struct here...

p4u (changes requested):

NateWilliams2:
> @NateWilliams2 I think before continuing the review we would need to iterate the new API methods.
> 
> Lets first do it in a HackMD or Notion so afterwars we can copy&paste it directly to docs.vocdoni.io

I'm working on this. I'll make the change from protobuf to JSON (this also helps by enabling a more human-readable api) and specify all the api changes here: https://hackmd.vocdoni.net/xtQcPL-2RgWK8NVa1p9Vzg#


> For future pull requests, I'd prefer to do it by smaller batches of code commits instead of 1 single pull requests.

Noted.

Also will keep addressing your smaller comments :)

p4u:
Thanks @NateWilliams2 great work in general :+1: that would be a nice improvement for the explorer

router/voteCallbacks.go
NateWilliams2: The Node isn't assigned in the App constructor, it has to be done at a second step. So we don't have access to the ChainID during app construction. We could create a SetNode method, though, which assigns the App.Node and also the App.ChainID

p4u: SGTM

vochain/scrutinizer/vote.go
```diff
@@ -59,7 +59,18 @@ func (s *Scrutinizer) GetEnvelope(nullifier []byte) (*models.VoteEnvelope, []byt
...
+	envelope.Nullifier = voteRef.Nullifier
+	return &types.EnvelopePackage{
+		Nonce:                envelope.Nonce,
+		ProcessId:            envelope.ProcessId,
+		VotePackage:          envelope.VotePackage,
+		Nullifier:            nullifier,
+		EncryptionKeyIndexes: envelope.EncryptionKeyIndexes,
+		Weight:               voteRef.Weight.Bytes(),
```
p4u: Make weight it a big.Int and do `new(big.Int).SetBytes(voteRef.Weight.Bytes()))`

vochain/scrutinizer/vote.go
```diff
@@ -59,7 +59,18 @@ func (s *Scrutinizer) GetEnvelope(nullifier []byte) (*models.VoteEnvelope, []byt
 	if envelope == nil {
 		return nil, nil, fmt.Errorf("transaction is not an Envelope")
 	}
-	return envelope, stx.Signature, nil
+	envelope.Nullifier = voteRef.Nullifier
+	return &types.EnvelopePackage{
+		Nonce:                envelope.Nonce,
+		ProcessId:            envelope.ProcessId,
+		VotePackage:          envelope.VotePackage,
+		Nullifier:            nullifier,
```
p4u: Do we need nullifier? It's the parameter used to fetch the envelope so the caller already have it.

vochain/scrutinizer/vote.go
```diff
@@ -59,7 +59,18 @@ func (s *Scrutinizer) GetEnvelope(nullifier []byte) (*models.VoteEnvelope, []byt
...
+		EncryptionKeyIndexes: envelope.EncryptionKeyIndexes,
+		Weight:               voteRef.Weight.Bytes(),
+		TxIndex:              voteRef.TxIndex,
```
p4u: Do we need index and height? Isn't TxHash enough?

vochain/scrutinizer/vote.go
```diff
@@ -35,19 +35,19 @@ const kvErrorStringForRetry = "Transaction Conflict"
 
 // GetVoteReference gets the reference for an AddVote transaction.
 // This reference can then be used to fetch the vote transaction directly from the BlockStore.
-func (s *Scrutinizer) GetEnvelopeReference(nullifier []byte) (*VoteReference, error) {
-	txRef := &VoteReference{}
+func (s *Scrutinizer) GetEnvelopeReference(nullifier []byte) (*types.VoteReference, error) {
+	txRef := &types.VoteReference{}
```
p4u: No need to declare a var here

vochain/scrutinizer/vote.go
```diff
@@ -59,7 +59,18 @@ func (s *Scrutinizer) GetEnvelope(nullifier []byte) (*models.VoteEnvelope, []byt
...
+		Nonce:                envelope.Nonce,
+		ProcessId:            envelope.ProcessId,
+		VotePackage:          envelope.VotePackage,
+		Nullifier:            nullifier,
+		EncryptionKeyIndexes: envelope.EncryptionKeyIndexes,
+		Weight:               voteRef.Weight.Bytes(),
+		TxIndex:              voteRef.TxIndex,
+		Height:               voteRef.Height,
+		TxHash:               txHash,
+	}, stx.Signature, nil
```
p4u: Let's make the signature a field of EnvelopePackage

vochain/scrutinizer/vote.go
```diff
@@ -121,11 +133,17 @@ func (s *Scrutinizer) GetEnvelopes(processId []byte) ([]*models.VoteEnvelope, []
 			if envelope == nil {
 				return fmt.Errorf("transaction is not an Envelope")
 			}
-			envelopes = append(envelopes, envelope)
-			weights = append(weights, txRef.Weight)
+			envelope.Nullifier = txRef.Nullifier
+			envelopes = append(envelopes, &types.EnvelopeMetadata{
```
p4u: Cannot we reuse EnvelopePackage type here?

vochain/scrutinizer/vote.go
```diff
@@ -290,7 +308,7 @@ func (s *Scrutinizer) addVoteIndex(nullifier, pid []byte, blockHeight uint32,
 			CreationTime: time.Now(),
 		})
 	}
-	return s.db.Insert(nullifier, &VoteReference{
+	return s.db.Insert(nullifier, &types.VoteReference{
```
p4u: I think it makes more sense to declare the scrutinizer types inside the scrutinizer package. In the future we should try to get rid of the types.go root file.

p4u (changes requested):

p4u:
@mvdan do you have buffer for review it by tomorrow? Else we'll go ahead once I finish my reviews (there is some urgency).

router/voteCallbacks.go
```diff
@@ -257,13 +283,114 @@ func (r *Router) getEnvelopeList(request routerRequest) {
...
+		r.sendError(request, fmt.Sprintf("cannot get envelope list: (%s)", err))
+		return
+	}
+	request.Send(r.buildReply(request, &response))
+}
+
+func (r *Router) getValidatorList(request routerRequest) {
+	var response types.MetaResponse
+	var err error
+	response.ValidatorList, err = r.vocapp.State.Validators(true)
```
p4u: `if response.ValidatorList, err = r.vocapp....`

router/voteCallbacks.go
```diff
@@ -257,13 +283,114 @@ func (r *Router) getEnvelopeList(request routerRequest) {
...
+	request.Send(r.buildReply(request, &response))
+}
+
+func (r *Router) getBlock(request routerRequest) {
+	var response types.MetaResponse
+	if request.Height > r.vocapp.Height() {
+		r.sendError(request, fmt.Sprintf("block height %d not valid for vochain with height %d", request.Height, r.vocapp.Height()))
+		return
+	}
+	response.Block = types.BlockMetadataFromBlockModel(r.Scrutinizer.App.GetBlockByHeight(int64(request.Height)))
```
p4u: `if response.Block = types...; response.Block == nil { ... }`

router/voteCallbacks.go
```diff
@@ -257,13 +283,114 @@ func (r *Router) getEnvelopeList(request routerRequest) {
...
+	request.Send(r.buildReply(request, &response))
+}
+
+func (r *Router) getBlockList(request routerRequest) {
+	var response types.MetaResponse
+	for i := 0; i < request.ListSize; i++ {
+		if int64(request.From)+int64(i) > r.vocinfo.Height() {
+			break
+		}
+		response.BlockList = append(response.BlockList, types.BlockMetadataFromBlockModel(r.Scrutinizer.App.GetBlockByHeight(int64(request.From)+int64(i))))
```
p4u: Split the lines :)

Are you using golang-ci with the configuration available on the repository? In theory 111 should be the maximum allowed.

router/voteCallbacks.go
```diff
@@ -257,13 +283,114 @@ func (r *Router) getEnvelopeList(request routerRequest) {
...
+	var response types.MetaResponse
+	response.Block = types.BlockMetadataFromBlockModel(r.Scrutinizer.App.GetBlockByHash(request.Payload))
+	if response.Block == nil {
+		r.sendError(request, fmt.Sprintf("cannot get block: no block with hash %x", request.Payload))
+		return
+	}
+	request.Send(r.buildReply(request, &response))
+}
+
+func (r *Router) getBlockList(request routerRequest) {
```
p4u: This is a very crazy function... Please let a TO-DO

p4u (changes requested):

vochain/scrutinizer/vote.go
mvdan: Or just `new(big.Int).Set(voteRef.Weight)`, assuming you just want to make a copy.

router/voteCallbacks.go
```diff
@@ -257,13 +283,114 @@ func (r *Router) getEnvelopeList(request routerRequest) {
...
+	}
+	if request.ListSize > MaxListSize || request.ListSize <= 0 {
+		request.ListSize = MaxListSize
+	}
+	maxIndex := request.From + request.ListSize
+	for i := request.From; i < maxIndex && i < len(block.Txs); i++ {
+		signedTx := new(models.SignedTx)
+		tx := new(models.Tx)
+		proto.Unmarshal(block.Txs[i], signedTx)
+		proto.Unmarshal(signedTx.Tx, tx)
```
mvdan: these can error, you should check for it

types/api.go
```diff
@@ -195,3 +211,124 @@ type Key struct {
...
+// TxMetadata contains tx information for the TransactionList api
+type TxMetadata struct {
+	Type        string
+	BlockHeight uint32
+	Index       int32
+	Hash        HexBytes
+}
+
+// BlockMetadata contains the metadata for a single tendermint block
+type BlockMetadata struct {
```
mvdan: don't we want these types to also have json field tags? otherwise the json keys will be like `FooBar` by default, instead of something more consistent with the existing types like `foo_bar`

types/api.go
```diff
@@ -195,3 +211,124 @@ type Key struct {
...
 }
+
+// VochainStats contains information about the current Vochain statistics and state
+type VochainStats struct {
+	BlockHeight      uint32    `json:"block_height"`
+	EntityCount      int64     `json:"entity_count"`
+	EnvelopeCount    uint64    `json:"envelope_count"`
+	ProcessCount     int64     `json:"process_count"`
+	ValidatorCount   int       `json:"validator_count"`
+	BlockTime        *[5]int32 `json:"block_time"`
```
mvdan: remove the pointer, since you don't need this to be omitempty?

types/api.go
```diff
@@ -195,3 +211,124 @@ type Key struct {
...
+	// AvgTxsPerBlock    float64   `json:"avg_txs_per_block"`
+	// AvgTxsPerMinute   float64   `json:"avg_txs_per_minute"`
+	// The hash of the block with the most txs
+	// MaxBytes          int64  `json:"max_bytes"`
+	// MaxTxsBlockHash   string `json:"max_txs_block_hash"`
+	// MaxTxsBlockHeight int64  `json:"max_txs_block_height"`
+	// The start of the minute with the most txs
+	// MaxTxsMinute    time.Time `json:"max_txs_minute"`
+	// MaxTxsPerBlock  int64     `json:"max_txs_per_block"`
+	// MaxTxsPerMinute int64     `json:"max_txs_per_minute"`
```
mvdan: what's the point of this chunk of commented out code? if you really want to keep it around, maybe add a TODO explaining why

vochain/scrutinizer/process.go
```diff
@@ -33,6 +40,15 @@ func (s *Scrutinizer) ProcessList(entityID []byte, namespace uint32,
...
 	}
+	searchMatchFunc := func(r *badgerhold.RecordAccess) (bool, error) {
+		if searchTerm == "" {
+			return true, nil
+		}
+		if strings.Contains(hex.EncodeToString(r.Field().(types.HexBytes)), searchTerm) {
+			return true, nil
+		}
+		return false, nil
+	}
```
mvdan: I know this func is small, but it's tricky and we have it in two places already - I'd make it a global func and reuse it.

I'd also add a TODO above the hex encoding stuff, because we can likely do better in the future, as we discussed before. For example, doing "raw bytes" search if the search term is complete hex.

vochain/scrutinizer/process.go
```diff
@@ -114,24 +134,41 @@ func (s *Scrutinizer) ProcessList(entityID []byte, namespace uint32,
...
 // ProcessCount return the number of processes indexed
-func (s *Scrutinizer) ProcessCount() int64 {
-	c, err := s.db.Count(&Process{}, nil)
+func (s *Scrutinizer) ProcessCount(entityID []byte) int64 {
+	var c int
+	var err error
+	if len(entityID) == 0 {
+		// If no entity ID, return the cached count of all processes
+		return atomic.LoadInt64(s.countTotalProcesses)
+	} else {
```
mvdan: you can drop this else and unindent the Count, as the "if" branch always returns. then you don't need the `var` lines above either.

mvdan (commented):
Some minor comments :)

router/voteCallbacks.go
mvdan: or, if you're sure this error can never happen:

```
if err := proto.Unmarshal(...); err != nil {
    panic(err) // should never happen, because <reason>
}
```

vochain/scrutinizer/vote.go
NateWilliams2: How so? `s.db.FindOne` needs a pointer to a VoteReference to store the query result. 

Are you saying it should use a named return variable?

vochain/scrutinizer/vote.go
NateWilliams2: We have `getBlockByHash` and `getEnvelope` by nullifier, but not `getTxByHash`. This is partially because we don't have an index of transactions, so we need to fetch a transaction directly from the block & index.

If we don't store `txIndex` and `height`, there's no way to tell the user which transaction contained their vote, which imo is necessary.

vochain/scrutinizer/vote.go
NateWilliams2: for the `getEnvelope` call we want `EnvelopePackage` because it contains the entire vote package itself (and other detailed information). for `getEnvelopeList`, we want to serve as little information as possible for each envelope, which means not sending the EncryptionKeyIndexes, Nonce, Signature, VotePackage, and Weight. 
I guess we could use a `EnvelopePackage` and simply leave some of the fields blank, I dislike this because I find it easy to mistakenly use a field that's never going to be set in this context. But I could easily make this change if it's better :)

vochain/scrutinizer/vote.go
NateWilliams2: The problem here is that `types.MetaResponse` has a `Process` field, so with Process declared in Scrutinizer there is a circular import. I agree this fix isn't great, though. 

For now I'll move the other types back to scrutinizer and leave Process in types with a todo.

p4u: We should check why the scrutinizer package needs types.go

vochain/scrutinizer/vote.go
p4u: There is another option which is keep EnvelopeMetadata{} but use it also as a field of EnvelopePackage{} so we are unifying and not repeating fields on two different structs.

vochain/scrutinizer/vote.go
NateWilliams2: It needs types.go directly for HexBytes, VotePackage (for creating test votes), ScrutinizerOnProcessData, KeyKeeperMaxKeyIndex, and the api types EnvelopePackage, EnvelopeMetadata. 

But also packages like vochain, config, testcommon, all also import types.

p4u: `EnvelopePackage, EnvelopeMetadata,  ScrutinizerOnProcessData` are specific for Scrutinizer, so we might move them.

`KeyKeeperMaxKeyIndex` and `VotePackage`could be moved to the vochain package.

I think with this we'll avoid circular conflicts.

NateWilliams2: Commit https://github.com/vocdoni/vocdoni-node/commit/cbedf52fdb9f6cf0121a8fa2e3237080ee8dc40d attempts to deal with this import cycle. 

Could you check it out @p4u ? Tried to make it as clean as possible..

NateWilliams2: and now here: https://github.com/vocdoni/vocdoni-node/commit/1affa20c8b0ea9ffcf126668587eddc814c1b3d7

NateWilliams2:
@p4u could you review my most recent changes in https://github.com/vocdoni/vocdoni-node/commit/4df35e12343b317db1294ccfab80d468a5c5dc9e ? this commit sorts envelope references by blockHeight and TxIndex in scrutinizer.EnvelopeList, rather than alphanumerically by nullifier. This makes much more sense from an api user perspective, but I imagine that this might affect performance too much to be worth it.

router/voteCallbacks.go
```diff
@@ -254,21 +287,126 @@ func (r *Router) getEnvelopeList(request routerRequest) {
...
+		var txType string
+		switch tx.Payload.(type) {
+		case *models.Tx_Vote:
+			txType = types.TxVote
+		case *models.Tx_NewProcess:
+			txType = types.TxNewProcess
+		case *models.Tx_Admin:
+			txType = tx.Payload.(*models.Tx_Admin).Admin.GetTxtype().String()
+		case *models.Tx_SetProcess:
+			txType = "setProcess"
```
p4u: Why this is hardcoded here an the others are defined in types?

types/types.go
```diff
@@ -3,3 +3,8 @@ package types
 type DataStore struct {
 	Datadir string
 }
+
+type Key struct {
```
p4u: What's this type for (add a description please). Shouldn't this be part of scrutinizer types?

types/vochain.go
```diff
@@ -90,8 +90,9 @@ type BlockParams struct {
 }
 
 type EvidenceParams struct {
```
p4u, line 99: These are Vochain specific types, right? Why not moving them to vochain/types.go ?

vochain/scrutinizer/db.go
```diff
@@ -2,61 +2,15 @@ package scrutinizer
...
-type VoteReference struct {
-	Nullifier    types.HexBytes `badgerholdKey:"Nullifier"`
-	ProcessID    types.HexBytes `badgerholdIndex:"ProcessID"`
-	Height       uint32
-	Weight       *big.Int
-	TxIndex      int32
-	CreationTime time.Time
-}
-
 type Results struct {
```
p4u: For consistency better move all scrutinizer types to the same package

vochain/scrutinizer/types/types.go
```diff
@@ -0,0 +1,171 @@
+package types
```
p4u: Don't call this package types please, call it "indexertypes" if you want (since we'll move from scrutinizer to indexer)

vochain/scrutinizer/types/types.go
```diff
@@ -0,0 +1,171 @@
...
+}
+
+// Entity holds the db reference for an entity
+type Entity struct {
+	ID           types.HexBytes `badgerholdKey:"ID"`
+	CreationTime time.Time
+}
+
+// VotePackage represents the payload of a vote (usually base64 encoded)
+type VotePackage struct {
```
p4u: This is not an scrutinizer specific type, that's actually a vochain type so it must not be here.

vochain/scrutinizer/types/types.go
```diff
@@ -0,0 +1,171 @@
...
+	EncryptionKeyIndexes []uint32         `json:"encryption_key_indexes"`
+	Meta                 EnvelopeMetadata `json:"meta"`
+	Nonce                types.HexBytes   `json:"nonce"`
+	Signature            types.HexBytes   `json:"signature"`
+	VotePackage          []byte           `json:"vote_package"`
+	Weight               string           `json:"weight"`
+}
+
+// VochainStats contains information about the current Vochain statistics and state
+type VochainStats struct {
```
p4u: I think that's not a scrutinizer type either

p4u (changes requested):

types/types.go
NateWilliams2: I'll add a description. I think it should be moved to api, not scrutinizer, because it's used just for sending keys over json and including an index with each key.

vochain/scrutinizer/process.go
```diff
@@ -138,7 +138,8 @@ func (s *Scrutinizer) ProcessCount(entityID []byte) int64 {
 		// If no entity ID, return the cached count of all processes
 		return atomic.LoadInt64(s.countTotalProcesses)
 	}
-	if c, err = s.db.Count(&indexertypes.Process{}, badgerhold.Where("EntityID").Eq(entityID)); err != nil {
+	if c, err = s.db.Count(&indexertypes.Process{},
+		badgerhold.Where("EntityID").Eq(entityID)); err != nil {
```
p4u: I think there is an index missing here

p4u: Please, review all indexes are used properly, that makes a big difference on the query cost

p4u (changes requested):

p4u:
> @p4u could you review my most recent changes in [4df35e1](https://github.com/vocdoni/vocdoni-node/commit/4df35e12343b317db1294ccfab80d468a5c5dc9e) ? this commit sorts envelope references by blockHeight and TxIndex in scrutinizer.EnvelopeList, rather than alphanumerically by nullifier. This makes much more sense from an api user perspective, but I imagine that this might affect performance too much to be worth it.

Yes, I think its probably too costly but we should check how is it implemented on badgerhold, maybe @mvdan can help us here.

Returning a list of envelope payloads is too expensive for the API. On the next iteration we need to get rid of these kind of API methods and allow only a specific envelope per query. Then GetEnvelopeList and GetEnvelope need to be used.

So, lets do like this, even if its expensive but:
1) Refactor it on the near future
2) Declare those methods as private, so by default they wont be accessible

mvdan:
> Yes, I think its probably too costly but we should check how is it implemented on badgerhold, maybe @mvdan can help us here.

badgerhold is pretty dumb - if we get many records by processID with some index, it will fetch all those records. Then, once in memory, the sorting will happen.

one thing to keep in mind is that queries are chained, so `SortBy(...).And(...).MatchFunc(...)` will do the sorting first, and matching after. assuming that the matching doesn't care about order, it's probably a bit better to filter first, and sort after. this might not matter much since we're not benchmarking, but just thinking outloud.

as far as I've seen in the badgerhold codebase, the only part that's clever and well-optimized is when indexes are used. Everything else is as you would expect - loops to filter records, sorting entire slices of records, etc.

p4u:
> > Yes, I think its probably too costly but we should check how is it implemented on badgerhold, maybe @mvdan can help us here.
> 
> badgerhold is pretty dumb - if we get many records by processID with some index, it will fetch all those records. Then, once in memory, the sorting will happen.
> 
> one thing to keep in mind is that queries are chained, so `SortBy(...).And(...).MatchFunc(...)` will do the sorting first, and matching after. assuming that the matching doesn't care about order, it's probably a bit better to filter first, and sort after. this might not matter much since we're not benchmarking, but just thinking outloud.
> 
> as far as I've seen in the badgerhold codebase, the only part that's clever and well-optimized is when indexes are used. Everything else is as you would expect - loops to filter records, sorting entire slices of records, etc.

So we should review that all queries are chaining correctly, first filter everything possible and later sort.

@mvdan have you identified any potential optimization on the badgerhold side that we might apply soon? In that case we might open an issue for that.

p4u (approved):

mvdan:
> @mvdan have you identified any potential optimization on the badgerhold side that we might apply soon? In that case we might open an issue for that.

Not really. The querying system is generic. You'd need to do pretty crazy "query optimization" stuff to make significant improvements, the kind of thing many SQL implementations do. I personally think we should deploy what we have, and take metrics/profiles to see what API calls or pieces of code are the slowest. If any of that is to blame on badgerhold, then we can dig deeper.

PR #144, opened 2021-03-31
Feature/scrutinizer optimizations

p4u:
see commit message

### DIFF
--- a/chain/ethevents/events.go
+++ b/chain/ethevents/events.go
@@ -150,60 +150,66 @@ func NewEthEvents(
 }
 
 // OnComputeResults is called once a process result is computed by the scrutinizer.
-func (ev *EthereumEvents) OnComputeResults(results *models.ProcessResult) {
-	// check vochain
+// The Oracle will build and send a RESULTS transaction to the Vochain.
+// The transaction includes the final results for the process.
+func (ev *EthereumEvents) OnComputeResults(results *scrutinizer.Results) {
 	// check vochain process status
-	vocProcessData, err := ev.VochainApp.State.Process(results.ProcessId, true)
+	vocProcessData, err := ev.VochainApp.State.Process(results.ProcessID, true)
 	if err != nil {
-		log.Errorf("error fetching process %x from the Vochain: %s", results.ProcessId, err)
+		log.Errorf("error fetching process %x from the Vochain: %v", results.ProcessID, err)
 		return
 	}
 	switch vocProcessData.Status {
 	case models.ProcessStatus_RESULTS:
 		// check vochain process results
 		if len(vocProcessData.Results.Votes) > 0 {
-			log.Infof("process %x results already added on the Vochain, skipping", results.ProcessId)
+			log.Infof("process %x results already added on the Vochain, skipping",
+				results.ProcessID)
 			return
 		}
 	case models.ProcessStatus_ENDED:
 		break
 	default:
-		log.Infof("invalid process %x status %s for setting the results, skipping", results.ProcessId, vocProcessData.Status)
+		log.Infof("invalid process %x status %s for setting the results, skipping",
+			results.ProcessID, vocProcessData.Status.String())
 		return
 	}
+
 	// create setProcessTx
 	setprocessTxArgs := &models.SetProcessTx{
-		ProcessId: results.ProcessId,
-		Results:   results,
+		ProcessId: results.ProcessID,
+		Results:   scrutinizer.BuildProcessResult(results, vocProcessData.EntityId),
 		Status:    models.ProcessStatus_RESULTS.Enum(),
 		Txtype:    models.TxType_SET_PROCESS_RESULTS,
 	}
 
 	stx := &models.SignedTx{}
-	stx.Tx, err = proto.Marshal(&models.Tx{Payload: &models.Tx_SetProcess{SetProcess: setprocessTxArgs}})
-	if err != nil {
-		log.Errorf("cannot marshal set process results tx: %s", err)
+	if stx.Tx, err = proto.Marshal(&models.Tx{
+		Payload: &models.Tx_SetProcess{
+			SetProcess: setprocessTxArgs,
+		},
+	}); err != nil {
+		log.Errorf("cannot marshal set process results tx: %v", err)
 		return
 	}
-	stx.Signature, err = ev.Signer.Sign(stx.Tx)
-	if err != nil {
-		log.Errorf("cannot sign oracle tx: %s", err)
+	if stx.Signature, err = ev.Signer.Sign(stx.Tx); err != nil {
+		log.Errorf("cannot sign oracle tx: %v", err)
 		return
 	}
 
 	txb, err := proto.Marshal(stx)
 	if err != nil {
-		log.Errorf("error marshaling set process results tx: %s", err)
+		log.Errorf("error marshaling set process results tx: %v", err)
 		return
 	}
-	log.Debugf("broadcasting Vochain Tx: %s", setprocessTxArgs.String())
+	log.Debugf("broadcasting Vochain Tx: %s", log.FormatProto(setprocessTxArgs))
 
 	res, err := ev.VochainApp.SendTX(txb)
 	if err != nil || res == nil {
-		log.Errorf("cannot broadcast tx: %s, res: %+v", err, res)
+		log.Errorf("cannot broadcast tx: %v, res: %+v", err, res)
 		return
 	}
-	log.Infof("oracle transaction sent, hash:%s", res.Hash)
+	log.Infof("oracle transaction sent, hash:%x", res.Hash)
 }
 
 // AddEventHandler adds a new handler even log function
--- a/chain/ethevents/events.go
+++ b/chain/ethevents/events.go
@@ -171,7 +171,7 @@ func (ev *EthereumEvents) OnComputeResults(results *scrutinizer.Results) {
 		break
 	default:
 		log.Infof("invalid process %x status %s for setting the results, skipping",
-			results.ProcessID, vocProcessData.Status.String())
+			results.ProcessID, vocProcessData.Status)
 		return
 	}
 
--- a/client/api.go
+++ b/client/api.go
@@ -403,30 +403,34 @@ func (c *Client) TestSendVotes(
 		}
 	}
 	votingElapsedTime := time.Since(start)
-	if checkNullifiers {
-		for {
-			for i, nullHex := range nullifiers {
-				if nullHex == "registered" {
-					registered++
-					continue
-				}
-				null, err := hex.DecodeString(nullHex)
-				if err != nil {
-					return 0, err
-				}
-				if es, _ := c.GetEnvelopeStatus(null, pid); es {
-					nullifiers[i] = "registered"
-				}
+
+	if !checkNullifiers {
+		return votingElapsedTime, nil
+	}
+	// If checkNullifiers, wait until all votes have been verified
+	for {
+		for i, nullHex := range nullifiers {
+			if nullHex == "registered" {
+				registered++
+				continue
 			}
-			if registered == len(nullifiers) {
-				break
+			null, err := hex.DecodeString(nullHex)
+			if err != nil {
+				return 0, err
 			}
-			registered = 0
-			if time.Since(checkStart) > timeDeadLine {
-				return 0, fmt.Errorf("check nullifier time took more than deadline, skipping")
+			if es, _ := c.GetEnvelopeStatus(null, pid); es {
+				nullifiers[i] = "registered"
 			}
 		}
+		if registered == len(nullifiers) {
+			break
+		}
+		registered = 0
+		if time.Since(checkStart) > timeDeadLine {
+			return 0, fmt.Errorf("check nullifier time took more than deadline, skipping")
+		}
 	}
+
 	return votingElapsedTime, nil
 }
 
--- a/client/api.go
+++ b/client/api.go
@@ -254,8 +254,7 @@ func (c *Client) TestSendVotes(
 	censusOrigin models.CensusOrigin,
 	caSigner *ethereum.SignKeys,
 	proofs [][]byte,
-	encrypted bool,
-	doubleVoting bool,
+	encrypted, doubleVoting, checkNullifiers bool,
 	wg *sync.WaitGroup) (time.Duration, error) {
 
 	var err error
@@ -392,7 +391,7 @@ func (c *Client) TestSendVotes(
 	for {
 		time.Sleep(time.Millisecond * 500)
 		if h, err := c.GetEnvelopeHeight(pid); err != nil {
-			log.Warnf("error getting envelope height")
+			log.Warnf("error getting envelope height: %v", err)
 			continue
 		} else {
 			if h >= uint32(len(signers)) {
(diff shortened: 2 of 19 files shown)


chain/ethevents/events.go
```diff
@@ -150,60 +150,66 @@ func NewEthEvents(
...
+			log.Infof("process %x results already added on the Vochain, skipping",
+				results.ProcessID)
 			return
 		}
 	case models.ProcessStatus_ENDED:
 		break
 	default:
-		log.Infof("invalid process %x status %s for setting the results, skipping", results.ProcessId, vocProcessData.Status)
+		log.Infof("invalid process %x status %s for setting the results, skipping",
+			results.ProcessID, vocProcessData.Status.String())
```
mvdan: no need to call String if the verb is already %s

client/api.go
```diff
@@ -404,27 +403,29 @@ func (c *Client) TestSendVotes(
 		}
 	}
 	votingElapsedTime := time.Since(start)
-	for {
-		for i, nullHex := range nullifiers {
-			if nullHex == "registered" {
-				registered++
-				continue
+	if checkNullifiers {
```
mvdan: invert and early return to unindent the rest

cmd/vochaintest/vochaintest.go
```diff
@@ -45,6 +45,7 @@ func main() {
 	parallelCons := flag.Int("parallelCons", 1, "parallel API connections")
 	procDuration := flag.Int("processDuration", 500, "voting process duration in blocks")
 	doubleVote := flag.Bool("doubleVote", true, "send every vote twice")
+	checkNullifiers := flag.Bool("checkNullifiers", false, "check all votes are correct one by one (slow)")
```
mvdan, line 48: So, turning this on is slower but... more secure? Or more reliable? I'm not sure if I get why anyone would ever want to turn it on.

If there's no reason to turn it on, I'm inclined to just not have this feature at all. And if the reason to turn it on is better security/reliability, I wonder why it's not on always.

vochain/state.go
```diff
@@ -45,17 +45,25 @@ var PrefixDBCacheSize = 0
 
 // EventListener is an interface used for executing custom functions during the
 // events of the block creation process.
-// The order in which events are executed is: Rollback, OnVote or OnProcess, Commit.
+// The order in which events are executed is: Rollback, OnVote, Onprocess, On..., Commit.
 // The process is concurrency safe, meaning that there cannot be two sequences
 // happening in parallel.
+//
+// If Commit() returns an error and isFatal set tu true, it is considered a consensus
+// failure and the blockchain will halt.
```
mvdan, line 53: typo: "tu true"

I also think that just an `error` return would be simpler and clearer here. You could then use a special error type to specify fatal errors. For example:

```
type ErrHaltVochain struct {
    reason error
}

func (e ErrHaltVochain) Error() string { return fmt.Sprintf("halting vochain: %v", e.reason) }
func (e ErrHaltVochain) Unwrap() error { return e.reason }
```

Then, instead of:

```
err, fatal := l.Commit(h.Height)
if err != nil {
	if fatal {
		panic(err)
	}
	log.Warn(err)
}
```

you could do

```
err := l.Commit(h.Height)
if err != nil {
	if _, fatal := err.(ErrHaltVochain); fatal {
		panic(err)
	}
	log.Warn(err)
}
```

Or, even simpler, just panic directly in Commit if there's a fatal error. Having the caller panic is really the same, but more complex.

vochain/state.go
```diff
@@ -100,19 +110,24 @@ func NewState(dataDir string) (*State, error) {
...
 	return vs, nil
 }
 
 func initStore(dataDir string, state *State) error {
+	log.Infof("initializing state db store")
 	state.Store = new(iavlstate.IavlState)
 	if err := state.Store.Init(dataDir, "disk"); err != nil {
 		return err
 	}
+	timer := time.Now()
```
mvdan: this is not a timer, so maybe call it `startTime`

vochain/state.go
```diff
@@ -122,6 +137,7 @@ func initStore(dataDir string, state *State) error {
 	if err := state.Store.AddTree(VoteTree); err != nil {
 		return err
 	}
+	log.Infof("state tree load took %d ms", time.Since(timer).Milliseconds())
```
mvdan: I'd just use `%s` with `time.Since(startTime)`, it will already do the right thing by printing e.g. `200ms`. It will also behave better if the duration is over 1s.

vochain/scrutinizer/db.go
```diff
@@ -47,9 +47,25 @@ type Entity struct {
...
 	ProcessID  types.HexBytes `badgerholdKey:"ProcessID"`
-	Votes      []*models.QuestionResult
+	Votes      [][]*big.Int
 	Weight     *big.Int
 	Signatures []types.HexBytes
+	Final      bool
+}
+
+func (r *Results) String() string {
+	results := ""
```
mvdan: use a `strings.Builder`; otherwise you're allocating and copying the entire string at ever `+=` statement. Since there could be thousands of votes, this could be *very* slow, and indeed quadratic.

vochain/scrutinizer/scrutinizer.go
```diff
@@ -34,16 +38,19 @@ func (s *Scrutinizer) AddEventListener(l EventListener) {
...
 	// updateProcessPool is the list of process IDs that require sync with the state database
 	updateProcessPool [][]byte
 	// resultsPool is the list of processes that finish on the current block
 	resultsPool []*types.ScrutinizerOnProcessData
+	// list of live processes (those on which the votes will be computed on arrival)
+	liveResultsProcs sync.Map
 	// eventListeners is the list of external callbacks that will be executed by the scrutinizer
 	eventListeners []EventListener
 	db             *badgerhold.Store
+	addVoteLock    sync.RWMutex
```
mvdan: please document what this mutex is actually guarding.

mvdan: For example, right now I don't get it - the DB is already atomic and thread-safe, so having a mutex around inserts/gets of votes seems entirely unnecessary. And if this is preventing data races, then we probably have a bug somewhere.

vochain/scrutinizer/scrutinizer.go
```diff
@@ -56,52 +63,103 @@ func NewScrutinizer(dbPath string, state *vochain.State) (*Scrutinizer, error) {
 		return nil, err
 	}
 	s.VochainState.AddEventListener(s)
+	go s.AfterSyncBootstrap()
 	return s, nil
 }
 
+func (s *Scrutinizer) AfterSyncBootstrap() {
+	// Some grace time to avoid false positive on IsSynchronizing()
```
mvdan: I don't get this...

vochain/scrutinizer/scrutinizer.go
```diff
@@ -56,52 +63,103 @@ func NewScrutinizer(dbPath string, state *vochain.State) (*Scrutinizer, error) {
...
+	time.Sleep(time.Second * 10)
+	for s.VochainState.IsSynchronizing() {
+		time.Sleep(time.Second * 2)
+	}
```
mvdan, line 100: is there no way to block until the state is synchronized? that would prevent the polling/sleeping, and supposedly also fix the 10s sleep at the beginning too.

For example, have one channel that is closed when the state is fully synced. Then, here we just need to `<-s.VochainState.Synced()` or something like that. When the channel is closed, that receive will unblock.

vochain/scrutinizer/scrutinizer.go
```diff
@@ -56,52 +63,103 @@ func NewScrutinizer(dbPath string, state *vochain.State) (*Scrutinizer, error) {
...
-	if nvotes > 0 {
-		log.Infof("added %d live votes from block %d", nvotes, height)
 	}
 
 	// Update existing processes
 	for _, p := range s.updateProcessPool {
 		if err := s.updateProcess(p); err != nil {
-			log.Warnf("commit: cannot update process %x: %v", p, err)
+			log.Errorf("commit: cannot update process %x: %v", p, err)
+			continue
```
mvdan: this loop does nothing else, so I guess you can remove the continue

vochain/scrutinizer/scrutinizer.go
```diff
@@ -56,52 +63,103 @@ func NewScrutinizer(dbPath string, state *vochain.State) (*Scrutinizer, error) {
...
+					nvotes++
+				}
+			}
+			if err := s.commitVotes([]byte(pid), results); err != nil {
+				log.Errorf("cannot commit live votes: (%v)", err)
+			}
+		}
+		if nvotes > 0 {
+			log.Infof("added %d live votes on block %d, took %d ms",
+				nvotes, height, time.Since(timer).Milliseconds())
```
mvdan: same comments about `timer` and `%s` like before :)

vochain/scrutinizer/scrutinizer.go
```diff
@@ -56,52 +63,103 @@ func NewScrutinizer(dbPath string, state *vochain.State) (*Scrutinizer, error) {
...
+	// Add votes collected by onVote (live results)
+	// this can be run async
+	go func() {
+		nvotes := 0
+		timer := time.Now()
+		for pid, votes := range s.votePool {
+			results := &Results{Weight: new(big.Int).SetUint64(0)}
+			for _, v := range votes {
+				if err := s.addLiveVote(v, results); err != nil {
+					log.Warnf("vote cannot be added: %v", err)
```
mvdan, line 215: shouldn't this be Errorf?

vochain/scrutinizer/scrutinizer.go
```diff
@@ -115,7 +173,11 @@ func (s *Scrutinizer) OnProcess(pid, eid []byte, censusRoot, censusURI string) {
 
 // OnVote scrutinizer stores the votes if liveResults enabled
 func (s *Scrutinizer) OnVote(v *models.Vote) {
-	s.votePool = append(s.votePool, v)
+	if _, ok := s.votePool[string(v.ProcessId)]; ok {
+		s.votePool[string(v.ProcessId)] = append(s.votePool[string(v.ProcessId)], v)
```
mvdan, line 250: I think you can just have this line. Because if the map key doesn't exist, you'll do `append` with the first argument being nil, which will work just fine.

vochain/scrutinizer/vote.go
```diff
@@ -163,80 +135,125 @@ func (s *Scrutinizer) addLiveVote(envelope *models.Vote) error {
...
+	s.addVoteLock.Lock()
+	defer s.addVoteLock.Unlock()
+	update := func(record interface{}) error {
+		update, ok := record.(*Results)
+		if !ok {
+			return fmt.Errorf("record isn't the correct type! Wanted Result, got %T", record)
+		}
+		update.Weight.Add(update.Weight, results.Weight)
+
+		if len(results.Votes) > 0 {
```
mvdan: invert and early return to unindent

vochain/scrutinizer/vote.go
```diff
@@ -163,80 +135,125 @@ func (s *Scrutinizer) addLiveVote(envelope *models.Vote) error {
...
+	var err error
+	// If badger returns "Transaction Conflict" we try again until it works
+	maxTries := 1000
+	for maxTries > 0 {
+		err = s.db.UpdateMatching(&Results{},
+			badgerhold.Where("ProcessID").Eq(pid).And("Final").Eq(false), update)
+		if err == nil {
+			break
+		}
+		if strings.Contains(err.Error(), "Transaction Conflict") {
```
mvdan: use badger.ErrConflict instead: https://pkg.go.dev/github.com/dgraph-io/badger/v2#pkg-variables

If the error is wrapped, then use `errors.Is(err, badger.ErrConflict)`.

vochain/scrutinizer/vote.go
```diff
@@ -163,80 +135,125 @@ func (s *Scrutinizer) addLiveVote(envelope *models.Vote) error {
...
-	return &models.ProcessResult{
-		Votes:     results.Votes,
-		ProcessId: processID,
-	}, nil
+	if maxTries == 0 {
+		err = fmt.Errorf("got too much transaction conflicts")
+	}
+	if err != nil {
+		log.Debugf("saved %d votes with total weight of %s on process %x", len(results.Votes),
+			results.Weight.String(), pid)
```
mvdan: no need to call String with %s

mvdan (commented):
I can't really approve because I fail to properly understand half of the changes :) If you have the time and want me to look more carefully, maybe we can do a 30m call and I'll ask questions while reading the changes.

vochain/scrutinizer/scrutinizer.go
p4u: Warning is fine since the user might provide a wrong Vote package and that should be considered as a "null vote", it's not an error and probably in the future (when we are sure everything works as expected) we can even remove the warning

vochain/scrutinizer/vote.go
p4u: Mmm, I did not add this error cause I did not want to import a specific KV implementation. So badgerhold can switch from badger v2 to v3 without any change in our side. Do you really think it is worth it? (importing badger v2 just for handling this error).

vochain/scrutinizer/scrutinizer.go
p4u: That's for avoiding the "Transaction Conflict" error which happens quite often. That's a non critical error and the badger developers recommend to try it again until the db transaction works. But I preferred to try to avoid directly the put/get conflict with a Mutex. And yes, with that Mutex the iteration over the "Transaction Conflict" error might be removed but I let it there for security since I'm still not 100% sure all conflicting calls are mutex.

vochain/scrutinizer/scrutinizer.go
p4u: During the first seconds/milliseconds of the Vochain startup, Tendermint might report that the chain is not synchronizing since it still does not have any peer and do not know the actual size of the blockchain. If AfterSyncBootStrap is executed on this specific moment, the Wait loop would pass.

vochain/scrutinizer/scrutinizer.go
p4u: With the channel we would still have the same problem.

Maye a more safe and dynamic approach would be something like:
```
        syncSignal := 5
	for {
	is !s.VochainState.IsSynchronizing() { syncSignal-- }
	if syncSignal == 0 { break }
        time.Sleep(time.Second * 1)
	}
```

p4u:
> I can't really approve because I fail to properly understand half of the changes :) If you have the time and want me to look more carefully, maybe we can do a 30m call and I'll ask questions while reading the changes.

Well I suppose you read the commit message already. I understand the changes are huge and there are plenty of design decisions on there kind of hidden for a third party developer. I have time for a short call this morning, else until 6th I won't be available.

vochain/scrutinizer/scrutinizer.go
mvdan: If retrying works, then why have the mutex? I don't know, having a mutex around the entire DB feels hacky. If we have good tests and they continuously pass without the mutex and the error retry, then I'd say that's good.

vochain/scrutinizer/scrutinizer.go
mvdan: Gotcha, maybe summarize that here?

vochain/scrutinizer/vote.go
mvdan: Right, I wasn't thinking about v2 vs v3. If we have good tests which cover the conflict case (e.g. with many reads/writes), then I imagine the tests would catch if they bump to v3 but our error points at v2.

Alternatively, add a short comment to explain the strings.Contains? Most times it's a mistake, so it would be good to clarify why it's wanted here.

mvdan:
No worries. Yeah, I read the commits and they make sense to me, though the code is kinda hard to follow. I think overall LGTM, the only major thing that worries me is why we want the checkNullifiers flag - left a comment about that earlier.

cmd/vochaintest/vochaintest.go
p4u: This is a testing tool. I think the full check of nullifiers could be useful depending on the purpose of the testing. I will keep it available, it's not a big chunk of code.

vochain/state.go
p4u: Great solution. Thanks.

p4u:
Thanks for the review. Some day I'd like to have a "mvdan-linter" to include on golangci-lint :)

vochain/scrutinizer/scrutinizer.go
p4u: Well, to me it looks that repeating a DB transaction N times until it success is kind of hacky. Using a Mutex on this scope makes sense and will release load on the KV context. 

In addition I added a `maxRetries=1000` mechanism which is kind of arbitrary (I tested different values until there were no more transactions failing). This value works in my local computer but a slower one could need more maxRetries. I don't like to block the goroutine forever if there is an issue on the KV (that won't accept new transactions) I prefer instead to try N times and raise an Error. So this is why I'm mixing both approaches: Mutex + Repeat Transaction N times

I'll add a to-do for removing it in the future, with proper research and testing involved. But for now I prefer to make it safe.

PR #147, opened 2021-04-08
WIP Feature/drop envelope database

NateWilliams2:
- [x] Index nullifiers -> voteTxs on scrutinizer
- [ ] Hash envelopes on VoteTree
- [ ] Remove `state.Envelope()` (or rename it to `state.EnvelopeHash()`) and switch to `scrutinizer.GetEnvelope()`  instead
- [ ] Rename scrutinizer to indexer `find -type f -exec sed -i s/scrutinizer/indexer/g {}\;`


router/voteCallbacks.go
```diff
@@ -106,21 +106,31 @@ func (r *Router) getEnvelopeStatus(request routerRequest) {
...
 		r.sendError(request, "cannot get envelope status: (malformed nullifier)")
 		return
 	}
-	envelope, err := r.vocapp.State.Envelope(request.ProcessID, request.Nullifier, true)
+	txRef, err := r.Scrutinizer.GetVoteReference(request.Nullifier)
 	if err != nil {
 		r.sendError(request, fmt.Sprintf("cannot get envelope: (%s)", err))
 		return
 	}
+	block := r.vocapp.Node.BlockStore().LoadBlock(txRef.BlockHeight)
```
p4u: I think the `State` should have this methods abstracted.
+ state.GetBlockByHeight(height)
+ state.GetBlockByHash(hash)
+ state.GetTx(height, txIndex)

I'm not sure the Tendermint  BlockStore is safe, if not we should add a `sync.Mutex`

The Scrutinizer then would have a method for geting a specific envelope (already Unmarshaled)

+ scrutinizer.GetEnvelope(nullifier)

vochain/scrutinizer/db.go
```diff
@@ -46,6 +46,14 @@ type Entity struct {
 	CreationTime time.Time
 }
 
+type VoteReference struct {
+	Nullifier    types.HexBytes `badgerholdKey:"Nullifier"`
+	ProcessId    types.HexBytes `badgerholdIndex:"ProcessID"`
+	BlockHeight  int64
```
p4u: Just Height should be fine

vochain/scrutinizer/vote.go
```diff
@@ -27,6 +27,14 @@ var ErrNoResultsYet = fmt.Errorf("no results yet")
 // implementation, thus let badgerhold abstract it.
 const kvErrorStringForRetry = "Transaction Conflict"
 
+// GetVoteReference gets the reference for an AddVote transaction.
+// This reference can then be used to fetch the vote transaction directly from the BlockStore.
+func (s *Scrutinizer) GetVoteReference(nullifier []byte) (*VoteReference, error) {
+	txRef := &VoteReference{}
+	err := s.db.FindOne(txRef, badgerhold.Where(badgerhold.Key).Eq(nullifier))
+	return txRef, err
```
p4u: You can do: `return txRef, s.db.FindOne(txRef, badgerhold.Where(badgerhold.Key).Eq(nullifier))`

vochain/scrutinizer/vote.go
```diff
@@ -174,6 +182,17 @@ func (s *Scrutinizer) addLiveVote(envelope *models.Vote, results *Results) error
 	return nil
 }
 
+// addVoteIndex adds the nullifier reference to the kv for fetching vote Txs from BlockStore.
+// This method is triggered by Commit callback for each vote added to the blockchain.
+func (s *Scrutinizer) addVoteIndex(nullifier, pid []byte, blockHeight int64, txIndex int32) error {
+	_, err := s.VochainState.Process(pid, false)
```
p4u: There is no need for this call. We should in general try to minimize the number of database transactions.

vochain/scrutinizer/vote.go
```diff
@@ -174,6 +182,17 @@ func (s *Scrutinizer) addLiveVote(envelope *models.Vote, results *Results) error
...
+	if err != nil {
+		return fmt.Errorf("cannot get process %x: %w", pid, err)
+	}
+	voteRef := &VoteReference{Nullifier: nullifier, ProcessId: pid, BlockHeight: blockHeight, TxIndex: txIndex}
+	return s.db.Insert(nullifier, voteRef)
```
p4u: no need to declare voteRef on a separated line

p4u (changes requested):

router/voteCallbacks.go
NateWilliams2: It seems like the BlockStore is safe for this:
```go
type BlockStore struct {
	db dbm.DB

	// mtx guards access to the struct fields listed below it. We rely on the database to enforce
	// fine-grained concurrency control for its data, and thus this mutex does not apply to
	// database contents. The only reason for keeping these fields in the struct is that the data
	// can't efficiently be queried from the database since the key encoding we use is not
	// lexicographically ordered (see https://github.com/tendermint/tendermint/issues/4567).
	mtx    tmsync.RWMutex
	base   int64
	height int64
}
```

mvdan:
Is the target branch here wrong? Seems like it should be master.

p4u:
Yes, it should be Master.

p4u:
Moved here https://github.com/vocdoni/vocdoni-node/pull/153

PR #156, opened 2021-04-27
update ethereum contracts bindings:

jordipainan:
- update protobuf pkg version
- update NewProcess event handler
- update GetBalanceMappingPosition


chain/contractWrappers.go
```diff
@@ -210,12 +210,22 @@ func (ph *VotingHandle) NewProcessTxArgs(ctx context.Context, pid [types.Process
...
 		}
 		iSlot32 := uint32(islot.Uint64())
 		processData.EthIndexSlot = &iSlot32
+		// decode owner
+		if processData.Owner, err = hex.DecodeString(util.TrimHex(processMeta.EntityAddressOwner[1].String())); err != nil {
+			return nil, fmt.Errorf("error decoding owner address: %w", err)
+		}
+		// save sourceBlockHeight
+		sourceBlockHeight64 := uint64(processMeta.SourceBlockHeight.Uint64())
+		if sourceBlockHeight64 == 0 {
```
p4u, line 225: Do we want to obligate the usage of sourceBlockHeight? I am not sure, I think that should be optional (as it is the field on protobuf).

chain/contractWrappers.go
```diff
@@ -122,7 +122,7 @@ func (ph *VotingHandle) NewProcessTxArgs(ctx context.Context, pid [types.Process
 
 	// entity id
 	// for evm censuses the entity id is the snapshoted contract address
-	if processData.EntityId, err = hex.DecodeString(util.TrimHex(processMeta.EntityAddress.String())); err != nil {
+	if processData.EntityId, err = hex.DecodeString(util.TrimHex(processMeta.EntityAddressOwner[0].String())); err != nil {
```
p4u, line 125: What's the type of EntityAddressOwner[0] ? Are we sure it can have the 0x behind the string?

chain/contractWrappers.go
jordipainan: Is optional on the protobuf because process can have on-chain / off-chain census. The code here only applies to processes that need IndexSlot (on-chain census) so I assume these processes will also require to have the sourceBlockHeight stored (if we really want to have it stored)

chain/contractWrappers.go
jordipainan: Returns 2 hex encoded strings.
Example:
  ```
  entityAddress_owner: [
    '0x9561C133DD8580860B6b7E504bC5Aa500f0f06a7',
    '0x90F8bf6A479f320ead074411a4B0e7944Ea8c9C1'
  ],
  ```

PR #157, opened 2021-04-27
Use dev tag

vdo:


p4u:
The idea behind splitting dev and master builds was to be able to have a specific target deployment branch.
This allows us to push stuff which is not yet commited to master in order to try it. Its been quite useful until now.

So I'd not apply this PR and instead push manually to dev whenever we want to deploy something.

PR #158, opened 2021-04-27
Feature/cost from weight

p4u:
See commit message.

### DIFF
--- a/vochain/scrutinizer/db.go
+++ b/vochain/scrutinizer/db.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
-	"math"
 	"math/big"
 	"sync"
 	"time"
@@ -84,6 +83,36 @@ func (r *Results) String() string {
 	return results.String()
 }
 
+// Add adds the total weight and votes from the given Results to the containing Results making the method call.
+func (r *Results) Add(new *Results) error {
+	r.Weight.Add(r.Weight, new.Weight)
+	if new.Height > r.Height {
+		r.Height = new.Height
+	}
+	if len(new.Votes) == 0 {
+		return nil
+	}
+	if len(new.Votes) != len(r.Votes) {
+		return fmt.Errorf("results.Add: different number of fields")
+	}
+	for i := range new.Votes {
+		if len(r.Votes[i]) < len(new.Votes[i]) {
+			return fmt.Errorf("results.Add: values overflow (%d)", i)
+		}
+		// Example:
+		//	[ [1,2,3], [0,1,5] ]
+		//	[ [2,1,1], [1,0,2] ]
+		//	+ -----------------
+		//	[ [3,3,4], [1,1,7] ]
+		for j := range new.Votes[i] {
+			r.Votes[i][j].Add(r.Votes[i][j], new.Votes[i][j])
+		}
+	}
+	return nil
+}
+
+// AddVote adds the voteValues and weight to the Results struct.
+// Checks are performed according the Ballot Protocol.
 func (r *Results) AddVote(voteValues []int, weight *big.Int, mutex *sync.Mutex) error {
 	if r.VoteOpts == nil {
 		return fmt.Errorf("addVote: processVoteOptions is nil")
@@ -113,7 +142,7 @@ func (r *Results) AddVote(voteValues []int, weight *big.Int, mutex *sync.Mutex) 
 		}
 	}
 
-	// Max Value
+	// Max Value, check it only if greather than zero
 	if r.VoteOpts.MaxValue > 0 {
 		for _, v := range voteValues {
 			if uint32(v) > r.VoteOpts.MaxValue {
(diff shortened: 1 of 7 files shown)


vochain/scrutinizer/db.go
```diff
@@ -83,6 +82,33 @@ func (r *Results) String() string {
 	}
 	return results.String()
 }
+func (r *Results) Add(new *Results) error {
+	r.Weight.Add(r.Weight, new.Weight)
```
NateWilliams2, line 88: This line adds the old results weight to the new partial results weight, whereas the old corresponding line simply sets the new weight. 
```
update.Weight.Set(results.Weight)
```

I remember we had a bug with this in the past that caused the weight to increase exponentially. Is this change (along with the renaming of `results` to `partialResults` clarifying that commitVotes() should use a partial results which will be added to, rather than replace, the current results?

NateWilliams2 (commented):
I left one clarification comment, otherwise looks good to me!

vochain/scrutinizer/db.go
```diff
@@ -122,34 +148,64 @@ func (r *Results) AddVote(voteValues []int, weight *big.Int, mutex *sync.Mutex)
...
 	if len(r.Votes) == 0 {
 		r.Votes = newEmptyVotes(int(r.VoteOpts.MaxCount), int(r.VoteOpts.MaxValue)+1)
 	}
-	for q, opt := range voteValues {
-		r.Votes[q][opt].Add(r.Votes[q][opt], weight)
-	}
 
+	// If MaxValue is zero, consider discrete value couting. So for each questoin, the value
+	// is aggregated. The weight is multiplied for the value if costFromWeight=False.
+	// This is a special case for Cuadratic Voting where maxValue should be 0 (no limit).
```
jordipainan: typo `Quadratic` ?

vochain/scrutinizer/scrutinizer_test.go
```diff
@@ -562,6 +570,58 @@ func TestBallotProtocolRateProduct(t *testing.T) {
 	qt.Assert(t, votes[0], qt.DeepEquals, []string{"1", "0", "1", "0", "2"})
 }
 
+func TestBallotProtocolCuadratic(t *testing.T) {
```
jordipainan: Also `Quadratic`

vochain/scrutinizer/scrutinizer_test.go
```diff
@@ -562,6 +570,58 @@ func TestBallotProtocolRateProduct(t *testing.T) {
...
+		BlockCount:   10,
+		Mode:         &models.ProcessMode{AutoStart: true},
+		VoteOptions:  &models.ProcessVoteOptions{MaxCount: 2, MaxValue: 0, CostExponent: 2},
+	}); err != nil {
+		t.Fatal(err)
+	}
+
+	qt.Assert(t, sc.newEmptyProcess(pid), qt.IsNil)
+
+	// Cuadratic voting, exepected result: [ [100016], [100033] ]
```
jordipainan: typo `Quadratic`

vochain/transaction.go
```diff
@@ -295,23 +298,21 @@ func AdminTxCheck(vtx *models.Tx, txBytes, signature []byte, state *State) error
...
 			// endblock is always greater than start block so that case is also included here
-			if header.Height > int64(process.StartBlock) {
+			if height > process.StartBlock {
 				return fmt.Errorf("cannot add process keys in a started or finished process")
 			}
 			// process is not canceled
-			if process.Status == models.ProcessStatus_CANCELED || process.Status == models.ProcessStatus_ENDED || process.Status == models.ProcessStatus_RESULTS {
+			if process.Status == models.ProcessStatus_CANCELED || process.Status == models.ProcessStatus_ENDED ||
+				process.Status == models.ProcessStatus_RESULTS {
 				return fmt.Errorf("cannot add process keys in a canceled process")
```
jordipainan: this error message can be misleading if the process status is ENDED or RESULTS, maybe we can return the process status formatted on the error

vochain/scrutinizer/db.go
p4u: AAAhhh, now I see! I got crazy trying to understand why commitVotes was doing Set instead of Add :)

The current description of commitVotes is as follows (it was actually like this, I did not modify it). Would you make it more clear? Can you propose an alternative? Thanks.
```
// commitVotes adds the votes and weight from results to the local database.
// It does not overwrite the stored results but update them by adding the new content.
func (s *Scrutinizer) commitVotes(pid []byte, partialResults *Results, height uint32) error 
```

vochain/scrutinizer/db.go
p4u: thanks

vochain/transaction.go
p4u: Well that's not from this commit but I'll fix it. Thanks.

vochain/scrutinizer/db.go
p4u: Btw @NateWilliams2 I think the actual bug was not in the commitVotes() method but in the scrutinizer_test, which I fixed on this PR.

NateWilliams2: I see! No, this looks clear to me. I "fixed" the bug according to the tests, rather than the specification for the method, assuming that the test was correct. Sounds like this is cleared up now :)

PR #159, opened 2021-04-27
Add prometheus and grafana to testsuite

vdo:
This allows the testsuite to be started with a prometheus+grafana instance, if added when doing the `docker-compose up` so benchmarks and manual tests can be monitored.

It's a bit of bloat to add to the vocdoni-node repo, we could consider splitting the testsuite to another repo and add it as a submodule (?)


p4u:
That's amazing! Will be very useful for testing and benchmarking.
I think its fine to have these compose files on the testsuit directory. However we could configure the github CI in order to not start it.

EDIT: I see that this is a different docker-compose file, so the CI won't start it. Great.

p4u (approved):

PR #170, opened 2021-05-11
IPFS/IPNS process archive

p4u:
See commit messages


cmd/dvotenode/dvotenode.go
```diff
@@ -389,9 +397,11 @@ func main() {
 	var ma *metrics.Agent
 
 	if globalCfg.Dev {
-		log.Warn("developer mode is enabled, I hope you know what you are doing ;)")
+		log.Warn("¡developer mode is enabled!")
```
mvdan: remove the leading one? this isn't spanish :)

cmd/dvotenode/dvotenode.go
```diff
@@ -464,9 +476,16 @@ func main() {
...
-		if err != nil {
+		scrutinizer := (globalCfg.Mode == types.ModeGateway && globalCfg.API.Results) ||
+			(globalCfg.Mode == types.ModeOracle)
+		if vnode, sc, vinfo, err = service.Vochain(
+			globalCfg.VochainConfig,
+			scrutinizer,
+			globalCfg.VochainConfig.NoWaitSync,
+			ma,
+			cm,
+			storage,
```
mvdan: I think this is a bit excessive :) 2-3 lines for this call would be plenty

data/ipfs.go
```diff
@@ -257,7 +262,76 @@ func (i *IPFSHandle) Retrieve(ctx context.Context, path string, maxSize int64) (
...
+		ck, err := i.CoreAPI.Key().Self(ctx)
+		if err != nil {
+			return nil, err
+		}
+		keyalias = ck.Name()
+	}
+	return i.CoreAPI.Name().Publish(
+		ctx,
+		ipath.IpfsPath(c),
+		options.Name.TTL(time.Duration(time.Minute*10)),
```
mvdan: this time.Duration conversion is unnecessary. I think you might spot more of these with https://github.com/mdempsky/unconvert

data/ipfs.go
```diff
@@ -257,7 +262,76 @@ func (i *IPFSHandle) Retrieve(ctx context.Context, path string, maxSize int64) (
...
+	}
+	return nil
+}
+
+// NewIPFSkey generates a new IPFS private key (ECDSA/256bit) and returns its
+// marshaled bytes representation.
+func NewIPFSkey() []byte {
+	// These functions must not return error since all input parameters
+	// are predefined, so we panic if an error returned.
+	pk, _, err := ipfscrypto.GenerateKeyPair(ipfscrypto.ECDSA, 256)
```
mvdan: I had to do a double take to think if pk was "private key" or "public key"; I'd maybe use `privKey` to be explicit. And if you have a marshaled version, then `encPrivKey` or something like that.

service/vochain.go
```diff
@@ -105,17 +111,35 @@ func Vochain(vconfig *config.VochainCfg, results, waitForSync bool, ma *metrics.
...
 		log.Infof("starting census downloader service")
 		censusdownloader.NewCensusDownloader(vnode, cm, !vconfig.ImportPreviousCensus)
 	}
 
+	// Process Archiver
+	if vconfig.EnableProcessArchive {
+		ipfs, ok := storage.(*data.IPFSHandle)
+		if !ok {
+			log.Warnf("storage is not IPFS, archive publishing disabled")
+			ipfs = nil
```
mvdan: pretty sure this is unnecessary. ipfs will always be nil if !ok.

vochain/processarchive/p2p.go
```diff
@@ -0,0 +1,87 @@
...
+const (
+	publishInterval = 10 * time.Minute
+	ipnsKeyAlias    = "processarchivekey"
+)
+
+// AddKey adds a base64 encoded ECDSA 256bit private key or generates
+// a new one.
+func (p *ProcessArchive) AddKey(b64key string) error {
+	var ipnsPk []byte
+	var err error
```
mvdan: move this err declaration closer to where it's needed, i.e. the else block

vochain/processarchive/p2p.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+// Publish is a blocking routine that reads ProcessArchive.publish channel
+// in order to trigger a new IPNS record announcement publishing the
+// process archive directory.
+func (p *ProcessArchive) Publish() {
+	log.Infof("starting process archive IPNS publish daemon with interval %s",
+		publishInterval)
+	// Wait for publishInterval and ensure a first execution
+	p.lastUpdate = time.Now().Add(-publishInterval)
```
mvdan, line 60: can you not set this to e.g. unix epoch, 1970? Then we'll always update, since the last update was decades ago (never).

I find it's a bit weird to say the last update was 10m ago, when the Go process just started.

vochain/processarchive/processarchive.go
```diff
@@ -0,0 +1,140 @@
...
+	js, err := NewJsonStorage(datadir)
+	if err != nil {
+		return nil, fmt.Errorf("could not create process archive: %w", err)
+	}
+	ir := &ProcessArchive{
+		vochain: v,
+		ipfs:    ipfs,
+		storage: js,
+		publish: make(chan (bool), 10),
+		close:   make(chan (bool), 1),
```
mvdan: I think it would be better to make NewProcesArchive take a context, and then plumb it down to Publish. Practically the same in practice, but more idiomatic, and we don't need to remember to call the Close method and write the channel stuff ourselves.

vochain/processarchive/p2p.go
```diff
@@ -0,0 +1,87 @@
...
+		if time.Since(p.lastUpdate) < publishInterval {
+			break
+		}
+		p.lastUpdate = time.Now()
+		// make it async so the channel does not get full and blocking
+		go func() {
+			p.publishLock.Lock()
+			defer p.publishLock.Unlock()
+			log.Infof("publishing process archive")
+			st := time.Now()
```
mvdan: `start`? :)

vochain/processarchive/p2p.go
```diff
@@ -0,0 +1,87 @@
...
+			ctx, cancel := context.WithTimeout(context.Background(), publishInterval)
+			ipnsentry, err := p.ipfs.PublishIPNSpath(ctx, p.storage.datadir, ipnsKeyAlias)
+			cancel()
```
mvdan: you could just put this in a defer, which is a bit safer in the long run in case the code gets more complex

vochain/processarchive/p2p.go
```diff
@@ -0,0 +1,87 @@
...
+	select {
+	case <-p.publish:
+		if time.Since(p.lastUpdate) < publishInterval {
+			break
+		}
+		p.lastUpdate = time.Now()
+		// make it async so the channel does not get full and blocking
+		go func() {
+			p.publishLock.Lock()
+			defer p.publishLock.Unlock()
```
mvdan, line 71: I don't get the point of this lock. If two goroutines call Publish concurrently, this func is racy because of how we use lastUpdate. And if we don't support calling Publish concurrently, then this lock is unnecessary.

vochain/processarchive/processarchive.go
```diff
@@ -0,0 +1,140 @@
...
+	if p == nil || p.Process == nil || len(p.Process.ProcessId) != types.ProcessIDsize {
+		return fmt.Errorf("process not valid")
+	}
+	data, err := json.Marshal(p)
+	if err != nil {
+		return err
+	}
+	js.lock.Lock()
+	defer js.lock.Unlock()
+	return os.WriteFile(filepath.Join(js.datadir, fmt.Sprintf("%x", p.Process.ProcessId)), data, 0o644)
```
mvdan, line 58: atomically writing files is possible with https://github.com/google/renameio, in which case you can get rid of the global locking.

vochain/processarchive/p2p.go
```diff
@@ -0,0 +1,87 @@
...
+func (p *ProcessArchive) GetKey() (string, error) {
+	pk, err := p.ipfs.Node.Repo.Keystore().Get(ipnsKeyAlias)
+	if err != nil {
+		return "", err
+	}
+	pkb, err := ipfscrypto.MarshalPrivateKey(pk)
+	if err != nil {
+		return "", err
+	}
+	return base64.StdEncoding.EncodeToString(pkb), nil
```
mvdan, line 50: I admit I'm not a fan of using base64 strings for keys everywhere. It would be a bit cleaner to just use the unmarshaled []byte directly internally, and only deal with marshaling and base64 when interacting with the user (reading/writing a config, reading a flag, writing a log...).

There are even nice and general APIs for this, which should be compatible with most flag/config libraries: https://golang.org/pkg/encoding/#TextMarshaler

You don't have to do this here if it feels like too much added work. I could take a look in the future as part of a "flag cleanup" refactor.

mvdan (approved):

vochain/processarchive/p2p.go
mvdan: > I could take a look in the future as part of a "flag cleanup" refactor.

if so, we should raise an issue about it, too

vochain/processarchive/processarchive.go
p4u: Will let a TO-DO.

vochain/processarchive/p2p.go
p4u: Its for safety. There might be some corner case where a new iteration happens while the current one has not yet finished.
I'm trying to ensure this does not happend by using the context.WithTimeout but some time adjustments on the host computer or something else might broke this logic. So that lock just tries to handle kind of safe this very unlikely situation.

vochain/processarchive/processarchive.go
p4u: TO-DO

vochain/processarchive/p2p.go
p4u: Yes, lets raise an Issue.

vochain/processarchive/p2p.go
mvdan: I still don't get it though. This doesn't run in a loop nor is it called many times. And, as said above, if concurrent calls to this func are happening, then the lock isn't saving you - the func is still racy.

PR #172, opened 2021-05-12
Feature/api improvements

NateWilliams2:
Make some small improvements to indexer api methods, mostly removing redundancies. 
Also improve formatting, check r.Send error values


jordipainan (approved):
LGTM :)

router/router.go
```diff
@@ -146,7 +146,8 @@ type routerRequest struct {
 }
 
 // semi-unmarshalls message, returns method name
-func (r *Router) getRequest(payload []byte, context transports.MessageContext) (request routerRequest, err error) {
+func (r *Router) getRequest(
```
p4u: Let's unify syntax model. We don't usually do this, instead:
```
getRequest( some parameters....,
      some more parameters....) {
```
So split the line in two or three lines of similar size.

p4u (approved):

router/indexerCallbacks.go
```diff
@@ -106,7 +106,7 @@ func (r *Router) getBlockList(request routerRequest) {
 			break
 		}
 		response.BlockList = append(response.BlockList,
-			indexertypes.BlockMetadataFromBlockModel(
+			blockMetadataFromBlockModel(
```
p4u: Same here with the syntax

PR #174, opened 2021-05-13
remove dependency on blindca, downgrade fuse

mvdan:


NateWilliams2:
It builds on my Mac with no issues :)

NateWilliams2 (approved):

PR #180, opened 2021-05-17
vochain: save process metadata

jordipainan:
- update proto version and change types accordingly

### DIFF
--- a/ethereum/handler/ethereumHandler.go
+++ b/ethereum/handler/ethereumHandler.go
@@ -223,6 +223,9 @@ func (eh *EthereumHandler) NewProcessTxArgs(ctx context.Context, pid [types.Proc
 		return nil, fmt.Errorf("error decoding entity address: %w", err)
 	}
 
+	// process metadata
+	processData.Metadata = &processMeta.MetadataCensusRootCensusUri[0]
+
 	// census root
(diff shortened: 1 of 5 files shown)


ethereum/handler/ethereumHandler.go
```diff
@@ -223,6 +223,12 @@ func (eh *EthereumHandler) NewProcessTxArgs(ctx context.Context, pid [types.Proc
 		return nil, fmt.Errorf("error decoding entity address: %w", err)
 	}
 
+	// process metadata
+	if len(processMeta.MetadataCensusRootCensusUri[0]) == 0 {
```
p4u: I'd allow creating a process without metadata, so I'd remove this check.

oracle/oracle.go
```diff
@@ -42,6 +42,9 @@ func (o *Oracle) NewProcess(process *models.Process) error {
 	if len(process.EntityId) != types.EntityIDsize {
 		return fmt.Errorf("entityId size is wrong")
 	}
+	if len(*process.Metadata) == 0 {
```
p4u: same

p4u (changes requested):

oracle/oracle.go
jordipainan: oki, thank you pau

jordipainan:
@p4u  all solved

p4u (approved):

p4u:
Great, thank you!

PR #181, opened 2021-05-18
Indexer: Update process directly on results

NateWilliams2:
With the old code, it is possible for RollBack() to be called before Commit(), refreshing the updateProcessPool before processing its contents. This leaves the given process non-updated, and the status remains ENDED. 
This change directly updates the process, ensuring its state will be set to RESULTS

### DIFF
--- a/vochain/scrutinizer/scrutinizer.go
+++ b/vochain/scrutinizer/scrutinizer.go
@@ -451,9 +451,9 @@ func (s *Scrutinizer) OnProcessResults(pid []byte, results []*models.QuestionRes
 				}
 			}
 		}
-		s.updateProcessPool = append(s.updateProcessPool, pid)
 		log.Infof("published results for process %x are correct!", pid)
 	}()
+	s.updateProcessPool = append(s.updateProcessPool, pid)
 	return nil
 }
 


p4u:
Rollback() is called (by BeginBlock) before Commit() for every block, the order is: `Rollback() => DeliverTx() => Commit()`

A block cycle can be reversed, if for instance a proposed block by a validator does not get enought votes during the consensus phase. This is why all operations performed by the blockchain and the event listenets can only be stored permanently on Commit(). So we use in-memory operation executing accumulators such as updateProcessPool that accumulates pending operations which are executed only on Commit.

Your fix breaks this logic and commits a change directly on DeliverTx() where the event listeners callbacks are triggered. So that cannot be the fix for the current race on set results status.

p4u:
We might move this line https://github.com/vocdoni/vocdoni-node/blob/master/vochain/scrutinizer/scrutinizer.go#L454 ouf of the goroutine, so it is executed sync.

The goroutine is currently only informing the user if a published results are valid or not, but it does not perform any real operation regarding the blockchain state.

NateWilliams2:
> We might move this line https://github.com/vocdoni/vocdoni-node/blob/master/vochain/scrutinizer/scrutinizer.go#L454 ouf of the goroutine, so it is executed sync.

Ok, I'll try this and test.

> The goroutine is currently only informing the user if a published results are valid or not, but it does not perform any real operation regarding the blockchain state.

I see, so even if the scrutinizer doesn't know yet if results are available, UpdateProcess() should still query the blockchain state to get the actual process status. Like I said, I'll try this out.

NateWilliams2:
The proposed fix works, force-pushed to this branch.

PR #186, opened 2021-05-19
vochain: use a better LRU

mvdan:
(see commit message)

### DIFF
--- a/db/lru/lru.go
+++ b/db/lru/lru.go
@@ -1,0 +1,67 @@
+package lru
+
+import (
+	"sync"
+
+	"github.com/hashicorp/golang-lru/simplelru"
+)
+
+// Cache implements a least-recently-used cache that is safe for concurrent use.
+type Cache struct {
+	lruMu sync.Mutex
+	lru   simplelru.LRU
+}
+
+// New creates a new Cache with a given maximum number of entries.
+func New(size int) *Cache {
+	lru, err := simplelru.NewLRU(size, nil)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


p4u:
Looks good. However I think we might have it as a standalone package so other packages can also use it. Maybe `db/lru` ?

mvdan:
Hm, it has little to do with databases though. Since we use LRUs only for the vochain right now, how about `vochain/lru` or `vochain/cache`?

mvdan:
Actually, I take that back - if only the vochain package needs this LRU, then I don't think we should split it into a separate package right now. We should only do that when we need it in the future. Prematurely splitting stuff into more packages is not a good idea.

p4u:
Also scrutinizer uses LRUs https://github.com/vocdoni/vocdoni-node/blob/master/vochain/scrutinizer/scrutinizer.go#L11

p4u:
And I bet we'll need LRU soon on packages such as censustree and statedb

mvdan:
Ah, fair enough - I'll move it under db/lru.

mvdan:
@p4u take a look, it's moved now. If this sounds good, I'll start moving the other LRU uses to this new package, also extending its API as needed.

p4u:
Yes, sounds good.// New creates a new Cache with a given maximum number of entries.
￼

mvdan:
Should we approve and merge, then? I thought I'd do one PR for each of the LRU caches I replace, to keep the changes smaller and easier to review and revert.

p4u (approved):

PR #189, opened 2021-05-19
vochain: also use our new LRU cache for envelope heights

mvdan:
(see commit message)

### DIFF
--- a/vochain/scrutinizer/vote.go
+++ b/vochain/scrutinizer/vote.go
@@ -209,25 +209,39 @@ func (s *Scrutinizer) GetEnvelopes(processId []byte, max, from int,
 
 // GetEnvelopeHeight returns the number of envelopes for a processId.
 // If processId is empty, returns the total number of envelopes.
-func (s *Scrutinizer) GetEnvelopeHeight(processId []byte) (uint64, error) {
-	if len(processId) > 0 {
-		cc, ok := s.envelopeHeightCache.Get(string(processId))
-		if ok {
-			return cc.(uint64), nil
-		}
-		// TODO: Warning, int can overflow
-		c, err := s.db.Count(&indexertypes.VoteReference{},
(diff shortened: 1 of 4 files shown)


mvdan:
@p4u please test this because I don't think `go test ./...` is covering these caches very well, and I could have made a mistake.

vochain/scrutinizer/vote.go
```diff
@@ -209,24 +209,33 @@ func (s *Scrutinizer) GetEnvelopes(processId []byte, max, from int,
...
+			count, err := s.db.Count(&indexertypes.VoteReference{},
+				badgerhold.Where("ProcessID").Eq(processID).Index("ProcessID"))
+			if err != nil {
+				return err
+			}
+			return uint64(count)
+
+		})
+		switch val := val.(type) {
+		case error:
```
NateWilliams2: In what case could GetOrUpdate return an error?

NateWilliams2 (commented):
Looks fine to me, although I'm confused why in the case of envelopes you check the type of the returned interface, and in the case of blocks you simply cast it.

vochain/scrutinizer/vote.go
mvdan: Thanks for flagging this - it's indeed a bit tricky, and it's on me to not have documented it better. It's now documented.

mvdan:
I've documented why the cached height count can be an error. The block cache can never contain errors because the way we build the values (GetBlockByHeight, i.e. `app.Node.BlockStore().LoadBlock`) can never error.

Arguably this could be a bug on the Tendermint side, as it sounds like an API that should return an error, but it doesn't. It seems to panic if something goes very wrong.

p4u:
Looks good to me, pretty smart. I have tested with dockersuit but for deeply test it we need to deploy it to the development network and run some benchmarking. I'll merge.

PR #191, opened 2021-05-20
Router/indexer: add getProcessMeta api

NateWilliams2:
Adds a `getProcessMetadata` api which contains only what's needed for the explorer's process list.


p4u:
LGTM. Remember to add the method to docs.vocdoni.io

Btw @emmdim I think the vocdoni.app UI could also use this method instead of `getProcessInfo()`, it's a kind of summary much faster to fetch and process.

p4u (approved):

PR #192, opened 2021-05-20
vochain: automatic start block for autostart processes

jordipainan:
Due to uncertainty of when an EVM based blockchain Tx
will be mined/sealed, let the user choose to put
`startBlock = 1` with the autostart mode in order
to start the process as soon as the tx is mined
on the EVM based blockchain plus some delay
required by the Vochain.

### DIFF
--- a/ethereum/ethevents/handlers.go
+++ b/ethereum/ethevents/handlers.go
@@ -69,6 +69,10 @@ var ethereumEventList = map[string]string{
 	"resultsAvailable": "0x5aff397e0d9bfad4e73dfd9c2da1d146ce7fe8cfd1a795dbf6b95b417236fa4c",
 }
 
+// Number of blocks to add to the process start block to start as soon as possible the process.
+// This start block addition takes into account that can be some delay on the Vochain commit operation.
+const ProcessStartBlockDelay = 10
+
 // HandleVochainOracle handles the events on ethereum for the Oracle.
 func HandleVochainOracle(ctx context.Context, event *ethtypes.Log, e *EthereumEvents) error {
 	tctx, cancel := context.WithTimeout(ctx, time.Minute)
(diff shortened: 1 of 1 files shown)


ethereum/ethevents/handlers.go
```diff
@@ -69,6 +69,10 @@ var ethereumEventList = map[string]string{
 	"resultsAvailable": "0x5aff397e0d9bfad4e73dfd9c2da1d146ce7fe8cfd1a795dbf6b95b417236fa4c",
 }
 
+// Number of blocks to add to the process start block to start as soon as possible the process.
+// This start block addition takes into account that can be some delay on the Vochain commit operation.
+const PROCESS_START_BLOCK_DELAY = 10
```
p4u: I think we use camelCase in general for the naming

ethereum/ethevents/handlers.go
```diff
@@ -109,6 +113,14 @@ func HandleVochainOracle(ctx context.Context, event *ethtypes.Log, e *EthereumEv
 			log.Infof("process already exists, skipping")
 			return nil
 		}
+
+		if processTx.Process.Status == models.ProcessStatus_READY &&
```
p4u, line 129: Add a comment here why are we doing that please. Also explain why we use `1` instead of `0`

ethereum/ethevents/handlers.go
```diff
@@ -109,6 +113,14 @@ func HandleVochainOracle(ctx context.Context, event *ethtypes.Log, e *EthereumEv
...
+			processTx.Process.StartBlock == 1 &&
+			processTx.Process.Mode.AutoStart {
+
```
p4u: no need to have an extra empty line here

p4u (approved):

PR #197, opened 2021-05-24
Scrutinizer: Fix ComputeResult race

NateWilliams2:
Adds a RWMutex for updating processes in the scrutinizer backend. 

Details:
- Remove vochain/scrutinizer/process.go lines 194-198- these lines are redundant, as this method calls `s.ComputeResult`, which executes the same exact lines of code.
- Add a RWMutex around process update database transactions. 
  - In `updateProcess()`, lines 296-345 of vochain/scrutinizer/process.go. This method loops through the `updateProcessPool`, updating each process. It's responsible for updating the process status to `RESULTS` when results are available. 
  - In `ComputeResult()`, lines 249-281 of vochain/scrutinizer/vote.go. This method fetches the process, then computes the results, then stores that process in the db again. This races with `updateProcess`, because it's possible for `updateProcess` to change the process status while `ComputeResult` computes the results. `ComputeResult` then re-stores the process (with the _old_ status), causing the status to be reverted. 

A RWMutex is used because ComputeResult may run concurrently over several processes, whereas updateProcess runs synchronously. ComputeResult therefore uses RLock/RUnlock, meaning many ComputeResult routines can update processes at the same time. updateProcess uses Lock/Unlock, meaning it can be the only routine updating a process and gets priority over results computation.

### DIFF
--- a/vochain/scrutinizer/vote.go
+++ b/vochain/scrutinizer/vote.go
@@ -254,31 +254,35 @@ func (s *Scrutinizer) ComputeResult(processID []byte) error {
 	if err != nil {
 		return fmt.Errorf("computeResult: cannot load processID %x from database: %w", processID, err)
 	}
+	// Compute the results
+	var results *indexertypes.Results
+	if results, err = s.computeFinalResults(p); err != nil {
+		return err
+	}
 
 	if err := s.db.UpdateMatching(&indexertypes.Process{},
 		badgerhold.Where("ID").Eq(processID),
 		func(record interface{}) error {
-			// Compute the results
-			var results *indexertypes.Results
-			if results, err = s.computeFinalResults(p); err != nil {
-				return err
+			update, ok := record.(*indexertypes.Process)
+			if !ok {
+				return fmt.Errorf("record isn't the correct type! Wanted Result, got %T", record)
 			}
-			p.HaveResults = true
-			p.FinalResults = true
-			s.addVoteLock.Lock()
-			defer s.addVoteLock.Unlock()
-			if err := s.db.Upsert(processID, results); err != nil {
-				return err
-			}
-
-			// Execute callbacks
-			for _, l := range s.eventListeners {
-				go l.OnComputeResults(results)
-			}
+			update.HaveResults = true
+			update.FinalResults = true
 			return nil
 		},
 	); err != nil {
 		return fmt.Errorf("computeResult: cannot update processID %x: %w, ", processID, err)
+	}
+	s.addVoteLock.Lock()
+	defer s.addVoteLock.Unlock()
+	if err := s.db.Upsert(processID, results); err != nil {
+		return err
+	}
+
+	// Execute callbacks
+	for _, l := range s.eventListeners {
+		go l.OnComputeResults(results)
 	}
 	return nil
 }
--- a/vochain/scrutinizer/vote.go
+++ b/vochain/scrutinizer/vote.go
@@ -255,25 +255,30 @@ func (s *Scrutinizer) ComputeResult(processID []byte) error {
 		return fmt.Errorf("computeResult: cannot load processID %x from database: %w", processID, err)
 	}
 
-	// Compute the results
-	var results *indexertypes.Results
-	if results, err = s.computeFinalResults(p); err != nil {
-		return err
-	}
-	p.HaveResults = true
-	p.FinalResults = true
-	if err := s.db.Update(processID, p); err != nil {
-		return fmt.Errorf("computeResult: cannot update processID %x: %w, ", processID, err)
-	}
-	s.addVoteLock.Lock()
-	defer s.addVoteLock.Unlock()
-	if err := s.db.Upsert(processID, results); err != nil {
-		return err
-	}
+	if err := s.db.UpdateMatching(&indexertypes.Process{},
+		badgerhold.Where("ID").Eq(processID),
+		func(record interface{}) error {
+			// Compute the results
+			var results *indexertypes.Results
+			if results, err = s.computeFinalResults(p); err != nil {
+				return err
+			}
+			p.HaveResults = true
+			p.FinalResults = true
+			s.addVoteLock.Lock()
+			defer s.addVoteLock.Unlock()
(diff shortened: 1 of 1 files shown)


NateWilliams2:
After writing this, I realized there may be a much more simple solution-

could `p, err := s.ProcessInfo(processID)` in ComputeResult be replaced with `p, err := s.App.State.Process(pid, false)`, thereby getting the status directly from the vochain state? then maybe updateProcess would not be needed on ProcessResults. I'll test this solution as well.

vochain/scrutinizer/vote.go
```diff
@@ -265,6 +266,7 @@ func (s *Scrutinizer) ComputeResult(processID []byte) error {
 	if err := s.db.Update(processID, p); err != nil {
 		return fmt.Errorf("computeResult: cannot update processID %x: %w, ", processID, err)
 	}
+	s.processUpdateMutex.RUnlock()
```
p4u: This must be `defer s.processUpdateMutex.RUnlock` on line 253, since there is a return in the middle

vochain/scrutinizer/process.go
```diff
@@ -333,13 +328,17 @@ func (s *Scrutinizer) updateProcess(pid []byte) error {
...
 		if err := s.db.UpdateMatching(&indexertypes.Process{},
 			badgerhold.Where(badgerhold.Key).Eq(pid), updateFunc); err != nil {
 			if strings.Contains(err.Error(), kvErrorStringForRetry) {
+				s.processUpdateMutex.Unlock()
 				continue
 			}
+			s.processUpdateMutex.Unlock()
 			return err
 		}
+		s.processUpdateMutex.Unlock()
```
p4u: Use defer here too.

vochain/scrutinizer/process.go
```diff
@@ -333,13 +328,17 @@ func (s *Scrutinizer) updateProcess(pid []byte) error {
 	}
 	// Retry if error is "Transaction Conflict"
 	for {
+		s.processUpdateMutex.Lock()
```
p4u: Why the Lock is inside the loop and not outside?

p4u (changes requested):

p4u:
> After writing this, I realized there may be a much more simple solution-
> 
> could `p, err := s.ProcessInfo(processID)` in ComputeResult be replaced with `p, err := s.App.State.Process(pid, false)`, thereby getting the status directly from the vochain state? then maybe updateProcess would not be needed on ProcessResults. I'll test this solution as well.

The only moment that the status of a process should be set to RESULTS is when the an Oracle transaction is executed. If not set by updateProcess when OnProcessresults() triggers it, where would you set it?

p4u:
I would propose another solution. On ComputeResults() you could use `s.db.ForEach(... func(p) { .... })` instead of Update(), ignoring then the process fetched on line 253. The transaction should run atomic, so the process status won't be touched.

p4u:
By the way, good catch!! I did not see that race at all.

NateWilliams2:
> I would propose another solution. On ComputeResults() you could use `s.db.ForEach(... func(p) { .... })` instead of Update(), ignoring then the process fetched on line 253. The transaction should run atomic, so the process status won't be touched.

This sounds good to me. Why not use UpdateMatching instead of ForEach?

vochain/scrutinizer/process.go
jordipainan: I proposed this because the for is executed per process

p4u:
> > I would propose another solution. On ComputeResults() you could use `s.db.ForEach(... func(p) { .... })` instead of Update(), ignoring then the process fetched on line 253. The transaction should run atomic, so the process status won't be touched.
> 
> This sounds good to me. Why not use UpdateMatching instead of ForEach?

I wanted to say UpdateMatching

vochain/scrutinizer/vote.go
```diff
@@ -255,25 +255,30 @@ func (s *Scrutinizer) ComputeResult(processID []byte) error {
...
-		return fmt.Errorf("computeResult: cannot update processID %x: %w, ", processID, err)
-	}
-	s.addVoteLock.Lock()
-	defer s.addVoteLock.Unlock()
-	if err := s.db.Upsert(processID, results); err != nil {
-		return err
-	}
+	if err := s.db.UpdateMatching(&indexertypes.Process{},
+		badgerhold.Where("ID").Eq(processID),
+		func(record interface{}) error {
```
p4u, line 265: The computeFinalResults operation is expensive and the one we don't want to have inside the atomic operation, so the code would be something like this:

```
diff --git a/vochain/scrutinizer/vote.go b/vochain/scrutinizer/vote.go
index 6ab2f373..4a39a5cc 100644
--- a/vochain/scrutinizer/vote.go
+++ b/vochain/scrutinizer/vote.go
@@ -260,9 +260,17 @@ func (s *Scrutinizer) ComputeResult(processID []byte) error {
        if results, err = s.computeFinalResults(p); err != nil {
                return err
        }
-       p.HaveResults = true
-       p.FinalResults = true
-       if err := s.db.Update(processID, p); err != nil {
+       if err := s.db.UpdateMatching(&indexertypes.Process{},
+               badgerhold.Where("ID").Eq(processID),
+               func(record interface{}) error {
+                       update, ok := record.(*indexertypes.Process)
+                       if !ok {
+                               return fmt.Errorf("record isn't the correct type! Wanted Result, got %T", record)
+                       }
+                       update.HaveResults = true
+                       update.FinalResults = true
+                       return nil
+               }); err != nil {
                return fmt.Errorf("computeResult: cannot update processID %x: %w, ", processID, err)
        }
        s.addVoteLock.Lock()
```

vochain/scrutinizer/process.go
p4u: Does not look correct to me, there is no need for getting and releasing the lock on each iteration. Anyway this is already outated code :)

p4u:
Shall we merge that already @NateWilliams2 ?

NateWilliams2:
@p4u yes, I don't have permissions but go ahead!

PR #198, opened 2021-05-24
update xdai ENS Registry contract addr

jordipainan:


jordipainan:
@p4u Go ahead, the domains for stg env are already updated

p4u (approved):

PR #204, opened 2021-05-25
Fix process status race (again)

NateWilliams2:


vochain/scrutinizer/vote.go
```diff
@@ -254,15 +254,24 @@ func (s *Scrutinizer) ComputeResult(processID []byte) error {
...
-
 	// Compute the results
 	var results *indexertypes.Results
 	if results, err = s.computeFinalResults(p); err != nil {
 		return err
 	}
-	p.HaveResults = true
-	p.FinalResults = true
-	if err := s.db.Update(processID, p); err != nil {
+
```
p4u, line 262: Can you add a comment here exlaining the RACE and why we need to use updateMatching? So future developments not get into the same issue

p4u (approved):

PR #205, opened 2021-05-25
Rename getProcessMeta -> getProcessSummary

### LINKED ISSUE #203: feature: rename and extend getProcessMeta()
**Describe the feature**
In order to make getProcessMeta() useful for the client, we'd need to extend the fields to:

- metadata
- startBlock
- endBlock
- sourceNetworkId
- entityIndex

Also rename getProcessMeta to getProcessHeader

**Motivation**
Make client more efficient.

NateWilliams2:
add fields:
- metadata
- startBlock
- blockCount
- sourceNetworkID
- entityIndex

rename:
- state -> status

modify:
- type from a string to a ProcessType

close https://github.com/vocdoni/vocdoni-node/issues/203

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -178,7 +178,6 @@ type ResponseMessage struct {
 type MetaResponse struct {
 	APIList              []string                         `json:"apiList,omitempty"`
 	Block                *indexertypes.BlockMetadata      `json:"block,omitempty"`
-	BlockCount           uint32                           `json:"blockCount,omitempty"`
 	BlockList            []*indexertypes.BlockMetadata    `json:"blockList,omitempty"`
 	BlockTime            *[5]int32                        `json:"blockTime,omitempty"`
 	BlockTimestamp       int32                            `json:"blockTimestamp,omitempty"`
@@ -194,7 +193,6 @@ type MetaResponse struct {
 	EncryptionPublicKeys []Key                            `json:"encryptionPubKeys,omitempty"`
 	EntityID             string                           `json:"entityId,omitempty"`
 	EntityIDs            []string                         `json:"entityIds,omitempty"`
-	EntityIndex          uint32                           `json:"entityIndex,omitempty"`
 	Envelope             *indexertypes.EnvelopePackage    `json:"envelope,omitempty"`
 	Envelopes            []*indexertypes.EnvelopeMetadata `json:"envelopes,omitempty"`
 	Files                []byte                           `json:"files,omitempty"`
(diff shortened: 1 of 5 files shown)


api/types.go
```diff
@@ -241,6 +246,14 @@ func (r *MetaResponse) SetError(v interface{}) {
 	r.Message = fmt.Sprintf("%s", v)
 }
 
+type ProcessType struct {
```
p4u: This is defined in the protobuf models, we should use that instead I think

client/api.go
```diff
@@ -68,9 +68,9 @@ func (c *Client) GetResults(pid []byte) ([][]string, string, bool, error) {
 		return nil, "", false, fmt.Errorf("cannot get results: (%s)", resp.Message)
 	}
 	if resp.Message == "no results yet" {
-		return nil, resp.State, false, nil
+		return nil, resp.Status, false, nil
```
p4u: Let's not change this now, it's breaking.

router/indexerCallbacks.go
```diff
@@ -195,22 +195,18 @@ func (r *Router) getProcessMeta(request RouterRequest) {
 		return
 	}
 
-	if procInfo.Envelope.Anonymous {
-		response.Type = "anonymous"
-	} else {
-		response.Type = "poll"
+	response.Type = api.ProcessType{
```
p4u: Whi not just returning procInfo.Envelope ??

router/indexerCallbacks.go
```diff
@@ -195,22 +195,18 @@ func (r *Router) getProcessMeta(request RouterRequest) {
...
-	} else {
-		response.Type += " open"
-	}
-	if procInfo.Envelope.Serial {
-		response.Type += " serial"
-	} else {
-		response.Type += " single"
-	}
-	response.State = models.ProcessStatus(procInfo.Status).String()
+	response.EntityIndex = procInfo.EntityIndex
```
p4u: I think instead of adding all these fields as new fields, we should use a struct, such as "ProcessHeader". Don't we have that already somewhere?

router/voteCallbacks.go
```diff
@@ -312,22 +312,14 @@ func (r *Router) getResults(request RouterRequest) {
...
-	}
-	response.State = models.ProcessStatus(procInfo.Status).String()
+	response.Type = api.ProcessType{
+		Anonymous:      procInfo.Envelope.Anonymous,
+		Encrypted:      procInfo.Envelope.EncryptedVotes,
+		CostFromWeight: procInfo.Envelope.CostFromWeight,
+		Serial:         procInfo.Envelope.Serial,
+		UniqueValues:   procInfo.Envelope.UniqueValues,
+	}
+	response.Status = models.ProcessStatus(procInfo.Status).String()
```
p4u: Not today

p4u (changes requested):

router/indexerCallbacks.go
```diff
@@ -195,23 +195,14 @@ func (r *Router) getProcessMeta(request RouterRequest) {
...
-		response.Type += " open"
-	}
-	if procInfo.Envelope.Serial {
-		response.Type += " serial"
-	} else {
-		response.Type += " single"
-	}
-	response.State = models.ProcessStatus(procInfo.Status).String()
-	response.EntityID = hex.EncodeToString(procInfo.EntityID)
+	response.ProcessSummary.Type = procInfo.Envelope
```
p4u:
```
response.ProcessSumarry = &ProcessSummary{
Type: procInfo.Envelope,
....
}
```

router/indexerCallbacks.go
```diff
@@ -223,7 +217,7 @@ func (r *Router) getProcessMeta(request RouterRequest) {
 		return
 	}
 	votes := uint32(eh)
-	response.Height = &votes
+	response.ProcessSummary.EnvelopeHeight = &votes
```
p4u: Lets add this to the ProcessSummary as the other fields.

p4u (approved):

PR #207, opened 2021-05-27
scrutinizer: retry if transaction conflict on ComputeResults

p4u:

### DIFF
--- a/vochain/scrutinizer/vote.go
+++ b/vochain/scrutinizer/vote.go
@@ -264,19 +264,32 @@ func (s *Scrutinizer) ComputeResult(processID []byte) error {
 	// updateProcess. If we fetch the process, compute the results, then update the process,
 	// ComputeResult can override the process status set by updateProcess. UpdateMatching
 	// gets rid of the time between fetching and updating the process, eliminating this race.
-	if err := s.db.UpdateMatching(&indexertypes.Process{},
-		badgerhold.Where("ID").Eq(processID),
-		func(record interface{}) error {
-			update, ok := record.(*indexertypes.Process)
-			if !ok {
-				return fmt.Errorf("record isn't the correct type! Wanted Result, got %T", record)
+	maxTries := 1000
+	for {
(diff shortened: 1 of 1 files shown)


mvdan:
How can we get to these conflicts, and why is sleeping a bit and retrying solving it? I'd imagine there's an underlying problem we need to fix, rather than adding sleeps.

p4u:
Well, it's like a "time based random access" mechanism, if other transactions are happening in the same time, soon or later all of them will probably success. We might add some random time field, such as `time.Milisecond() * 20 + rand(100)`

The underlaying problem is not a real problem since Transaction Conflicts are expected and normal in Badger. We might use a Lock to avoid this but IMHO the loop works too and might be even faster.

mvdan:
Huh, I find it really weird that it's expected to have transaction conflicts with badger. Should we not wrap every badger transaction with a retry loop, then?

p4u:
Well, that's a characteristic of so called [ACID transactions](https://dgraph.io/blog/post/badger-txn/) 

> Finally, when Commit is called, we send our read fingerprints to Oracle and ask for a commit timestamp. Oracle will check if the read rows have since been updated (row commit ts > txn read ts). If so, it would return a conflict, and the transaction would abort. This is indicated by Commit method returning an ErrConflict error.

See also this issue https://github.com/dgraph-io/badger/issues/539

The Badger founder states: 

> I don't know your code, but if you're using db.Update, running that in a loop would do.

p4u:
> Huh, I find it really weird that it's expected to have transaction conflicts with badger. Should we not wrap every badger transaction with a retry loop, then?

Yes, we should probably wrap all badger transactions in order to retry in a loop or to use mutex to avoid conflicts if we want to stay 100% sure it does not happend. However conflicts only appear on those db transactions which are very concurrent.

p4u:
I'll merge that since we need to fix this issue ASAP.

Afterwars we can see what's the best approach to solve this kind of issues.

PR #208, opened 2021-05-27
ethereum/ethevents: go back to a map+mutex

mvdan:
(see commit message)

### DIFF
--- a/ethereum/ethevents/events.go
+++ b/ethereum/ethevents/events.go
@@ -65,7 +65,7 @@ type EthereumEvents struct {
 	ContractsInfo map[string]*ethereumhandler.EthereumContract
 }
 
-type logEvent struct {
+type timedEvent struct {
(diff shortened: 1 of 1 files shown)


ethereum/ethevents/events.go
```diff
@@ -326,76 +333,66 @@ func (ev *EthereumEvents) processEventLogsFromTo(ctx context.Context,
...
+				ev.EventProcessor.add(&event)
 			}
 		}
 	}()
 
 	for {
+		// TODO(mvdan): Perhaps refactor this so we don't need a sleep.
+		// Maybe use a queue-like structure sorted by added time,
+		// and separately mark the removed events to ignore them.
 		time.Sleep(2 * time.Second)
```
p4u, line 391: since this next is now safe we could reduce the sleep to something like `time.Milisecond*200`

PR #209, opened 2021-05-27
vochain: allow signatures over base64

p4u:


mvdan:
I don't see anything wrong with this change, but I can't really understand it without an explanation for the reason in the commit message :)

p4u:
That's a bit complex. We believe there are some Web based signers (such as WalletConnect) which are failing to sign raw byte payloads. This commit introduces an optional flag on the SignedTx protobuf message, that if true, the signature over the whole transaction is done on the `base64(bytesPayload)` instead of `bytesPayload`.

However that's something we have not yet tested so it's just a testing commit for now.

PR #210, opened 2021-05-28
Scrutinizer: Don't use badgerhold.Key

NateWilliams2:
the use of Badgerhold.Key can lead to really bad performance issues. This PR replaces Badgerhold.Key with the key itself (eg. "ID") for all badgerhold queries.

### DIFF
--- a/vochain/scrutinizer/process.go
+++ b/vochain/scrutinizer/process.go
@@ -19,7 +19,7 @@ import (
 // ProcessInfo returns the available information regarding an election process id
 func (s *Scrutinizer) ProcessInfo(pid []byte) (*indexertypes.Process, error) {
 	proc := &indexertypes.Process{}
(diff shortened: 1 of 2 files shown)


p4u:
That was changed in purpose when @mvdan introduced this commmit on badgerhold https://github.com/vocdoni/badgerhold/commit/2d704df3456f01f13f7dfcd8a3ab1c909a19281d

So AFAIK only if we use `badgerhold.Key` we will be using that workaround

mvdan:
Right, FindOne queries are linear unless we use Key with my hack on badgerhold. How did you find that the performance was really bad with Key?

I could make badgerhold not linear with explicit key strings, but that would be even more hacks on badgerhold, so I'd like to avoid that.

NateWilliams2:
I see, so if anything, we should make sure that FindOne() is using badgerhold.Key wherever possible.

PR #213, opened 2021-05-31
Feature/faster envelope height

p4u:
see commit message

### DIFF
--- a/vochain/scrutinizer/vote.go
+++ b/vochain/scrutinizer/vote.go
@@ -38,6 +38,8 @@ const kvErrorStringForRetry = "Transaction Conflict"
 // Getindexertypes.VoteReference gets the reference for an AddVote transaction.
 // This reference can then be used to fetch the vote transaction directly from the BlockStore.
 func (s *Scrutinizer) GetEnvelopeReference(nullifier []byte) (*indexertypes.VoteReference, error) {
+	startTime := time.Now()
+	defer func() { log.Debugf("GetEnvelopeReference took %s", time.Since(startTime)) }()
 	txRef := &indexertypes.VoteReference{}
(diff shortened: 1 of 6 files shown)


vochain/scrutinizer/vote.go
```diff
@@ -220,34 +220,21 @@ func (s *Scrutinizer) GetEnvelopeHeight(processID []byte) (uint64, error) {
...
-		panic(fmt.Sprintf("unexpected type: %T", val))
+	// Check if the envelope height is cached
+	val := s.envelopeHeightCache.Get(string(processID))
+	if val != nil {
+		return val.(uint64), nil
+	}
+	// If not cached, make the expensive query
+	results := &indexertypes.Results{}
+	if err := s.db.FindOne(results, badgerhold.Where(badgerhold.Key).Eq(processID)); err != nil {
+		return 0, err
```
NateWilliams2, line 231: @p4u it doesn't make sense to me that we would only be able to fetch envelope height from a process that has results. For an encrypted process that hasn't completed, for example, will results be available to fetch the envelope height?

In either case I think we should check for ErrNotFound here and either manually count the envelopes, return 0, or return an error message that gives more context

p4u: I don't follow you @NateWilliams2, if ErrNotFound a 0 and the error are returned

Are you saying that Encrypted processes do not have a Results entry? In that case yes, we should create also a results entry for Encrypted results and count the envelopes. But I think (need to check the code to be sure) that results are created since we need to count the Weight too.

NateWilliams2: @p4u right, i guess what I'm describing has more to do with how this method is used. I brought this up because of the error on friday we had with getProcessSummary returning an ErrNotFound due to the results not being available. It seems like this should be the expected behavior though, so never mind :)

PR #215, opened 2021-06-01
Tx indexer

NateWilliams2:
Indexes transactions on the scrutinizer.

Store `TxReference`s containing each tx's height, blockHeight, and txIndex. 
Also provides the getTxByHeight api, allowing transactions to be queried by transaction height, not just block/txIndex

### DIFF
--- a/vochain/app.go
+++ b/vochain/app.go
@@ -351,9 +351,7 @@ func (app *BaseApplication) DeliverTx(req abcitypes.RequestDeliverTx) abcitypes.
 			return abcitypes.ResponseDeliverTx{Code: 1, Data: []byte(err.Error())}
 		}
 		for _, e := range app.State.eventListeners {
-			if err := e.OnNewTx(app.Height()+1, app.State.TxCounter()); err != nil {
-				log.Warnf("deliverTx: %v", err)
-			}
+			e.OnNewTx(app.Height()+1, app.State.TxCounter())
 		}
 	} else {
 		return abcitypes.ResponseDeliverTx{Code: 1, Data: []byte(err.Error())}
--- a/vochain/app.go
+++ b/vochain/app.go
@@ -353,7 +353,7 @@ func (app *BaseApplication) DeliverTx(req abcitypes.RequestDeliverTx) abcitypes.
 		for _, e := range app.State.eventListeners {
 			err := e.OnNewTx(app.State.height+1, uint32(app.State.TxCounter()))
 			if err != nil {
-				log.Errorf("DeliverTx: %v", err)
+				log.Errorf("deliverTx: %v", err)
 			}
 		}
 	} else {
--- a/vochain/app.go
+++ b/vochain/app.go
@@ -351,9 +351,8 @@ func (app *BaseApplication) DeliverTx(req abcitypes.RequestDeliverTx) abcitypes.
 			return abcitypes.ResponseDeliverTx{Code: 1, Data: []byte(err.Error())}
 		}
 		for _, e := range app.State.eventListeners {
-			err := e.OnNewTx(app.State.height+1, uint32(app.State.TxCounter()))
-			if err != nil {
-				log.Errorf("deliverTx: %v", err)
+			if err := e.OnNewTx(app.Height()+1, app.State.TxCounter()); err != nil {
+				log.Warnf("deliverTx: %v", err)
 			}
 		}
 	} else {
--- a/vochain/app.go
+++ b/vochain/app.go
@@ -350,6 +350,9 @@ func (app *BaseApplication) DeliverTx(req abcitypes.RequestDeliverTx) abcitypes.
 			log.Debugf("rejected tx: %v", err)
 			return abcitypes.ResponseDeliverTx{Code: 1, Data: []byte(err.Error())}
 		}
+		for _, e := range app.State.eventListeners {
+			e.OnNewTx(uint32(app.State.TxCounter()), app.State.height)
+		}
 	} else {
 		return abcitypes.ResponseDeliverTx{Code: 1, Data: []byte(err.Error())}
 	}
--- a/vochain/app.go
+++ b/vochain/app.go
@@ -351,7 +351,10 @@ func (app *BaseApplication) DeliverTx(req abcitypes.RequestDeliverTx) abcitypes.
 			return abcitypes.ResponseDeliverTx{Code: 1, Data: []byte(err.Error())}
 		}
 		for _, e := range app.State.eventListeners {
-			e.OnNewTx(app.State.height+1, uint32(app.State.TxCounter()))
+			err := e.OnNewTx(app.State.height+1, uint32(app.State.TxCounter()))
+			if err != nil {
+				log.Errorf("DeliverTx: %v", err)
+			}
 		}
 	} else {
 		return abcitypes.ResponseDeliverTx{Code: 1, Data: []byte(err.Error())}
--- a/vochain/app.go
+++ b/vochain/app.go
@@ -351,7 +351,7 @@ func (app *BaseApplication) DeliverTx(req abcitypes.RequestDeliverTx) abcitypes.
 			return abcitypes.ResponseDeliverTx{Code: 1, Data: []byte(err.Error())}
 		}
 		for _, e := range app.State.eventListeners {
-			e.OnNewTx(uint32(app.State.TxCounter()), app.State.height)
+			e.OnNewTx(app.State.height+1, uint32(app.State.TxCounter()))
 		}
 	} else {
 		return abcitypes.ResponseDeliverTx{Code: 1, Data: []byte(err.Error())}
--- a/vochain/censusdownloader/censusdownloader.go
+++ b/vochain/censusdownloader/censusdownloader.go
@@ -81,7 +81,7 @@ func (c *CensusDownloader) OnProcess(pid, eid []byte, censusRoot, censusURI stri
 // NOT USED but required for implementing the interface
 func (c *CensusDownloader) OnCancel(pid []byte, txindex int32)                       {}
 func (c *CensusDownloader) OnVote(v *models.Vote, txindex int32)                     {}
-func (c *CensusDownloader) OnNewTx(blockHeight, txIndex uint32)                      {}
+func (c *CensusDownloader) OnNewTx(blockHeight, txIndex uint32) error                { return nil }
 func (c *CensusDownloader) OnProcessKeys(pid []byte, pub, com string, txindex int32) {}
 func (c *CensusDownloader) OnRevealKeys(pid []byte, priv, rev string, txindex int32) {}
 func (c *CensusDownloader) OnProcessStatusChange(pid []byte,
--- a/vochain/censusdownloader/censusdownloader.go
+++ b/vochain/censusdownloader/censusdownloader.go
@@ -81,7 +81,7 @@ func (c *CensusDownloader) OnProcess(pid, eid []byte, censusRoot, censusURI stri
 // NOT USED but required for implementing the interface
 func (c *CensusDownloader) OnCancel(pid []byte, txindex int32)                       {}
 func (c *CensusDownloader) OnVote(v *models.Vote, txindex int32)                     {}
-func (c *CensusDownloader) OnNewTx(blockHeight uint32, txIndex int32) error          { return nil }
+func (c *CensusDownloader) OnNewTx(blockHeight uint32, txIndex int32)                {}
 func (c *CensusDownloader) OnProcessKeys(pid []byte, pub, com string, txindex int32) {}
 func (c *CensusDownloader) OnRevealKeys(pid []byte, priv, rev string, txindex int32) {}
 func (c *CensusDownloader) OnProcessStatusChange(pid []byte,
--- a/vochain/censusdownloader/censusdownloader.go
+++ b/vochain/censusdownloader/censusdownloader.go
@@ -81,7 +81,7 @@ func (c *CensusDownloader) OnProcess(pid, eid []byte, censusRoot, censusURI stri
 // NOT USED but required for implementing the interface
 func (c *CensusDownloader) OnCancel(pid []byte, txindex int32)                       {}
 func (c *CensusDownloader) OnVote(v *models.Vote, txindex int32)                     {}
-func (c *CensusDownloader) OnNewTx(blockHeight, txIndex uint32) error                { return nil }
+func (c *CensusDownloader) OnNewTx(blockHeight uint32, txIndex int32) error          { return nil }
 func (c *CensusDownloader) OnProcessKeys(pid []byte, pub, com string, txindex int32) {}
 func (c *CensusDownloader) OnRevealKeys(pid []byte, priv, rev string, txindex int32) {}
 func (c *CensusDownloader) OnProcessStatusChange(pid []byte,
--- a/vochain/censusdownloader/censusdownloader.go
+++ b/vochain/censusdownloader/censusdownloader.go
@@ -81,6 +81,7 @@ func (c *CensusDownloader) OnProcess(pid, eid []byte, censusRoot, censusURI stri
 // NOT USED but required for implementing the interface
 func (c *CensusDownloader) OnCancel(pid []byte, txindex int32)                       {}
 func (c *CensusDownloader) OnVote(v *models.Vote, txindex int32)                     {}
+func (c *CensusDownloader) OnNewTx(blockHeight, txIndex uint32)                      {}
 func (c *CensusDownloader) OnProcessKeys(pid []byte, pub, com string, txindex int32) {}
 func (c *CensusDownloader) OnRevealKeys(pid []byte, priv, rev string, txindex int32) {}
 func (c *CensusDownloader) OnProcessStatusChange(pid []byte,
--- a/vochain/scrutinizer/indexertypes/types.go
+++ b/vochain/scrutinizer/indexertypes/types.go
@@ -14,13 +14,13 @@ import (
 
 const (
 	// CountStore_Entities is the key for the entity count db reference
-	CountStore_Entities = 0
+	CountStore_Entities uint8 = 0
 	// CountStore_Processes is the key for the process count db reference
-	CountStore_Processes = 1
+	CountStore_Processes uint8 = 1
 	// CountStore_Envelopes is the key for the envelope count db reference
-	CountStore_Envelopes = 2
+	CountStore_Envelopes uint8 = 2
 	// CountStore_Transactions is the key for the transaction count db reference
-	CountStore_Transactions = 3
+	CountStore_Transactions uint8 = 3
 )
 
 // CountStore holds the count of envelopes, processes, entities, or transactions
--- a/vochain/scrutinizer/indexertypes/types.go
+++ b/vochain/scrutinizer/indexertypes/types.go
@@ -13,10 +13,14 @@ import (
 )
 
 const (
-	Entities     = 0
-	Processes    = 1
-	Envelopes    = 2
-	Transactions = 3
+	// CountStore_Entities is the key for the entity count db reference
+	CountStore_Entities = 0
+	// CountStore_Processes is the key for the process count db reference
+	CountStore_Processes = 1
+	// CountStore_Envelopes is the key for the envelope count db reference
+	CountStore_Envelopes = 2
+	// CountStore_Transactions is the key for the transaction count db reference
+	CountStore_Transactions = 3
 )
 
 // CountStore holds the count of envelopes, processes, entities, or transactions
--- a/vochain/scrutinizer/indexertypes/types.go
+++ b/vochain/scrutinizer/indexertypes/types.go
@@ -13,14 +13,14 @@ import (
 )
 
 const (
-	// CountStore_Entities is the key for the entity count db reference
-	CountStore_Entities uint8 = 0
-	// CountStore_Processes is the key for the process count db reference
-	CountStore_Processes uint8 = 1
-	// CountStore_Envelopes is the key for the envelope count db reference
-	CountStore_Envelopes uint8 = 2
-	// CountStore_Transactions is the key for the transaction count db reference
-	CountStore_Transactions uint8 = 3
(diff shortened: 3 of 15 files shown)


vochain/app.go
```diff
@@ -350,6 +350,12 @@ func (app *BaseApplication) DeliverTx(req abcitypes.RequestDeliverTx) abcitypes.
 			log.Debugf("rejected tx: %v", err)
 			return abcitypes.ResponseDeliverTx{Code: 1, Data: []byte(err.Error())}
 		}
+		for _, e := range app.State.eventListeners {
+			err := e.OnNewTx(app.State.height+1, uint32(app.State.TxCounter()))
```
p4u: use `app.Height()` instead

vochain/app.go
```diff
@@ -350,6 +350,12 @@ func (app *BaseApplication) DeliverTx(req abcitypes.RequestDeliverTx) abcitypes.
...
+			if err != nil {
```
p4u: `if err := e.OnNewTx()...`

vochain/app.go
```diff
@@ -350,6 +350,12 @@ func (app *BaseApplication) DeliverTx(req abcitypes.RequestDeliverTx) abcitypes.
...
+				log.Errorf("deliverTx: %v", err)
```
p4u: I think that should be a `Warnf` since the error is from a eventlistener

vochain/censusdownloader/censusdownloader.go
```diff
@@ -81,6 +81,7 @@ func (c *CensusDownloader) OnProcess(pid, eid []byte, censusRoot, censusURI stri
 // NOT USED but required for implementing the interface
 func (c *CensusDownloader) OnCancel(pid []byte, txindex int32)                       {}
 func (c *CensusDownloader) OnVote(v *models.Vote, txindex int32)                     {}
+func (c *CensusDownloader) OnNewTx(blockHeight, txIndex uint32) error                { return nil }
```
p4u: I think its more consistent to use the same type for all `txIndex`. I think all of them should be uint32 but if you don't want to change so many code (its fine, we can do it afterwars in another commit) just use int32

vochain/scrutinizer/indexertypes/types.go
```diff
@@ -12,6 +12,19 @@ import (
...
+const (
+	Entities     = 0
+	Processes    = 1
+	Envelopes    = 2
+	Transactions = 3
+)
+
+// CountStore holds the count of envelopes, processes, entities, or transactions
+type CountStore struct {
+	Type  int `badgerholdKey:"Type"`
```
p4u: uint8 ?

vochain/scrutinizer/scrutinizer.go
```diff
@@ -97,26 +97,44 @@ func NewScrutinizer(dbPath string, app *vochain.BaseApplication) (*Scrutinizer,
 		return nil, err
 	}
 	startTime := time.Now()
-	envelopes, err := s.db.Count(&indexertypes.VoteReference{}, &badgerhold.Query{})
+
+	txCountStore := new(indexertypes.CountStore)
+	err = s.db.FindOne(txCountStore,
+		badgerhold.Where(badgerhold.Key).Eq(indexertypes.Transactions))
+	if err != nil {
```
p4u, line 103: `if err = s.db...; err != nil {}`

vochain/scrutinizer/transaction.go
```diff
@@ -0,0 +1,38 @@
...
+	txReference := &indexertypes.TxReference{}
+	err := s.db.FindOne(txReference, badgerhold.Where(badgerhold.Key).Eq(height))
+	if err != nil {
+		return nil, fmt.Errorf("tx height %d not found: %v", height, err)
+	}
+	return txReference, nil
+}
+
+// OnNewTx stores the transaction reference in the indexer database
+func (s *Scrutinizer) OnNewTx(blockHeight, txIndex uint32) error {
```
p4u: We cannot store data on OnNewTx since its executed in deliverTx where the block can still be discarted. So you need to queue it and store the data on `Commit` instead

vochain/scrutinizer/transaction.go
```diff
@@ -0,0 +1,38 @@
...
+	txCount := atomic.AddUint64(&s.countTotalTransactions, 1)
+	err := s.db.Insert(txCount, &indexertypes.TxReference{
+		Index:        txCount,
+		BlockHeight:  blockHeight,
+		TxBlockIndex: txIndex,
+	})
+	if err != nil {
```
p4u: `if err := s.db.Insert...`

p4u (changes requested):

p4u:
@mvdan please also review

vochain/scrutinizer/scrutinizer_test.go
```diff
@@ -953,3 +953,28 @@ func TestCountVotes(t *testing.T) {
...
+			qt.Assert(t, err, qt.IsNil)
+		}
+	}
+
+	for i := 0; i < 10; i++ {
+		for j := 0; j < 10; j++ {
+			ref, err := sc.GetTxReference(uint64(i*10 + j + 1))
+			qt.Assert(t, err, qt.IsNil)
+			qt.Assert(t, ref.BlockHeight, qt.CmpEquals(), uint32(i))
+			qt.Assert(t, ref.TxBlockIndex, qt.CmpEquals(), uint32(j))
```
mvdan: you can just use qt.Equals here, since these are just integers.

NateWilliams2:
@p4u I addressed the comments. Pending is to change all tx indexes to uint32

vochain/scrutinizer/indexertypes/types.go
```diff
@@ -12,6 +12,19 @@ import (
 	"go.vocdoni.io/proto/build/go/models"
 )
 
+const (
```
p4u, line 15: These types should have a prefix and also a comment, else its hard to understand how to use them

vochain/scrutinizer/process.go
```diff
@@ -153,20 +153,20 @@ func (s *Scrutinizer) ProcessList(entityID []byte,
 	return procs, err
 }
 
-// ProcessCount return the number of processes indexed
-func (s *Scrutinizer) ProcessCount(entityID []byte) int64 {
+// ProcessCount returns the number of processes indexed
+func (s *Scrutinizer) ProcessCount(entityID []byte) uint64 {
```
p4u, line 156: We should try to unify integer types, I think for ProcessCount() uint32 should mot more than enough

vochain/scrutinizer/scrutinizer.go
```diff
@@ -97,26 +99,40 @@ func NewScrutinizer(dbPath string, app *vochain.BaseApplication) (*Scrutinizer,
...
 	}
 	startTime := time.Now()
-	envelopes, err := s.db.Count(&indexertypes.VoteReference{}, &badgerhold.Query{})
-	if err != nil {
-		return nil, fmt.Errorf("could not count the total envelopes: %w", err)
+
+	txCountStore := new(indexertypes.CountStore)
+	if err = s.db.FindOne(txCountStore,
+		badgerhold.Where(badgerhold.Key).Eq(indexertypes.Transactions)); err != nil {
+		log.Warnf("could not get the transaction count: %v", err)
```
p4u, line 127: If the counters cannot be recovered from the KV, we should calculate them (using badgerhold.Count()) and store

vochain/scrutinizer/scrutinizer.go
```diff
@@ -97,26 +99,40 @@ func NewScrutinizer(dbPath string, app *vochain.BaseApplication) (*Scrutinizer,
...
+		badgerhold.Where(badgerhold.Key).Eq(indexertypes.Processes)); err != nil {
+		log.Warnf("could not get the process count: %v", err)
 	}
+	entityCountStore := new(indexertypes.CountStore)
+	if err = s.db.FindOne(entityCountStore,
+		badgerhold.Where(badgerhold.Key).Eq(indexertypes.Entities)); err != nil {
+		log.Warnf("could not get the entity count: %v", err)
+	}
+
+	s.countTotalTransactions = txCountStore.Count
```
p4u: Since now we have the counters stored on the persistent KV, I don't think its worth to keep these in-memory caches.

vochain/scrutinizer/scrutinizer.go
```diff
@@ -238,6 +254,16 @@ func (s *Scrutinizer) Commit(height uint32) error {
 		}
 	}
 
+	// Index new transactions
+	for _, tx := range s.newTxPool {
```
p4u: It would be great if we can run this async, but we should be sure that the code is safe. Also I think we should abstract this code in a function (and just call something like `go updateTxIndex(s.newTxPool)`)

p4u (changes requested):

vochain/scrutinizer/process.go
NateWilliams2: should I change the CountStore Count type to be uint32? This would include processes, entitites, envelopes, txs

vochain/scrutinizer/scrutinizer.go
NateWilliams2: Are you proposing to have a db transaction every time we want to access or update these values, i.e. every block?

vochain/scrutinizer/scrutinizer.go
NateWilliams2: I've updated the code to run async, and I've also tested it with the race flag a few times without any races.

NateWilliams2: This update actually was causing many database transaction conflicts. Rather than handling them only for this case, I submitted https://github.com/vocdoni/vocdoni-node/pull/216 which handles transaction conflicts for most update functions. I've rebased this branch on that branch locally and verified that it fixes the problem.

The other issue with running this async is that there's no guarantee that txs will all be indexed after commit is called. For testing, this requires adding a time delay after calling Commit().

vochain/scrutinizer/process.go
p4u: You already changed it from int64 to uint64, no?

vochain/scrutinizer/process.go
```diff
@@ -266,8 +273,15 @@ func (s *Scrutinizer) newEmptyProcess(pid []byte) error {
 	}
 	s.addVoteLock.Unlock()
 
-	// Increment the total process count cache
-	atomic.AddInt64(&s.countTotalProcesses, 1)
+	// Increment the total process count storage
+	s.db.UpdateMatching(&indexertypes.CountStore{}, badgerhold.Where(badgerhold.Key).Eq(indexertypes.CountStore_Processes), func(record interface{}) error {
```
p4u: break lines

vochain/scrutinizer/scrutinizer.go
```diff
@@ -97,33 +94,91 @@ func NewScrutinizer(dbPath string, app *vochain.BaseApplication) (*Scrutinizer,
...
+				log.Warnf("could not store envelope count: %v", err)
+			}
+		}
+	}
+	processCountStore := new(indexertypes.CountStore)
+	if err = s.db.Get(indexertypes.CountStore_Processes, processCountStore); err != nil {
+		log.Warnf("could not get the process count: %v", err)
+		count, err := s.db.Count(&indexertypes.Process{}, &badgerhold.Query{})
+		if err != nil {
+			log.Warnf("could not count total processes: %v", err)
```
p4u: That should either return an error or end up in a Fatal

p4u (changes requested):

vochain/scrutinizer/scrutinizer.go
p4u: Why do we have transaction conflicts here? I don't get it. We running the Inserts sync (not in parallel), so there should not be transaction conflicts.

NateWilliams2: The conflict is actually on storing the transaction count. There aren't conflicts within each indexNewTxs call, but there could be multiple calls to this method running at the same time (ie. block 200 has 25 transactions and is indexing them asynchronously, then block 201 has only 1 transaction and indexes it much faster,  catching up with the previous call to indexNewTxs). The conflicts actually stop if I test and only update the txCount if there are txs in the block (rather than updating the txCount from x to x). We should do this anyways to avoid making unnecessary transactions, but the point is a race between transactions is possible here.

NateWilliams2: This is also fixed by adding a mutex within this method, so only one block's transactions will be indexed at once

vochain/scrutinizer/process.go
NateWilliams2: yes, for processes. But now processes, entities, envelopes, and tx counts will all share the same type- CountStore. Do we want those all to be uint32?

vochain/scrutinizer/indexertypes/types.go
```diff
@@ -12,6 +12,23 @@ import (
...
+	// CountStore_Entities is the key for the entity count db reference
+	CountStore_Entities uint8 = 0
```
p4u: Use camel case please

vochain/scrutinizer/process.go
p4u: Better not make changes not related with this commit, so lets just change the minimum values. However uint32 makes sense in my opinion here

vochain/scrutinizer/process.go
```diff
@@ -198,8 +201,12 @@ func (s *Scrutinizer) EntityProcessCount(entityId []byte) (uint32, error) {
 }
 
 // EntityCount return the number of entities indexed by the scrutinizer
-func (s *Scrutinizer) EntityCount() int64 {
-	return atomic.LoadInt64(&s.countTotalEntities)
+func (s *Scrutinizer) EntityCount() uint64 {
+	entityCountStore := &indexertypes.CountStore{}
+	if err := s.db.Get(indexertypes.CountStore_Entities, entityCountStore); err != nil {
+		log.Warnf("could not get the process count: %v", err)
```
p4u: This should return 0

vochain/scrutinizer/scrutinizer.go
```diff
@@ -97,33 +96,90 @@ func NewScrutinizer(dbPath string, app *vochain.BaseApplication) (*Scrutinizer,
...
+// retrieveCounts returns a count for txs, envelopes, processes, and entities, in that order.
+// If no CountStore model is stored for the type, it counts all db entries of that type.
+func (s *Scrutinizer) retrieveCounts() (*indexertypes.CountStore,
+	*indexertypes.CountStore, *indexertypes.CountStore, *indexertypes.CountStore) {
+	var err error
+	txCountStore := new(indexertypes.CountStore)
+	if err = s.db.Get(indexertypes.CountStore_Transactions, txCountStore); err != nil {
+		log.Warnf("could not get the transaction count: %v", err)
+		count, err := s.db.Count(&indexertypes.TxReference{}, &badgerhold.Query{})
+		if err != nil {
```
p4u, line 129: This method should return an Error and instead of printing it and keep the execution we should return the error (and stop execution).

vochain/scrutinizer/scrutinizer.go
```diff
@@ -97,33 +96,90 @@ func NewScrutinizer(dbPath string, app *vochain.BaseApplication) (*Scrutinizer,
...
+			log.Warnf("could not count total transactions: %v", err)
+		} else {
```
p4u: No need for this else, instead we should check if the error is of type "key Not Found" (else return the error and stop the execution).

vochain/scrutinizer/scrutinizer.go
```diff
@@ -271,11 +330,18 @@ func (s *Scrutinizer) Commit(height uint32) error {
...
 		})
 		wg.Wait()
-		s.indexTxLock.Unlock()
+		s.voteTxLock.Unlock()
 		log.Infof("indexed %d new envelopes, took %s",
 			len(s.voteIndexPool), time.Since(startTime))
-		// Add the envelopes to the countTotalEnvelopes var
-		atomic.AddUint64(&s.countTotalEnvelopes, uint64(len(s.voteIndexPool)))
+		envelopeCount := &indexertypes.CountStore{}
+		if err := s.db.Get(indexertypes.CountStore_Envelopes, envelopeCount); err != nil {
```
p4u: Cannot we assume this index exist already (was created by scrutinizer initialization) and avoid doing two db operations here?

vochain/scrutinizer/vote.go
```diff
@@ -218,7 +218,11 @@ func (s *Scrutinizer) GetEnvelopeHeight(processID []byte) (uint64, error) {
 	defer func() { log.Debugf("GetEnvelopeHeight took %s", time.Since(startTime)) }()
 	if len(processID) == 0 {
 		// If no processID is provided, count all envelopes
-		return atomic.LoadUint64(&s.countTotalEnvelopes), nil
+		envelopeCountStore := &indexertypes.CountStore{}
+		if err := s.db.Get(indexertypes.CountStore_Envelopes, envelopeCountStore); err != nil {
+			return envelopeCountStore.Count, err
```
p4u: If error return 0 not the value

p4u (changes requested):

vochain/scrutinizer/scrutinizer.go
NateWilliams2: You mean by using updateMatching instead?

p4u: Yes

vochain/scrutinizer/indexertypes/types.go
```diff
@@ -12,6 +12,23 @@ import (
 	"go.vocdoni.io/proto/build/go/models"
 )
 
+const (
+	// CountStoreEntities is the key for the entity count db reference
```
p4u, line 16: Use `iota`

vochain/scrutinizer/process.go
```diff
@@ -153,20 +152,24 @@ func (s *Scrutinizer) ProcessList(entityID []byte,
...
 	defer func() { log.Debugf("ProcessCount took %s", time.Since(startTime)) }()
 	var c uint32
 	var err error
 	if len(entityID) == 0 {
-		// If no entity ID, return the cached count of all processes
-		return atomic.LoadInt64(&s.countTotalProcesses)
+		// If no entity ID, return the stored count of all processes
+		processCountStore := &indexertypes.CountStore{}
+		if err = s.db.Get(indexertypes.CountStoreProcesses, processCountStore); err != nil {
+			log.Warnf("could not get the process count: %v", err)
```
p4u: That should be an `Error()` since its something that should never happen (actually it should be a Panic() but in general we are not making the components Panic)

vochain/scrutinizer/scrutinizer.go
```diff
@@ -97,33 +98,109 @@ func NewScrutinizer(dbPath string, app *vochain.BaseApplication) (*Scrutinizer,
...
 	s.App.State.AddEventListener(s)
 	s.envelopeHeightCache = lru.New(countEnvelopeCacheSize)
 	s.resultsCache = lru.New(resultsCacheSize)
 	return s, nil
 }
 
+// retrieveCounts returns a count for txs, envelopes, processes, and entities, in that order.
+// If no CountStore model is stored for the type, it counts all db entries of that type.
+func (s *Scrutinizer) retrieveCounts() (*indexertypes.CountStore,
+	*indexertypes.CountStore, *indexertypes.CountStore, *indexertypes.CountStore, error) {
```
p4u: Wouldn't it make more sense to return a `map[uint8]uint64` ?

vochain/scrutinizer/scrutinizer.go
```diff
@@ -97,33 +98,109 @@ func NewScrutinizer(dbPath string, app *vochain.BaseApplication) (*Scrutinizer,
...
+		txCountStore.Type = indexertypes.CountStoreTransactions
+		if err := s.db.Upsert(txCountStore.Type, txCountStore); err != nil {
+			return nil, nil, nil, nil, fmt.Errorf("could not store transaction count: %v", err)
+		}
+	}
+	envelopeCountStore := new(indexertypes.CountStore)
+	if err = s.db.Get(indexertypes.CountStoreEnvelopes, envelopeCountStore); err != nil {
+		log.Warnf("could not get the envelope count: %v", err)
+		count, err := s.db.Count(&indexertypes.VoteReference{}, &badgerhold.Query{})
+		if err != nil {
```
p4u: count is 0 in case of error (zero-value for uint64) so you can simplify this logic:

```if err != nil &&  err != badger.ErrKeyNotFound { return .... }```

Comment (p4u) [reply vochain/scrutinizer/scrutinizer.go]: Same on all cases.

File: vochain/scrutinizer/transaction.go [hunk 27]
```diff
@@ -0,0 +1,61 @@
...
+	s.newTxPool = append(s.newTxPool, &indexertypes.TxReference{
+		BlockHeight:  blockHeight,
+		TxBlockIndex: txIndex,
+	})
+}
+
+// indexNewTxs indexes the txs pending in the newTxPool and updates the transaction count
+// this function should only be called within Commit(), on a new block.
+func (s *Scrutinizer) indexNewTxs(txList []*indexertypes.TxReference) {
+	if len(txList) > 0 {
```
p4u: `if len(txList) == 0 { return }`

vochain/scrutinizer/transaction.go
```diff
@@ -0,0 +1,61 @@
...
+		s.txIndexLock.Lock()
```
p4u: use defer for Unlock()

vochain/scrutinizer/transaction.go
```diff
@@ -0,0 +1,61 @@
...
+		txCount := &indexertypes.CountStore{}
+		if err := s.db.Get(indexertypes.CountStoreTransactions, txCount); err != nil {
+			log.Errorf("could not get tx count: %v", err)
```
p4u: Are we sure we want to keep the execution in case of error??

Please review all this cases of your PR, either "return" or "continue" or something is required if a query returns an error.

p4u (changes requested):

NateWilliams2:
addressed most recent comments @p4u

p4u (approved):

PR #216, opened 2021-06-03
Indexer db transaction wrapper

NateWilliams2:
Implements wrappers for the most commonly used update/insert db transactions. These wrappers check for transaction conflict error and re-try transactions with these errors. Transaction conflicts are caused by timing and are normal, but they transactions need to be attempted again. 
This was previously done in many places, or sometimes just ignored, leaving the possibility for db transactions that didn't happen due to conflicts.

### DIFF
--- a/vochain/scrutinizer/db.go
+++ b/vochain/scrutinizer/db.go
@@ -29,14 +29,10 @@ func InitDB(dataDir string) (*badgerhold.Store, error) {
 	return badgerhold.Open(opts)
 }
 
-func (s *Scrutinizer) updateMatchingWithoutTxConflicts(datatype interface{},
-	query *badgerhold.Query, update func(record interface{}) error) error {
+func (s *Scrutinizer) queryWithRetries(query func() error) error {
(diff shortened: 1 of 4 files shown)


mvdan (commented):
can we not do this in a generic way? like:

```
func queryWithRetries(query func() error) error {
    for {
        if err := query(); err != nil && strings.Contains(...) { ... }
```

Then you can use it like:

```
err := queryWithRetries(func() error {
    return s.db.UpdateMatching(&dst, badgerhold.Where(...).Eq(...)
}
```

Otherwise you need one extra helper func for every edge case: different operation, different parameters, etc.

NateWilliams2:
@mvdan I made a commit following your suggestion

p4u:
@mvdan can you review again?

mvdan (approved):

mvdan:
@NateWilliams2 apologies for the delay, I saw your comment and meant to re-review but it must have slipped my mind.

PR #217, opened 2021-06-04
Add interface check to db & censustree impl

arnaucube:
Adds interface checks to ensure that the different implementations of
`db.Database` & `censustree.Tree` match the interfaces.

### DIFF
--- a/censustree/gravitontree/trie.go
+++ b/censustree/gravitontree/trie.go
@@ -30,6 +30,8 @@ type Tree struct {
 	lastAccessUnix int64 // a unix timestamp, used via sync/atomic
 }
 
+var _ censustree.Tree = (*Tree)(nil)
+
 type exportElement struct {
 	Key   []byte `bare:"key"`
 	Value []byte `bare:"value"`
--- a/censustree/iden3tree/tree.go
+++ b/censustree/iden3tree/tree.go
@@ -25,6 +25,8 @@ type Tree struct {
 	lastAccessUnix int64 // a unix timestamp, used via sync/atomic
 }
 
+var _ censustree.Tree = (*Tree)(nil)
+
 type exportElement struct {
 	Key   []byte `bare:"key"`
 	Value []byte `bare:"value"`
--- a/db/badger.go
+++ b/db/badger.go
@@ -13,6 +13,8 @@ type BadgerDB struct {
 	db   *badger.DB
 }
 
+var _ Database = (*BadgerDB)(nil)
+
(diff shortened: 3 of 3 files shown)


censustree/gravitontree/trie_test.go
```diff
@@ -6,9 +6,25 @@ import (
 	"testing"
 	"time"
 
+	"go.vocdoni.io/dvote/censustree"
 	"go.vocdoni.io/dvote/util"
 )
 
+func TestInterface(t *testing.T) {
+	storage := t.TempDir()
```
p4u: no need to declare storage as a var

censustree/gravitontree/trie_test.go
```diff
@@ -6,9 +6,25 @@ import (
...
+	tree := &Tree{}
+	err := tree.Init("test", storage)
+	if err != nil {
```
p4u: please, use syntax `if err := tree.Init()...` when possible

censustree/gravitontree/trie_test.go
```diff
@@ -6,9 +6,25 @@ import (
...
+		t.Fatal(err)
+	}
+
+	var i interface{} = tree
+	_, ok := i.(censustree.Tree)
```
p4u: same, use inline `if` declaration

p4u (changes requested):

p4u (approved):

mvdan:
Why do we need actual tests here, though? Usually one just does `var _ SomeInterface = (*SomeImplementation)(nil)` as a static assertion that SomeImplementation implements SomeInterface. The package will stop building if the assertion is false.

arnaucube:
> Why do we need actual tests here, though? Usually one just does `var _ SomeInterface = (*SomeImplementation)(nil)` as a static assertion that SomeImplementation implements SomeInterface. The package will stop building if the assertion is false.

True! Amended the commit changing the added tests with the test that you've mentioned.
The motivation to add these tests is that if in the future the implementation is changed it can be automatically detected if it breaks the matching with the interface. As an example, the current [db.Iden3Storage](https://github.com/vocdoni/vocdoni-node/blob/master/db/iden3.go) does not match the `db.Database` interface, but as there is no test checking it, is not detected.

mvdan (approved):

arnaucube:
After @mvdan comment and chat, amended the commit to move the checks from test files to the actual implementation files. Much better in this way ^^

PR #221, opened 2021-06-11
Router: count api calls with atomic methods

### LINKED ISSUE #220: Data race
I found a race in dvotenode:
```
WARNING: DATA RACE
Read at 0x00c003638f60 by goroutine 611:
  go.vocdoni.io/dvote/service.API.func1()
      /Users/natewilliams/Vocdoni/go/vocdoni-node/service/api.go:86 +0x8f

Previous write at 0x00c003638f60 by goroutine 610:
  go.vocdoni.io/dvote/router.(*Router).Route()
      /Users/natewilliams/Vocdoni/go/vocdoni-node/router/router.go:251 +0x76e
```
router/router.go:
`r.PublicCalls++`

service/api.go:
`routerAPI.PrivateCalls, routerAPI.PublicCalls)`

This would be fixed by using atomic methods for r.PublicCalls and r.PrivateCalls

NateWilliams2:
Use atomic methods to update/access r.PublicCalls, r.PrivateCalls.

fixes #220

### DIFF
--- a/router/router.go
+++ b/router/router.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"strings"
+	"sync/atomic"
 	"time"
 
 	ethcommon "github.com/ethereum/go-ethereum/common"
@@ -246,9 +247,9 @@ func (r *Router) Route() {
 		}
 		log.Debugf("api query %s", request.MetaRequest.String())
 		if request.private {
-			r.PrivateCalls++
+			atomic.AddUint64(&r.PrivateCalls, 1)
 		} else {
-			r.PublicCalls++
+			atomic.AddUint64(&r.PublicCalls, 1)
 		}
 
 		if r.metricsagent != nil {
--- a/service/api.go
+++ b/service/api.go
@@ -2,6 +2,7 @@ package service
 
 import (
 	"fmt"
+	"sync/atomic"
 	"time"
 
 	"go.vocdoni.io/dvote/census"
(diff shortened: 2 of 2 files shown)


p4u (approved):

mvdan (approved):
Yikes, this is a pretty bad data race :)

p4u:
Totally! Good catch @NateWilliams2

I wonder why we did not catch that race until now... Of course, we need more tests and probably a server for run "go test -race" continuously. What do you think @mvdan ? Any tool for that?

NateWilliams2:
> I wonder why we did not catch that race until now... 

I did catch this by running dvotenode with "-race" and using it as an explorer gateway, something not standard to running normal tests.

mvdan:
> I wonder why we did not catch that race until now... Of course, we need more tests and probably a server for run "go test -race" continuously. What do you think @mvdan ? Any tool for that?

We could and probably should write more tests, but what we could also do is what @NateWilliams2 says: in the development/staging deployment, make one of the gateways run a binary built with the race detector. It will be roughly twice as slow when doing CPU work due to the overhead of the race detector, but it will be the best way to catch data races when we run realistic workloads.

I've seen plenty of other companies do this to catch bugs. That is, have 1-5% of the development/staging machines run with extra sanity checks. The number should be small enough to not impact performance or stability too much, but high enough to actually be likely to flag bugs.

p4u:
That makes sense to me. We could have this enabled on the DEV environment by default. Can you provide instructions so @vdo can add this flag to the DEV gateways?

mvdan:
I wouldn't enable it for every deployed gateway in the dev environment, though, because that could really slow down the environment when compared to production. I think one of them (an approximation of ~5%) would be a good start. Doing all of them can also give stability problems, because when a race is encountered, the Go process panics and would get restarted.

The only thing to do is to `go build` with `-race`, really. Assuming we want to use Docker images, we could build "race" docker images alongside our regular ones, and instruct one of the dev machines to pull that instead of the regular image.

vdo:
Got it, we can add

`ARG BUILDARGS`

to the Dockerfile, and then pass it as an extra var in the github actions, so we can re-use it for future settings.
Does it sound ok?

mvdan:
SGTM. We're all on the same page that a different BUILDARGS will mean a different docker image tag name, right? E.g. via a suffix.

vdo:
yes, I will add `-race` to `dev` and `master` generated images

PR #223, opened 2021-06-14
Add step for race builds

vdo:


p4u:
In general this makes sense to me, the only question is if adding -race to the CI actions (test) would make the CI too slow.

vdo:
Well, the push for the non race builds and tests will take the same time, and it will only work for `master` and `dev`. We can limit this to `dev` only though, if it makes sense.

vdo:
Just to be clear, the `-race`  flag was already present in the tests, I only linted them because of a newer version of the YAML vscode plugin

mvdan (approved):

PR #227, opened 2021-06-18
Indexer: GetEnvelopeHeight: return 0 for process without results

NateWilliams2:


vochain/scrutinizer/vote.go
```diff
@@ -224,6 +224,11 @@ func (s *Scrutinizer) GetEnvelopeHeight(processID []byte) (uint64, error) {
 	// If not cached, make the expensive query
 	results := &indexertypes.Results{}
 	if err := s.db.FindOne(results, badgerhold.Where(badgerhold.Key).Eq(processID)); err != nil {
+		// If results are not found, return envelope height of 0.
```
p4u: With this approach we cannot differentiate between canceled processes and not existing processes.

I think return en ErrNotFoundInDatabase if a process is Canceled is the right approach. There are not results and that's what we are returning.

We could modify the scrutinizer it norder to not remove Canceled processes but set it to 0 instead. However I still don't understand what's the problem behind returnin ErrNotFoundInDatabase. Can you clarify?

NateWilliams2: To me, in the case of a cancelled process, it seems really messy to return an ErrNotFoundInDatabase when querying for envelope count, if the envelope count is in fact known. Whoever is invoking the GetEnvelopeCount shouldn't need to expect an error as standard, non-error behavior.

p4u: Ok, then let's modify the scrutinizer logic to not remove the results entry (just clean Votes but let EnvelopeHeight and Weight) and set Final to true. Does that work?

NateWilliams2: that works for me! I'll implement that

vochain/scrutinizer/process.go
```diff
@@ -385,7 +385,21 @@ func (s *Scrutinizer) updateProcess(pid []byte) error {
...
+					results, ok := record.(*indexertypes.Results)
+					if !ok {
+						return fmt.Errorf("record isn't the correct type! Wanted Result, got %T", record)
+					}
+					// On cancelled process, remove all results except for envelope height, weight, pid
+					results.Votes = [][]*big.Int{}
+					results.EnvelopeType = &models.EnvelopeType{}
+					results.VoteOpts = &models.ProcessVoteOptions{}
+					results.Signatures = []types.HexBytes{}
+					results.Final = *types.False
```
p4u: That's weird, I'd just set it to "false"

p4u (approved):

PR #228, opened 2021-06-18
oracle: add oracle web3 endpoint pool

jordipainan:
If an Oracle that is subscribed to one or more
ethereum smart contract events
have any problem with the endpoint that is
connected with, after some retries, another
web3 endpoint will be used.

If there is not any web3 endpoint working
the application will terminate.

When a new connection is created the ethevents
service will be restarted as well as the
PrintInfo go routine.

A context.WithCancel is being used to terminate
all the goroutines created by the ethereum events
service and the contracts will be resolved on
the ENS an instanciated again. That way if any
event is lost during the time that there
are connection issues, will be catch again
on restart.

### DIFF
--- a/cmd/dvotenode/dvotenode.go
+++ b/cmd/dvotenode/dvotenode.go
@@ -110,7 +110,7 @@ func newConfig() (*config.DvoteCfg, config.Error) {
 	// ethereum web3
 	globalCfg.W3Config.ChainType = *flag.StringP("ethChain", "c", "goerli",
 		fmt.Sprintf("Ethereum blockchain to use: %s", ethchain.AvailableChains))
-	globalCfg.W3Config.W3External = *flag.StringP("w3External", "w", "",
+	globalCfg.W3Config.W3External = *flag.StringArrayP("w3External", "w", []string{},
 		"ethereum web3 endpoint. Supported protocols: http(s)://, ws(s):// and IPC filepath")
 	// ipfs
 	globalCfg.Ipfs.NoInit = *flag.Bool("ipfsNoInit", false,
@@ -561,37 +561,48 @@ func main() {
 				go kk.RevealUnpublished()
 			}
 
-			// Start ethereum events (if web3 endpoint configured)
-			if globalCfg.W3Config.W3External != "" {
-				var evh []ethevents.EventHandler
-				var w3uri string
-				switch {
-				case strings.HasPrefix(globalCfg.W3Config.W3External, "ws"):
-					w3uri = globalCfg.W3Config.W3External
-				case strings.HasSuffix(globalCfg.W3Config.W3External, "ipc"):
-					w3uri = globalCfg.W3Config.W3External
-				default:
-					log.Fatal("web3 external must be websocket or IPC for event subscription")
+			var w3uris []string
+			if globalCfg.W3Config.W3External != nil {
+				for idx, web3Endpoint := range globalCfg.W3Config.W3External {
+					web3EndpointTrimmed := strings.Trim(web3Endpoint, `"[]`)
+					log.Debugf("web3endpoint %d: %s", idx, web3EndpointTrimmed)
+					switch {
+					case strings.HasPrefix(web3EndpointTrimmed, "ws"):
+						w3uris = append(w3uris, web3EndpointTrimmed)
+					case strings.HasSuffix(web3EndpointTrimmed, "ipc"):
+						w3uris = append(w3uris, web3EndpointTrimmed)
+					default:
+						log.Warnf(`invalid web3 endpoint %s must be websocket or IPC
+						for event subscription`, web3EndpointTrimmed)
+					}
 				}
+			}
+			// Start ethereum events (if at least one web3 endpoint configured)
+			if len(w3uris) > 0 {
+				log.Infof("using %+v web3 endpoints", w3uris)
 
+				var evh []ethevents.EventHandler
 				evh = append(evh, ethevents.HandleVochainOracle)
 
-				var initBlock *int64
+				var initBlock int64
 				if !globalCfg.EthEventConfig.SubscribeOnly {
-					initBlock = new(int64)
 					chainSpecs, err := ethchain.SpecsFor(globalCfg.W3Config.ChainType)
 					if err != nil {
-						log.Warn("cannot get chain block to start looking for events, using 0")
-						*initBlock = 0
-					} else {
-						*initBlock = chainSpecs.StartingBlock
+						log.Fatal("cannot get chain block to start looking for events")
 					}
+					initBlock = chainSpecs.StartingBlock
 				}
 
 				whiteListedAddr := []string{}
 				for _, addr := range globalCfg.VochainConfig.EthereumWhiteListAddrs {
+					if addr == "[]" {
+						// avoid empty address and pflag autofill in case of empty array
+						continue
+					}
 					if ethcommon.IsHexAddress(addr) {
 						whiteListedAddr = append(whiteListedAddr, addr)
+					} else {
+						log.Warnf("whitelist address %s is not a valid hex address", addr)
 					}
 				}
 				if len(whiteListedAddr) > 0 {
@@ -599,7 +610,7 @@ func main() {
 				}
 				if err := service.EthEvents(
 					context.Background(),
-					w3uri,
+					w3uris,
 					globalCfg.W3Config.ChainType,
 					initBlock,
 					cm,
@@ -633,8 +644,7 @@ func main() {
 			if err != nil {
 				log.Fatal(err)
 			}
-			if err := apior.EnableERC20(globalCfg.W3Config.ChainType,
-				globalCfg.W3Config.W3External); err != nil {
+			if err := apior.EnableERC20(globalCfg.W3Config.ChainType, globalCfg.W3Config.W3External); err != nil {
 				log.Fatal(err)
 			}
 		}
--- a/ethereum/ethevents/events.go
+++ b/ethereum/ethevents/events.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"math/big"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	eth "github.com/ethereum/go-ethereum"
@@ -64,7 +65,7 @@ type EthereumEvents struct {
 	// ContractsInfo holds useful info for working with the desired contracts
 	ContractsInfo map[string]*ethereumhandler.EthereumContract
 	// EthereumLastKnownBlock keeps track of the latest Ethereum known block
-	EthereumLastKnownBlock int64
+	EthereumLastKnownBlock atomic.Value
 }
 
 type timedEvent struct {
@@ -109,15 +110,6 @@ func NewEthEvents(
 	vocapp *vochain.BaseApplication,
 	ethereumWhiteList []string,
 ) (*EthereumEvents, error) {
-	// try to connect to default addr if w3Endpoint is empty
-	if len(w3Endpoints) == 0 {
-		return nil, fmt.Errorf("no w3Endpoint specified on Ethereum Events")
-	}
-	ph, err := ethereumhandler.NewEthereumHandler(contracts, srcNetworkId, w3Endpoints)
-	if err != nil {
-		return nil, fmt.Errorf("cannot create voting handle: %w", err)
-	}
-	ph.WaitSync()
 	secureAddrList := make(map[common.Address]bool, len(ethereumWhiteList))
 	if len(ethereumWhiteList) != 0 {
 		for _, sAddr := range ethereumWhiteList {
@@ -128,7 +120,6 @@ func NewEthEvents(
 			secureAddrList[addr] = true
 		}
 	}
-
 	contractsAddress := []common.Address{}
 	for _, contract := range contracts {
 		if !bytes.Equal(contract.Address.Bytes(), common.Address{}.Bytes()) {
@@ -137,7 +128,6 @@ func NewEthEvents(
 			}
 		}
 	}
-
 	confirmThreshold := blockConfirmThreshold[0]
 	if _, ok := blockConfirmThreshold[srcNetworkId]; ok {
 		confirmThreshold = blockConfirmThreshold[srcNetworkId]
@@ -145,11 +135,10 @@ func NewEthEvents(
 	log.Infof("chain %s found, block confirmation threshold set to %s",
 		srcNetworkId, confirmThreshold)
 	ethev := &EthereumEvents{
-		VotingHandle: ph,
-		Signer:       signer,
-		DialAddrs:    w3Endpoints,
-		Census:       cens,
-		VochainApp:   vocapp,
+		Signer:     signer,
+		DialAddrs:  w3Endpoints,
+		Census:     cens,
+		VochainApp: vocapp,
 		EventProcessor: &EventProcessor{
 			Events:                make(chan ethtypes.Log),
 			EventProcessThreshold: confirmThreshold,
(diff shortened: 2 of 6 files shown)


cmd/dvotenode/dvotenode.go
```diff
@@ -543,6 +543,28 @@ func main() {
...
+					w3uris = append(w3uris, web3EndpointTrimmed)
+				case strings.HasSuffix(web3EndpointTrimmed, "ipc"):
+					w3uris = append(w3uris, web3EndpointTrimmed)
+				default:
+					log.Warnf(`invalid web3 endpoint %s must be websocket or IPC
+						for event subscription`, web3EndpointTrimmed)
+				}
+			}
+		} else {
+			log.Errorf(`No web3 endpoint found, oracle needs at least
```
p4u: That should probably be a Fatal... or if its an Error do something

cmd/dvotenode/dvotenode.go
```diff
@@ -543,6 +543,28 @@ func main() {
...
+			one valid web3 endpoint to handle Ethereum contract events`)
+		}
+
+		log.Debugf("Using %+v web3 endpoints", w3uris)
```
p4u: capital letter

cmd/dvotenode/dvotenode.go
```diff
@@ -543,6 +543,28 @@ func main() {
 			log.Fatal(err)
 		}
 
+		w3uris := []string{}
+		if len(globalCfg.W3Config.W3External) > 0 {
```
p4u: I'd revert that logic... its a bit messy.

Remove the `if len(globalCfg.W3Config.W3External) > 0 {` and also the else and just add after the for

`if len(w3uris) == 0 { log.Fatal("no valid web3 endpoint found") }`

cmd/dvotenode/dvotenode.go
p4u: superseded by the previous comment

p4u (changes requested):

p4u:
@arnaucube and @mvdan can you both review too? thanks!

cmd/dvotenode/dvotenode.go
```diff
@@ -543,6 +543,27 @@ func main() {
...
+				log.Warnf(`invalid web3 endpoint %s must be websocket or IPC
+						for event subscription`, web3EndpointTrimmed)
+			}
+		}
+
+		if len(w3uris) == 0 {
+			log.Fatal(`no web3 endpoint found, oracle needs at least
+			one valid web3 endpoint to handle Ethereum contract events`)
+		}
+		log.Debugf("using %+v web3 endpoints", w3uris)
```
p4u: this should be info IMO

cmd/dvotenode/dvotenode.go
```diff
@@ -559,56 +580,45 @@ func main() {
...
-					}
-				}
-				if len(whiteListedAddr) > 0 {
-					log.Infof("ethereum whitelisted addresses %+v", whiteListedAddr)
+			var initBlock *int64
+			if !globalCfg.EthEventConfig.SubscribeOnly {
+				initBlock = new(int64)
+				chainSpecs, err := ethchain.SpecsFor(globalCfg.W3Config.ChainType)
+				if err != nil {
+					log.Warn("cannot get chain block to start looking for events, using 0")
```
p4u: I think initBlock should not be a pointer, what's the reason? 

Also I'm not sure we should uset initBlock=0 if error. I think its safer to do a Fatal() and print the error instead

cmd/dvotenode/dvotenode.go
```diff
@@ -559,56 +580,45 @@ func main() {
...
+					*initBlock = 0
+				} else {
```
p4u: No need for this else if err executes Fatal()

cmd/dvotenode/dvotenode.go
```diff
@@ -559,56 +580,45 @@ func main() {
...
-					vnode,
-					evh,
-					sc,
-					whiteListedAddr); err != nil {
-					log.Fatal(err)
+			}
+
+			whiteListedAddr := []string{}
+			for _, addr := range globalCfg.VochainConfig.EthereumWhiteListAddrs {
+				if ethcommon.IsHexAddress(addr) {
```
p4u: If `addr` is not Hex, shouldn't we at least print a Warning?

cmd/dvotenode/dvotenode.go
```diff
@@ -632,7 +642,7 @@ func main() {
 				log.Fatal(err)
 			}
 			if err := apior.EnableERC20(globalCfg.W3Config.ChainType,
```
p4u: I think breaking lines here is not needed

ethereum/ethevents/events.go
```diff
@@ -63,6 +63,7 @@ type EthereumEvents struct {
 	ContractsAddress []common.Address
 	// ContractsInfo holds useful info for working with the desired contracts
 	ContractsInfo map[string]*ethereumhandler.EthereumContract
+	StartBlock    *int64
```
p4u: why a pointer?

p4u: If you thinking about saving memory, take into account you are actually doing the opposite sice a pointer will require 2xint64 (one for the address and another for the value)

ethereum/ethevents/events.go
```diff
@@ -290,13 +290,36 @@ func (ev *EthereumEvents) SubscribeEthereumEventLogs(ctx context.Context, fromBl
 		go ev.runEventProcessor(ctx)
 	}
 
+	// check eth connection
+	connCh := make(chan bool, 1)
+	go func(chan bool) {
+		for {
+			time.Sleep(time.Second * 5)
+			blockTctx, cancel := context.WithTimeout(ctx, types.EthereumReadTimeout*2)
```
p4u: Why EthereumReadTimeout x2 ?

p4u: Just call it ctx (more standard)

ethereum/ethevents/events.go
```diff
@@ -290,13 +290,36 @@ func (ev *EthereumEvents) SubscribeEthereumEventLogs(ctx context.Context, fromBl
...
+			if _, err := ev.VotingHandle.EthereumClient.BlockByNumber(blockTctx, nil); err != nil {
+				log.Warnf("cannot check ethereum connection while running the event processor, %s", err)
+				connCh <- false
+				cancel()
```
p4u: cancel() should also be called if not error

ethereum/ethevents/events.go
```diff
@@ -290,13 +290,36 @@ func (ev *EthereumEvents) SubscribeEthereumEventLogs(ctx context.Context, fromBl
 		go ev.runEventProcessor(ctx)
 	}
 
+	// check eth connection
+	connCh := make(chan bool, 1)
+	go func(chan bool) {
+		for {
+			time.Sleep(time.Second * 5)
+			blockTctx, cancel := context.WithTimeout(ctx, types.EthereumReadTimeout*2)
+			if _, err := ev.VotingHandle.EthereumClient.BlockByNumber(blockTctx, nil); err != nil {
```
p4u: Shouldn't the block number be saved and check that it is actually being increased? Else we won't detect a frozen web3 endpoint

ethereum/ethevents/events.go
```diff
@@ -290,13 +290,36 @@ func (ev *EthereumEvents) SubscribeEthereumEventLogs(ctx context.Context, fromBl
 		go ev.runEventProcessor(ctx)
 	}
 
+	// check eth connection
+	connCh := make(chan bool, 1)
```
p4u: This channel should have a more descriptive name, such as "connFailure"

ethereum/ethevents/events.go
```diff
@@ -377,20 +400,28 @@ func (ep *EventProcessor) next() *ethtypes.Log {
...
+				if err != nil {
+					log.Error(err)
+					continue
+				}
+				if event.Removed {
+					log.Warnf("removing reversed log event: %s", evtJSON)
+					ev.EventProcessor.del(&event)
+				} else {
+					if event.BlockNumber == 0 {
+						log.Fatalf("this event %+v must not be arribe here (BOMB)", event)
```
p4u: what's the meaning of this message? I don't get it. Can you rephrase?

ethereum/handler/ethereumHandler.go
```diff
@@ -84,32 +87,29 @@ func NewEthereumHandler(contracts map[string]*EthereumContract, srcNetworkId mod
...
-			maxtries--
-			continue
+	for _, endpoint := range dialEndpoints {
+		maxtries := 5
+		for {
+			if maxtries == 0 {
+				log.Warnf("could not connect to %s endpoint, trying the next one ...", endpoint)
+				break
+			}
+			eh.EthereumRPC, err = ethrpc.Dial(endpoint)
```
p4u: Use inline if

ethereum/handler/ethereumHandler.go
```diff
@@ -84,32 +87,29 @@ func NewEthereumHandler(contracts map[string]*EthereumContract, srcNetworkId mod
...
-		if err != nil || eh.EthereumRPC == nil {
-			log.Warnf("cannot create an ethereum rpc connection: (%s), trying again ...", err)
-			time.Sleep(time.Second * 2)
-			maxtries--
-			continue
+	for _, endpoint := range dialEndpoints {
+		maxtries := 5
+		for {
+			if maxtries == 0 {
+				log.Warnf("could not connect to %s endpoint, trying the next one ...", endpoint)
```
p4u: "one..." (or even better, remove the "..." since we are not using that in general)

ethereum/handler/ethereumHandler.go
```diff
@@ -84,32 +87,29 @@ func NewEthereumHandler(contracts map[string]*EthereumContract, srcNetworkId mod
...
+				break
+			}
+			eh.EthereumRPC, err = ethrpc.Dial(endpoint)
+			if err != nil || eh.EthereumRPC == nil {
+				log.Warnf("cannot create an ethereum rpc connection with : (%s), trying again ...", err)
```
p4u: `"with %s: (%v)", endpoint, err)`

ethereum/handler/ethereumHandler.go
```diff
@@ -84,32 +87,29 @@ func NewEthereumHandler(contracts map[string]*EthereumContract, srcNetworkId mod
...
+				time.Sleep(time.Second * 5)
```
p4u: Why sleeping 5 seconds? Seems too much

ethereum/handler/ethereumHandler.go
```diff
@@ -163,7 +166,10 @@ func (eh *EthereumHandler) PrintInfo(ctx context.Context, seconds time.Duration)
 		info, err = eh.SyncInfo(tctx)
 		cancel()
 		if err != nil {
-			log.Warn(err)
+			log.Warnf("error getting ethereum info: %s", err)
+			if err := eh.Connect(eh.Endpoints); err != nil {
+				log.Warnf("cannot connect to any web3 endpoint: %v", err)
```
p4u: Shouldn't this Fatal?

service/ethevents.go
```diff
@@ -40,20 +41,37 @@ func EthEvents(
...
 		return fmt.Errorf("couldn't create ethereum events listener: %w", err)
 	}
 	for _, e := range evh {
 		ev.AddEventHandler(e)
 	}
-	go func() {
-		ev.SubscribeEthereumEventLogs(ctx, startBlock)
-	}()
+
+	initialized := new(bool)
```
p4u: Why a pointer?

service/ethevents.go
```diff
@@ -40,20 +41,37 @@ func EthEvents(
...
+	go func(initialized *bool) {
+		for {
+			if ctx.Err() != nil {
+				return
+			}
+			ctx, cancel := context.WithCancel(ctx)
+			if *initialized {
+				if ev.VotingHandle, err = ethereumhandler.NewEthereumHandler(ev.ContractsInfo, specs.NetworkSource, w3uris); err != nil {
```
p4u: This line is 126 chars

service/ethevents.go
```diff
@@ -40,20 +41,37 @@ func EthEvents(
...
+					log.Fatalf("cannot restart ethereum events, cannot create ethereum handler: %v", err)
+				}
+			}
+			*initialized = true
```
p4u: I don't understand the logic behind this routine. Why a for loop? "initialized" is set to true on each iteration, why? Also why ev.VotingHandle is inside the for loop?

p4u (changes requested):

ethereum/handler/ethereumHandler.go
```diff
@@ -84,32 +87,29 @@ func NewEthereumHandler(contracts map[string]*EthereumContract, srcNetworkId mod
...
+			}
+			eh.EthereumRPC, err = ethrpc.Dial(endpoint)
+			if err != nil || eh.EthereumRPC == nil {
+				log.Warnf("cannot create an ethereum rpc connection with : (%s), trying again ...", err)
+				time.Sleep(time.Second * 5)
+				maxtries--
+				continue
+			}
+			// if RPC connection established, create an ethereum client using the RPC client
+			eh.EthereumClient = ethclient.NewClient(eh.EthereumRPC)
```
mvdan, line 107: are you sure that the EthereumClient field isn't racy? it's used from other goroutines as far as I can tell.

ethereum/ethevents/events.go
```diff
@@ -290,13 +290,36 @@ func (ev *EthereumEvents) SubscribeEthereumEventLogs(ctx context.Context, fromBl
...
+		case <-sub.Err():
 			// TODO: @jordipainan do not fatal here, handle error on event subscription channel
-			log.Fatalf("ethereum events subscription error on channel: %s", err)
+			log.Warn("ethereum events subscription error on channel")
+			log.Infof("restarting ethereum events subscription with web3: %+v", ev.DialAddrs)
+			ev.EventProcessor.eventProcessorRunning = false
+			return
 		case event := <-logs:
 			ev.EventProcessor.Events <- event
+		case <-connCh:
```
mvdan: we have two identical cases in this select now; why do we have two channels instead of one?

cmd/dvotenode/dvotenode.go
jordipainan: changed to info :)

cmd/dvotenode/dvotenode.go
jordipainan: yes agree, at some point we discussed to not fatal in this kind of cases, that is why most of this errors are handled with a warning. In any case I also prefer to not use initBlock=0

cmd/dvotenode/dvotenode.go
jordipainan: yes, added

ethereum/ethevents/events.go
jordipainan: copy paste err, changed

ethereum/ethevents/events.go
jordipainan: moved and added a defer that will be executed in the very end

ethereum/ethevents/events.go
jordipainan: there is an else clause that will be executed when restarting the subscription (100 blocks), I think is enough for this purpose. What do you think?

ethereum/ethevents/events.go
jordipainan: changed :)

ethereum/ethevents/events.go
jordipainan: just a sanity check cause I do not expect having block number 0 on event logs. Changed the message for more clarity

ethereum/handler/ethereumHandler.go
jordipainan: lowered to 3

ethereum/handler/ethereumHandler.go
jordipainan: yes assuming last comment made on this, changed to a fatal

service/ethevents.go
jordipainan: no need, only one goroutine use the var, changed and no longer using a pointer

service/ethevents.go
jordipainan: kind of a workaround because new EthEvents already invokes new ethereum handler and I wanted to avoid this. Solving properly requires some refactor on the new EthEvents function. This is out of the scope of this commit.
WIll add a TODO explaining more about this for adding clarity and i'll make another PR improving this.

ethereum/ethevents/events.go
jordipainan: discussed privately and solved

jordipainan: btw, the fallthrough clause cannot be applied here afaik. The compiler complains about it and I only solved it with a weird if clause.

ethereum/handler/ethereumHandler.go
jordipainan: yes it is, adding a TODO for having 2 different connections:
- One for the subscription
- One for the ethereum sync status

cmd/dvotenode/dvotenode.go
p4u: If an error is something that does not allow the execution, it should be a Fatal. And still, why initBlock is a pointer?

cmd/dvotenode/dvotenode.go
```diff
@@ -545,6 +545,27 @@ func main() {
...
+				w3uris = append(w3uris, web3EndpointTrimmed)
+			default:
+				log.Warnf(`invalid web3 endpoint %s must be websocket or IPC
+						for event subscription`, web3EndpointTrimmed)
+			}
+		}
+
+		if len(w3uris) == 0 {
+			log.Fatal(`no web3 endpoint found, oracle needs at least
+			one valid web3 endpoint to handle Ethereum contract events`)
```
mvdan: FYI this log line will have a newline and lots of whitespace. I'd just join it as a single line, but if you really want to split it:

```
log.Fatal("foo bar " + 
    "baz")
```

cmd/dvotenode/dvotenode.go
```diff
@@ -561,56 +582,48 @@ func main() {
...
+				context.Background(),
+				w3uris,
+				globalCfg.W3Config.ChainType,
+				initBlock,
+				cm,
+				signer,
+				vnode,
+				evh,
+				sc,
+				whiteListedAddr); err != nil {
```
mvdan: I really think we're going way, way overkill with avoiding long lines. You're literally writing multiple lines which are >80% whitespace :) This entire call can very easily fit in 2-3 lines.

ethereum/ethevents/events.go
```diff
@@ -290,13 +291,38 @@ func (ev *EthereumEvents) SubscribeEthereumEventLogs(ctx context.Context, fromBl
...
+			}
+			ev.EthereumLastKnownBlock = block.Number().Int64()
+		}
+	}(connFailure)
+
 	for {
 		select {
-		case err := <-sub.Err():
+		case <-sub.Err():
 			// TODO: @jordipainan do not fatal here, handle error on event subscription channel
```
mvdan: I think this TODO is outdated now

ethereum/handler/ethereumHandler.go
mvdan: Hm, OK. Seems like a pretty big TODO, but I guess "it usually works" is better than delaying this PR.

mvdan (approved):
SGTM, just a few minor nits

ethereum/ethevents/events.go
mvdan: You're right, I forgot that fallthroughs can't be used in selects. I guess not worth doing anything more magic here, but we could at least put both cases next to each other.

jordipainan:
thanks for the re-review @mvdan

ethereum/ethevents/events.go
jordipainan: will tackle on the next PR, added a better approach :)

service/ethevents.go
jordipainan: will be better addressed on the next pr

ethereum/ethevents/events.go
```diff
@@ -290,13 +291,37 @@ func (ev *EthereumEvents) SubscribeEthereumEventLogs(ctx context.Context, fromBl
...
+			time.Sleep(time.Second * 5)
+			tctx, cancel := context.WithTimeout(ctx, types.EthereumReadTimeout)
+			defer cancel()
+			block, err := ev.VotingHandle.EthereumClient.BlockByNumber(tctx, nil)
+			if err != nil {
+				log.Warnf("cannot check ethereum connection while running the event processor, %s", err)
+				connFailure <- false
+				return
+			}
+			ev.EthereumLastKnownBlock = block.Number().Int64()
```
p4u: This var is never used and also it looks like racy, I think sync.Atomic should be used here

jordipainan: yep, so finally adding the commit I mentioned regarding this case here. And yes, is used on the resubscribe routine

p4u (approved):

ethereum/ethevents/events.go
```diff
@@ -172,155 +161,81 @@ func (ev *EthereumEvents) AddEventHandler(handler EventHandler) {
...
-			time.Sleep(time.Second * 2)
-			if err := ev.VotingHandle.Connect(ev.DialAddrs); err != nil {
-				log.Warn(err)
-			}
-			continue
-		}
-		lastBlock = blk.Number().Int64()
-		break
+	var err error
+	uLastBlockNumber, err := ev.VotingHandle.EthereumClient.BlockNumber(tctx)
```
mvdan, line 170: you only need to do this remote ethereum request if fromBlock==0, right? I'd move it inside the "if".

ethereum/ethevents/events.go
```diff
@@ -172,155 +161,81 @@ func (ev *EthereumEvents) AddEventHandler(handler EventHandler) {
...
-		// For security, even if subscribe only, force to process at least the past `readBlockPast`
-		// Same as the last processEventLogsFromTo function call
-		if err := ev.processEventLogsFromTo(ctx, blk.Number().Int64()-readBlocksPast,
-			blk.Number().Int64(), ev.VotingHandle.EthereumClient); err != nil {
-			log.Errorf("cannot process event logs: %v", err)
-		}
+	lastBlockNumber := int64(uLastBlockNumber)
+	fromBlock := ev.EthereumLastKnownBlock.Load().(int64)
+	if fromBlock == 0 {
+		ev.EthereumLastKnownBlock.Store(lastBlockNumber)
```
mvdan, line 177: note that this Store is separate from the Load, so it's not atomic. I imagine we don't mind that. If the original atomic value was 0 and two goroutines both try to update it at the same time, I imagine they'd both try to update it to the same new value. If it's possible that those two goroutines would end up with different "latest block" values, then that might be problematic.

ethereum/handler/ethereumHandler.go
```diff
@@ -161,15 +161,15 @@ func (eh *EthereumHandler) PrintInfo(ctx context.Context, seconds time.Duration)
 	var err error
 	var syncingInfo string
 	for {
+		if ctx.Err() != nil {
+			return
+		}
 		time.Sleep(seconds)
```
mvdan, line 167: nit: do a select between ctx.Done and time.After, to stop sleeping if the context is cancelled

mvdan (commented):
Sorry for the late review. Better late than never :) Mostly just nits.

p4u:
@mvdan that's old code, there is another commit that modifies the behavior of this PR. I'd propose (if you want) to review the code in master and create a new PR if you have suggestions.

mvdan:
Yep I know - Jordi had asked me to review the last commit two weeks ago, but I was away for some time, so I'm only getting to it now :)

PR #229, opened 2021-06-25
WIP: censusTree: add support for multiple census

p4u:
The census manager handler now receives a constructor function from the
censustree factory which allows to add and manage multiple census types.

The API exposes a new field named "censusType" which can be used by the
caller in order to choose the census type.

The default census (if not escpecified) is Graviton in order to keep
backwards compatibility.

### DIFF
--- a/api/types.go
+++ b/api/types.go
@@ -95,6 +95,7 @@ type MetaRequest struct {
 	CensusValue  []byte                         `json:"censusValue,omitempty"`
 	CensusValues [][]byte                       `json:"censusValues,omitempty"`
 	CensusDump   []byte                         `json:"censusDump,omitempty"`
+	CensusType   int                            `json:"censusType,omitempty"`
 	Content      []byte                         `json:"content,omitempty"`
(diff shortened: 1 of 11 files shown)


p4u:
TODO: use protobuf types instead of internal int identifier for census

arnaucube:
I'm in a branch where I'm adding a new parameter to the censustree factory to define how many levels will have the tree.
This is necessary due the censustrees that need to be compatible with snarks, need to have a fixed number of levels, which will depend on the circuit used.
The current changes are at this branch https://github.com/vocdoni/vocdoni-node/commit/60dcffb7c1895c0379a4d99fbe0b31c6a922e0ed.
Maybe an option can be to merge first the branch [feature/censustree-circom-comp](https://github.com/vocdoni/vocdoni-node/tree/feature/censustree-circom-comp), and then rebase this current PR to that branch merged into master, which will include the parameter of `nLevels`.

arnaucube:
As https://github.com/vocdoni/vocdoni-node/pull/231 has been merged into `master` and included the changes of this PR, we can close this PR.

PR #232, opened 2021-07-07
ethevents: improve web3 failure management

p4u:
- fix data race and make it race free (hopefully)
- control over the block number to detect frozen web3 endpoints
- clean a bit the code (remove all legacy stuff)
- manage web3 endpoints in a better way

If a web3 connection fails, then the next one is directly choosen (if
any), instead of keeping to reconnect with the one already failed.

TODO: apply the same mechanism on the signaling oracle


p4u:
I have been testing this on the development environment and works well. The previous implementation had a race that end up on a nil pointer situation, so in order to fix that ASAP, merging now.

PR #234, opened 2021-07-13
Feature/update storage proofs eth go

ed255:
### vocchain: Remove unecessary len(proofKey) check

In VoteTxCheck remove the check that len(proofKey) must be >= 20.
proofKey can be one of the tree:

1. `pubKey` (censusOrigin_OFF_CHAIN_TREE)
Comes from:
`pubKey, err = ethereum.PubKeyFromSignature(txBytes, signature)`
so it will be 33 bytes (compressed ethereum key) if err == nil.

2. `addr.Bytes()` (CensusOrigin_OFF_CHAIN_CA)
Comes from:
`addr, err := ethereum.AddrFromPublicKey(pubKey)`
so it will be 20 bytes (ethereum Address) if err == nil

3. `slot` (CensusOrigin_ERC20)
Comes from:
`sphelpers.GetMapSlot(addr, int(*process.EthIndexSlot))`
so it will be 32 bytes (ethereum Hash) if err == nil

There's no case in which len(proofKey) can be < 20, so we remove the
check.

### Split vochain CheckProof into smaller fns

### DIFF
--- a/vochain/apputils.go
+++ b/vochain/apputils.go
@@ -20,7 +20,6 @@ import (
 	blind "github.com/arnaucube/go-blindsecp256k1"
 	ethcommon "github.com/ethereum/go-ethereum/common"
 
-	"github.com/ethereum/go-ethereum/common/hexutil"
 	cfg "github.com/tendermint/tendermint/config"
 	crypto25519 "github.com/tendermint/tendermint/crypto/ed25519"
 	tmjson "github.com/tendermint/tendermint/libs/json"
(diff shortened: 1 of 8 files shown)


vochain/apputils.go
```diff
@@ -29,102 +28,118 @@ import (
...
-				return false, nil, fmt.Errorf("graviton proof is empty")
-			}
-			valid, err := gravitontree.CheckProof(key, []byte{}, censusRoot, p.Siblings)
-			return valid, big.NewInt(1), err
-		case *models.Proof_Iden3:
-			// NOT IMPLEMENTED
-			return false, nil, fmt.Errorf("iden3 proof not implemented")
+// VerifyProofOffChainTree verifies a proof with census origin OFF_CHAIN_TREE.
+// Returns verification result and weight.
+func VerifyProofOffChainTree(process *models.Process, proof *models.Proof, censusOrigin models.CensusOrigin,
```
p4u: Since we are refactoring this I think we should also move the Proof verification functions to a Go file (such as proof.go) instead of "apputils.go"

vochain/apputils.go
```diff
@@ -29,102 +28,118 @@ import (
...
+	default:
+		return false, nil, fmt.Errorf("ca proof %s type not supported", p.Type.String())
 	}
-	return false, nil, fmt.Errorf("proof type not supported for census origin %d", censusOrigin)
+	return true, big.NewInt(1), nil
+}
+
+// VerifyProofERC20 verifies a proof with census origin ERC20 (mapbased).
+// Returns verification result and weight.
+func VerifyProofERC20(process *models.Process, proof *models.Proof, censusOrigin models.CensusOrigin,
```
p4u: I also wonder if it would make sense to write an interface around this functions, so it will be more modular and easy to extend with new proofs. What do you think?

ed255: I have introduced the `VerifyProofFunc` function type to define a prototype of the verification function that all verifyProof functions follow.  I haven't followed the interface approach because interfaces work for methods on structs, but in this case we only have standalone functions whose prototype can be defined by a function type.  Let me know what you think about it!

p4u: Yes, that's a better approach IMHO. Thanks

p4u (approved):

PR #238, opened 2021-07-22
Update statedb Get to return error

arnaucube:


vochain/process.go
```diff
@@ -64,9 +64,15 @@ func (v *State) Process(pid []byte, isQuery bool) (*models.Process, error) {
...
 	v.RLock()
 	if isQuery {
-		processBytes = v.Store.ImmutableTree(ProcessTree).Get(pid)
+		processBytes, err = v.Store.ImmutableTree(ProcessTree).Get(pid)
+		if err != nil {
+			return nil, err
+		}
 	} else {
-		processBytes = v.Store.Tree(ProcessTree).Get(pid)
+		processBytes, err = v.Store.Tree(ProcessTree).Get(pid)
```
p4u: Use `if processBytes, err = .... ; err != nil {} `

p4u: Same for all other cases

arnaucube: Right! Updated.

arnaucube: Would love to have a linter for this hehe

p4u (approved):

PR #240, opened 2021-07-22
vochain: introduce new transaction registerKeyTx

p4u:
This is an initial commit in order to support voting with zkSnark census

### DIFF
--- a/vochain/census.go
+++ b/vochain/census.go
@@ -1,0 +1,104 @@
+package vochain
+
+import (
+	"fmt"
+	"math/big"
+
+	"github.com/ethereum/go-ethereum/common"
+	"go.vocdoni.io/dvote/crypto/ethereum"
+	models "go.vocdoni.io/proto/build/go/models"
+)
+
+// AddToRollingCensus adds a new key to an existing rolling census.
+// If census does not exist yet it will be created.
+func (s *State) AddToRollingCensus(pid []byte, key []byte, weight *big.Int) error {
+	/*
+		// In the state we only store the last census root (as value) using key as index
+		s.Lock()
+		err := s.Store.Tree(CensusTree).Add(p.ProcessId, root)
+		s.Unlock()
+		if err != nil {
+			return err
+		}
+	*/
+	return fmt.Errorf("TODO")
+}
+
+// PurgeRollingCensus removes a rolling census from the permanent store
+// If the census does not exist, it does nothing.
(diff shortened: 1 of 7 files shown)


vochain/census.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+// PurgeRollingCensus removes a rolling census from the permanent store
+// If the census does not exist, it does nothing.
+func (s *State) PurgeRollingCensus(pid []byte) error {
+	return nil
+}
+
+// GetRollingCensusRoot returns the last rolling census root for a process id
+func (s *State) GetRollingCensusRoot(pid []byte, isQuery bool) ([]byte, error) {
```
ed255, line 34: What is `isQuery` used for?

vochain/census.go
```diff
@@ -0,0 +1,104 @@
...
+	"math/big"
+
+	"github.com/ethereum/go-ethereum/common"
+	"go.vocdoni.io/dvote/crypto/ethereum"
+	models "go.vocdoni.io/proto/build/go/models"
+)
+
+// AddToRollingCensus adds a new key to an existing rolling census.
+// If census does not exist yet it will be created.
+func (s *State) AddToRollingCensus(pid []byte, key []byte, weight *big.Int) error {
```
ed255, line 14: I suggest returning `fmt.Errorf("TODO")` in all unimplemented functions to avoid unexpected behaviors (like someone using this function and forgetting that it was not implemented)

ed255 (changes requested):

vochain/census.go
p4u: isQuery is a boolean which is passed to all State query methods. If true, the Inmutable state will be queried. This avoids querying information which is not yet part of the blockchain (from an ongoing new block which can still be discarted).

vochain/census.go
p4u: done

vochain/census.go
```diff
@@ -0,0 +1,104 @@
...
+	/*
+		// In the state we only store the last census root (as value) using key as index
+		s.Lock()
+		err := s.Store.Tree(CensusTree).Add(p.ProcessId, root)
+		s.Unlock()
+		if err != nil {
+			return err
+		}
+	*/
+	return nil
```
ed255: I think you missed this one return to be `fmt.Errorf("TODO")`.  Otherwise LGTM

ed255 (approved):

PR #244, opened 2021-07-27
multirpc/subpub: set limits on libp2p discovery

mvdan:
(see commit message)

### DIFF
--- a/multirpc/subpub/discovery.go
+++ b/multirpc/subpub/discovery.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"time"
 
+	corediscovery "github.com/libp2p/go-libp2p-core/discovery"
 	"github.com/libp2p/go-libp2p-core/protocol"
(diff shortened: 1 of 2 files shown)


multirpc/subpub/discovery.go
```diff
@@ -79,7 +81,9 @@ func (ps *SubPub) Subscribe(ctx context.Context) {
 		case <-ps.close:
 			return
 		default:
-			ps.discover(ctx)
+			pctx, cancel := context.WithTimeout(ctx, ps.DiscoveryPeriod)
+			ps.discover(pctx)
+			cancel()
```
mvdan, line 86: I added this because we call discover in a loop, so with just the parent ctx, the ps.discover call could hang or get stuck for a long time. I assume we want to set some relatively low timeout here, like DiscoveryPeriod.

p4u:
LGTM

p4u (approved):

PR #247, opened 2021-08-02
Refactor db package with new interface & badgerdb impl

arnaucube:


db/badgerdb/badger.go
```diff
@@ -0,0 +1,145 @@
...
+// method passed into Iterate requires it
+var stopIterations = fmt.Errorf("stop iterations")
+
+func (db *BadgerDB) Iterate(prefix []byte, callback func(k, v []byte) bool) error {
+	err := db.db.View(func(txn *badger.Txn) error {
+		it := txn.NewIterator(badger.DefaultIteratorOptions)
+		defer it.Close()
+		for it.Rewind(); it.Valid(); it.Next() {
+			item := it.Item()
+			k := item.Key()
```
p4u: No need here to store Key() into a var

db/badgerdb/badger.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+}
+
+func (db *BadgerDB) Close() error {
+	return db.db.Close()
+}
+
+// stopIterations is used to stop from iterating in the case that the callback
+// method passed into Iterate requires it
+var stopIterations = fmt.Errorf("stop iterations")
```
p4u: use the `err` prefix for global error vars please

db/badgerdb/badger.go
```diff
@@ -0,0 +1,145 @@
...
+
+func (db *BadgerDB) Iterate(prefix []byte, callback func(k, v []byte) bool) error {
+	err := db.db.View(func(txn *badger.Txn) error {
```
p4u: `return db.db.View(func...)`

vochain/keykeeper/keykeeper.go
```diff
@@ -433,12 +448,21 @@ func (k *KeyKeeper) publishKeys(pk *processKeys, pid string) error {
...
 	defer k.lock.Unlock()
 	dbKey := []byte(dbPrefixProcess + pid)
-	// TODO(mvdan): replace Has with just Get
-	if exists, err := k.storage.Has(dbKey); exists || err != nil {
-		return fmt.Errorf("keys for process %x already exist or error fetching storage: (%s)", pid, err)
+
+	wTx := k.storage.WriteTx()
+	defer wTx.Discard()
+
+	_, err := wTx.Get(dbKey)
```
p4u: `if _, err := wtx.Get(dbKey); err == db.ErrKeynotFound {`

p4u (changes requested):

vochain/keykeeper/keykeeper.go
arnaucube: Ok, but outside the if, there is this line:
```go
return fmt.Errorf("keys for process %x already exist or error fetching storage: (%s)", pid, err)
```
I think that is less simple, but I've added a `var err error` before this line, so the `if _, err:=...; err!=nil` can be used and the error can be returned later outside this if.

db/badgerdb/badger.go
arnaucube: In this case, is not an 'error' as the concept of error, but an aux variable that is returned to indicate to stop iterating, but not necessary because of an error, it can be a desired behavior, like for example wanting to iterate until finding a key X, so when that key X is found it stops iterating by using this aux variable.
Anyway, I've updated the name of the variable to `errStopIterations` as suggested, but is important to have clear that is not used as an error. The description of the variable in the inline code comment should give a notion on how this 'error' is used (not as an error).

db/interface.go
```diff
@@ -1,62 +1,47 @@
...
-	NewBatch() Batch
-	Path() string
-	NewIterator() Iterator
+	// ReadTx returns a ReadTx
+	ReadTx() ReadTx
+	// WriteTx returns a WriteTx
+	WriteTx() WriteTx
+	// Iterate iterates over the key-value pairs executing the given
+	// callback function for each pair. If a prefix is given, it will only
+	// iterate over the key-values with the prefix
```
ed255: Could you extend the documentation saying that when the callback returns false, the iteration is stopped?

db/badgerdb/badger.go
```diff
@@ -0,0 +1,167 @@
...
+	return item.ValueCopy(nil)
+}
+
+// Discard implements the db.ReadTx.Discard interface method
+func (tx ReadTx) Discard() {
+	tx.tx.Discard()
+}
+
+// Get implements the db.WriteTx.Get interface method
+func (tx WriteTx) Get(k []byte) ([]byte, error) {
```
ed255, line 50: This method is already defined for ReadTx but you are not using it.  Yet WriteTx contains ReadTx inside, unused!
The same happens with the Commit method.
I propose two solutions:

1. Keep `ReadTx` inside `WriteTx` and remove the WriteTx.Get and WriteTx.Commit declarations.  For this to work you need to modify the WriteTx constructor to be like this:
```go
// WriteTx returns a db.WriteTx
func (db *BadgerDB) WriteTx() db.WriteTx {
	tx := db.db.NewTransaction(true)
	return WriteTx{
		ReadTx: ReadTx{tx: tx},
		tx:     tx,
	}
}
```

2. Remove `ReadTx` from `WriteTx` because it's not being used, and redeclare `WriteTx.Get` like this (and the same for Commit):
```go
// Get implements the db.WriteTx.Get interface method
func (tx WriteTx) Get(k []byte) ([]byte, error) {
	return ReadTx{tx: tx.tx}.Get(k)
}
```

db/badgerdb/badger.go
```diff
@@ -0,0 +1,167 @@
...
+type BadgerDB struct {
+	db *badger.DB
+}
+
+// check that BadgerDB implements the db.Database interface
+var _ db.Database = (*BadgerDB)(nil)
+
+// NewBadgerDB returns a BadgerDB using the given path, which implements the
+// db.Database interface
+func NewBadgerDB(path string) (*BadgerDB, error) {
```
ed255: I think now is a good moment to replace this single parameter `path` by an options struct that we can grow in the future.  For now this options would have a single field `Path string`, but in the future it can have more fields (like `Compress bool`) without breaking the method interface.

Also we could use this Options struct to have a default MemTableSize that can be changed via the constructor parameters.  Just following the same API design as Badger, but with our own `Options` type.

db/badgerdb/badger.go
```diff
@@ -0,0 +1,167 @@
...
+		if prefix != nil {
+			opts.Prefix = prefix
+		}
+		it := txn.NewIterator(opts)
+		defer it.Close()
+		for it.Rewind(); it.Valid(); it.Next() {
+			item := it.Item()
+			err := item.Value(func(v []byte) error {
+				if cont := callback(item.Key(), v); !cont {
+					return errStopIterations
```
ed255: I think it's weird to introduce this global private variable just to break from the iterations.

Why not do it like this?  I think it's simpler.
```go
// Iterate implements the db.Database.Iterate interface method
func (db *BadgerDB) Iterate(prefix []byte, callback func(k, v []byte) bool) error {
	return db.db.View(func(txn *badger.Txn) error {
		opts := badger.DefaultIteratorOptions
		if prefix != nil {
			opts.Prefix = prefix
		}
		it := txn.NewIterator(opts)
		defer it.Close()
		for it.Rewind(); it.Valid(); it.Next() {
			item := it.Item()
			stopIter := false
			err := item.Value(func(v []byte) error {
				if cont := callback(item.Key(), v); !cont {
					stopIter = true
				}
				return nil
			})
			if stopIter {
				break
			}
			if err != nil {
				return err
			}
		}
		return nil
	})
}
```

db/badgerdb/badger_test.go
```diff
@@ -0,0 +1,129 @@
...
+import (
+	"bytes"
+	"strconv"
+	"testing"
+
+	"go.vocdoni.io/dvote/db"
+)
+
+func TestWriteTx(t *testing.T) {
+	database, err := NewBadgerDB(t.TempDir())
```
ed255: I really like this `t.TempDir()`! :heart:

db/badgerdb/badger_test.go
```diff
@@ -0,0 +1,129 @@
...
+	if err != nil {
```
ed255, line 13: In `storage-proofs-eth-go` we recently started using the `quicktest` library in tests: https://github.com/vocdoni/storage-proofs-eth-go/blob/675817be681e9f3abfb5c385772aa80d2a88bbdb/helpers/helpers_test.go#L9
I also see that it's being used in `vocdoni-node` as well: https://github.com/vocdoni/vocdoni-node/blob/f329f467f6e60ac8fa7f6122d71b51de70895f25/census/census_test.go#L7

I think it's more comfortable for the person that writes the test to use `quicktest`, so I just wanted to make you aware of this in case you want to use `quicktest` for future tests!  (Or this one if you feel like updating it)

db/badgerdb/badger_test.go
ed255: A minor comment: you added the `newTestDB` function but you are not using it here.

vochain/keykeeper/keykeeper.go
```diff
@@ -330,18 +335,23 @@ func (k *KeyKeeper) generateKeys(pid []byte) (*processKeys, error) {
...
 	for pid, height := range k.blockPool {
 		pids := models.StoredKeys{}
 		pkey := []byte(dbPrefixBlock + fmt.Sprintf("%d", height))
 		var data []byte
 		var err error
-		// TODO(mvdan): replace Has+Get with just Get
-		if has, _ := k.storage.Has(pkey); has {
-			data, err = k.storage.Get(pkey)
+		data, err = wTx.Get(pkey)
+		if err != db.ErrKeyNotFound {
```
ed255: Here's a small suggestion (I think it makes the handling of possible cases more clear), but feel free to ignore it if you like the way it is now :smiley::
```go
		data, err = wTx.Get(pkey)
		if err == nil {
			if err := proto.Unmarshal(data, &pids); err != nil {
				log.Errorf("cannot unmarshal process pids for block %d: (%s)", height, err)
			}
		} else if err != db.ErrKeyNotFound {
			// case where Get returns an error, and is not that ErrKeyNotFound
			log.Errorf("cannot get existing list of scheduled reveal processes for block %d", height)
			continue
		}
```

ed255 (changes requested):

db/badgerdb/badger_test.go
arnaucube: Nop, I didn't added, it was there before this PR: https://github.com/vocdoni/vocdoni-node/blob/master/db/badger_test.go#L57
I think that I'll delete the `NewTestDB` method.

db/badgerdb/badger_test.go
arnaucube: Agree, in fact [arbo](https://github.com/vocdoni/arbo) uses only `quicktest` for the tests ^^
In this case, in some `vocdoni-node` packages there is usage of `quicktest` and in some others there is not. In this PR I continued with the approach of each particular package, meaning that if a package was not using quicktest I updated the tests in that approach, and if a package was using quicktest the updates that I did were using quicktest also.
Maybe we can setup a task (low priority) that would be migrate all the tests of vocdoni-node to use quicktest.

db/badgerdb/badger.go
arnaucube: Right! Happened for the Get & Discard methods (not for Commit, that ReadTx does not have it).
Updated the code with the first proposed option.

ed255 (approved):

p4u (approved):

censustree/arbotree/wrapper.go
```diff
@@ -73,7 +73,7 @@ func (t *Tree) TypeString() string {
 // to choose the hash function to be used in the Tree.
 func (t *Tree) Init(name, storageDir string) error {
 	dbDir := filepath.Join(storageDir, "arbotree.db."+strings.TrimSpace(name))
-	database, err := db.NewBadgerDB(dbDir)
+	database, err := badgerdb.NewBadgerDB(badgerdb.Options{Path: dbDir})
```
mvdan: nit: if you rename the package to badgerdb, might as well do `badgerdb.New` to prevent stutter

db/badgerdb/badger.go
```diff
@@ -0,0 +1,153 @@
...
+type ReadTx struct {
+	tx *badger.Txn
+}
+
+// check that ReadTx implements the db.ReadTx interface
+var _ db.ReadTx = (*ReadTx)(nil)
+
+// WriteTx implements the interface db.WriteTx
+type WriteTx struct {
+	ReadTx
```
mvdan: hmm, why would a WriteTx contain a ReadTx? that means a WriteTx actually has two `*badger.Txn` pointers, but I assume it will only use one.

I think it would make sense to avoid this embedding, unless we require it for some reason.

db/badgerdb/badger.go
```diff
@@ -0,0 +1,153 @@
...
+	"os"
+
+	"github.com/dgraph-io/badger/v3"
+	"go.vocdoni.io/dvote/db"
+)
+
+// MEMTABLESIZE defines the BadgerDB maximum size in bytes for memtable table.
+// The default is 64<<20 (64MB), this does not pre-allocate enough memory for
+// big Txs, that's why we use 128<<20.
+const MEMTABLESIZE = 128 << 20
```
mvdan: ALL_CAPS_CONSTANTS is a C/Java thing, in Go you should just go with MemTableSize :)

db/interface.go
```diff
@@ -1,62 +1,51 @@
...
-type Writer interface {
-	Put(key []byte, value []byte) error
+type WriteTx interface {
+	ReadTx
+
+	// Set adds a key-value pair. If the key already exists, it's value is
+	// updated.
+	Set(key []byte, value []byte) error
+	// Del deletes a key and its value.
 	Del(key []byte) error
```
mvdan: I think I'm changing my mind on this name; Delete is barely any longer, and easier to read.

It's not like we are requiring all methods to be super short, e.g. Commit.

mvdan (approved):
SGTM.

I remember seeing some comments about quicktest - I think we agreed we can start using it, and we've gradually been doing that. But there's also no rush to refactor all tests to use it ASAP. So it will be a gradual process as we modify more of the codebase.

mvdan:
Also beware with `it's` vs `its` typos :)

db/badgerdb/badger.go
```diff
@@ -0,0 +1,153 @@
...
+func (tx ReadTx) Discard() {
+	tx.tx.Discard()
+}
+
+// Set implements the db.WriteTx.Set interface method
+func (tx WriteTx) Set(k, v []byte) error {
+	if err := tx.tx.Set(k, v); err != nil {
+		return err
+	}
+	return nil
```
mvdan: nit: you can just `return tx.tx.Set(k, v)`

PR #251, opened 2021-08-12
Feature/tree

arnaucube:
- Add new tree.Tree implementation
- Update the usage of last arbo version (new Root() returned params)
- Add new CensusTree impl, move old censustree to censustreelegacy

The old `censustreelegacy` package is already used in vocdoni-node on this
commit.  It will be deleted in a near future commit, and all the vocdoni-node
code moved to use the new `censustree` package.


tree/tree.go
```diff
@@ -0,0 +1,158 @@
...
+	return &Tree{
+		t:  tree,
+		db: t.db,
+	}, nil
+}
+
+// Dump exports all the Tree leafs in a byte array.
+func (t *Tree) Dump() ([]byte, error) {
+	// TODO NOTE maybe here we want to allow to specify which root to be
+	// used for the Dump, not just the last one
```
arnaucube: @ed255 check note

tree/tree.go
```diff
@@ -0,0 +1,158 @@
...
+// Size returns the number of leafs under the current root
+func (t *Tree) Size(rTx db.ReadTx) uint64 {
+	return 0
+}
+
+// GenProof returns a byte array with the necessary data to verify that the
+// key&value are in a leaf under the current root
+func (t *Tree) GenProof(rTx db.ReadTx, key []byte) ([]byte, error) {
+	// TODO NOTE leafK & leafV will be needed for verifying the proof,
+	// maybe is worth to return them
```
arnaucube: @ed255 check note

db/badgerdb/badger.go
```diff
@@ -48,6 +48,7 @@ func (tx ReadTx) Discard() {
 
 // Get implements the db.WriteTx.Get interface method
 func (tx WriteTx) Get(k []byte) ([]byte, error) {
+	//lint:ignore S1016 TMP
```
arnaucube: @mvdan I've added this *linter ignore* comment as staticcheck was complaining for the `ReadTx{tx: tx.tx}` of the next line. Same for line 72.

mvdan: That lint error seems valid though. You can just do `ReadTx(tx).Get(k)`.

arnaucube: Great, thanks!

p4u (approved):
LGTM (quick check)

tree/tree.go
```diff
@@ -0,0 +1,160 @@
...
+type Tree struct {
+	t  *arbo.Tree
+	db db.Database
+}
+
+// Options is used to pass the parameters to load a new Tree
+type Options struct {
+	// Path of the db.Database to be used
+	Path string
+	// MaxLevels that will have the Tree
```
ed255:
```suggestion
	// MaxLevels that the Tree will have
```
wording edit

tree/tree.go
```diff
@@ -0,0 +1,160 @@
...
+		return nil, err
+	}
+
+	return &Tree{
+		t:  tree,
+		db: database,
+	}, nil
+}
+
+func (t *Tree) DB() db.Database {
```
ed255, line 54: I suggest removing this method.  Do you think this will be used outside of a test?

tree/tree.go
```diff
@@ -0,0 +1,160 @@
...
+// Tree defines the struct that implements the MerkleTree functionalities
+type Tree struct {
+	t  *arbo.Tree
+	db db.Database
+}
+
+// Options is used to pass the parameters to load a new Tree
+type Options struct {
+	// Path of the db.Database to be used
+	Path string
```
ed255: I think passing a db.Database here will suit our needs better.  For example, in the StateDB we will have different trees that use Databases with prefix, so we need to build the Database object outside.

tree/tree.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+
+	return v, nil
+}
+
+// Set adds or updates a key. First performs a Get, and if the key does not
+// exist yet it will call Add, while if the key already exists, it will perform
+// an Update. If the non-existence or existence of the key is already known, is
+// more efficient to directly call Add or Update.
+func (t *Tree) Set(wTx db.WriteTx, key, value []byte) error {
```
ed255, line 76: Right now when you do a Set for a key that already exists, you traverse the tree 2 times: 1st for a get, and then for update.  I suggest changing the code like this as an optimization:
1. Try to call `UpdateWithTx`.  If no error return
2. If error == KeyNotFound, call `Add`.

This way, if the key exists the tree is only traversed once.  I'm not sure if `UpdateWithTx` returns `ErrKeyNotFound` when the key doesn't exist.  If not, I think it should!

tree/tree.go
```diff
@@ -0,0 +1,160 @@
...
+// the indexes of the keys failed to add. Supports empty values as input
+// parameters, which is equivalent to 0 valued byte array.
+func (t *Tree) AddBatch(wTx db.WriteTx, keys, values [][]byte) ([]int, error) {
+	return t.t.AddBatchWithTx(wTx, keys, values)
+}
+
+func (t *Tree) Iterate(callback func(key, value []byte) bool) error {
+	callbackWrapper := func(currLvl int, k, v []byte) bool {
+		callback(k, v)
+		return true
```
ed255:
```suggestion
		return callback(k, v)
```
Otherwise you're ignoring the callback return value, which means the iteration can't be stopped in the middle!

tree/tree.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+// AddBatch adds a batch of key-values to the Tree. Returns an array containing
+// the indexes of the keys failed to add. Supports empty values as input
+// parameters, which is equivalent to 0 valued byte array.
+func (t *Tree) AddBatch(wTx db.WriteTx, keys, values [][]byte) ([]int, error) {
+	return t.t.AddBatchWithTx(wTx, keys, values)
+}
+
+func (t *Tree) Iterate(callback func(key, value []byte) bool) error {
```
ed255: I see that we specified this function the way you implemented.  But I'm thinking maybe it makes more sense to pass a readTx as well?  Like this:
```
Iterate(rTx db.ReadTx, callback func(key, value []byte) bool) error
```
This way we can also iterate on snapshots.  What do you think?

tree/tree.go
```diff
@@ -0,0 +1,160 @@
...
+	"encoding/hex"
+	"fmt"
+
+	"github.com/vocdoni/arbo"
+	"go.vocdoni.io/dvote/db"
+	"go.vocdoni.io/dvote/db/badgerdb"
+)
+
+// Tree defines the struct that implements the MerkleTree functionalities
+type Tree struct {
```
ed255, line 12: I've checked our spec and what I haven't found this, but I think we mentioned it at some point.  Right now this implementation doesn't allow you to pass a nil Tx, so the db inside the tree is never used (except for `Iterate`).  I think we mentioned the idea of allowing passing a nil Tx, and in such a case a Tx would be created internally.  What do you think about this?  I think this would make it more comfortable when we do read queries to the tree.  For example, when querying a censusTree to return via API, or when querying the LastBlock StateDB.

tree/tree.go
```diff
@@ -0,0 +1,160 @@
...
+// Root returns the current root.
+func (t *Tree) Root(rTx db.ReadTx) ([]byte, error) {
+	return t.t.RootWithTx(rTx)
+}
+
+// Size returns the number of leafs under the current root
+func (t *Tree) Size(rTx db.ReadTx) uint64 {
+	// TODO we need to review where this would be used and in which
+	// context, and define the way how this is computed
+	return 0
```
ed255: Considering that this is not yet implemented, I suggest doing a panic here, to make sure that we don't use this function yet, and if we do, detect it immediately, instead of letting this 0 propagate with unknown consequences leading to harder to track bugs.

tree/tree.go
ed255: I understand that right now this function returns only the siblings serialized right?  And that it only returns proofs of existence.

Is `leafK == key`? (Where `key` is in the function arguments).  If so, I don't think we need to return the key, because the user that generates the proof already has the key.

About the `leafV` I think it's worth returning it, as it's needed for verification.

And in any case, I think these extra values should be in different return values (that is, not inside the `[]byte` proof).

tree/tree.go
```diff
@@ -0,0 +1,160 @@
...
+
+// New returns a new Tree, if there already is a Tree in the database, it will
+// load it
+func New(opts Options) (*Tree, error) {
+	database, err := badgerdb.New(badgerdb.Options{Path: opts.Path})
+	if err != nil {
+		return nil, err
+	}
+
+	tree, err := arbo.NewTree(database, opts.MaxLevels, opts.HashFunc)
```
ed255: I'm thinking that maybe we will need the `arbo.NewTree` to optionally accept a writeTx.  Like `arbo.NewTreeWithTx`.  The use case will be the StateDB.  Imagine in two different transactions in the same block you add two leafs in a new tree.  If you open the tree from the DB, the root will not exists, but in the StateDB tx after the first transaction the root already exists.

tree/tree.go
```diff
@@ -0,0 +1,160 @@
...
+	if !existence {
+		// proof of non-existence currently not needed in vocdoni-node
+		return nil, fmt.Errorf("key does not exist %s", hex.EncodeToString(key))
+	}
+	return s, nil
+}
+
+// VerifyProof checks the proof for the given key, value and root, using the
+// hash function of the Tree
+func (t *Tree) VerifyProof(key, value, proof, root []byte) (bool, error) {
```
ed255, line 199: I think it would be nice if we find a way to allow calling this function without using an already opened tree.  For example, I suggest adding a method like this:
```
func VerifyProof(hashFunc HashFunction, key, value, proof, root []byte) (bool, error)
```
And `Tree.VerifyProof` just calls it.  This way we can verify proofs without an opened tree.

tree/tree.go
ed255: Well, right now we can do this:
```
snap, err := tree.FromRoot(root)
if err != nil {
  return err
}
dump, err := snap.Dump()
if err != nil {
  return err
}
```
I think it's good enough!

Note: Would it make sense to pass a readTx to Dump?  Just for consistency?

tree/tree.go
```diff
@@ -0,0 +1,160 @@
...
+// Dump exports all the Tree leafs in a byte array.
+func (t *Tree) Dump() ([]byte, error) {
+	// TODO NOTE maybe here we want to allow to specify which root to be
+	// used for the Dump, not just the last one
+	return t.t.Dump(nil)
+}
+
+// ImportDump imports the leafs (that have been exported with the Dump method)
+// in the Tree.
+func (t *Tree) ImportDump(b []byte) error {
```
ed255, line 223: I have a question regarding how this is supposed to be used.  I understand that the use case is that you open a clean tree for the first time, and then you call ImportDump.  Is that right?

What happens if the tree already has leaves and you do ImportDump?  I think this would be very problematic right?  Because you will end up with a different tree than the exported tree.  I've checked the Arbo implementation and it seems there's no special check.  What about checking that the tree has root == emptyHash, and otherwise return error?  And then specify in the function documentation that the tree must be empty before calling ImportDump?

census/handler.go
```diff
@@ -312,12 +338,20 @@ func (m *Manager) Handler(ctx context.Context, r *api.MetaRequest, isAuth bool,
...
+		dump.RootHash, err = tr.Root()
+		if err != nil {
+			resp.SetError(err.Error())
+			return resp
+		}
+		root, err := tr.Root()
+		if err != nil {
+			resp.SetError(err.Error())
+			return resp
+		}
```
ed255:
```suggestion
		root, err := tr.Root()
		if err != nil {
			resp.SetError(err.Error())
			return resp
		}
		dump.RootHash = root
```

census/handler.go
```diff
@@ -335,7 +369,11 @@ func (m *Manager) Handler(ctx context.Context, r *api.MetaRequest, isAuth bool,
 		}
 		resp.URI = m.RemoteStorage.URIprefix() + cid
 		log.Infof("published census at %s", resp.URI)
-		resp.Root = tr.Root()
+		resp.Root, err = tr.Root()
+		if err != nil {
+			resp.SetError(err.Error())
+			return resp
+		}
```
ed255:
```suggestion
		resp.Root = root
```
I think you can reuse the variable from before

ed255 (changes requested):

tree/tree.go
arnaucube: `leafK` is not the same than the `key` passed as input in the cases where the key does not exist, but the path ends in another leaf, in that case. This is the case to generate a proof of non-existence. So for these cases, the user does not know the key that will be needed to verify the proof before hand.

But, currently we don't use proofs of non-existence. So for the moment we might not need the `leafK`. I'll add the `leafV` that currently is needed to verify the proof. And if in the future we need non-existence proofs we add the `leafK` in the return.

tree/tree.go
arnaucube: Agree

tree/tree.go
arnaucube: Right! Much better this way ^^

tree/tree.go
arnaucube: Right! Updated

tree/tree.go
arnaucube: Agree, thought about this but stuck to the spec. Added the function now

tree/tree.go
arnaucube: Agree, updated

tree/tree.go
arnaucube: I put it in order to create new WriteTx & ReadTx, avoiding to have the db at some struct in upper levels.

census/handler.go
arnaucube: True, updated

tree/tree.go
```diff
@@ -0,0 +1,184 @@
...
+	HashFunc arbo.HashFunction
+}
+
+// New returns a new Tree, if there already is a Tree in the database, it will
+// load it
+func New(wTx db.WriteTx, opts Options) (*Tree, error) {
+	givenTx := true
+	if wTx == nil {
+		givenTx = false
+		wTx = opts.Db.WriteTx()
```
ed255: After every `db.WriteTx()` you should add a `defer wTx.Discard()`.  Otherwise if the function returns with an error before the commit, the resources hold by the wTx are not released!

tree/tree.go
```diff
@@ -0,0 +1,184 @@
...
+
+// DB returns the db.Database from the Tree
+func (t *Tree) DB() db.Database {
+	return t.db
+}
+
+// Get returns the value for a given key.
+func (t *Tree) Get(rTx db.ReadTx, key []byte) ([]byte, error) {
+	if rTx == nil {
+		rTx = t.DB().ReadTx()
```
ed255, line 61: After every `db.ReadTx()` you should add a `defer rTx.Discard()`.  Otherwise after the function returns, the resources hold by the rTx are not released!

tree/tree.go
```diff
@@ -0,0 +1,184 @@
...
+	return v, nil
+}
+
+// Set adds or updates a key. First performs a Get, and if the key does not
+// exist yet it will call Add, while if the key already exists, it will perform
+// an Update. If the non-existence or existence of the key is already known, is
+// more efficient to directly call Add or Update.
+func (t *Tree) Set(wTx db.WriteTx, key, value []byte) error {
+	if wTx == nil {
+		wTx = t.DB().WriteTx()
```
ed255, line 80: The `wTx.Commit()` is missing!  Same with all the other functions that create a `WriteTx`!

ed255 (changes requested):

tree/tree.go
arnaucube: Right!! My fault, updated

tree/tree.go
arnaucube: Right!! My fault, updated

tree/tree.go
arnaucube: Right!! My fault, updated

ed255 (approved):

censustreelegacy/censusTree.go
```diff
@@ -1,4 +1,4 @@
-package censustree
+package censustreelegacy
```
mvdan, line 1: should this have a TODO or issue to be removed, given the "legacy" name?

censustree/censustree.go
```diff
@@ -0,0 +1,41 @@
...
+func New(wTx db.WriteTx, opts tree.Options) (*CensusTree, error) {
+	t, err := tree.New(wTx, opts)
+	if err != nil {
+		return nil, err
+	}
+	return &CensusTree{t, 0}, nil
+}
+
+// Publish makes a merkle tree available for queries.  Application layer should
+// check IsPublish() before considering the Tree available.
```
mvdan: typo: IsPublic

also, you should refer to func names by their name, like FooBar, not their "call expression", like FooBar(). otherwise you'd have to always include the parameters for the sake of consistency.

censustree/censustree.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+// New returns a new CensusTree, if there already is a CensusTree in the
+// database, it will load it.
+func New(wTx db.WriteTx, opts tree.Options) (*CensusTree, error) {
+	t, err := tree.New(wTx, opts)
+	if err != nil {
+		return nil, err
+	}
+	return &CensusTree{t, 0}, nil
```
mvdan: nit: `&CensusTree{Tree: t}`, so you don't have to spell out the rest of the zero fields

tree/tree.go
```diff
@@ -0,0 +1,225 @@
...
+// Tree defines the struct that implements the MerkleTree functionalities
+type Tree struct {
+	t  *arbo.Tree
+	db db.Database
+}
+
+// Options is used to pass the parameters to load a new Tree
+type Options struct {
+	// DB defines the database that will be used for the tree
+	Db db.Database
```
mvdan: nit: either call this Database or DB, but Db is not a common abbreviation

tree/tree.go
```diff
@@ -0,0 +1,225 @@
...
+	"encoding/hex"
+	"fmt"
+
+	"github.com/vocdoni/arbo"
+	"go.vocdoni.io/dvote/db"
+)
+
+// Tree defines the struct that implements the MerkleTree functionalities
+type Tree struct {
+	t  *arbo.Tree
```
mvdan: I'd avoid single-letter fields in general. If I just see `foo.t` 200 lines after the type declaration, it's hard to know if `t` is a tree, or a timestamp, or a token, etc. And the name `tree` is barely any longer.

tree/tree.go
```diff
@@ -0,0 +1,225 @@
...
+		wTx = opts.Db.WriteTx()
+		defer wTx.Discard()
+	}
+	tree, err := arbo.NewTreeWithTx(wTx, opts.Db, opts.MaxLevels, opts.HashFunc)
+	if err != nil {
+		return nil, err
+	}
+
+	if !givenTx {
+		if err = wTx.Commit(); err != nil {
```
mvdan: nit: use `:=` here to keep the err value to a smaller scope

mvdan (approved):
I guess we no longer have a need for a "Tree" interface, as we just have the tree.Tree struct type with some methods?

PR #252, opened 2021-08-19
Remove explicit file logging

vdo:


vdo:
This is needed to use different logging drivers (like fluentd), and implement centralized logging.
File size limits can still be configured in a host level (docker's daemon.json)

p4u (approved):

PR #253, opened 2021-08-30
Add prefixed Database, ReadTx, WriteTx

ed255:


arnaucube (approved):

db/prefixeddb/prefixeddb.go
```diff
@@ -0,0 +1,124 @@
...
+
+// check that PrefixedDatabase implements the db.Database interface
+var _ db.Database = (*PrefixedDatabase)(nil)
+
+// NewPrefixedDatabase creates a new PrefixedDatabase.  If the db is already a
+// PrefixedDatabase, instead of wrapping again, the prefixes are appended to
+// avoid unecessay layers.
+func NewPrefixedDatabase(db db.Database, prefix []byte) *PrefixedDatabase {
+	if pdb, ok := db.(*PrefixedDatabase); ok {
+		return &PrefixedDatabase{append(pdb.prefix, prefix...), pdb.db}
```
mvdan: note that you should make a copy of pdb.prefix, otherwise calling NewPrefixDatabase twice on the same "base" prefix database could overwrite the same underlying array.

something like:

```
// If the underlying DB is already prefixed, join the prefixes to minimise overhead.
if pdb, ok := db.(*PrefixedDatabase); ok {
    // Copy pdb.prefix to ensure we don't reuse the slice's backing array.
    jointPrefix := make([]byte, 0, len(pdb.prefix)+len(prefix))
    jointPrefix = append(jointPrefix, pdb.prefix)
    jointPrefix = append(jointPrefix, prefix)
    prefix = jointPrefix
}

// Make sure the prefix's capacity is fixed, to prevent future appends from sharing memory.
prefix = prefix[:len(prefix):len(prefix)]
return &PrefixedDatabase{prefix: prefix, db: db}
```

That last bit is important to avoid a form of data race.

db/prefixeddb/prefixeddb.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+	return &PrefixedDatabase{prefix, db}
+}
+
+// Notice that this method doesn't close the wrapped db.Database.
```
mvdan: Hmm, why? So it's not possible to close a database once it's prefixed? Seems to me like this should forward the call to the underlying database

db/prefixeddb/prefixeddb.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+// WriteTx returns a db.WriteTx
+func (d *PrefixedDatabase) WriteTx() db.WriteTx {
+	return NewPrefixedWriteTx(d.db.WriteTx(), d.prefix)
+}
+
+// Iterate implements the db.Database.Iterate interface method
+func (d *PrefixedDatabase) Iterate(prefix []byte, callback func(key, value []byte) bool) error {
+	return d.db.Iterate(append(d.prefix, prefix...), func(key, value []byte) bool {
```
mvdan: note that upstream badger does support iterating over a prefix natively. I guess we lose that ability given that our DB interface is too high level. maybe leave a TODO here to support badgerdb shortcuts in the future?

mvdan (approved):
Be careful with reusing slices :)

db/prefixeddb/prefixeddb.go
mvdan: Here's an example: https://play.golang.org/p/e1Slm5Ai-ug

You'll see that the output sometimes has duplicates, because the goroutines "collide" when reusing the same backing array for the prefix slice, since it has enough capacity that the appends happen in-place:

```
foo04
foo06
foo05
foo01
foo09
foo07
foo08
foo02
foo03
foo03
```

This does not hapen if you fix the capacity to its length, so that all appends must copy.

db/prefixeddb/prefixeddb.go
ed255: The Iterate method as implemented for badgerdb uses the native prefix: https://github.com/vocdoni/vocdoni-node/blob/698b3434f1381c656e3907ac60f35461da7aa3a0/db/badgerdb/badger.go#L131
So we are already using the prefix Iteration of badgerdb here.  Am I missing something?

db/prefixeddb/prefixeddb.go
ed255: I wasn't too sure about what to do here.  My original reasoning was that the proper usage should be to always close from the parent Database, because you may close the wrapped Database thinking that the original is still opened, and then you're unable to operate on the original one.  But on the other hand, for Commit and Discard I do the action on the wrapped Tx.  So I'll just apply your suggestion.

db/prefixeddb/prefixeddb.go
ed255: Thanks for spotting this!

db/prefixeddb/prefixeddb.go
mvdan: Yup, as you say, either all these methods should apply to the parent, or none of them should.

db/prefixeddb/prefixeddb.go
mvdan: Ah, you're right, I forgot this.

PR #255, opened 2021-09-02
Add crypo/zk/artifacts circuit files downloader

arnaucube:
#### Directories structure
Path: circuitname/network/parameters/file.extension
- Circuit Name: the circuit name. Example: `zkcensusproof`
- Network: The network name. Example: `main, dev, stage`
- Circuit Parameters: numbers in decimal separated by underscore. Example: `1024`, `1024_3_500`
- Files:
	- circuit.zkey: contains the provingKey & verificationKey (big size, linear with circuit
	size) [needed by the user]
	- verificationKey.json: extracted from circuit.zkey (small size, constant size) [needed
	by the vochain]
	- witness.wasm: (medium size, linear with circuit size) [needed by the user]
- Examples:
	- zkcensusproof/stage/1024/circuit.zkey
	- zkcensusproof/dev/128/witness.wasm

#### Files hash
Each file is hashed with sha256, which output is publicly known. The hash
values of each file are hardcoded in the VochainGenesis.

### DIFF
--- a/crypto/zk/artifacts/artifacts.go
+++ b/crypto/zk/artifacts/artifacts.go
@@ -1,0 +1,199 @@
+package artifacts
+
+import (
+	"bytes"
+	"context"
+	"crypto/sha256"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"os"
+	"path"
+	"path/filepath"
+)
+
+/*
+# Directories structure
+=======================
+Path: circuitname/network/parameters/file.extension
+- Circuit Name: the circuit name. Example: `zkcensusproof`
+- Network: The network name. Example: `main, dev, stage`
+- Circuit Parameters: numbers in decimal separated by underscore. Example: `1024`, `1024_3_500`
+- Files:
+	- circuit.zkey: contains the provingKey & verificationKey (big size, linear with circuit
+	size) [needed by the user]
+	- verificationKey.json: extracted from circuit.zkey (small size, constant size) [needed
+	by the vochain]
+	- witness.wasm: (medium size, linear with circuit size) [needed by the user]
+- Examples:
+	- zkcensusproof/stage/1024/circuit.zkey
+	- zkcensusproof/dev/128/witness.wasm
+
+# Files hash
+=============
+Each file is hashed with sha256, which output is publicly known. The hash
+values of each file are hardcoded in the VochainGenesis.
+
+*/
+
+const (
+	// FilenameWitness defines the name of the file of the WitnessCalculator
+	FilenameWitness = "witness.wasm"
+	// FilenameZKey defines the name of the file of the circom ZKey
+	FilenameZKey = "circuit.zkey"
+	// FilenameVK defines the name of the VerificationKey.json
+	FilenameVK = "verificationKey.json"
+)
+
+// CircuitsConfig defines the configuration of the files to be downloaded
+type CircuitsConfig struct {
+	// URL defines the URL from where to download the files
+	URL string
+	// CircuitsPath defines the path from where the files are downloaded
+	CircuitsPath string
+	// LocalDir defines in which directory will be the files
+	// downloaded, under that directory it will follow the CircuitsPath
+	// directories structure
+	LocalDir string
+
+	// WitnessHash contains the expected hash for the file filenameWitness
+	WitnessHash []byte // witness.wasm
+	// ZKeyHash contains the expected hash for the file filenameZKey
+	ZKeyHash []byte // circuit.zkey
+	// VKHash contains the expected hash for the file filenameVK
+	VKHash []byte // verificationKey.json
+}
+
+// DownloadCircuitFiles will download the circuits in the specified path,
+// checking the expected sha256 hash of each file. If the files already exist
+// in the path, it will not download them again but it will check the hash of
+// the existing files.
+func DownloadCircuitFiles(ctx context.Context, c CircuitsConfig) error {
+	// check if files already exist, if files already exist, check hashes
+	err := checkHashes(c)
+	if err == nil {
+		// files already exist, and match the expected hashes
+		return nil
+	}
+	if err != nil && !os.IsNotExist(err) {
+		return err
+	}
+
+	// at this point, or files do not exist, or exist with a not expected
+	// hash
+	return downloadFiles(ctx, c)
+}
+
+func checkHashes(c CircuitsConfig) error {
+	err := checkHash(filepath.Join(c.LocalDir, c.CircuitsPath, FilenameWitness), c.WitnessHash)
+	if err != nil {
+		return err
+	}
+
+	err = checkHash(filepath.Join(c.LocalDir, c.CircuitsPath, FilenameZKey), c.ZKeyHash)
+	if err != nil {
+		return err
+	}
+
+	err = checkHash(filepath.Join(c.LocalDir, c.CircuitsPath, FilenameVK), c.VKHash)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func checkHash(path string, expected []byte) error {
+	f, err := os.Open(path)
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+
+	h := sha256.New()
+	if _, err := io.Copy(h, f); err != nil {
+		return err
+	}
+	hash := h.Sum(nil)
+
+	if !bytes.Equal(hash[:], expected) {
+		return fmt.Errorf("path: %s, expected hash: %x, computed hash: %x",
+			path, expected, hash)
+	}
+	return nil
+}
+
+// downloadFiles downloads the files, and once downloaded calls checkHashes
+func downloadFiles(ctx context.Context, c CircuitsConfig) error {
+	if err := os.MkdirAll(filepath.Join(c.LocalDir, c.CircuitsPath), os.ModePerm); err != nil {
+		return err
+	}
+
+	u, err := url.Parse(c.URL)
+	if err != nil {
+		return err
+	}
+	u.Path = path.Join(u.Path, c.CircuitsPath, FilenameWitness)
+	if err := downloadFile(ctx,
+		u.String(),
+		filepath.Join(c.LocalDir, c.CircuitsPath, FilenameWitness),
+	); err != nil {
+		return err
+	}
+
+	u, err = url.Parse(c.URL)
+	if err != nil {
+		return err
+	}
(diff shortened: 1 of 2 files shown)


crypto/zk/artifacts/artifacts.go
```diff
@@ -0,0 +1,175 @@
...
+
+*/
+
+const (
+	// FilenameWitness defines the name of the file of the WitnessCalculator
+	FilenameWitness string = "witness.wasm"
+	// FilenameZkey defines the name of the file of the circom ZKey
+	FilenameZkey string = "circuit.zkey"
+	// FilenameVk defines the name of the VerificationKey.json
+	FilenameVk string = "verificationKey.json"
```
ed255:
```suggestion
	// FilenameVK defines the name of the VerificationKey.json
	FilenameVK string = "verificationKey.json"
```
I think we should follow https://github.com/golang/go/wiki/CodeReviewComments#initialisms

crypto/zk/artifacts/artifacts.go
```diff
@@ -0,0 +1,175 @@
...
+Each file is hashed with sha256, which output is publicly known. The hash
+values of each file are hardcoded in the VochainGenesis.
+
+*/
+
+const (
+	// FilenameWitness defines the name of the file of the WitnessCalculator
+	FilenameWitness string = "witness.wasm"
+	// FilenameZkey defines the name of the file of the circom ZKey
+	FilenameZkey string = "circuit.zkey"
```
ed255:
```suggestion
	// FilenameZKey defines the name of the file of the circom ZKey
	FilenameZKey string = "circuit.zkey"
```
I think this was a typo, because you then wrote: "...file of the circom **ZKey**" (emphasis mine)

crypto/zk/artifacts/artifacts.go
```diff
@@ -0,0 +1,175 @@
...
+	CircuitsPath string
+	// LocalDir defines in which directory will be the files
+	// downloaded, under that directory it will follow the CircuitsPath
+	// directories structure
+	LocalDir string
+
+	// WitnessHash contains the expected hash for the file filenameWitness
+	WitnessHash []byte // witness.wasm
+	// ZkeyHash contains the expected hash for the file filenameZkey
+	ZkeyHash []byte // circuit.zkey
```
ed255:
```suggestion
	// ZKeyHash contains the expected hash for the file filenameZKey
	ZKeyHash []byte // circuit.zkey
```

crypto/zk/artifacts/artifacts.go
```diff
@@ -0,0 +1,175 @@
...
+	// VkHash contains the expected hash for the file filenameVk
+	VkHash []byte // verificationKey.json
```
ed255:
```suggestion
	// VKHash contains the expected hash for the file filenameVK
	VKHash []byte // verificationKey.json
```

crypto/zk/artifacts/artifacts.go
```diff
@@ -0,0 +1,175 @@
...
+// in the path, it will not download them again but it will check the hash of
+// the existing files.
+func DownloadCircuitFiles(c CircuitsConfig) error {
+	// check if files already exist, if files already exist, check hashes
+	err := checkHashes(c)
+	if err == nil {
+		// files already exist, and match the expected hashes
+		return nil
+	}
+	if err != nil && !strings.Contains(err.Error(), "no such file or directory") {
```
ed255: Please, use https://pkg.go.dev/os#IsNotExist instead of string matching.

crypto/zk/artifacts/artifacts.go
```diff
@@ -0,0 +1,175 @@
...
+
+	h := sha256.New()
+	if _, err := io.Copy(h, f); err != nil {
+		return err
+	}
+	hash := h.Sum(nil)
+
+	if !bytes.Equal(hash[:], expected) {
+		return fmt.Errorf("path: %s, expected hash: %s, computed hash: %s",
+			path, hex.EncodeToString(expected), hex.EncodeToString(hash[:]))
```
ed255:
```suggestion
		return fmt.Errorf("path: %s, expected hash: %x, computed hash: %x",
			path, expected, hash[:])
```
Did you know about this?

crypto/zk/artifacts/artifacts.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+	return nil
+}
+
+// downloadFiles downloads the files, and once downloaded calls checkHashes
+func downloadFiles(c CircuitsConfig) error {
+	err := os.MkdirAll(c.LocalDir+c.CircuitsPath, os.ModePerm)
+	if err != nil {
```
ed255: why not like this? (same for the following ones):
```go
if err := foo(); err ! nil {
```

crypto/zk/artifacts/artifacts.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+
+	// at this point, or files do not exist, or exist with a not expected
+	// hash
+	return downloadFiles(c)
+}
+
+func checkHashes(c CircuitsConfig) error {
+	err := checkHash(c.LocalDir+c.CircuitsPath+FilenameWitness, c.WitnessHash)
+	if err != nil {
```
ed255: why not like this? (same for the following ones):
```go
if err := foo(); err ! nil {
```

crypto/zk/artifacts/artifacts.go
```diff
@@ -0,0 +1,175 @@
...
+}
+
+func downloadFile(fromUrl, toPath string) error {
+	out, err := os.Create(toPath)
+	if err != nil {
+		return err
+	}
+	defer out.Close()
+
+	resp, err := http.Get(fromUrl)
```
ed255: I think it would be much better if we can have a timeout (maybe 1 minute?  Or a reasonable value for the expected size of the files, we can adjust it in the future) for the http request.

I think this should work https://stackoverflow.com/a/25344458

Another option would be to pass a context and decide the timeout somewhere else.  Like this:
```go
req, _ := http.NewRequest("GET", fromUrl, nil)
req = req.WithContext(ctx)
res, error := client.Do(req)
```
And somewhere else there would be:
```go
ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
defer cancel()
downloadFile(ctx, fromUrl, toPath)
```

ed255 (changes requested):

crypto/zk/artifacts/artifacts.go
arnaucube: I didn't put it in this way in purpose because if not the lines get too long and need to be cropped at some point and gets even more confusing, for example:
```go
if err := downloadFile(ctx, c.URL+c.CircuitsPath+FilenameWitness, c.LocalDir+c.CircuitsPath+FilenameWitness); err!=nil {
```
this is too long (116 characters), then could be cropped like:
```go
if err := downloadFile(ctx, c.URL+c.CircuitsPath+FilenameWitness,
        c.LocalDir+c.CircuitsPath+FilenameWitness); err!=nil {
````
but then is like it takes too long and it's harder to follow.

Then, there are cases where is quite direct to use `if err:= foo(); err!=nil {` as the line is quite short, but then which criteria we use? Like, in the short lines we use it and not in the long ones? or we use it on all the lines even if the code gets more obfuscated and harder to read?

File: crypto/zk/artifacts/artifacts.go (hunk 8 above)
Comment (arnaucube) [reply crypto/zk/artifacts/artifacts.go]: https://github.com/vocdoni/vocdoni-node/pull/255#discussion_r701018497

File: crypto/zk/artifacts/artifacts.go (hunk 7 above)
Comment (ed255) [reply crypto/zk/artifacts/artifacts.go]: I personally prefer always using `if err:= foo(); err!=nil {` even if the line becomes long and I have to split it.  But I totally understand your point.  So for me this conversation is resolved in the sense that we don't have a specific criteria.  I feel this is a very personal opinion.  Maybe @mvdan can give an opinion here?

File: crypto/zk/artifacts/artifacts.go (hunk 9 above)
Comment (arnaucube) [reply crypto/zk/artifacts/artifacts.go]: Cool! Done (using directly `http.NewRequestWithContext` https://pkg.go.dev/net/http#NewRequestWithContext )

**Review by ed255 — APPROVED** [review]

File: crypto/zk/artifacts/artifacts.go [hunk 10]
```diff
@@ -0,0 +1,179 @@
...
+
+*/
+
+const (
+	// FilenameWitness defines the name of the file of the WitnessCalculator
+	FilenameWitness string = "witness.wasm"
+	// FilenameZKey defines the name of the file of the circom ZKey
+	FilenameZKey string = "circuit.zkey"
+	// FilenameVK defines the name of the VerificationKey.json
+	FilenameVK string = "verificationKey.json"
```
mvdan: nit: best to leave constants untyped in general, it's easier and more flexible - just `const foo = "value"`

crypto/zk/artifacts/artifacts.go
```diff
@@ -0,0 +1,179 @@
...
+		return err
+	}
+
+	// at this point, or files do not exist, or exist with a not expected
+	// hash
+	return downloadFiles(ctx, c)
+}
+
+func checkHashes(c CircuitsConfig) error {
+	err := checkHash(c.LocalDir+c.CircuitsPath+FilenameWitness, c.WitnessHash)
```
mvdan: you should be using filepath.Join for paths on disk; otherwise this is not very portable outside unix-like systems

crypto/zk/artifacts/artifacts.go
mvdan: I agree with @ed255. If the line is too long, the added `if ...; err != nil {` doesn't really make it significantly worse. And you should split up the func call between multiple lines, anyway.

You can also do it in a way that's readable:

```
if err := downloadFile(ctx,
        c.URL+c.CircuitsPath+FilenameWitness,
        c.LocalDir+c.CircuitsPath+FilenameWitness,
); err != nil {
```

crypto/zk/artifacts/artifacts.go
```diff
@@ -0,0 +1,179 @@
...
+	if err != nil {
+		return err
+	}
+	defer out.Close()
+
+	req, err := http.NewRequestWithContext(ctx, "GET", fromUrl, nil)
+	if err != nil {
+		return err
+	}
+	client := &http.Client{}
```
mvdan: just use http.DefaultClient if it's empty?

mvdan (commented):
SGTM

arnaucube:
Amended with @mvdan suggestions

PR #256, opened 2021-09-02
Implement dynamic StateDB using Arbo

ed255:
Also:
- Move current statedb to statedblegacy

Note1: I've left the following functions in statedb_test.go: `treePrint` and `dumpPrint`.  I've used these functions during testing to verify that the key-values in the db are stored as I expect them via manual inspection.  I haven't deleted the functions because I think they can be useful in the future (for debugging if we add more tests), and because they can be useful to the reviewer to understand what's going on in the tests.  But I'm open to suggestions of a better approach about what to do with these functions.

Note2: @arnaucube please while reviewing this PR, check the document we made with the specification of the StateDB to verify that I've followed the agreements that we wrote down and that I haven't changed anything important.


statedb/statedb.go
```diff
@@ -1,35 +1,720 @@
+/*
+Package statedb contains the implementation of the StateDB, a database backed
```
mvdan:
```suggestion
Package statedb contains the implementation of StateDB, a database backed
```

statedb/statedb.go
```diff
@@ -1,35 +1,720 @@
...
+	return b
+}
+
+func bytesToUint32(b []byte) uint32 {
+	return binary.LittleEndian.Uint32(b)
+}
+
+// join two paths with a '/' in between
+func join(pathA, pathB []byte) []byte {
+	return appendCopy(appendCopy(pathA, separator), pathB...)
```
mvdan: you're calling `join` from many places, so it's unfortunate that it allocates multiple times. you should be able to allocate once, for `len(pathA)+len(pathB)+1`.

statedb/statedb.go
```diff
@@ -1,35 +1,720 @@
...
-}
-
-type StateTree interface {
+func uint32ToBytes(v uint32) []byte {
+	b := make([]byte, 4)
+	binary.LittleEndian.PutUint32(b, v)
+	return b
+}
+
+func bytesToUint32(b []byte) uint32 {
```
mvdan: this func just saves you a few keystrokes, and it's only used once :) inline it?

statedb/statedb.go
```diff
@@ -1,35 +1,720 @@
...
+	return appended[:len(appended):len(appended)]
+}
+
+var separator byte = '/'
+
+var (
+	subKeyTree    []byte = []byte("t")
+	subKeyMeta    []byte = []byte("m")
+	subKeyNoState []byte = []byte("n")
+	subKeySubTree []byte = []byte("s")
```
mvdan: you can omit the first `[]byte` type here in all these

statedb/statedb.go
```diff
@@ -1,35 +1,720 @@
...
+func appendCopy(s []byte, v ...byte) []byte {
+	// Copy s to ensure we don't reuse the slice's backing array.
+	appended := make([]byte, 0, len(s)+len(v))
+	appended = append(appended, s...)
+	appended = append(appended, v...)
+	// Make sure the appended's capacity is fixed, to prevent future appends from sharing memory.
+	return appended[:len(appended):len(appended)]
+}
+
+var separator byte = '/'
```
mvdan: might as well make this a const

statedb/statedb.go
```diff
@@ -1,35 +1,720 @@
...
+// affect the cryptographic integrity of the StateDB.  Writing to this database
+// won't change the StateDB.Root.
+func (u *TreeUpdate) NoState() Updater {
+	return subWriteTx(u.tx, subKeyNoState)
+}
+
+// Add a new key-value to this tree.  `key` is the path of the leaf, and
+// `value` is the content of the leaf.
+func (u *TreeUpdate) Add(key, value []byte) error {
+	u.dirtyTree = true
```
mvdan: if these methods are meant to be safe for concurrent use (e.g. calling Add/Set from many goroutines), you could use sync/atomic for this field

statedb/statedb.go
mvdan: also, a thought for you: if you used string instead of []byte for prefixes and separators/paths, then your code would get simpler. you'd use `path.Join` instead of `join`, and `appendCopy` would be unnecessary as Go strings are immutable so additions won't overwrite memory.

statedb/statedb.go
```diff
@@ -1,35 +1,720 @@
...
+
+// TreeUpdate is an opened tree that can be updated.  All updates are stored in
+// an internal transaction (shared with all the opened subTrees) and will only
+// be commited with the TreeTx.Commit function is called.
+type TreeUpdate struct {
+	tx        db.WriteTx
+	txTree    db.WriteTx
+	dirtyTree bool
+	tree      *tree.Tree
+	openSubs  map[string]*TreeUpdate
```
mvdan: I think dirtyTree and openSubs need documenting, as it's not clear why they are needed

statedb/statedb.go
```diff
@@ -1,35 +1,720 @@
...
+func (v *TreeView) SubTree(key []byte, c *SubTreeConfig) (*TreeView, error) {
+	return v.subTree(c.treeConfig(key))
+}
+
+// TreeUpdate is an opened tree that can be updated.  All updates are stored in
+// an internal transaction (shared with all the opened subTrees) and will only
+// be commited with the TreeTx.Commit function is called.
+type TreeUpdate struct {
+	tx        db.WriteTx
+	txTree    db.WriteTx
```
mvdan: if TreeUpdate has two WriteTx fields, I'd perhaps not call one of them tx, as that feels like a footgun. txTree also doesn't tell me much about how it's different. If we need two for a good reason, give them good names?

statedb/statedb.go
ed255: The TreeUpdate and TreeTx are not safe for concurrent use (it's not in our use case); so I've added that to the documentation of the types.

statedb/statedb.go
ed255: I've moved the txTree to a new structure:
```go
// treeWithTx is an Arbo merkle tree with the db.WriteTx used for updating it.
type treeWithTx struct {
	*tree.Tree
	tx db.WriteTx
}
```
So now we have:
```go
TreeUpdate.tx // the tx of the TreeUpdate
TreeUpdate.tree.tx // the tx of the Arbo Tree in TreeUpdate
```
NOTE: [`Arbo`](https://github.com/vocdoni/arbo) is package that implements the tree type we have wrapped in `tree.Tree`

statedb/statedb.go
ed255: I've replaced the exposed `[]byte` types by strings, and simplified a few things following your advice :)

statedb/statedb.go
ed255: these are now const strings.

statedb/statedb.go
```diff
@@ -1,35 +1,741 @@
...
+used for zkSNARKS, where the number of levels is of critical matter.  For
+example, we may want to build a census tree of babyjubjub public keys that will
+be used to prove ownership of the public key via a SNARK in order to vote.  If
+we build the merkle tree using the public key as path, we will have an
+unbalanced tree which requires more levels than strictly necessary.  On the
+other hand, if we use a sequential index as path and set the value to the
+public key, we achieve maximum balancing reducing the number of tree levels.
+But then we can't easily query for the existence of a public key in the tree to
+generate a proof.  In such a case, we can store the mapping of public key to
+index in the NoState database.
```
arnaucube: Small clarification, that can not easily be queried because would be accessed by index and not bye public key:
```suggestion
But then we can't easily query for the existence of a public key in the tree to
generate a proof, as it requires to know its index.  In such a case, we can
store the mapping of public key to index in the NoState database.
```
(the only addition of the suggestion is `, as it requires to know its index` the rest is diff because of the jump lines)

ed255:
@arnaucube currently We have the following method for StateDB:
```go
// Root returns the root of the StateDB, which corresponds to the root of the
// mainTree.  This root is a hash that cryptographically represents the entire
// StateDB (except for the NoState databases).
func (s *StateDB) Root() ([]byte, error) {
```

That's following the spec we agreed upon.  I'm thinking maybe it makes more sense to rename this to Hash.  So the StateDB.Hash is the Root of the mainTree.  What do you think?

statedb/statedb.go
```diff
@@ -1,35 +1,741 @@
...
+type TreeTx struct {
+	sdb *StateDB
+	// TreeUpdate contains the mainTree opened for updates.
+	TreeUpdate
+}
+
+// update is a helper struct used to collect subTree updates that need to
+// update the parent's corresponding leaf with the new root.
+type update struct {
+	// key     []byte
```
arnaucube: is `key` unused?

arnaucube (commented):
`statedb.go` is 760 lines with lot of different structs and methods, maybe a future iteration (not necessarily in this PR) could be organizing the package in different files grouping structs&methods, like for example all the structs & methods related to trees in one files and the structs & methods related to statedb (without trees) in another file.

arnaucube:
> @arnaucube currently We have the following method for StateDB:
> 
> ```go
> // Root returns the root of the StateDB, which corresponds to the root of the
> // mainTree.  This root is a hash that cryptographically represents the entire
> // StateDB (except for the NoState databases).
> func (s *StateDB) Root() ([]byte, error) {
> ```
> 
> That's following the spec we agreed upon. I'm thinking maybe it makes more sense to rename this to Hash. So the StateDB.Hash is the Root of the mainTree. What do you think?

Make sense, agree (explaining the reasoning in the method comment)

arnaucube (approved):

PR #257, opened 2021-09-08
Add missing dump.Type to census publish

arnaucube:
This branch is an attempt to hot-fix a bug found by @NateWilliams2 & spotted by @emmdim .

This branch adds the `dump.Type = tr.Type()` inside the `publish` case.
But, I'm not sure which is the expected behavior & connection between the `dump` & `publish` & `importDump`, as in the `publish` the dump specifies the type of the tree, but in the import the tree type is defined outside the scope, thus if the dump type is different than the existing one in that scope would not match and the import dump would not be correct.

### DIFF
--- a/census/handler.go
+++ b/census/handler.go
@@ -320,6 +320,7 @@ func (m *Manager) Handler(ctx context.Context, r *api.MetaRequest, isAuth bool,
 			log.Warnf("cannot dump census with root %x: %s", tr.Root(), err)
 			return resp
 		}
+		dump.Type = tr.Type()
 		dumpBytes, err := json.Marshal(dump)
 		if err != nil {
 			resp.SetError(err)


emmdim:
@arnaucube  great. Maybe @NateWilliams2  can review for you?
In order to test it i would cherry-pick it at stage. Do you think it makes sense?

arnaucube:
> @arnaucube great. Maybe @NateWilliams2 can review for you?
> In order to test it i would cherry-pick it at stage. Do you think it makes sense?

Agree with all of this

emmdim:
> But, I'm not sure which is the expected behavior & connection between the `dump` & `publish` & `importDump`, as in the `publish` the dump specifies the type of the tree, but in the import the tree type is defined outside the scope, thus if the dump type is different than the existing one in that scope would not match and the import dump would not be correct.

@arnaucube 
Example of the error log:
`2021-09-08T11:58:07Z	WARN	census/importqueue.go:143	cannot import census 1c37356bdf2c8ae17ba1ad673f895fe3aa83c47dcd2d2d8e905c90c42c91f170: (cannot create new census namespace: (unrecognized tree type (0)))`

I assume this in the end leads to this line:
https://github.com/vocdoni/vocdoni-node/blob/698b3434f1381c656e3907ac60f35461da7aa3a0/census/importqueue.go#L33

arnaucube:
Added an issue to add tests around this behavior to try to avoid the current situation in the future https://github.com/vocdoni/vocdoni-node/issues/258

NateWilliams2 (approved):

PR #260, opened 2021-09-10
Integrate Arbo-based StateDB in vochain

ed255:
Integrate Arbo-based StateDB in vochain

Replace the IAVL-based StateDB used to store the vochain state by the new
Arbo-based StateDB.  This change doesn't introduce new features.

Due to the change of the vochain state underlying structure, even if the same
data is stored in the StateDB, the calculated hash of the state will be
completely different and incompatible with the previous one.  This means that
this change is not compatible with a chain built with the old StateDB, and
requires a chain reset to work.

Change notes:
- In statedb, remove SubTreeSingleConfig type and SubTreeSingle method
  in favour of a single config type (renamed to TreeConfig) and method.
  Introduced a new type TreeNonSingleConfig that becomes a TreeConfig
  with the method WithKey, in order to keep different types for
  singleton trees configuration and non-singleton trees configuration.
  So now, we open a singleton tree like this:
  `mainTree.SubTree(OraclesCfg)` and a non-singleton tree like this:
  `processesTree.SubTree(CensusCfg.Key(pid))`.  This type unification
  allows creating a slice of trees configuration (no matter if they are
  singleton or non-singleton) to easily access nested subtrees.
- Added new methods to the TreeUpdate/TreeView in statedb to make it
  simpler to query nested trees: `DeepSubTree`, `DeepGet`, `DeepAdd` and
  `DeepSet`.
- Replace VoteID = processID + nullifier by hash(processID + nullifier).  The
  old VoteID was longer than 32 bytes, and that made it not friendly to be used
  as a key in Arbo.  By hashing we obtain a 32byte that suits an Arbo key.
- In VotesTree, replace storage format from (key:VoteID, value:VoteHash) to
  (key:VoteID, value:StateDBVote) where StateDBVote is a new protobuf model
  which contains { VoteHash, ProcessID, Nullifier }.  This is required because
  now the VoteID is a hash of processID and nullifier, and thus from the VoteID
  we can't recover the nullifier, and we need a way to obtain the nullifier of
  a stored vote.
- In ProcessesTree, replace storage format from (key: ProcessID, value:
  Process) to (key: ProcessID, value: StateDBProcess) where StateDBProcess is a
  new protobuf model which contains { Process, VotesRoot }.  This is required
  because now there is a VotesTree for each Processes stored in the StateDB.
  To query the VotesTree at diferent blocks we need to store it's root next to
  the Process.
- New hierarchy in the StateDB. In bold are trees.
    - Before:
        - 1 **AppTree**
            - 1 OracleList
            - 1 ValidatorList
            - 1 Header
        - 1 **ProcessTree**
            - N Processes indexed by ProcessID
        - 1 **VoteTree**
            - N Votes indexed by ProcessID + Nullifier
    - After:
        - 1 **MainTree**
            - 1 **OracleTree**
                - N Oracles indexed by address
            - 1 **ValidatorTree**
                - N Validators indexed by address
            - 1 Header
            - 1 **ProcessesTree**
                - N Processes indexed by ProcessID
                    - 1 **VotesTree**
                        - N Votes indexed by VoteID

### DIFF
--- a/statedb/statedb.go
+++ b/statedb/statedb.go
@@ -95,13 +95,6 @@ func subReadTx(tx db.ReadTx, path string) db.ReadTx {
 	return prefixeddb.NewPrefixedReadTx(tx, []byte(path+"/"))
 }
 
-// treeConfig is a unified configuration for a subTree.
-type treeConfig struct {
-	parentLeafKey []byte
-	prefix        string
-	*SubTreeConfig
-}
-
 // GetRootFn is a function type that takes a leaf value and returns the contained root.
 type GetRootFn func(value []byte) ([]byte, error)
 
@@ -109,8 +102,28 @@ type GetRootFn func(value []byte) ([]byte, error)
 // leaf value with the new root and returns it.
 type SetRootFn func(value []byte, root []byte) ([]byte, error)
 
-// SubTreeConfig
-type SubTreeConfig struct {
+// TreeParams are the parameters used in the constructor for a tree configuration.
+type TreeParams struct {
+	// HashFunc is the hash function used in the merkle tree
+	HashFunc arbo.HashFunction
+	// KindID is a unique identifier that specifies what kind of tree this
+	// is.  Diferent kind of trees under the same parent tree must have
+	// different KindIDs, as this parameter is used to build a disjoint
+	// database prefix for each subTree under the same parent tree.
+	KindID string
+	// MaxLevels is the maximum number of merkle tree levels allowed.
+	MaxLevels int
+	// ParentLeafGetRoot is the function that takes a leaf value of the
+	// parent at which this tree hangs, and returns the root of this tree.
+	ParentLeafGetRoot GetRootFn
+	// ParentLeafSetRoot is the function that takes a leaf value of the
+	// parent at which this tree hangs, and updates it (returning it) with
+	// the new root of this tree.
+	ParentLeafSetRoot SetRootFn
+}
+
+// TreeNonSingletonConfig contains the configuration used for a non-singleton subTree.
+type TreeNonSingletonConfig struct {
 	hashFunc          arbo.HashFunction
 	kindID            string
 	parentLeafGetRoot GetRootFn
@@ -118,61 +131,70 @@ type SubTreeConfig struct {
 	maxLevels         int
 }
 
-// SubTreeSingleConfig contains the configuration used for a non-singleton subTree.
-func NewSubTreeConfig(hashFunc arbo.HashFunction, kindID string, maxLevels int,
-	parentLeafGetRoot GetRootFn, parentLeafSetRoot SetRootFn) *SubTreeConfig {
-	return &SubTreeConfig{
-		hashFunc:          hashFunc,
-		kindID:            kindID,
-		parentLeafGetRoot: parentLeafGetRoot,
-		parentLeafSetRoot: parentLeafSetRoot,
-		maxLevels:         maxLevels,
+// NewTreeNonSingletonConfig creates a new configuration for a non-singleton subTree.
+func NewTreeNonSingletonConfig(params TreeParams) *TreeNonSingletonConfig {
+	return &TreeNonSingletonConfig{
+		hashFunc:          params.HashFunc,
+		kindID:            params.KindID,
+		parentLeafGetRoot: params.ParentLeafGetRoot,
+		parentLeafSetRoot: params.ParentLeafSetRoot,
+		maxLevels:         params.MaxLevels,
 	}
 }
 
-// treeConfig returns a unified configuration type for opening a singleton
-// subTree that is identified by `id`.  `id` is the path in the parent tree to
-// the leaf that contains the subTree root.
-func (c *SubTreeConfig) treeConfig(id []byte) *treeConfig {
-	return &treeConfig{
-		parentLeafKey: id,
-		prefix:        c.kindID + string(id),
-		SubTreeConfig: c,
+// HashFunc returns the hashFunc set for this SubTreeConfig
+func (c *TreeNonSingletonConfig) HashFunc() arbo.HashFunction {
+	return c.hashFunc
+}
+
+// WithKey returns a unified subTree configuration type for opening a singleton
+// subTree that is identified by `key`.  `key` is the path in the parent tree
+// to the leaf that contains the subTree root.
+func (c *TreeNonSingletonConfig) WithKey(key []byte) TreeConfig {
+	return TreeConfig{
+		parentLeafKey:          key,
+		prefix:                 c.kindID + string(key),
+		TreeNonSingletonConfig: c,
 	}
 }
 
-// SubTreeSingleConfig contains the configuration used for a singleton subTree.
-type SubTreeSingleConfig struct {
-	*SubTreeConfig
+// TreeConfig contains the configuration used for a subTree.
+type TreeConfig struct {
+	parentLeafKey []byte
+	prefix        string
+	*TreeNonSingletonConfig
 }
 
-// NewSubTreeSingleConfig creates a new SubTreeSingleConfig.
-func NewSubTreeSingleConfig(hashFunc arbo.HashFunction, kindID string, maxLevels int,
-	parentLeafGetRoot GetRootFn, parentLeafSetRoot SetRootFn) *SubTreeSingleConfig {
-	return &SubTreeSingleConfig{
-		NewSubTreeConfig(hashFunc, kindID, maxLevels, parentLeafGetRoot, parentLeafSetRoot),
+// NewTreeSingletonConfig creates a new configuration for a singleton subTree.
+func NewTreeSingletonConfig(params TreeParams) TreeConfig {
+	return TreeConfig{
+		parentLeafKey:          []byte(params.KindID),
+		prefix:                 params.KindID,
+		TreeNonSingletonConfig: NewTreeNonSingletonConfig(params),
 	}
 }
 
 // Key returns the key used in the parent tree in which the value that contains
 // the subTree root is stored.  The key is the path of the parent leaf with the root.
-func (c *SubTreeSingleConfig) Key() []byte {
+func (c *TreeConfig) Key() []byte {
 	return []byte(c.kindID)
 }
 
-// treeConfig returns a unified configuration type for opening a subTree.
-func (c *SubTreeSingleConfig) treeConfig() *treeConfig {
-	return &treeConfig{
-		parentLeafKey: []byte(c.kindID),
-		prefix:        c.kindID,
-		SubTreeConfig: c.SubTreeConfig,
-	}
+// HashFunc returns the hashFunc set for this SubTreeSingleConfig
+func (c *TreeConfig) HashFunc() arbo.HashFunction {
+	return c.hashFunc
 }
 
 // mainTreeCfg is the subTree configuration of the mainTree.  It doesn't have a
 // kindID because it's the top level tree.  For the same reason, it doesn't
 // contain functions to work with the parent leaf: it doesn't have a parent.
-var mainTreeCfg = NewSubTreeSingleConfig(arbo.HashFunctionSha256, "", 256, nil, nil).treeConfig()
+var mainTreeCfg = NewTreeSingletonConfig(TreeParams{
+	HashFunc:          arbo.HashFunctionSha256,
+	KindID:            "",
+	MaxLevels:         256,
+	ParentLeafGetRoot: nil,
+	ParentLeafSetRoot: nil,
+})
 
 // StateDB is a database backed structure that holds a dynamic hierarchy of
 // linked merkle trees with the property that the keys and values of all merkle
--- a/statedb/treeupdate.go
+++ b/statedb/treeupdate.go
@@ -48,7 +48,7 @@ type TreeUpdate struct {
 	// corresponding parent leafs up to the mainTree.
 	openSubs map[string]*TreeUpdate
 	// cfg points to this TreeUpdate configuration.
-	cfg *treeConfig
+	cfg TreeConfig
 }
 
 // Get returns the value at key in this tree.  `key` is the path of the leaf,
@@ -57,9 +57,17 @@ func (u *TreeUpdate) Get(key []byte) ([]byte, error) {
 	return u.tree.Get(u.tree.tx, key)
 }
 
-// Iterate iterates over all nodes of this tree.
-func (u *TreeUpdate) Iterate(callback func(key, value []byte) bool) error {
+// IterateNodes iterates over all nodes of this tree.  The key and value are
+// the Arbo database representation of a node, and don't match the key value
+// used in Get, Add and Set.
+func (u *TreeUpdate) IterateNodes(callback func(key, value []byte) bool) error {
 	return u.tree.Iterate(u.tree.tx, callback)
+}
+
+// Iterate iterates over all leafs of this tree.  When callback returns true,
+// the iteration is stopped and this function returns.
+func (u *TreeUpdate) Iterate(callback func(key, value []byte) bool) error {
+	return u.tree.IterateLeaves(u.tree.tx, callback)
 }
 
 // Root returns the root of the tree, which cryptographically summarises the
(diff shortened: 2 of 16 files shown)


statedb/statedb.go
```diff
@@ -95,33 +95,26 @@ func subReadTx(tx db.ReadTx, path string) db.ReadTx {
...
 type GetRootFn func(value []byte) ([]byte, error)
 
 // SetRootFn is a function type that takes a leaf value and a root, updates the
 // leaf value with the new root and returns it.
 type SetRootFn func(value []byte, root []byte) ([]byte, error)
 
-// SubTreeConfig
-type SubTreeConfig struct {
+// TreeNonSingleConfig contains the configuration used for a non-singleton subTree.
+type TreeNonSingleConfig struct {
```
arnaucube: The naming of `NonSingleConfig` may bring confusion about `non single config` (like a config that is not a single one), when in reality is for `non-singleton config` (for non singleton subtrees).
I would propose to rename all the `NonSingle` for `NonSingleton` (it's just 3 chars more, and helps to have clearer naming)

statedb/treeview.go
```diff
@@ -58,11 +91,26 @@ func (v *TreeView) Get(key []byte) ([]byte, error) {
...
+// the Arbo database representation of a node, and don't match the key value
+// used in Get, Add and Set.  When callback returns true, the iteration is
+// stopped and this function returns.
+func (v *TreeView) IterateNodes(callback func(key, value []byte) bool) error {
 	return v.tree.Iterate(nil, callback)
 }
 
+// Iterate iterates over all leafs of this tree.  When callback returns true,
+// the iteration is stopped and this function returns.
+func (v *TreeView) Iterate(callback func(key, value []byte) bool) error {
```
arnaucube, line 103: This method also is implemented in `TreeUpdate` (`statedb/treeupdate.go`) with the exact same code, maybe can be abstracted in a way that there is only one implementation used by both structs/interfaces?

test/iavl_test.go
```diff
@@ -6,9 +6,11 @@ import (
 	"testing"
 
 	"github.com/ethereum/go-ethereum/common"
+	qt "github.com/frankban/quicktest"
```
arnaucube, line 9: This file (`test/iavl_test.go`) is no longer testing anything related to `iavl`, maybe it can be renamed to a more up-to-date filename?

vochain/process.go
```diff
@@ -63,49 +68,66 @@ func (v *State) Process(pid []byte, isQuery bool) (*models.Process, error) {
...
 }
 
 // CountProcesses returns the overall number of processes the vochain has
-func (v *State) CountProcesses(isQuery bool) int64 {
-	v.RLock()
-	defer v.RUnlock()
-	if isQuery {
-		return int64(v.Store.ImmutableTree(ProcessTree).Count())
+func (v *State) CountProcesses(isQuery bool) (uint64, error) {
+	// TODO: Once statedb.TreeView.Size() works, replace this by that.
```
arnaucube, line 87: Side note: we will need to take this into account depending on the usage of this method, as it can be slow to compute with the current approach.

vochain/state.go
```diff
@@ -2,28 +2,154 @@ package vochain
...
 	ed25519 "github.com/tendermint/tendermint/crypto/ed25519"
+	"github.com/vocdoni/arbo"
 	"go.vocdoni.io/dvote/crypto/ethereum"
+	"go.vocdoni.io/dvote/db/badgerdb"
 	"go.vocdoni.io/dvote/log"
-	statedb "go.vocdoni.io/dvote/statedblegacy"
-	"go.vocdoni.io/dvote/statedblegacy/iavlstate"
+	"go.vocdoni.io/dvote/statedb"
+
+	// "go.vocdoni.io/dvote/statedblegacy/iavlstate"
```
arnaucube: Is this ready to be removed?

vochain/state.go
```diff
@@ -58,15 +184,21 @@ func (e ErrHaltVochain) Unwrap() error { return e.reason }
...
+	mainTreeViewValue atomic.Value
+	voteCache         *lru.Cache
 	ImmutableState
 	mempoolRemoveTxKeys func([][32]byte, bool)
 	txCounter           int32
 	eventListeners      []EventListener
 	height              uint32
 }
 
+// NOTE(Edu): It is unclear to me why we use locking for write operations here.
```
arnaucube: Agree, also, the line 204 comment makes reference to IAVL trees, which are no longer there.

vochain/state.go
```diff
@@ -424,50 +552,57 @@ func (v *State) AddVote(vote *models.Vote) error {
...
-	defer func() {
-		if r := recover(); r != nil {
-			err = fmt.Errorf("recovered panic: %v", r)
-		}
-	}()
+	// defer func() {
+	// 	if r := recover(); r != nil {
+	// 		err = fmt.Errorf("recovered panic: %v", r)
+	// 	}
+	// }()
```
arnaucube: :wave:

arnaucube (commented):
I really like how the statedb new design looks ^^

vochain/state.go
ed255: Right. I've updated the comment, stating as a TODO to remove the mutex once we're ready to do proper integration tests.

tree/tree.go
```diff
@@ -154,6 +156,18 @@ func (t *Tree) Iterate(rTx db.ReadTx, callback func(key, value []byte) bool) err
 	return t.tree.IterateWithStopWithTx(rTx, root, callbackWrapper)
 }
 
+// Iterate iterates over all leafs of the tree.  When callback returns true,
+// the iteration is stopped and this function returns.
+func (t *Tree) IterateLeafs(rTx db.ReadTx, callback func(key, value []byte) bool) error {
```
mvdan: nit: plural of leaf is leaves

vochain/state.go
```diff
@@ -2,28 +2,153 @@ package vochain
...
+var CensusPoseidonCfg = statedb.NewTreeNonSingletonConfig(
+	arbo.HashFunctionPoseidon,
+	"cenPos",
+	64,
+	processGetCensusRoot,
+	processSetCensusRoot,
+)
+
+// VotesCfg is the Votes subTree (found under a Process leaf) configuration.
+var VotesCfg = statedb.NewTreeNonSingletonConfig(
```
mvdan: consider using a var group for these, like:

```
var (
    foo = ...
    baz = ...
)
```

statedb/treeupdate.go
```diff
@@ -140,16 +147,78 @@ func (u *TreeUpdate) subTree(cfg *treeConfig) (treeUpdate *TreeUpdate, err error
 	return treeUpdate, nil
 }
 
-// SubTreeSingle returns a TreeUpdate of a singleton SubTree whose root is stored
-// in the leaf with `cfg.Key()`, and is parametrized by `cfg`.
-func (u *TreeUpdate) SubTreeSingle(cfg *SubTreeSingleConfig) (*TreeUpdate, error) {
-	return u.subTree(cfg.treeConfig())
+// DeepSubTree allows opening a nested subTree by passing the list of tree
+// configurations.
+func (u *TreeUpdate) DeepSubTree(cfgs []*TreeConfig) (treeUpdate *TreeUpdate, err error) {
```
mvdan: would be nicer to use if it was like `cfgs ...*TreeConfig`, then you can call it like `DeepSubTree(cfg1)` o `DeepSubTree(cfg1, cfg2)` without having to build a slice explicitly.

mvdan: The other funcs could do the same if you move the cfgs to the last parameter.

statedb/treeupdate.go
```diff
@@ -140,16 +147,78 @@ func (u *TreeUpdate) subTree(cfg *treeConfig) (treeUpdate *TreeUpdate, err error
...
+// TreeViewer interface.
+func (u *TreeUpdate) AsTreeView() TreeViewer {
+	return &treeUpdateView{u}
+}
+
+// treeUpdateView is a wrapper over TreeUpdate that fulfills the TreeViewer
+// interface.
+type treeUpdateView struct {
+	*TreeUpdate
+}
```
mvdan: note that treeUpdateView also has all of TreeUpdate's methods; you might not want this.

if you do not want it, another plus is that you can avoid an allocation and extra pointer by doing `type treeUpdateView TreeUpdate`, then you convert between them like `(*treeUpdateView)(tu)`.

test/statedb_test.go
```diff
@@ -126,81 +110,68 @@ func TestAddProcess(t *testing.T) {
...
+	err := s.AddVote(testcommon.NewVoteHardcoded())
+	qt.Assert(t, err, qt.IsNil)
+	_, err = s.Envelope(testutil.Hex2byte(t, "e9d5e8d791f51179e218c606f83f5967ab272292a6dbda887853d81f7a1d5105"),
+		testutil.Hex2byte(t, "5592f1c18e2a15953f355c34b247d751da307338c994000b9a65db1dc14cc6c0"), false)
+	qt.Assert(t, err, qt.IsNil)
 	c := s.CountVotes(testutil.Hex2byte(t, "e9d5e8d791f51179e218c606f83f5967ab272292a6dbda887853d81f7a1d5105"), false)
-	if c != 1 {
-		t.Errorf("number of votes should be 1, received %d", c)
-	}
+	qt.Assert(t, c, qt.Equals, uint32(1))
```
mvdan, line 161: I like how qt is improving the tests :)

vochain/state.go
mvdan: I also wonder if using composite literals for these would be better; right now I have no idea what `256` means, but if you had `HashSize: 256` it would be a lot clearer. (I'm just making up that this means a hash size)

arnaucube (approved):

vochain/state.go
```diff
@@ -424,50 +554,48 @@ func (v *State) AddVote(vote *models.Vote) error {
...
 	vid.Write(pid)
 	vid.Write(nullifier)
-	return vid.Bytes(), nil
+	return vid.Sum(nil), nil
 }
 
 // Envelope returns the hash of a stored vote if exists.
 func (v *State) Envelope(processID, nullifier []byte, isQuery bool) (_ []byte, err error) {
-	// TODO(mvdan): remove the recover once
-	// https://github.com/tendermint/iavl/issues/212 is fixed
```
mvdan, line 444: they still haven't fixed this :see_no_evil: good that we're moving away from IAVL I think.

vochain/state.go
```diff
@@ -482,24 +610,39 @@ func (v *State) EnvelopeExists(processID, nullifier []byte, isQuery bool) (bool,
...
-		v.Store.ImmutableTree(VoteTree).Iterate(processID, fn)
-	} else {
-		v.Store.Tree(VoteTree).Iterate(processID, fn)
+	votesTree, err := v.mainTreeViewer(isQuery).DeepSubTree(
+		[]*statedb.TreeConfig{ProcessesCfg, VotesCfg.WithKey(processID)})
+	if err != nil {
+		return err
 	}
-	return true
+	var callbackErr error
```
mvdan, line 621: +100 to giving this variable a name different than `err`, because it helps clarify its purpose and avoid typos

statedb/statedb.go
```diff
@@ -130,49 +123,55 @@ func NewSubTreeConfig(hashFunc arbo.HashFunction, kindID string, maxLevels int,
...
 
-// SubTreeSingleConfig contains the configuration used for a singleton subTree.
-type SubTreeSingleConfig struct {
-	*SubTreeConfig
+// TreeConfig contains the configuration used for a subTree.
+type TreeConfig struct {
+	parentLeafKey []byte
+	prefix        string
+	*TreeNonSingletonConfig
 }
```
mvdan, line 166: TreeConfig is just a pointer, a string, and a slice: roughly the size of six pointers. I'd just use it without a pointer like `TreeConfig` rather than `*TreeConfig`, because unnecessary pointers to small types just adds overhead. Unless you want functions to be able to omit the config via `fn(nil)` rather than `fn(TreeConfig{})`, but I don't think that is the case.

This would be different if TreeConfig implemented an interface - since interfaces in Go always contain pointers, trying to avoid the pointer is futile. But I also don't think that's the case here.

This might apply to other small types in this PR, I'm not sure. Just something for you to think about. If a type is small, and doesn't benefit from being used as a pointer, don't use one. And "small" is bigger than you think these days, thanks to the larger caches and number of registers. 32, 64, or even 128 bytes without a pointer are all reasonable.

vochain/state.go
mvdan: If you need constructors, another way is to use named types for the parameters, so then you can end up with `statedb.HashSize(256)`.

vochain/state.go
```diff
@@ -58,15 +183,24 @@ func (e ErrHaltVochain) Unwrap() error { return e.reason }
...
 	height              uint32
 }
 
+// NOTE(Edu): It is my understanding that all write operations to the State
+// come from processing transaction, which are always processed serially, and
+// thus no thread-safety is required.  The arbo-based StateDB supports
+// concurrent reads to the TreeView while a new transaction is opened and
+// written serially, so there should not be need for this mutex anymore.
+// TODO: Remove the mutex and the locking once we're ready to do integration
+// tests.
```
mvdan, line 205: another note for you: when you do that, also test with `go test -race`, because it will do a better job at finding data races (e.g. concurrent non-atomic writes to the same memory address).

vochain/state.go
```diff
@@ -78,65 +212,115 @@ type ImmutableState struct {
...
+		make([]byte, ValidatorsCfg.HashFunc().Len())); err != nil {
+		return nil, err
+	}
+	if err := update.Add(ProcessesCfg.Key(),
+		make([]byte, ProcessesCfg.HashFunc().Len())); err != nil {
+		return nil, err
+	}
+	header := models.TendermintHeader{
+		Height:  0,
+		AppHash: []byte{},
```
mvdan: nit: in general you don't need to specify zero values in fields. you could drop both Height and AppHash here.

vochain/state.go
```diff
@@ -78,65 +212,115 @@ type ImmutableState struct {
...
+}
+
+// mainTreeView is a thread-safe function to obtain a pointer to the last
+// opened mainTree as a TreeView.
+func (v *State) mainTreeView() *statedb.TreeView {
+	return v.mainTreeViewValue.Load().(*statedb.TreeView)
+}
+
+// TODO: @mvdan is the usage of atomic.Value appropiate here?  Or should I
+// better use atomic.LoadPointer & atomic.StorePointer?
```
mvdan: your use of atomic.Value is correct, I think. if you used pointers, you would be forced into using unsafe, which is... less safe. If you mess up with `interface{}`, you will get a helpful runtime panic (in the line 303 above, when you type assert). If you mess up with `unsafe.Pointer`, you may end up with memory corruption or worse.

vochain/state.go
```diff
@@ -78,65 +212,115 @@ type ImmutableState struct {
...
+// StateDB transaction.  When isQuery is false, the mainTree returned is the
+// last commited version.
+func (v *State) mainTreeViewer(isQuery bool) statedb.TreeViewer {
+	var mainTree statedb.TreeViewer
+	if isQuery {
+		mainTree = v.mainTreeView()
+	} else {
+		mainTree = v.Tx.AsTreeView()
+	}
+	return mainTree
```
mvdan: nit, but it can be shorter:

```
if isQuery {
    return v.mainTreeView()
}
return v.Tx.AsTreeView()
```

vochain/state.go
```diff
@@ -147,90 +331,49 @@ func (v *State) AddEventListener(l EventListener) {
...
-	}
-	oracles = append(oracles, address.Bytes())
-	oracleList.Oracles = oracles
-	newOraclesBytes, err := proto.Marshal(&oracleList)
-	if err != nil {
-		return fmt.Errorf("cannot marshal oracles: (%s)", err)
-	}
-	return v.Store.Tree(AppTree).Add(oracleKey, newOraclesBytes)
+	return v.Tx.DeepSet([]*statedb.TreeConfig{OraclesCfg},
+		address.Bytes(), []byte{1})
```
mvdan: maybe define `[]byte{1}` as a global saying what it means

mvdan (approved):

vochain/state.go
ed255: > I also wonder if using composite literals for these would be better; right now I have no idea what `256` means, but if you had `HashSize: 256` it would be a lot clearer. (I'm just making up that this means a hash size)

Yeah, this is something that bothered me a bit as well.  I wanted to use constructors so that the generated type can't be modified externally, but it was not clear what each parameter was.  I added a new struct type `TreeParams` to pass the parameters to the constructor and now it's much more clear!  thanks!

Now:
```go
	ValidatorsCfg = statedb.NewTreeSingletonConfig(statedb.TreeParams{
		HashFunc:          arbo.HashFunctionSha256,
		KindID:            "valids",
		MaxLevels:         256,
		ParentLeafGetRoot: rootLeafGetRoot,
		ParentLeafSetRoot: rootLeafSetRoot,
	})
```

PR #267, opened 2021-09-17
Integrate new censustree (with arbo) into census package

arnaucube:


census/handler.go
```diff
@@ -315,13 +314,20 @@ func (m *Manager) Handler(ctx context.Context, r *api.MetaRequest, isAuth bool,
...
+			resp.CensusDump, err = snapshot.Dump()
+		} else {
+			// TODO the method DumpPlain no longer exists, clarify
+			// the difference of usage between 'dump' & 'dumpPlain'.
+
+			// var vals [][]byte
+			// resp.CensusKeys, vals, err = tr.DumpPlain(root)
+			// for _, v := range vals {
+			//         resp.CensusValues = append(resp.CensusValues, types.HexBytes(v))
+			// }
```
arnaucube: @p4u Which was the purpose of having `Dump` & `DumpPlain`? Could the `DumpPlain` be removed?

test/census_test.go
```diff
@@ -202,10 +203,11 @@ func TestCensus(t *testing.T) {
 	resp = doRequest("getRoot", nil)
 	qt.Assert(t, resp.Root, qt.DeepEquals, root)
 
+	// TODO WIP getSize method is not currently supported
```
arnaucube, line 208: The `getSize` will be supported once https://github.com/vocdoni/vocdoni-node/issues/271 is done (soon)

benchmark/api_benchmark_test.go
```diff
@@ -34,7 +35,7 @@ func BenchmarkCensus(b *testing.B) {
 	host := *hostFlag
 	if host == "" {
 		var server testcommon.DvoteAPIServer
-		server.CensusBackend = *censusBackend
+		server.CensusBackend = *censusBackend // NOTE: CensusBackend is not used
```
ed255, line 38: Maybe the field CensusBackend can be deleted from DvoteAPIServer?

.gitignore
```diff
@@ -12,6 +12,7 @@
 
 # Output of the go coverage tool, specifically when used with LiteIDE
 *.out
+covprofile
```
mvdan: or just use `-coverprofile=cover.out`? :)

censustree/censustree.go
```diff
@@ -1,27 +1,88 @@
...
 	"go.vocdoni.io/dvote/db"
+	"go.vocdoni.io/dvote/db/badgerdb"
 	"go.vocdoni.io/dvote/tree"
+	"go.vocdoni.io/proto/build/go/models"
+)
+
+const (
+	// TreeUnknown is the default value used for censusTree implementations
+	// which are not part of the this factory.
+	TreeTypeUnknown = models.Census_UNKNOWN
```
mvdan: the qualified name here will be `censustree.TreeTypeUnknown`, so I think you can drop the redundant "Tree" and just make it `censustree.TypeUnknown`. Same with the others.

(this problem exists elsewhere, e.g. censustree.CensusTree is very redundant, but that's in master)

censustree/censustree.go
```diff
@@ -1,27 +1,88 @@
...
+	*CensusTree, error) {
+	var hashFunc arbo.HashFunction = arbo.HashFunctionBlake2b
+
+	switch censusType {
+	case 0:
+		hashFunc = arbo.HashFunctionBlake2b
+		censusType = models.Census_ARBO_BLAKE2B
+	case models.Census_ARBO_BLAKE2B:
+		hashFunc = arbo.HashFunctionBlake2b
+	case models.Census_ARBO_POSEIDON:
```
mvdan, line 43: use your constants here? would also be nice to use a constant for 0

censustree/censustree.go
```diff
@@ -1,27 +1,88 @@
...
+// TMP to be defined the production circuit nLevels
+const nLevels = 32
+
 // New returns a new CensusTree, if there already is a CensusTree in the
 // database, it will load it.
-func New(wTx db.WriteTx, opts tree.Options) (*CensusTree, error) {
-	t, err := tree.New(wTx, opts)
+func New(wTx db.WriteTx, censusType models.Census_Type, name, storageDir string) (
+	*CensusTree, error) {
+	var hashFunc arbo.HashFunction = arbo.HashFunctionBlake2b
```
mvdan: I think this can just be `hashFunc := arbo.HashFunctionBlake2b`

vochain/proof.go
```diff
@@ -70,15 +71,26 @@ func VerifyProofOffChainTree(process *models.Process, proof *models.Proof,
...
+		return false, nil, fmt.Errorf("graviton proof no longer supported")
 	case *models.Proof_Iden3:
 		// NOT IMPLEMENTED
 		return false, nil, fmt.Errorf("iden3 proof not implemented")
+	case *models.Proof_Arbo:
+		p := proof.GetArbo()
+		if p == nil {
+			return false, nil, fmt.Errorf("arbo proof is empty")
+		}
+		var hashFunc arbo.HashFunction = arbo.HashFunctionBlake2b
```
mvdan, line 88: same as the other line; use `:=`

vochain/proof.go
```diff
@@ -70,15 +71,26 @@ func VerifyProofOffChainTree(process *models.Process, proof *models.Proof,
...
+		switch p.Type {
+		case models.ProofArbo_BLAKE2B:
+			hashFunc = arbo.HashFunctionBlake2b
+		case models.ProofArbo_POSEIDON:
+			hashFunc = arbo.HashFunctionPoseidon
+		default:
+			return false, nil, fmt.Errorf("not recognized ProofArbo type: %s", p.Type)
+		}
+		valid, err := tree.VerifyProof(hashFunc, key, []byte{}, p.Siblings, censusRoot)
+		return valid, big.NewInt(1), err
```
mvdan: nit: do you need to return a new big int every time? if the caller does not modify the value, maybe you could use a global big int that's always 1.

mvdan (approved):

census/handler.go
```diff
@@ -277,24 +277,23 @@ func (m *Manager) Handler(ctx context.Context, r *api.MetaRequest, isAuth bool,
...
 		return resp
 
 	case "getSize":
-		root, err := tr.Root()
-		if err != nil {
-			resp.SetError(err.Error())
-			return resp
-		}
-		size, err := tr.Size(root)
+		size, err := tr.Size(nil)
```
ed255, line 300: NOTE: This will currently return an error.

census/handler.go
```diff
@@ -339,13 +345,17 @@ func (m *Manager) Handler(ctx context.Context, r *api.MetaRequest, isAuth bool,
...
+		root, err := tr.Root(nil)
 		if err != nil {
 			resp.SetError(err.Error())
 			return resp
 		}
 		dump.RootHash = root
-		dump.Data, err = tr.Dump(root)
+		snapshot, err := tr.FromRoot(root)
+		if err != nil {
+			resp.SetError(err)
```
ed255, line 373: missing return.

census/handler.go
```diff
@@ -315,13 +314,20 @@ func (m *Manager) Handler(ctx context.Context, r *api.MetaRequest, isAuth bool,
...
 		if r.Method == "dump" {
-			resp.CensusDump, err = tr.Dump(root)
-		} else {
-			var vals [][]byte
-			resp.CensusKeys, vals, err = tr.DumpPlain(root)
-			for _, v := range vals {
-				resp.CensusValues = append(resp.CensusValues, types.HexBytes(v))
+			snapshot, err := tr.FromRoot(root)
+			if err != nil {
+				resp.SetError(err)
```
ed255, line 329: missing return.

census/handler.go
```diff
@@ -277,24 +277,23 @@ func (m *Manager) Handler(ctx context.Context, r *api.MetaRequest, isAuth bool,
 		//if !r.Digested {
 		//	data = snarks.Poseidon.Hash(data)
 		//}
-		siblings, err := tr.GenProof(data, r.CensusValue)
+		leafV, siblings, err := tr.GenProof(nil, data)
 		if err != nil {
 			resp.SetError(err)
```
ed255, line 289: missing return

census/handler.go
```diff
@@ -151,11 +151,11 @@ func (m *Manager) Handler(ctx context.Context, r *api.MetaRequest, isAuth bool,
 			//if !r.Digested {
 			// data = snarks.Poseidon.Hash(data)
 			//}
-			err := tr.Add(data, r.CensusValue)
+			err := tr.Add(nil, data, r.CensusValue)
 			if err != nil {
 				resp.SetError(err)
```
ed255, line 157: missing return

censustree/censustree.go
```diff
@@ -1,27 +1,88 @@
...
 // CensusTree implements the Tree used for census
 type CensusTree struct {
 	*tree.Tree
 
-	public uint32
+	public     uint32
+	censusType models.Census_Type
 }
 
+// TMP to be defined the production circuit nLevels
```
ed255, line 31: ALERT: Note that this nLevels is also for trees that are not related to circuits!  I understand that 32 is a reasonable number for poseidon trees where the key is a sequential number, but the previous common use case of the censustree was to build a census before the process and send proofs to the vochain without zkSNARK.  In that case, the keys are not sequential, and thus I think 32 levels is very low.  For non-poseidon trees I would set nLevels = 256 (considering that the key is not sequential)

censustree/censustree.go
```diff
@@ -1,27 +1,88 @@
...
 // New returns a new CensusTree, if there already is a CensusTree in the
 // database, it will load it.
-func New(wTx db.WriteTx, opts tree.Options) (*CensusTree, error) {
-	t, err := tree.New(wTx, opts)
+func New(wTx db.WriteTx, censusType models.Census_Type, name, storageDir string) (
+	*CensusTree, error) {
+	var hashFunc arbo.HashFunction = arbo.HashFunctionBlake2b
+
+	switch censusType {
+	case 0:
```
ed255: I think setting the censusType to a value that is not one of the defined constants should be an error.  What's the use case of setting a 0 here?  The caller to censustree.New should decide a censuType != 0 IMHO.  This case should go to the default and return an error IMO.

censustree/censustree.go
ed255, line 43: Note for myself: Does it make sense for the census manager (via API) to start a census with type different than ARBO_BLAKE2B?  If it doesn't make sense, I think we should enforce the type (by checking the query)

censustree/censustree.go
```diff
@@ -1,27 +1,88 @@
...
 }
 
+// TMP to be defined the production circuit nLevels
+const nLevels = 32
+
 // New returns a new CensusTree, if there already is a CensusTree in the
 // database, it will load it.
-func New(wTx db.WriteTx, opts tree.Options) (*CensusTree, error) {
-	t, err := tree.New(wTx, opts)
+func New(wTx db.WriteTx, censusType models.Census_Type, name, storageDir string) (
```
ed255: I think the previous API of passing an Option struct to the constructor was nice.  Have you considered adding an Option type for the censustree?

vochain/proof.go
ed255, line 88:
```suggestion
		var hashFunc arbo.HashFunction
```
Why not like that?

ed255 (changes requested):

censustree/censustree.go
ed255: That would be here https://github.com/vocdoni/vocdoni-node/blob/d5eb3aa0155f34e0b55905e6e1a4d033e67f3570/census/handler.go#L52
I think we should only accept `r.CensusType == models.Census_ARBO_BLAKE2B`

censustree/censustree.go
arnaucube: Right, updated both things that you mention:
- `TreeTypeUnknown` -> `TypeUnknown` (the same with the other types)
- `CensusTree` -> `Tree`

censustree/censustree.go
arnaucube: In fact, I've added 'custom constants' that where an alias over the protobuff constants, updated it to not have the custom constants and to use directly the protobuff constants in this package.

censustree/censustree.go
arnaucube: I think that in this case not, because `arbo.HashFunction` is an interface, while `arbo.HashFunctionBlake2b` and `arbo.HashFunctionPoseidon` are implementations of that interface that can be placed in this `hashFunc` var.

vochain/proof.go
arnaucube: Same reasoning than in https://github.com/vocdoni/vocdoni-node/pull/267#discussion_r714625848

vochain/proof.go
arnaucube: Agree, this was in this way previous to this PR, but agree to use a global big int. Updated it, and also for the `big.Int(0)` that is used in a couple of places in the same file.

benchmark/api_benchmark_test.go
arnaucube: I think so, that's why I've added that `note`, but as is out of the scope of this PR, and I was not 100% sure if I was missing some argument I decided to leave it as it was with the note, and we can update it in a near future.

census/handler.go
arnaucube: Thanks for spotting all the missing returns! My bad!

censustree/censustree.go
arnaucube: Right, agree with the reasoning. Updated it to 256, and we will add the logic for the cases of Poseidon to use the specific circuit maxlength.

censustree/censustree.go
arnaucube: Yes, this is because there is a test in `census/census_test.go` that checks that if the TreeType is set to 0 it will use the 'default' tree type: https://github.com/vocdoni/vocdoni-node/blob/3f4d59e2fd93fcaa2ed018d49ba7ab29d4b737b3/census/census_test.go#L34

I don't know from which reason this comes from.

In fact, this reasoning (the one from that already existing test) means that when the type is using the protobuf `models.Census_UNKNOWN`, this is expected to be equivalent to use the 'default' tree type.

censustree/censustree.go
arnaucube: I've updated it into `var hashFunc arbo.HashFunction` as the specific hash function is set in the following lines after this one.

censustree/censustree.go
arnaucube: Right, updated with `censustree.Option` new struct.

censustree/censustree.go
arnaucube: Update: removed the `case 0`, this was set for backwards compatibility, not needed now as the current changes will break backwards compatibility.

ed255 (approved):

PR #268, opened 2021-09-20
Fix block.Height appHash disparity

ed255:
Don't store the block request header in the state.  Previously the block
request header was stored in the state.  I found this confusing because then
the state hash is updated at every block even if there are no transactions that
update the state, and only the height and timestamp were ever queried.

Add startBlockTimestamp in BaseApplication so that it can be consumed by the
scrutinizer (previously this was stored in the header in the state).

Store the last commited block height in the NoState of the StateDB, so that on
init we can query the height of the last stored height (required for Tendermint
Info method).  Also keep the current block in a variable in State.

Move State RWMutex into a wrapper with the Tx to make it explicit that
only the State.Tx requires locking.

### DIFF
--- a/router/indexerCallbacks.go
+++ b/router/indexerCallbacks.go
@@ -15,15 +15,23 @@ import (
 func (r *Router) getStats(request RouterRequest) {
 	var err error
 	stats := new(api.VochainStats)
-	stats.BlockHeight = r.vocapp.Height()
-	stats.BlockTimeStamp = int32(r.vocapp.State.Header(true).Timestamp)
+	height := r.vocapp.Node.BlockStore().Height()
+	stats.BlockHeight = uint32(height)
+	block := r.vocapp.Node.BlockStore().LoadBlockMeta(height)
+	if block == nil {
+		r.SendError(request, fmt.Sprintf("could not get block at height %v", height))
+		return
+	}
+	stats.BlockTimeStamp = int32(block.Header.Time.Unix())
 	stats.EntityCount = int64(r.Scrutinizer.EntityCount())
 	if stats.EnvelopeCount, err = r.Scrutinizer.GetEnvelopeHeight([]byte{}); err != nil {
 		r.SendError(request, fmt.Sprintf("could not count vote envelopes: (%s)", err))
+		return
 	}
 	stats.ProcessCount = int64(r.Scrutinizer.ProcessCount([]byte{}))
 	if stats.TransactionCount, err = r.Scrutinizer.TransactionCount(); err != nil {
 		r.SendError(request, fmt.Sprintf("could not count transactions: (%s)", err))
+		return
 	}
 	vals, _ := r.vocapp.State.Validators(true)
 	stats.ValidatorCount = len(vals)
--- a/vochain/app.go
+++ b/vochain/app.go
@@ -39,8 +39,12 @@ type BaseApplication struct {
 	fnSendTx           func(tx []byte) (*ctypes.ResultBroadcastTx, error)
 	blockCache         *lru.AtomicCache
 	height             uint32
-	timestamp          int64
-	chainId            string
+	// endBlockTimestamp is the last block end timestamp calculated from local time.
+	endBlockTimestamp int64
+	// startBlockTimestamp is the current block timestamp from tendermint's
+	// abcitypes.RequestBeginBlock.Header.Time
+	startBlockTimestamp int64
+	chainId             string
 	// ZkVks contains the VerificationKey for each circuit parameters index
 	ZkVks []*snarkTypes.Vk
 }
@@ -123,9 +127,14 @@ func (app *BaseApplication) Height() uint32 {
 	return atomic.LoadUint32(&app.height)
 }
 
-// Timestamp returns the last block timestamp
+// Timestamp returns the last block end timestamp
 func (app *BaseApplication) Timestamp() int64 {
-	return atomic.LoadInt64(&app.timestamp)
+	return atomic.LoadInt64(&app.endBlockTimestamp)
+}
+
+// TimestampStartBlock returns the current block start timestamp
+func (app *BaseApplication) TimestampStartBlock() int64 {
+	return atomic.LoadInt64(&app.startBlockTimestamp)
 }
 
 // ChainID returns the Node ChainID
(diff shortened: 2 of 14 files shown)


vochain/app.go
```diff
@@ -125,7 +126,12 @@ func (app *BaseApplication) Height() uint32 {
 
 // Timestamp returns the last block timestamp
 func (app *BaseApplication) Timestamp() int64 {
-	return atomic.LoadInt64(&app.timestamp)
+	return atomic.LoadInt64(&app.endBlockTimestamp)
+}
+
+// Timestamp returns the current block start timestamp
```
mvdan:
```suggestion
// TimestampStartBlock returns the current block start timestamp
```

vochain/app.go
```diff
@@ -139,7 +145,7 @@ func (app *BaseApplication) MempoolRemoveTx(txKey [32]byte) {
 	app.Node.Mempool().(*mempl.CListMempool).RemoveTxByKey(txKey, false)
 }
 
-// GetBlockByHeight retreies a full Tendermint block indexed by its height.
+// GetBlockByHeight retreives a full Tendermint block indexed by its height.
```
mvdan:
```suggestion
// GetBlockByHeight retrieves a full Tendermint block indexed by its height.
```

vochain/state.go
```diff
@@ -664,34 +679,34 @@ func (v *State) EnvelopeList(processID []byte, from, listSize int,
...
+// 	v.Tx.RUnlock()
+// 	if err != nil {
+// 		log.Fatalf("cannot get headerKey from mainTree: %s", err)
+// 	}
+// 	var header models.TendermintHeader
+// 	if err := proto.Unmarshal(headerBytes, &header); err != nil {
+// 		log.Fatalf("cannot get proto.Unmarshal header: %s", err)
+// 	}
+// 	return &header
+// }
```
mvdan: just remove it? IMO leaving uncommented code isn't helpful, because we can always retrieve it from git history

vochain/app.go
```diff
@@ -285,27 +298,22 @@ func (app *BaseApplication) InitChain(req abcitypes.RequestInitChain) abcitypes.
...
+	// of the State.voteCache while there are other goroutines that write
+	// to elements in State.voteCache (for example
+	// BaseApplication.VoteEnvelopeCheck calls State.CacheGet and modifies
+	// the returned element).  I think a possible solution would be for
+	// State.CacheGet to return a copy of the element, but the returned
+	// type is protobuf generated and is not copy friendly.  Can we create
+	// a new element manually copying the fields?  (it's cumbersome and can
+	// be error prone if more fields are added in the future).  Can we find
+	// another solution?  @mvdan got any idea?  At the end I decided to use
+	// proto.Clone which seems appropiate.  Please comment!
```
mvdan: I think your spotted race is right. We probably haven't found it because none of our tests end up purging the cache while VoteEnvelopeCheck runs.

Your solution seems right, it's just slightly inefficient. We're going to do a full generic copy at every Get, even for the cases that don't need it, like CacheDel.

An alternative is to make both ends use sync/atomic, since the field is just a uint32, and we only seem to access it in two places. That would definitely be cheaper, with the downside that we always need to remember to use atomics when accessing it. That actually seems like a problem, because then any calls to `proto.Marshal` or `proto.Unmarshal` would be unsafe since they would not use atomics.

There's also a middle ground: use `proto.Clone`, but only when we're about to modify a vote struct: in VoteEnvelopeCheck. There you can explain why we need it.

mvdan (approved):

vochain/app.go
mvdan: There's also a subtle behavior change here: in master right now, VoteEnvelopeCheck updates the height of the vote in the cache, so other goroutines could see that update in-place. If we clone, then that would never happen.

VoteEnvelopeCheck seems to do a CacheDel right after modifying the vote, so I think we are not meant to share this update. So, in that sense, your Clone is more correct than my atomics, since my atomics could still lead to other goroutines seeing the update when they don't want to.

p4u (approved):
Quick review, looks good.

mvdan (approved):

vochain/state.go
```diff
@@ -709,34 +709,61 @@ func (v *State) Save() ([]byte, error) {
 			log.Warnf("event callback error on commit: %v", err)
 		}
 	}
-	atomic.StoreUint32(&v.height, height)
-	return v.Store.Hash()
+	hash, err := v.Store.Hash()
+	if err != nil {
+		return nil, err
+	}
+	return hash, nil
```
arnaucube:
```suggestion
	return v.Store.Hash()
```

arnaucube (approved):

router/indexerCallbacks.go
```diff
@@ -15,8 +15,12 @@ import (
 func (r *Router) getStats(request RouterRequest) {
 	var err error
 	stats := new(api.VochainStats)
-	stats.BlockHeight = r.vocapp.Height()
-	stats.BlockTimeStamp = int32(r.vocapp.State.Header(true).Timestamp)
+	height := r.vocapp.Node.BlockStore().Height()
+	stats.BlockHeight = uint32(height)
+	block := r.vocapp.Node.BlockStore().LoadBlockMeta(height)
+	if block != nil {
```
jordipainan: Shouldn't we return an error on the request if the block is nil ?

vochain/process.go
```diff
@@ -66,9 +66,13 @@ func (v *State) CancelProcess(pid []byte) error { // LEGACY
 func (v *State) Process(pid []byte, isQuery bool) (*models.Process, error) {
 	var processBytes []byte
 	var err error
-	v.RLock()
+	if !isQuery {
+		v.Tx.RLock()
+	}
 	processBytes, err = v.mainTreeViewer(isQuery).DeepGet(pid, ProcessesCfg)
-	v.RUnlock()
+	if !isQuery {
```
jordipainan: The same second if here, having the same condition, seems strange to me. Can we do ?
```golang
if !isQuery { 
    v.Tx.RLock()
    defer v.Tx.RUnlock()
}
```

vochain/state.go
```diff
@@ -183,38 +185,41 @@ type ErrHaltVochain struct {
...
-// written serially, so there should not be need for this mutex anymore.
-// TODO: Remove the mutex and the locking once we're ready to do integration
-// tests.
-// ImmutableState holds the latest trees version saved on disk
-type ImmutableState struct {
-	// Note that the mutex locks the entirety of the three IAVL trees, both
-	// their mutable and immutable components. An immutable tree is not safe
-	// for concurrent use with its parent mutable tree.
-	sync.RWMutex
+	curHeight           uint32 // current height
```
jordipainan: I would prefer to have `currentHeight`, but just a personal preference :)

jordipainan:
Despite the last comments, LGTM :) Thank you @ed255

router/indexerCallbacks.go
ed255: Right! I've updated the function to call SendError when block is nil.  I've also added returns after two `r.SendErrors` that were missing them.

PR #269, opened 2021-09-21
pull WIP fix for go-libp2p-kad-dht request goroutine bug

mvdan:
(see commit message)


vdo (approved):

mvdan:
I'm following the relevant issue and PR, so I'll update this if there are any news.

We should be careful about updates to libp2p versions, because if we bump to kad-dht v0.13, we'll lose this unmerged fix.

PR #274, opened 2021-09-22
all: upgrade to Go 1.17

mvdan:
(see commit message)


mvdan:
I forgot to say; `-mod=readonly` has been a default since 1.16, so we no longer need the explicit flag.

arnaucube (approved):

ed255 (approved):

Issue #280, opened 2021-09-24
bug: transaction too big on block commit


p4u:
**Describe the bug**
When benchmarking with a huge voting process (20k voters), the gateway crashes.

```
2021-09-24T16:40:44+02:00       FATAL   vochain/app.go:367      cannot save state: cannot commit statedb tx: Txn is too big to fit into one request
go.vocdoni.io/dvote/vochain.(*BaseApplication).Commit
        /home/vocdoni/go-dvote/vochain/app.go:367
github.com/tendermint/tendermint/abci/client.(*localClient).CommitSync
        /home/vocdoni/go/pkg/mod/github.com/tendermint/tendermint@v0.34.13/abci/client/local_client.go:266
github.com/tendermint/tendermint/proxy.(*appConnConsensus).CommitSync
        /home/vocdoni/go/pkg/mod/github.com/tendermint/tendermint@v0.34.13/proxy/app_conn.go:93
github.com/tendermint/tendermint/state.(*BlockExecutor).Commit
        /home/vocdoni/go/pkg/mod/github.com/tendermint/tendermint@v0.34.13/state/execution.go:228
github.com/tendermint/tendermint/state.(*BlockExecutor).ApplyBlock
        /home/vocdoni/go/pkg/mod/github.com/tendermint/tendermint@v0.34.13/state/execution.go:180
github.com/tendermint/tendermint/consensus.(*State).finalizeCommit
        /home/vocdoni/go/pkg/mod/github.com/tendermint/tendermint@v0.34.13/consensus/state.go:1635
github.com/tendermint/tendermint/consensus.(*State).tryFinalizeCommit
        /home/vocdoni/go/pkg/mod/github.com/tendermint/tendermint@v0.34.13/consensus/state.go:1546
github.com/tendermint/tendermint/consensus.(*State).enterCommit.func1
        /home/vocdoni/go/pkg/mod/github.com/tendermint/tendermint@v0.34.13/consensus/state.go:1481
github.com/tendermint/tendermint/consensus.(*State).enterCommit
        /home/vocdoni/go/pkg/mod/github.com/tendermint/tendermint@v0.34.13/consensus/state.go:1519
github.com/tendermint/tendermint/consensus.(*State).addVote
        /home/vocdoni/go/pkg/mod/github.com/tendermint/tendermint@v0.34.13/consensus/state.go:2132
github.com/tendermint/tendermint/consensus.(*State).tryAddVote
[...]

PR #282, opened 2021-09-27
Feature/batchtx

ed255:
# Close Database on failure

# Batch transactions in the StateDB

Introduce db/batch.Batch: a wrapper over a db.WriteTx that implements
db.WriteTx.  The batch acts as a WriteTx but when it becomes too big, it
automatically commits and continues with a new internal db.WriteTx
transparently.

Use this Batch instead of a regular WriteTx in the statedb.TreeTx to
avoid hitting the db.WriteTx size limit when the block contains many
updates.  Since we are now writting to disk between StateDB versions,
when opening TreeUpdates we need to make sure we start always from the
last commited version, so we query the expected root and set it via
tree.SetRoot if necessary.

This should fix the observed error "Txn is too big to fit into one
request" seen in #280

### DIFF
--- a/censustree/censustree.go
+++ b/censustree/censustree.go
@@ -33,8 +33,7 @@ const nLevels = 256
 
 // New returns a new Tree, if there already is a Tree in the
 // database, it will load it.
-func New(wTx db.WriteTx, opts Options) (
-	*Tree, error) {
+func New(wTx db.WriteTx, opts Options) (*Tree, error) {
 	var hashFunc arbo.HashFunction
 
 	switch opts.CensusType {
@@ -54,6 +53,7 @@ func New(wTx db.WriteTx, opts Options) (
 
 	t, err := tree.New(wTx, tree.Options{DB: database, MaxLevels: nLevels, HashFunc: hashFunc})
 	if err != nil {
+		database.Close()
 		return nil, err
 	}
 	return &Tree{Tree: t, censusType: opts.CensusType}, nil
--- a/censustreelegacy/arbotree/wrapper.go
+++ b/censustreelegacy/arbotree/wrapper.go
@@ -42,6 +42,7 @@ func NewTree(name, storageDir string, nLevels int, hashFunc arbo.HashFunction) (
 
 	mt, err := arbo.NewTree(database, nLevels, hashFunc)
 	if err != nil {
+		database.Close()
 		return nil, err
 	}
 	tree := &Tree{
@@ -80,6 +81,7 @@ func (t *Tree) Init(name, storageDir string) error {
 
 	mt, err := arbo.NewTree(database, 140, arbo.HashFunctionBlake2b)
 	if err != nil {
+		database.Close()
 		return err
 	}
 	t.Tree = mt
--- a/db/badgerdb/badger.go
+++ b/db/badgerdb/badger.go
@@ -1,6 +1,7 @@
 package badgerdb
 
 import (
+	"errors"
 	"os"
 
 	"github.com/dgraph-io/badger/v3"
@@ -53,12 +54,20 @@ func (tx WriteTx) Get(k []byte) ([]byte, error) {
 
 // Set implements the db.WriteTx.Set interface method
 func (tx WriteTx) Set(k, v []byte) error {
-	return tx.tx.Set(k, v)
+	if err := tx.tx.Set(k, v); errors.Is(err, badger.ErrTxnTooBig) {
+		return db.ErrTxnTooBig
+	} else {
+		return err
+	}
 }
 
 // Delete implements the db.WriteTx.Delete interface method
 func (tx WriteTx) Delete(k []byte) error {
-	return tx.tx.Delete(k)
+	if err := tx.tx.Delete(k); errors.Is(err, badger.ErrTxnTooBig) {
+		return db.ErrTxnTooBig
+	} else {
+		return err
+	}
 }
 
 // Commit implements the db.WriteTx.Commit interface method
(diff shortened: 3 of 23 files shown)


censustree/censustree.go
```diff
@@ -51,6 +51,11 @@ func New(wTx db.WriteTx, opts Options) (
 	if err != nil {
 		return nil, err
 	}
+	defer func() {
+		if err != nil {
```
mvdan: what err are you reading here? supposedly the one from tree.New, but that's a `:=`, so you're actually shadowing.

IMO it's much better if you avoid a named return parameter and defer func, and just call `database.Close` in the only case that matters: when `tree.New` fails.

censustreelegacy/arbotree/wrapper.go
```diff
@@ -33,12 +33,17 @@ var _ censustree.Tree = (*Tree)(nil)
...
-	censustree.Tree, error) {
+	_ censustree.Tree, err error) {
 	dbDir := filepath.Join(storageDir, "arbotree.db."+strings.TrimSpace(name))
 	database, err := badgerdb.New(badgerdb.Options{Path: dbDir})
 	if err != nil {
 		return nil, err
 	}
+	defer func() {
+		if err != nil {
+			database.Close()
```
mvdan: same as before.

db/badgerdb/badger.go
```diff
@@ -53,12 +53,20 @@ func (tx WriteTx) Get(k []byte) ([]byte, error) {
 
 // Set implements the db.WriteTx.Set interface method
 func (tx WriteTx) Set(k, v []byte) error {
-	return tx.tx.Set(k, v)
+	if err := tx.tx.Set(k, v); err == badger.ErrTxnTooBig {
```
mvdan: use `errors.Is` rather than `==`

tree/tree.go
```diff
@@ -227,6 +227,24 @@ func (t *Tree) FromRoot(root []byte) (*Tree, error) {
 	}, nil
 }
 
+func (t *Tree) SetRoot(wTx db.WriteTx, root []byte) error {
+	givenTx := true
+	if wTx == nil {
+		givenTx = false
```
mvdan: nit: easier if you do

```
givenTx := wTx != nil
if !givenTx {
```

this way it also mirrors the other `if !givenTx` check, so it's easier to understand they are matching each other

vochain/state.go
```diff
@@ -634,14 +634,15 @@ func (v *State) Envelope(processID, nullifier []byte, isQuery bool) (_ []byte, e
...
 func (v *State) EnvelopeExists(processID, nullifier []byte, isQuery bool) (bool, error) {
-	e, err := v.Envelope(processID, nullifier, isQuery)
-	if err != nil && err != ErrVoteDoesNotExist {
-		return false, err
-	}
-	if err == ErrVoteDoesNotExist {
+	_, err := v.Envelope(processID, nullifier, isQuery)
+	if err == ErrProcessNotFound {
+		return false, nil
+	} else if err == ErrVoteDoesNotExist {
```
mvdan: `errors.Is` for these too; this helps ensure that future error wrapping won't break our code.

db/batch/batch.go
```diff
@@ -0,0 +1,77 @@
+package batch
+
+import "go.vocdoni.io/dvote/db"
+
+// Batch is a wrapper over a db.WriteTx that implements db.WriteTx.  The batch
+// acts as a WriteTx but when it becomes too big, it automatically commits and
+// continues with a new internal db.WriteTx transparently.
```
mvdan: > Batch is a wrapper over a db.WriteTx that implements db.WriteTx

That feels a bit redundant :P How about just keeping the second part, with a small modification at the start:

```
// Batch wraps a WriteTx to automatically commit when it becomes too big,
// causing a new internal WriteTx to be created.
```

db/batch/batch.go
```diff
@@ -0,0 +1,77 @@
+package batch
```
mvdan: why make this a new package? tiny packages are usually not a good idea, as they just make your API harder to use and the code harder to develop. Plus, `db.Batch` is IMO a much nicer name than `batch.Batch` - less repetitive, and it clarifies what its purpose is and where it fits.

db/batch/batch.go
```diff
@@ -0,0 +1,77 @@
...
+func (t *Batch) Discard() {
+	t.tx.Discard()
+}
+
+// Set implements the db.WriteTx.Set interface method.  If during this
+// operation, the internal tx becomes too big, all the pending writes will be
+// committed and a new db.WriteTx will be created to continue with this and
+// future writes.
+func (t *Batch) Set(key []byte, value []byte) error {
+	if err := t.tx.Set(key, value); err == db.ErrTxnTooBig {
```
mvdan: `errors.Is` here too and any other places I might have missed

mvdan (approved):
Very nice tests!

arnaucube (approved):

censustree/censustree.go
```diff
@@ -34,7 +34,7 @@ const nLevels = 256
 // New returns a new Tree, if there already is a Tree in the
 // database, it will load it.
 func New(wTx db.WriteTx, opts Options) (
-	*Tree, error) {
+	_ *Tree, err error) {
```
mvdan: probably undo these changes now :) also no need to split these func decls into two lines, they're fairly short

mvdan (approved):

db/batch/batch.go
ed255: I've moved Batch to the db package.  In `batch_test.go` I was using the badgerdb implementation, so having the batch tests in the db package created an import cycle, so I moved `batch_test.go` to the badgerdb package (I think it makes some sense, as it's testing the Batch over badgerdb).

mvdan: That makes sense. Another possibility is to write tests in a `db_test` package, so then you can have db_test import badgerdb, and badgerdb importing db - no cycle.

PR #283, opened 2021-09-28
Store all censuses in same Database

ed255:
Previously the census manager used graviton to store the censues, and
there was a different graviton database for each census.  Now we have
moved to BadgerDB with Arbo; and BadgerDB has a significant memory usage
for each opened Database.  For this reason we switch to having a single
BadgerDB database where all censuses are stored (using Arbo) under
different prefixes.  This should solve the Out of Memory issue observed
in nodes with many censues.

Introduce the concept of census version.  We define censuses previous to
the update to Arbo + BadgerDB to be version 0, and the current
implementation of censuses to be version 1.  The vocdoni-node stores
and loads censuses indexed in the namespaces.json file.  Censuses with
version 0 (note that by using json.Unmarshal, entries in the
namespaces.json that don't have the version key will default to version
= 0) are incompatible with the current implementation and will be
removed from the namespaces.json file and their corresponding database
directories will be discarded.  For safety, if entries in namespace.json
are deleted, a backup will be made with the current time timestamp, and
the corresponding database directories will be moved to a folder for
outdated censues with the current time timestamp.

Add locking on update operations in the census Trees to avoid
inconsisntencies.  Without locking, in the best case a concurrent update
would fail, and in the worst case a concurrent update would overwrite a
previous update.

For update operations in the cenuses, use db.Batch instead of db.WriteTx
to avoid encountering "txn too big" errors, which could happen with big
censues during "AddBatch" or "ImportDump" methods.  We do this by
wrapping BadgerDB via a struct that returns a db.Batch for every WriteTx
requested.

### DIFF
--- a/census/census.go
+++ b/census/census.go
@@ -2,9 +2,11 @@
 package census
 
 import (
+	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strings"
@@ -14,6 +16,8 @@ import (
 	"go.vocdoni.io/dvote/censustree"
 	"go.vocdoni.io/dvote/crypto/ethereum"
 	"go.vocdoni.io/dvote/data"
+	"go.vocdoni.io/dvote/db"
+	"go.vocdoni.io/dvote/db/badgerdb"
 	"go.vocdoni.io/dvote/log"
 	"go.vocdoni.io/proto/build/go/models"
 )
@@ -40,19 +44,30 @@ type Namespaces struct {
 	Namespaces []Namespace `json:"namespaces"`
 }
 
+// CurrentCensusVersion history:
+// - 0: First Version, Graviton based with a separate database for each census.
+// - 1: Arbo + BadgerDB based, single database with all censuses separated by
+//      Database prefixes (using Namespace.Name).
+const CurrentCensusVersion = 1
+
 // Namespace is composed by a list of keys which are capable to execute private operations
 // on the namespace.
 type Namespace struct {
 	Type models.Census_Type `json:"type"`
 	Name string             `json:"name"`
 	Keys []string           `json:"keys"`
+	// Version of the census in this Namespace.  If the version doesn't
+	// correspond to the CurrentCensusVersion, this census is outdated and
+	// will not be used.
+	Version int `json:"version"`
 }
 
 // Manager is the type representing the census manager component
 type Manager struct {
-	StorageDir string     // Root storage data dir for LocalStorage
-	AuthWindow int32      // Time window (seconds) in which TimeStamp will be accepted if auth enabled
-	Census     Namespaces // Available namespaces
+	StorageDir string      // Root storage data dir for LocalStorage
+	db         db.Database // Database where all census trees are stored under different prefixes
+	AuthWindow int32       // Time window (seconds) in which TimeStamp will be accepted if auth enabled
+	Census     Namespaces  // Available namespaces
 
 	// TODO(mvdan): should we protect Census with the mutex too?
 	TreesMu sync.RWMutex
@@ -65,15 +80,30 @@ type Manager struct {
 	failedQueueLock sync.RWMutex
 	failedQueue     map[string]string
 	compressor
+	cancel         context.CancelFunc
+	wgQueueDaemons sync.WaitGroup
 }
 
 // Data helps satisfy an ethevents interface.
 func (m *Manager) Data() data.Storage { return m.RemoteStorage }
 
-// Init creates a new census manager.
+// BadgerDBBatched wraps badgerdb.BadgerDB turning all requested db.WriteTx
+// into db.Batch so that commits happen automatically when the Tx becomes too
+// big.
+type BadgerDBBatched struct {
+	*badgerdb.BadgerDB
+}
+
+// WriteTx returns a db.Batch that will commit automatically with the Tx
+// becomes too big.
+func (d *BadgerDBBatched) WriteTx() db.WriteTx {
+	return db.NewBatch(d.BadgerDB)
+}
+
+// Start creates a new census manager.
 // A constructor function for the interface censustree.Tree must be provided.
-func (m *Manager) Init(storageDir, rootAuthPubKey string) error {
-	nsConfig := fmt.Sprintf("%s/namespaces.json", storageDir)
+func (m *Manager) Start(storageDir, rootAuthPubKey string) (err error) {
+	nsConfig := filepath.Join(storageDir, "namespaces.json")
 	m.StorageDir = storageDir
 	m.Trees = make(map[string]*censustree.Tree)
 	m.failedQueue = make(map[string]string)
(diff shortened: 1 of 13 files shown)


census/census.go
```diff
@@ -82,6 +105,18 @@ func (m *Manager) Init(storageDir, rootAuthPubKey string) error {
...
 	m.compressor = newCompressor()
 
+	dbDir := filepath.Join(m.StorageDir, fmt.Sprintf("v%v", CensusVersion))
+	db, err := badgerdb.New(badgerdb.Options{Path: dbDir})
+	if err != nil {
+		return err
+	}
+	m.db = &BadgerDBBatched{db}
+	defer func() {
+		if err != nil {
```
ed255, line 122: @mvdan My intention here is to handle the case where the function Init returns err != nil.  I believe this will work, because in the function scope, the variable "err" refers to the return value, and on any return, whatever is returned is assigned to err, and afterwards the defer is called.  Here are some test that make me think this works: https://play.golang.org/p/2Hj7inmOWV5

census/importqueue.go
```diff
@@ -37,11 +37,12 @@ func (m *Manager) importTree(tree []byte, cid string) error {
 	} else if err != nil {
 		return fmt.Errorf("cannot create new census namespace: (%s)", err)
 	}
+	// FIXME: Can cause Txn too big
```
arnaucube: I understand that this is what is solved in this PR

censustree/censustree.go
```diff
@@ -2,28 +2,33 @@ package censustree
...
 
 // Tree implements the Merkle Tree used for census
+// Concurrent updates to the tree.Tree can lead to losing some of the updates,
+// so we don't expose the tree.Tree directly, and lock it in every method that
+// updates it.
 type Tree struct {
-	*tree.Tree
+	tree *tree.Tree
 
+	sync.Mutex
```
arnaucube, line 22: arbo tree already has a mutex for the write operations, maybe this Tree wrapper over arbo does not need it? (and consequently the new methods of the wrapper are not needed?)

ed255: When I made this PR I didn't consider that Arbo already had mutex for write operations.  Nevertheless I've reviewed the way we are using it and we are creating the WriteTx outside the mutex, which could lead to conflicts, so I think the mutex I added is necessary.  It's not ideal because then we have double locking, but as we discussed AFK, this could be addressed in Arbo by having the FooWithTx functions not lock, and Foo functions lock.  I've added a badgerdb test to validate what happens on concurrent WriteTx that write to the same key.

census/census.go
```diff
@@ -40,19 +44,26 @@ type Namespaces struct {
 	Namespaces []Namespace `json:"namespaces"`
 }
 
+const CensusVersion = 1
```
mvdan: perhaps call this CurrentCensusVersion to clarify what it means, and document the historical versions here (including 0 aka the default)

census/census.go
mvdan: Yup, this looks right. Just be careful with the `err :=` on line 112, because that could be shadowing the parent `(err error)`, in which case you might be checking the wrong error variable - the inner one instead of the outer one.

I think it's going to be easiest to reason about this if you always assign to err, instead of shadowing/redeclaring it.

census/census.go
```diff
@@ -99,20 +141,63 @@ func (m *Manager) Init(storageDir, rootAuthPubKey string) error {
...
+	var namespaces Namespaces
+	if err := json.Unmarshal(jsonBytes, &namespaces); err != nil {
 		log.Warn("could not unmarshal json config file, probably empty. Skipping")
 		return nil
 	}
+
+	// Copy censuses of the expected version into m.Census and remove old ones
+	m.Census.RootKey = namespaces.RootKey
+	cleanOldCensues := false
+	now := time.Now()
```
mvdan: perhaps easier to understand it if you call this `start`; it's technically not "now" whenever you use it.

census/census.go
```diff
@@ -99,20 +141,63 @@ func (m *Manager) Init(storageDir, rootAuthPubKey string) error {
...
+			// Version 0 are graviton censuses.  These censuses are
+			// no longer supported and belong to legacy processes.
+			cleanOldCensues = true
+			outdatedDir := filepath.Join(storageDir,
+				fmt.Sprintf("outdated.%v", now.Format(time.RFC3339)), "v0")
+			if err := os.MkdirAll(outdatedDir, 0o700); err != nil {
+				return err
+			}
+			log.Infof("Moving outdated census %v to %v", v.Name, outdatedDir)
+			dir := strings.Split(v.Name, string(os.PathSeparator))[0]
```
mvdan, line 181: a bit easier to use filepath.SplitList. Or, if there's just one separator, filepath.Dir is the equivalent of getting the parent directory.

census/census.go
```diff
@@ -99,20 +141,63 @@ func (m *Manager) Init(storageDir, rootAuthPubKey string) error {
...
+			src := filepath.Join(storageDir, dir)
+			dst := filepath.Join(outdatedDir, dir)
+			if err := os.Rename(src, dst); err != nil {
```
mvdan, line 184: do you expect both to be in the same filesystem? because if not, this rename might fail

census/census.go
```diff
@@ -65,15 +76,30 @@ type Manager struct {
...
+func (d *BadgerDBBatched) WriteTx() db.WriteTx {
+	return db.NewBatch(d.BadgerDB)
+}
+
 // Init creates a new census manager.
 // A constructor function for the interface censustree.Tree must be provided.
-func (m *Manager) Init(storageDir, rootAuthPubKey string) error {
-	nsConfig := fmt.Sprintf("%s/namespaces.json", storageDir)
+func (m *Manager) Init(storageDir, rootAuthPubKey string) (err error) {
+	nsConfig := filepath.Clean(filepath.Join(storageDir, "namespaces.json"))
```
mvdan: not sure why we have all these `filepath.Clean` invocations. if we're worried that `storageDir` has bits that cancel out, like `foo/bar/..`, we should clean those at startup and avoid calling Clean again in five separate places.

census/census.go
```diff
@@ -127,15 +212,24 @@ func (m *Manager) Init(storageDir, rootAuthPubKey string) error {
 	return nil
 }
 
+// Deinit terminates the Manager goroutines and closes the database.  The
+// Manager can't be used after Deinit is called.
+func (m *Manager) Deinit() error {
```
mvdan: Usually this is called Shutdown :)

mvdan: Speaking of which, Init should be called Start if it fires goroutines. Because Init has a connotation that it just initializes memory or storage, not that it starts background processes.

census/census.go
```diff
@@ -65,15 +76,30 @@ type Manager struct {
 	failedQueueLock sync.RWMutex
 	failedQueue     map[string]string
 	compressor
+	cancel context.CancelFunc
+	wg     sync.WaitGroup
```
mvdan: I'd give this a better name, like `wgQueueDaemons`, to clarify what it's used for

mvdan (approved):
Regarding the mutex, remember that if it only protects one field, you could use atomics instead.

census/census.go
ed255: I've settled with `timestamp` because there's no notion of "start/stop", and it's a timestamp to annotate when something has happened.

census/census.go
ed255: Renamed Init to Start, and Deinit to Stop.

census/census.go
ed255: I don't expect the dst and src to be in a different filesystem, because both dst and src share the same ancestor directory (storageDir).  It could happen, but it would be a strange situation.  Nevertheless, I haven't found any function in the standard library that does the same as `mv`; and I don't think it's worth adding a dependency to have this function as it would be a very strange case.

census/census.go
ed255: Makes sense.  I didn't remove it because I though it was there for some reason, I'll see if I can move it upwards.

census/census.go
mvdan: Yep agreed. I mentioned this in case the user could configure src and dst to end up in separate filesystems in a reasonable way, but it seems to not be the case.

census/census.go
mvdan: I realise some of what I bring up is not your own doing. But since we're here... :)

census/census.go
ed255: Ah! That's what I thought!  But then realized that `SplitList` separates by `:` in Unix, and it's meant to separate lists of paths found in `$PATH` for example.
In this case there are either 1 or 2 elements in the Name, and `Dir` would only work when there are 2 elements.  I just wanted a function that returns the list of elements in a path, and didn't find such function, so I used `strings.Split` with `os.PathSeparator`

mvdan: Oops, you're absolutely right.

How about:

```
if dir := filepath.Dir(v.Name); dir != "." {
    v.Name = dir
}
```

Slightly longer, but also more obvious with its logic, if you ask me.

ed255: Basically there are two cases:
- a: `v.Name = "foo"` -> `mv src/foo dst/foo`
- b: `v.Name = "bar/baz"` -> `mv src/bar dst/bar`

I wouldn't want to change the content of `v.Name`.
I just want to take the first entry (as `dir`) in the path and move it from `src` to `dst`.  I don't see why it's not obvious the way I implemented it: splitting the path by '/' and taking the first element.

mvdan: I botched my example above, yeah, it shouldn't be updating `v.Name`. Your method sounds fine.

arnaucube (approved):

PR #284, opened 2021-09-28
Integrate zk-artifacts download on dvote-node init

arnaucube:
Integrate zk-artifacts download on dvote-node init to download the
verification_key.json defined in the Vochain genesis, and load it in
memory for vochain usage.

### DIFF
--- a/crypto/zk/artifacts/artifacts.go
+++ b/crypto/zk/artifacts/artifacts.go
@@ -11,6 +11,8 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+
+	"go.vocdoni.io/dvote/log"
 )
 
 /*
@@ -42,18 +44,18 @@ const (
 	FilenameWitness = "witness.wasm"
 	// FilenameZKey defines the name of the file of the circom ZKey
 	FilenameZKey = "circuit.zkey"
-	// FilenameVK defines the name of the VerificationKey.json
-	FilenameVK = "verificationKey.json"
+	// FilenameVK defines the name of the verification_key.json
+	FilenameVK = "verification_key.json"
 )
 
-// CircuitsConfig defines the configuration of the files to be downloaded
-type CircuitsConfig struct {
+// CircuitConfig defines the configuration of the files to be downloaded
+type CircuitConfig struct {
 	// URL defines the URL from where to download the files
 	URL string
-	// CircuitsPath defines the path from where the files are downloaded
-	CircuitsPath string
+	// CircuitPath defines the path from where the files are downloaded
+	CircuitPath string
 	// LocalDir defines in which directory will be the files
-	// downloaded, under that directory it will follow the CircuitsPath
+	// downloaded, under that directory it will follow the CircuitPath
 	// directories structure
 	LocalDir string
 
@@ -69,11 +71,12 @@ type CircuitsConfig struct {
 // checking the expected sha256 hash of each file. If the files already exist
 // in the path, it will not download them again but it will check the hash of
 // the existing files.
-func DownloadCircuitFiles(ctx context.Context, c CircuitsConfig) error {
+func DownloadCircuitFiles(ctx context.Context, c CircuitConfig) error {
 	// check if files already exist, if files already exist, check hashes
 	err := checkHashes(c)
 	if err == nil {
 		// files already exist, and match the expected hashes
+		log.Info("files already exist and match the expected hashes")
 		return nil
 	}
 	if err != nil && !os.IsNotExist(err) {
@@ -82,6 +85,7 @@ func DownloadCircuitFiles(ctx context.Context, c CircuitsConfig) error {
 
 	// at this point, or files do not exist, or exist with a not expected
 	// hash
+	log.Info("files or don't exist locally or exist with a not expected hash. Proceed with fresh download")
 	return downloadFiles(ctx, c)
 }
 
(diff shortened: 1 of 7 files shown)


crypto/zk/artifacts/artifacts.go
```diff
@@ -82,57 +85,61 @@ func DownloadCircuitFiles(ctx context.Context, c CircuitsConfig) error {
...
 	if err == nil {
 		// VK file already exist, and match the expected hash
+		log.Info("VK file already exist and match the expected hash")
 		return err
 	}
 	if err != nil && !os.IsNotExist(err) {
 		return err
 	}
 
+	log.Info("VK file or does not exist locally or exist with a not expected hash. Proceed with freash download")
```
ed255:
```suggestion
	log.Info("VK file or does not exist locally or exist with a not expected hash. Proceed with fresh download")
```

crypto/zk/artifacts/artifacts.go
```diff
@@ -227,6 +229,16 @@ func downloadFile(ctx context.Context, fromUrl, toPath string) error {
 	}
 	defer resp.Body.Close()
 
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("error on download file %s: %d", fromUrl, resp.StatusCode)
```
ed255:
```suggestion
		return fmt.Errorf("error on download file %s: http status is %d", fromUrl, resp.StatusCode)
```
I think indicating what the number is will make it more clear.

crypto/zk/artifacts/artifacts_test.go
```diff
@@ -95,26 +95,48 @@ func TestDownloadCircuitFiles(t *testing.T) {
...
+		CircuitPath: "/zkcensusproof/test/1024/",
+		LocalDir:    path,
+		WitnessHash: nil,
+		ZKeyHash:    nil,
+		VKHash:      nil,
+	}
+	ctx := context.Background()
+	err := DownloadCircuitFiles(ctx, c)
+	qt.Assert(t, err, qt.IsNotNil)
+	qt.Assert(t, strings.Contains(err.Error(), "error on download file"), qt.IsTrue)
```
ed255: Check this out https://pkg.go.dev/github.com/frankban/quicktest#readme-errormatches

service/vochain.go
```diff
@@ -146,6 +150,30 @@ func Vochain(vconfig *config.VochainCfg, waitForSync bool,
 		)
 	}
 
+	// get the zk Circuits VerificationKey files
+	vnode.ZkVKs = []*snarkTypes.Vk{}
+	for i := 0; i < len(vochain.Genesis[vconfig.Chain].CircuitsConfig); i++ {
```
ed255: You could do:
```go
for i, cc := range vochain.Genesis[vconfig.Chain].CircuitsConfig {
```

service/vochain.go
```diff
@@ -146,6 +150,30 @@ func Vochain(vconfig *config.VochainCfg, waitForSync bool,
...
+		log.Infof("downloading zk-circuits-artifacts index: %d", i)
+		cc := vochain.Genesis[vconfig.Chain].CircuitsConfig[i]
+
+		// download VKs from CircuitsConfig
+		ctx := context.Background()
```
ed255: I suggest using a timeout here, of maybe 1 to 5 minutes

vochain/genesis.go
```diff
@@ -143,6 +147,20 @@ var Genesis = map[string]VochainGenesis{
...
 			"7440a5b086e16620ce7b13198479016aa2b07988@seed.dev.vocdoni.net:26656"},
+		CircuitsConfig: []artifacts.CircuitConfig{
+			{ // index: 0, size: 1024
+				URL:         "https://raw.githubusercontent.com/vocdoni/zk-circuits-artifacts/master/",
+				CircuitPath: "/zkcensusproof/dev/1024",
+				LocalDir:    "./circuits",
+				WitnessHash: nil,
+				ZKeyHash:    nil,
+				// hex VkHash: 0x3669e12ea939564b59b995b9067eab83c8ebb09f5a83ad4aa3d6d6f90c1b0fc4
+				VKHash: []byte{54, 105, 225, 46, 169, 57, 86,
```
ed255: maybe you can use helper function here that takes a `string` in hex and returns `[]byte`, and panics on invalid inputs.  I'm saying this because in the future there will be at least 3 hashes (dev, stage and main), and we may update some of them.  I think this would make it more readable and avoid needing an extra comment.

Something like this:
```go
func Hex2Bytes(s string) []byte {
  b, err := hex.DecodeString(s)
  if err != nil {
    log.Fatalf("Error decoding hex string %v: %v", s, err)
  }
  return b
}
```

ed255 (approved):

ed255 (approved):

crypto/zk/artifacts/artifacts.go
```diff
@@ -227,6 +229,16 @@ func downloadFile(ctx context.Context, fromUrl, toPath string) error {
...
 
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("error on download file %s: http status: %d", fromUrl, resp.StatusCode)
+	}
+
+	out, err := os.Create(toPath)
+	if err != nil {
+		return err
+	}
+	defer out.Close()
```
mvdan, line 240: nit: you should check this file error close by leaving the defer here and adding this at the end of the func:

```
if err := out.Close(); err != nil {
    return err
}
```

This works because closing a file is idempotent: you can do it twice and it's okay.

crypto/zk/artifacts/artifacts_test.go
```diff
@@ -95,26 +95,49 @@ func TestDownloadCircuitFiles(t *testing.T) {
...
+	defer ts.Close()
+
+	path := t.TempDir()
+	c := CircuitConfig{
+		URL:         ts.URL,
+		CircuitPath: "/zkcensusproof/test/1024/",
+		LocalDir:    path,
+		WitnessHash: nil,
+		ZKeyHash:    nil,
+		VKHash:      nil,
```
mvdan: nit: all these fields which have zero value can be ommitted :)

service/vochain.go
```diff
@@ -146,6 +150,30 @@ func Vochain(vconfig *config.VochainCfg, waitForSync bool,
...
+		defer cancel()
+		err = zkartifacts.DownloadVKFile(ctx, cc)
+		if err != nil {
+			return
+		}
+
+		// parse VK and store it into vnode.ZkVKs
+		log.Infof("parse VK from file into memory. CircuitArtifact index: %d", i)
+		var vk *snarkTypes.Vk
+		vk, err = zk.LoadVkFromFile(filepath.Join(cc.LocalDir, cc.CircuitPath, zkartifacts.FilenameVK))
```
mvdan, line 169: use `:=` and avoid the `var`?

vochain/genesis.go
```diff
@@ -13,6 +22,7 @@ var Genesis = map[string]VochainGenesis{
 	"main": {
 		AutoUpdateGenesis: false,
 		SeedNodes:         []string{"121e65eb5994874d9c05cd8d584a54669d23f294@seed.vocdoni.net:26656"},
+		CircuitsConfig:    nil,
```
mvdan: ditto with the zero value

mvdan (commented):
just a few extra nits :)

crypto/zk/artifacts/artifacts.go
mvdan: why this matters: if there's an error when flushing writes to disk, Close might error, so you don't want to ignore that.

service/vochain.go
arnaucube: I used the `var` because the `err` is the one returned in the function, and if I use `vk, err :=` the compiler complains that 'result parameter err not in scope at return`. But I'm down to use a better way instead than the current one

mvdan: oh, right, named result parameters and naked returns. I'd strongly recommend avoiding those :) See https://github.com/golang/go/wiki/CodeReviewComments#named-result-parameters and https://github.com/golang/go/wiki/CodeReviewComments#naked-returns.

If I see a naked return 50 or 100 lines after the func declaration, I have no idea what I'm returning.

arnaucube: Agree with what you mention, in this case this *naked return* was there prior to this PR, but I'm very agree with what you mentioned, so I'll update the function to not use the naked return.

PR #296, opened 2021-10-04
vochain: stop using tendermint's tempfile package

mvdan:
(see commit message)

### DIFF
--- a/vochain/apputils.go
+++ b/vochain/apputils.go
@@ -5,6 +5,7 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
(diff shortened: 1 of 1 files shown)


mvdan:
How important is it that we write this file atomically? For example, that we don't leave a partial file written.

If that's important, then my suggestion is to use a maintained library for it, like https://github.com/google/renameio. I'm happy to update the PR to use that. We could also make other bits of code use it, if they also need atomic file writes.

p4u:
I don't think it needs to be atomic but I'm not sure. We can merge and see if this generate some error while still in development.

p4u (approved):

PR #301, opened 2021-10-05
Feature/pebbledb

arnaucube:
- Abstract db tests
    - no new tests where created, where abstracted into `db/dbtest` (and updated to use `quicktest` for checks)
    - in badgerdb, the test that was in `db/badgerdb/batch_test.go` has been moved into `db/badgerdb/badger_test.go`
- Add [pebbledb](https://github.com/cockroachdb/pebble) db interface implementation


vochain/types.go
```diff
@@ -23,8 +23,8 @@ var (
 	ErrVoteDoesNotExist = fmt.Errorf("vote does not exist")
 	ErrProcessNotFound  = fmt.Errorf("process not found")
 	// keys; not constants because of []byte
-	headerKey    = []byte("header")
-	heightKey    = []byte("height")
+	headerKey = []byte("header")
+	// heightKey    = []byte("height") // unused
```
mvdan: remove it then?

db/badgerdb/badger_test.go
```diff
@@ -1,124 +1,57 @@
...
 
 import (
-	"bytes"
-	"strconv"
+	"encoding/binary"
 	"testing"
 
+	qt "github.com/frankban/quicktest"
 	"go.vocdoni.io/dvote/db"
+	"go.vocdoni.io/dvote/db/dbtest"
```
mvdan: move this package to `db/internal/dbtest`. we don't want it to be importable by others.

db/badgerdb/badger_test.go
```diff
@@ -1,124 +1,57 @@
...
-	if err := d.Iterate(nil, func(k, v []byte) bool {
-		noPrefixKeysFound++
-		return true
-	}); err != nil {
-		t.Fatal(err)
-	}
-	if noPrefixKeysFound != (prefix0NumKeys + prefix1NumKeys) {
-		t.Errorf("expected %d keys, found %d", prefix0NumKeys+prefix1NumKeys, noPrefixKeysFound)
-	}
+func TestBatch(t *testing.T) {
```
mvdan, line 26: why not put this one in dbtest too?

db/pebbledb/pebledb.go
```diff
@@ -0,0 +1,158 @@
...
+var _ db.ReadTx = (*ReadTx)(nil)
+
+// WriteTx implements the interface db.WriteTx
+type WriteTx struct {
+	batch *pebble.Batch
+}
+
+// check that WriteTx implements the db.ReadTx & db.WriteTx interfaces
+var _ db.ReadTx = (*WriteTx)(nil)
+var _ db.WriteTx = (*WriteTx)(nil)
```
mvdan, line 24: writetx is already a readtx, so the first line here is redundant

db/pebbledb/pebledb.go
```diff
@@ -0,0 +1,158 @@
...
+}
+
+// Get implements the db.WriteTx.Get interface method
+func (tx WriteTx) Get(k []byte) ([]byte, error) {
+	return ReadTx(tx).Get(k)
+}
+
+// Set implements the db.WriteTx.Set interface method
+func (tx WriteTx) Set(k, v []byte) error {
+	return tx.batch.Set(k[:], v, nil)
```
mvdan: I think that `[:]` expression is a no-op, since `k` is already a slice.

db/pebbledb/pebledb.go
```diff
@@ -0,0 +1,158 @@
...
+		LowerBound: prefix,
+		UpperBound: keyUpperBound(prefix),
+	}
+	iter := db.db.NewIter(iterOptions)
+	defer func() {
+		errC := iter.Close()
+		if err != nil {
+			return
+		}
+		err = errC
```
mvdan, line 146: I don't think this error ends up anywhere, unless you use `) (err error) {` for the return parameter.

mvdan (approved):

db/pebbledb/pebledb.go
arnaucube: Right! Updated.

vochain/types.go
arnaucube: Initially I did not remove it as was 'out of the scope' of this PR, but appeared in the goanalyze checks. But I'm agree with what you say, so I've remove it in the last update.

db/badgerdb/badger_test.go
arnaucube: This test checks that the badgerdb implementation gets an error of `badger.ErrTxnTooBig` (in the form of `db.ErrTxnTooBig`), so for the moment I thought to leave it as specific test of badgerdb package.

mvdan: Hm, I see. I'll probably try something here in my follow-up PR. SGTM

db/badgerdb/badger.go
```diff
@@ -27,7 +27,6 @@ type WriteTx struct {
 }
 
 // check that WriteTx implements the db.ReadTx & db.WriteTx interfaces
-var _ db.ReadTx = (*WriteTx)(nil)
```
ed255, line 30: why?

db/pebbledb/pebledb.go
```diff
@@ -0,0 +1,156 @@
+package pebbledb
```
ed255, line 1: the filename should be `pebbledb.go`

db/pebbledb/pebledb.go
```diff
@@ -0,0 +1,156 @@
...
+
+// check that ReadTx implements the db.ReadTx interface
+var _ db.ReadTx = (*ReadTx)(nil)
+
+// WriteTx implements the interface db.WriteTx
+type WriteTx struct {
+	batch *pebble.Batch
+}
+
+// check that WriteTx implements the db.ReadTx & db.WriteTx interfaces
```
ed255, line 23: missing check for interface ReadTx

db/badgerdb/badger.go
arnaucube: https://github.com/vocdoni/vocdoni-node/pull/301#discussion_r723009131

db/pebbledb/pebledb.go
arnaucube: https://github.com/vocdoni/vocdoni-node/pull/301#discussion_r723009131

db/pebbledb/pebledb.go
arnaucube: Right! I've missed a `b` :S

db/badgerdb/badger.go
ed255: Ok, got it from https://github.com/vocdoni/vocdoni-node/pull/301#discussion_r723009131

db/pebbledb/pebledb.go
ed255: Ok, got it from https://github.com/vocdoni/vocdoni-node/pull/301#discussion_r723009131

PR #304, opened 2021-10-06
allow configure key value database backend, use pebble by default

p4u:
Current implementations: pebble (default), badger v2 and badger v3
New flag added "--dbBackend" for specifying backend to use with the Vochain state

### DIFF
--- a/cmd/dvotenode/dvotenode.go
+++ b/cmd/dvotenode/dvotenode.go
@@ -22,6 +22,7 @@ import (
 	"go.vocdoni.io/dvote/config"
 	"go.vocdoni.io/dvote/crypto/ethereum"
 	"go.vocdoni.io/dvote/data"
+	"go.vocdoni.io/dvote/db"
 	ethchain "go.vocdoni.io/dvote/ethereum"
 	"go.vocdoni.io/dvote/ethereum/ethevents"
 	"go.vocdoni.io/dvote/internal"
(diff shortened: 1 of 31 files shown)


p4u:
ping @mvdan

cmd/dvotenode/dvotenode.go
```diff
@@ -126,6 +127,8 @@ func newConfig() (*config.DvoteCfg, config.Error) {
 		"external address:port to announce to other peers (automatically guessed if empty)")
 	globalCfg.VochainConfig.RPCListen = *flag.String("vochainRPCListen", "127.0.0.1:26657",
 		"rpc host and port to listen to for the voting chain")
+	globalCfg.VochainConfig.DBbackend = db.StorageType(*flag.Int("dbBackend", 0,
+		"database backend for storing the vochain state (0:pebble 1:badger2 2:badger3"))
```
mvdan, line 131: how about a string with those three options? much more usable and readable IMO :)

vochain/state.go
```diff
@@ -217,13 +219,23 @@ type State struct {
...
+	switch storageType {
+	case db.StoragePebble:
+		kv, err = pebbledb.New(pebbledb.Options{Path: filepath.Join(dataDir, "vcstate")})
+	case db.StorageBadgerv2:
+		kv, err = badgerdbv2.New(badgerdbv2.Options{Path: filepath.Join(dataDir, "vcstate")})
+	case db.StorageBadgerv3:
+		kv, err = badgerdbv3.New(badgerdbv3.Options{Path: filepath.Join(dataDir, "vcstate")})
+	default:
+		return nil, fmt.Errorf("storage type %d not recognized", storageType)
+	}
```
mvdan, line 234: if you put this "meta constructor" somewhere reusable, like in `db/metadb.New(storageType string)`, then the tests that currently hard-code pebble or badger could do something like:

```
metadb.New(os.Getenv("DB_BACKEND"))
```

`metadb.New` could also have a default when the string is empty, e.g. pebble.

mvdan (commented):
I'm reluctant to pull in badger/v2, because it's basically unmaintained at this point. Pulling in badger once as a DB backend seems enough to me :)

vochain/state.go
mvdan: I know that you went with an enum on purpose, but I think a string is a bit better in terms of UX, and allows us to use env vars directly too. And the switch in `metadb.New` can simply reject invalid/unknown strings.

p4u:
superseded by #305

PR #305, opened 2021-10-06
Integrate pebbledb

arnaucube:
Integrate pebbledb:
- Integrate pebbledb in the census & statedb
- Add flag to define which db implementation to use: PebbleDB or
BadgerDB

Currently the default db is pebble, but can be set to use badger by a flag. Maybe before merging this PR would be worth to test it with in the dev environment.

### DIFF
--- a/db/interface.go
+++ b/db/interface.go
@@ -5,12 +5,24 @@ import (
 	"io"
 )
 
+const (
+	// TypePebble defines the type of db that uses PebbleDB
+	TypePebble = "pebble"
+	// TypeBadger defines the type of db that uses BadgerDB
+	TypeBadger = "badger"
+)
+
 // ErrKeysNotFound is used to indicate that a key does not exist in the db.
 var ErrKeyNotFound = fmt.Errorf("key not found")
 
 // ErrTxnTooBig is used to indicate that a WriteTx is too big and can't include
 // more writes.
 var ErrTxnTooBig = fmt.Errorf("txn too big")
+
+// Options defines params for creating a new Database
+type Options struct {
+	Path string
+}
 
 // Database wraps all database operations. All methods are safe for concurrent
 // use.
--- a/db/metadb/metadb.go
+++ b/db/metadb/metadb.go
@@ -1,0 +1,30 @@
+package metadb
+
+import (
+	"fmt"
+
+	"go.vocdoni.io/dvote/db"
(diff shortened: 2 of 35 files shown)


db/metadb/metadb.go
```diff
@@ -0,0 +1,31 @@
...
+		if err != nil {
+			return nil, err
+		}
+	case types.DBTypeBadger:
+		database, err = pebbledb.New(opts)
+		if err != nil {
+			return nil, err
+		}
+	default:
+		return nil, fmt.Errorf("invalid dbType: %s. Available types: %s, %s", typ, types.DBTypePebble, types.DBTypeBadger)
```
mvdan: use `%q` to quote the strings, which will be better for e.g. empty strings or strings with spaces

db/interface.go
```diff
@@ -12,6 +12,11 @@ var ErrKeyNotFound = fmt.Errorf("key not found")
 // more writes.
 var ErrTxnTooBig = fmt.Errorf("txn too big")
 
+// Options defines params for creating a new Database
+type Options struct {
```
mvdan, line 23: I see why you wanted to move the type here, since it's the same across all db implementations, but note that some implementations might want to have custom extra options at a later time. so this might be a bad idea.

types/consts.go
```diff
@@ -25,6 +25,11 @@ const (
 	// ModeEthAPIoracle start an Ethereum signaling API oracle
 	ModeEthAPIoracle = "ethApiOracle"
 
+	// DBType pebble
+	DBTypePebble = "pebble"
+	// DBType badger
+	DBTypeBadger = "badger"
```
mvdan: define these consts in the `db` package, like `db.TypePebble`? I'd prefer that over placing them in the types package, because the types package is a mixed bag of random stuff :)

mvdan (approved):
Perhaps wait for Pau's review too, given the PR conflict

db/interface.go
arnaucube: Agree, my reasoning has been to not fall into 'premature abstractions' too early, and if at some point we need some parameters only for one specific implementation we can define a way to abstract/resolve it.

types/consts.go
arnaucube: Right, updated as you suggest.

p4u (approved):

PR #306, opened 2021-10-06
db: use finalizers for sanity checks, add test APIs

mvdan:
(see commit message)


mvdan:
I can try to add similar "leak checks" to pebbledb later, though I don't see any API to detect if the database is closed.

p4u (approved):

PR #312, opened 2021-10-11
[DO NOT MERGE] Tmp update Tree.AddBatch to use Add loop

arnaucube:
Temporary update while `arbo.AddBatch` works full on memory. The idea is this week to have the AddBatch working also in disk, so this can be updated into the new version of `arbo`, but meanwhile we can test the node without 'noise' of high memory consumption of the AddBatch.

### DIFF
--- a/tree/tree.go
+++ b/tree/tree.go
@@ -1,9 +1,11 @@
 package tree
 
 import (
+	"bytes"
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"io"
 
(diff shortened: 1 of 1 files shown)


ed255 (approved):

arnaucube:
Closing this 'temporary PR', as this will be covered by the new arbo version from https://github.com/vocdoni/arbo/tree/feature/addbatch-bigtrees-indisk

PR #313, opened 2021-10-11
IPv6 support

altergui:
this is a work-in-progress exploring how would dualstack support mean in ipfsSync.go
```
2021-10-11T14:43:34+02:00	INFO	ipfssync/ipfssync.go:412	my multiaddress ipv4: /ip4/95.xxx.xxx.2/tcp/4001/p2p/QmTj3Zrrk56jYYpwtdRWzKFFHCZM1HdujTVRRTrm6UfRn9
2021-10-11T14:43:34+02:00	INFO	ipfssync/ipfssync.go:415	my multiaddress ipv6: /ip6/2a02:xxxx:fede::feca/tcp/4001/p2p/QmTj3Zrrk56jYYpwtdRWzKFFHCZM1HdujTVRRTrm6UfRn9
2021-10-11T14:43:34+02:00	DEBUG	ipfssync/ipfssync.go:223	broadcasting message HELLO {Address:QmRH37orTHrt3e8xcE3rFi7jdHBw3tJf1AThgU9GyrmwsD Hash: MA:/ip4/95.xxx.xxx.2/tcp/4001/p2p/QmTj3Zrrk56jYYpwtdRWzKFFHCZM1HdujTVRRTrm6UfRn9 PL:[] Ts:1633956214}
2021-10-11T14:43:34+02:00	DEBUG	subpubtransport/subpub.go:88	sending 135 bytes to broadcast channel
2021-10-11T14:43:34+02:00	DEBUG	ipfssync/ipfssync.go:223	broadcasting message HELLO {Address:QmRH37orTHrt3e8xcE3rFi7jdHBw3tJf1AThgU9GyrmwsD Hash: MA:/ip6/2a02:xxxx:fede::feca/tcp/4001/p2p/QmTj3Zrrk56jYYpwtdRWzKFFHCZM1HdujTVRRTrm6UfRn9 PL:[] Ts:1633956214}
2021-10-11T14:43:34+02:00	DEBUG	subpubtransport/subpub.go:88	sending 152 bytes to broadcast channel
2021-10-11T14:43:34+02:00	DEBUG	subpub/discovery.go:43	found peer: QmatPDpxw8ovckRZ7vJkp7aPXuv5xxFeBXaCTtL78QtG6r
2021-10-11T14:43:34+02:00	INFO	subpub/stream.go:40	connected to peer QmatPDpxw8ovckRZ7vJkp7aPXuv5xxFeBXaCTtL78QtG6r: /ip6/2a02:xxxx::1/tcp/4171
```

### DIFF
--- a/ipfssync/ipfssync.go
+++ b/ipfssync/ipfssync.go
@@ -28,6 +28,8 @@ import (
 
 const (
 	MaxKeySize = 64
+	IPv4       = 4
+	IPv6       = 6
 )
 
 type Message struct {
@@ -54,12 +56,13 @@ type IPFSsync struct {
 	Timeout         time.Duration
 	TimestampWindow int32
 
-	hashTree    statedb.StateTree
-	state       statedb.StateDB
-	updateLock  sync.RWMutex
-	myMultiAddr ma.Multiaddr // The IPFS multiaddress
-	lastHash    []byte
-	private     bool
+	hashTree        statedb.StateTree
+	state           statedb.StateDB
+	updateLock      sync.RWMutex
+	myMultiAddrIPv4 ma.Multiaddr // The IPFS multiaddress (IPv4)
+	myMultiAddrIPv6 ma.Multiaddr // The IPFS multiaddress (IPv6)
+	lastHash        []byte
+	private         bool
 }
 
 // NewIPFSsync creates a new IPFSsync instance. Transports supported are "libp2p" or "privlibp2p"
@@ -90,8 +93,8 @@ func NewIPFSsync(dataDir, groupKey, privKeyHex, transport string, storage data.S
 }
 
 // shity function to workaround NAT problems (hope it's temporary)
-func guessMyAddress(port int, id string) string {
-	ip, err := util.PublicIP()
+func guessMyAddress(ipversion uint, port int, id string) string {
+	ip, err := util.PublicIP(ipversion)
 	if err != nil {
 		log.Warnf("guessMyAddress: %v", err)
 		return ""
@@ -100,7 +103,7 @@ func guessMyAddress(port int, id string) string {
 		return fmt.Sprintf("/ip4/%s/tcp/%d/ipfs/%s", ip4, port, id)
 	}
 	if ip6 := ip.To16(); ip6 != nil {
-		return fmt.Sprintf("/ip6/[%s]/tcp/%d/ipfs/%s", ip6, port, id)
+		return fmt.Sprintf("/ip6/%s/tcp/%d/ipfs/%s", ip6, port, id)
 	}
 	return ""
 }
@@ -309,14 +312,25 @@ func (is *IPFSsync) sendUpdate() {
 	}
 }
 
-func (is *IPFSsync) sendHello() {
+func (is *IPFSsync) sendHelloWithAddr(multiaddress string) {
 	var msg models.IpfsSync
 	msg.Msgtype = models.IpfsSync_HELLO
 	msg.Address = is.Transport.Address()
-	msg.Multiaddress = is.myMultiAddr.String()
+	msg.Multiaddress = multiaddress
 	err := is.broadcastMsg(&msg)
 	if err != nil {
 		log.Warnf("sendHello: %v", err)
+	}
+}
+
+func (is *IPFSsync) sendHello() {
+	// send HELLO with IPv4 address, if defined
+	if is.myMultiAddrIPv4 != nil {
+		is.sendHelloWithAddr(is.myMultiAddrIPv4.String())
+	}
+	// send HELLO with IPv6 address, if defined
+	if is.myMultiAddrIPv6 != nil {
+		is.sendHelloWithAddr(is.myMultiAddrIPv6.String())
 	}
 }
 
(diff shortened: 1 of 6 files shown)


ipfssync/ipfssync.go
```diff
@@ -388,11 +402,18 @@ func (is *IPFSsync) Start() {
...
 
-	is.myMultiAddr, err = ma.NewMultiaddr(guessMyAddress(4001, is.Storage.Node.PeerHost.ID().String()))
-	if err != nil {
-		log.Fatal(err)
+	var err4, err6 error
+	is.myMultiAddrIPv4, err4 = ma.NewMultiaddr(guessMyAddress(IPv4, 4001, is.Storage.Node.PeerHost.ID().String()))
+	is.myMultiAddrIPv6, err6 = ma.NewMultiaddr(guessMyAddress(IPv6, 4001, is.Storage.Node.PeerHost.ID().String()))
+	if err4 != nil && err6 != nil {
+		log.Fatal("ipv4: %s; ipv6: %s", err4, err6)
+	}
```
mvdan, line 410: do we want to allow having both be non-nil? your code does that, just want to double check.

util/net.go
```diff
@@ -10,16 +10,18 @@ import (
...
 // If a nil error is returned, the returned IP must be valid.
-func PublicIP() (net.IP, error) {
+
+func PublicIP(ipversion uint) (net.IP, error) {
 	consensus := externalip.DefaultConsensus(nil, nil)
-	ip, err := consensus.ExternalIP(4)
-	// if the IP isn't a valid ipv4, To4 will return nil
-	if ip = ip.To4(); ip == nil {
-		return nil, fmt.Errorf("public IP discovery failed: %v", err)
+	consensus.UseIPProtocol(ipversion)
```
mvdan, line 19: Hmm, where is this defined? Shouldn't the `ipversion` be passed to `ExternalIP`?

altergui:
hey, thanks a lot for the review!! i'm just slowly starting to get a grasp on all this, so getting some feedback makes me feel a little bit less lost :) and is much appreciated

ipfssync/ipfssync.go
altergui: yes, if both ipv4 and ipv6 are discovered, it will send two HELLOs, one with /ip4/ and one with /ip6/, but pointing to the same nodeID (last part of the multiaddr), which in essence means "you can reach me over these two addresses". 

it should only fail if *no* addresses are discovered (so both err4 and err6 are non-nil)
if either ipv4 or ipv6 is available, that's OK.

util/net.go
altergui: i'm using @jc21 syntax since i `replace`d go-external-ip in go.mod
https://github.com/vocdoni/vocdoni-node/pull/313/files/c871e212feb89c1cf3d26795850c771ab63e50ab#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R12

mvdan: Ah I see. Given that upstream will likely review/merge that PR in the next few days, I'd say we should wait until that happens to merge this, because we should hopefully be able to get rid of our fork :)

mvdan:
Looks like upstream did a release with this feature :) https://github.com/GlenDC/go-external-ip/releases/tag/v0.1.0

altergui:
rebased, squashed, pointed go-external-ip to upstream, and retested, still good :)

ipfssync/ipfssync.go
```diff
@@ -57,7 +59,8 @@ type IPFSsync struct {
 	hashTree    statedb.StateTree
 	state       statedb.StateDB
 	updateLock  sync.RWMutex
-	myMultiAddr ma.Multiaddr // The IPFS multiaddress
+	myMultiAddrIPv4 ma.Multiaddr // The IPFS multiaddress (IPv4)
+	myMultiAddrIPv6 ma.Multiaddr // The IPFS multiaddress (IPv6)
```
mvdan, line 63: re-run gofmt?

ipfssync/ipfssync.go
```diff
@@ -28,6 +28,8 @@ import (
 
 const (
 	MaxKeySize = 64
+	IPv4 = 4
+	IPv6 = 6
```
mvdan: Hmm, would be nice if we defined these constants in a single place, but I guess we don't have a good place in this module. Perhaps if upstream exposed the constants it would be better.

mvdan (approved):
SGTM!

ipfssync/ipfssync.go
altergui: right

altergui:
gofmt'd

mvdan:
SGTM. We should probably squash-merge, unless you want to squash your commits yourself.

altergui:
sounds good, squashed

mvdan (approved):
LGTM! Up to you if you want a review from @p4u as well. If not, we can merge on green.

emmdim:
@mvdan @altergui @jordipainan  any concrete reason the compose-test is fail or should we simply just re-run the test?

p4u:
I am re-running the compose test, lets see.

@altergui you can run the compose test in your local machine like this too:
```
cd dockerfiles/testsuite
bash start_test.sh
```

EDIT: you will need support for docker buildKit ( `{ "features": { "buildkit": true } }` >> /etc/docker/daemon.json )

altergui:
i did introduce a small (but critical) regression in this PR:
setting ListenAddr to `/ip6/::/tcp/%d` does NOT make it dualstack but ipv6-only!
for listening on ipv4 as well, `/ip4/0.0.0.0/tcp/%d` needs to be added as well. 

Sorry about that, I squashed this bug in https://github.com/vocdoni/vocdoni-node/pull/363 (specifically in [this line](https://github.com/vocdoni/vocdoni-node/commit/936aa8dd9182ebb9584d0b079cf1fa7201a0bf5f#diff-e28cfee6e67f3ae610cb00a944329f65d646e1b36c5a356a18b8c25df8cc8ca5R129))

PR #314, opened 2021-10-11
Feature/db batch max size

ed255:
- Generalize db ConcurrentWriteTx test
  - I have commented the test for pebble because it doesn't pass.  We implemented the WriteTx with pebble.Batch, and it doesn't heave like a Transaction.  I think we don't depend on the Transaction behavior for WriteTx, but we should have this behavior in mind! See `db/pebbledb/pebledb_test.go`:
```
// NOTE: This test fails.  pebble.Batch doesn't detect conflicts.  Moreover,
// reads from a pebble.Batch return the last version from the Database, even if
// the update was made after the pebble.Batch was created.  Basically it's not
// a Transaction, but a Batch of write operations.
```
- Add maxSize in db.Batch


db/internal/dbtest/dbtest.go
```diff
@@ -91,3 +92,64 @@ func TestIterate(t *testing.T, d db.Database) {
...
+
+	wgInc.Wait()
+	rTx := database.ReadTx()
+	defer rTx.Discard()
+	val, err := rTx.Get(key)
+	qt.Assert(t, err, qt.IsNil)
+	qt.Assert(t, val, qt.DeepEquals, []byte{1})
+
+	if errA == nil {
+		qt.Assert(t, errA, qt.IsNil)
```
mvdan, line 149: nit: isn't this line redundant?

db/batch.go
```diff
@@ -4,22 +4,39 @@ import (
...
+	count   uint
 }
 
 // check that Batch implements the ReadTx & WriteTx interfaces
 var _ ReadTx = (*Batch)(nil)
 var _ WriteTx = (*Batch)(nil)
 
-// NewBatch creates a new Batch from a Database
+// DefaultBatchMaxSize is the default value for maxSize used in NewBatch.
+const DefaultBatchMaxSize uint = 1024 * 1024
```
mvdan, line 26: It would be nice to document how we arrived at this value. Presumably it's a nice tradeoff between performance and peak latency or memory usage.

mvdan (approved):

arnaucube (approved):

db/internal/dbtest/dbtest.go
ed255: Yes, I left it to make it explicit that this case is a dual of the other case.  Either A is nil and B is err, or A is err and B is nil.

db/batch.go
ed255: This value is the maximum number of writes, not the actual memory size that the Tx will hold (because that will depend on the key-value sizes).  I think keeping a Tx at around 100 MiB is a good start.  If we assume keys and values of 64 bytes (which I think is not too far from most of our key-values), we get 128 MiB with this value.  I'll add a comment about it.

PR #316, opened 2021-10-13
fix dualstack support of ExternalIP()

altergui:
the current implementation of https://github.com/vocdoni/go-external-ip fork does not work as expected
(sometimes returns ipv4 when asked for ipv6, for example)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -20,6 +20,7 @@ require (
 	github.com/dgraph-io/badger/v3 v3.2103.1
 	github.com/ethereum/go-ethereum v1.10.8
 	github.com/frankban/quicktest v1.13.0
+	github.com/glendc/go-external-ip v0.1.0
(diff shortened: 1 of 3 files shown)


mvdan:
I'd like to wait a week to see if upstream can merge ipv6 support first. Having the replace directive in master isn't a good idea - anything using vocdoni-node as a library (such as the explorer) will need to copy the same replace directive, or otherwise break.

If upstream doesn't merge soon enough, then option B is to continue maintaining our hard fork in vocdoni, such as that PR's HEAD rebased on upstream's latest master. That still wouldn't require a replace directive, though - we can just `require` our fork directly.

altergui:
i think this PR alone is now kind of pointless, since `vocdoni/go-external-ip` will remain untouched.
i can just point the `require` to (fixed) `glendc/go-external-ip` in #313

mvdan:
Sounds good!

PR #319, opened 2021-10-14
Add db.WriteTx.Apply method

arnaucube:
Apply applies the value-passed WriteTx into the given WriteTx,
copying the key-values from the original WriteTx into the one from
which the method is called.
For PebbleDB case, it uses the original Pebble `Apply` method. For the
BadgerDB case, as Badger does not offer this functionallity, our wrapper
uses reflect to access to the keys&values of the tx.

It is needed for arbo, in order to be able to work with a WriteTx from
different goroutines, where it would be done as: create a WriteTx for
each goroutine, use Apply to copy the content from the main WriteTx into
each goroutine WriteTx, proceed with the goroutines, and once completed,
use Apply to copy the content from each goroutine-WriteTx into the main
WriteTx.

~**Warning**:~
~PebbleDB has an `Apply` method, but BadgerDB hasn't. I've tried to do access to private values of the badgerdb Tx, in order to implement the `Apply` method for the badgerdb wrapper, running tests it seems to work, but does not seem a solid solution, and currently GoAnalyzer gives an error due `db/badgerdb/badger.go:84:23: call of reflect.ValueOf copies lock value: github.com/dgraph-io/badger/v3.Txn contains sync.Mutex`, as the `badger.Txn` internally has a mutex. I don't know if we could use a different approach or some different 'trick', @mvdan I would appreciate your opinion ^^~
**Update**: the BadgerDB tx 'Apply' has been done as stated in https://github.com/vocdoni/vocdoni-node/pull/319#issuecomment-949711467

Here can be found a use of the WriteTx.Apply in arbo when parallelizing while adding stuff in the tree (in the WriteTx): https://github.com/vocdoni/arbo/commit/8b2be1eb82849c438994e4ef732f9b41a473d534#diff-050830ac4334170e8d335e1b95519052a3461c5b57a4ca38a0531e35a8f3d388R307

### DIFF
--- a/db/badgerdb/badger.go
+++ b/db/badgerdb/badger.go
@@ -4,8 +4,10 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"reflect"
 	"runtime"
 	"runtime/debug"
+
 	"strings"
 
 	"github.com/dgraph-io/badger/v3"
@@ -73,6 +75,25 @@ func (tx WriteTx) Delete(k []byte) error {
 	} else {
 		return err
 	}
+}
+
+// Apply implements the db.WriteTx.Apply interface method
+func (tx WriteTx) Apply(other db.WriteTx) (err error) {
+	otherBadger := other.(WriteTx)
+
+	valueOfTx := reflect.ValueOf(otherBadger.tx)
+	pendingWrites := valueOfTx.Elem().FieldByName("pendingWrites")
+
+	for _, mKey := range pendingWrites.MapKeys() {
+		e := pendingWrites.MapIndex(mKey).Elem()
+		k := e.FieldByName("Key").Bytes()
+		v := e.FieldByName("Value").Bytes()
+
+		if err := tx.Set(k, v); err != nil {
+			return err
+		}
+	}
+	return nil
 }
 
 // Commit implements the db.WriteTx.Commit interface method
--- a/db/interface.go
+++ b/db/interface.go
@@ -67,6 +67,13 @@ type WriteTx interface {
 	Set(key []byte, value []byte) error
 	// Delete deletes a key and its value.
 	Delete(key []byte) error
+	// Apply applies the value-passed WriteTx into the given WriteTx,
+	// copying the key-values from the original WriteTx into the one from
+	// which the method is called.
+	// TODO Review once generics are ready: WriteTx is an interface, so
+	// Apply internally needs type assertions, revisit this once generics
+	// are ready.
+	Apply(WriteTx) error
 	// Commit commits the transaction into the db
 	Commit() error
 }
--- a/db/badgerdb/badger_test.go
+++ b/db/badgerdb/badger_test.go
@@ -30,6 +30,13 @@ func TestConcurrentWriteTx(t *testing.T) {
 	dbtest.TestConcurrentWriteTx(t, database)
 }
 
+func TestWriteTxApply(t *testing.T) {
+	database, err := New(db.Options{Path: t.TempDir()})
+	qt.Assert(t, err, qt.IsNil)
+
+	dbtest.TestWriteTxApply(t, database)
+}
+
 func TestBatch(t *testing.T) {
 	database, err := New(db.Options{Path: t.TempDir()})
 	qt.Assert(t, err, qt.IsNil)
--- a/db/batch.go
+++ b/db/batch.go
@@ -56,6 +56,11 @@ func (t *Batch) Discard() {
 	t.tx.Discard()
 }
 
+// Apply implements the WriteTx.Apply interface method
+func (t *Batch) Apply(other WriteTx) (err error) {
+	return t.tx.Apply(other)
+}
+
 // Set implements the WriteTx.Set interface method.  If during this
 // operation, the internal tx becomes too big, all the pending writes will be
 // committed and a new WriteTx will be created to continue with this and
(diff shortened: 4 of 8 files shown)


p4u:
what about this? @arnaucube @ed255 @mvdan

ed255:
> what about this? @arnaucube @ed255 @mvdan

I'm waiting for an update that makes the tests pass to review this PR.

arnaucube:
Currently the tests break because of a 'trick' on BadgerDB, as BadgerDB does not support merging two txs (and the 'trick' suffers from the BadgerDB Txs Mutex, and that's why goanalyzer fails).
I said to @mvdan that there was no hurry into reviewing this these last days, as I was going to be mostly afk this week. I'm back online now, I'll talk with @mvdan to see if we can find a way to support merges of txs in BadgerDB.
This PR will enable this branch of `arbo`: https://github.com/vocdoni/arbo/commit/8b2be1eb82849c438994e4ef732f9b41a473d534 (which adds in AddBatch the approach that does not load the full tree on memory and it works on disk, but still parallelize for each CPU (that's why is needed the 'merge of dbtxs')).

arnaucube:
Update: talked through chat with @mvdan and seems that we can keep the current approach.

Updated the code of badgerdb wrapper for `tx.Apply` to not have the conflict with the tx internal mutex.

@ed255 @mvdan the PR is now ready for review

db/badgerdb/badger.go
```diff
@@ -75,6 +77,25 @@ func (tx WriteTx) Delete(k []byte) error {
...
+func (tx WriteTx) Apply(other db.WriteTx) (err error) {
+	otherBadger := other.(WriteTx)
+
+	v := reflect.ValueOf(otherBadger.tx)
+	pendingWrites := reflect.Indirect(v).FieldByName("pendingWrites")
+
+	for _, mKey := range pendingWrites.MapKeys() {
+		mValue := pendingWrites.MapIndex(mKey)
+		k := mValue.Elem().FieldByName("Key").Bytes()
+		v := mValue.Elem().FieldByName("Value").Bytes()
```
mvdan: deduplicate the `Elem()` call via `mValue` perhaps

db/badgerdb/badger.go
```diff
@@ -75,6 +77,25 @@ func (tx WriteTx) Delete(k []byte) error {
 	}
 }
 
+// Apply implements the db.WriteTx.Apply interface method
+func (tx WriteTx) Apply(other db.WriteTx) (err error) {
+	otherBadger := other.(WriteTx)
+
+	v := reflect.ValueOf(otherBadger.tx)
+	pendingWrites := reflect.Indirect(v).FieldByName("pendingWrites")
```
mvdan: nit: you know `v` is a pointer, so just do `v.Elem().FieldByName(...)`.

db/interface.go
```diff
@@ -67,6 +67,10 @@ type WriteTx interface {
 	Set(key []byte, value []byte) error
 	// Delete deletes a key and its value.
 	Delete(key []byte) error
+	// Apply applies the value-passed WriteTx into the given WriteTx,
+	// copying the key-values from the original WriteTx into the one from
+	// which the method is called.
+	Apply(WriteTx) error
```
mvdan, line 76: Unfortunate that WriteTx here is an interface, so you need type assertions in every implementation.
Perhaps add a TODO to revisit this once generics lands in a few months.

mvdan (approved):

db/badgerdb/badger.go
```diff
@@ -75,6 +77,25 @@ func (tx WriteTx) Delete(k []byte) error {
...
+	for _, mKey := range pendingWrites.MapKeys() {
+		e := pendingWrites.MapIndex(mKey).Elem()
+		k := e.FieldByName("Key").Bytes()
+		v := e.FieldByName("Value").Bytes()
+
+		if err := tx.Set(k, v); err != nil {
+			return err
+		}
+	}
+	return nil
```
ed255, line 96: Shoudn't `other` be Discarded after this?

ed255 (approved):
I'm approving this PR to move forward but I think that currently we have a messy design for the Database implementation with Pebble because our current Database implementation has Transactions and Pebble doesn't have transactions.  So with Badger a db.WriteTx is a badger.Transaction but in Pebble a db.WriteTx is a Batch.

Moreover, on the reading side, Badger requires a ReadTx to read, but Pebble doesn't, and our Database interface uses ReadTx.

So we have two implementations that behave very differently.

And then we introduce the Apply, which in my opinion, makes sense for a Batch but not for a Transaction.
A Transaction is more than a list of writes; it also keeps a point in time where the Tx was started, and every pending write has associated with it the value found in the DB before the Tx was started.  A Batch doesn't have this.

I think we could implement a Batch type which just accumulates writes and deletes, and allows reading from an underlying Database / ReadTx, and then can be Committed by applying all the reads and writes to a new WriteTx.  For Pebble we would just wrap the Batch type, for Badger we would create a new type with all the necessary fields.  And a proper Batch is what Arbo should use in this case.

arnaucube:
Agree, and, for arbo we don't need the Tx features (the point in time), I don't know if is a requirement in the rest of vocdoni-node (it does not seem so).

Opened a issue to track & talk about this: https://github.com/vocdoni/vocdoni-node/issues/328

db/badgerdb/badger.go
arnaucube: The idea was that `other` can still be used after this method is called

PR #324, opened 2021-10-18
vochain: add balances support

p4u:
Allow transfering balances among addresses.
This commit introduces the feature only at State level.


vochain/balances.go
```diff
@@ -0,0 +1,123 @@
...
+	"errors"
+	"fmt"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/vocdoni/arbo"
+)
+
+// Balance represents an amount of tokens, usually attached to an address.
+// Balance includes a Nonce which needs to be incremented by 1 on each transfer.
+type Balance struct {
```
ed255: I suggest renaming this to Account.  I think it makes more sense because balance is the Amount, but we also have Nonce, so to me it seems more natural to have Account.Nonce and Account.Amount / Account.Balance.

vochain/balances.go
```diff
@@ -0,0 +1,123 @@
...
+	b.Nonce = binary.LittleEndian.Uint32(data[8:])
+	return nil
+}
+
+// Transfer moves amount from the origin Balance to the dest Balance.
+func (b *Balance) Transfer(dest *Balance, amount uint64, nonce uint32) error {
+	if dest == nil {
+		return fmt.Errorf("destination balance wallet nil")
+	}
+	if b.Nonce+1 != nonce {
```
ed255: If I understand it correctly, in Ethereum the Nonce of a given account is the value that will be accepted for a transaction.  So if the Account.Nonce = 0, then the tx.Nonce should be 0, and then Account.Nonce = 1.
So the first Tx of an account will have Nonce = 0.  See for example the first tx of a random account in rinkeby: https://rinkeby.etherscan.io/tx/0x79f0bdf0df93845ead993f9020f5d5da3ac1a907219a88002e549b5ff5b9efb8 (Nonce is 0).

I think that following the same pattern here will make it easier to understand for people that already know how Ethereum works.  So I suggest changing the implementation so that the tx.Nonce must match the account.Nonce. (instead of the account.Nonce+1)

vochain/balances.go
```diff
@@ -0,0 +1,123 @@
...
+		return ErrBalanceNonceInvalid
+	}
+	if b.Amount < amount {
+		return ErrNotEnoughBalance
+	}
+	dest.Amount += amount
```
ed255: I think adding an overflow check here would be wise to avoid surprises in the future.

vochain/balances.go
```diff
@@ -0,0 +1,123 @@
...
+	b.Amount -= amount
+	b.Nonce++
```
ed255: I think adding an overflow check here would be wise to avoid surprises in the future.

vochain/balances.go
```diff
@@ -0,0 +1,123 @@
...
+	defer v.Tx.Unlock()
+	raw, err := v.Tx.DeepGet(address.Bytes(), BalancesCfg)
+	if err != nil && !errors.Is(err, arbo.ErrKeyNotFound) {
+		return err
+	} else if err == nil {
+		if err := balance.Unmarshal(raw); err != nil {
+			return err
+		}
+	}
+	balance.Amount += amount
```
ed255: I think adding an overflow check here would be wise to avoid surprises in the future.

vochain/state.go
```diff
@@ -319,7 +332,7 @@ func (v *State) setMainTreeView(treeView *statedb.TreeView) {
 
 // mainTreeViewer returns the mainTree as a treeViewer.  When isQuery is false,
 // the mainTree returned is the not yet commited one from the currently open
-// StateDB transaction.  When isQuery is false, the mainTree returned is the
+// StateDB transaction.  When isQuery is true, the mainTree returned is the
```
ed255, line 335: :flushed:

vochain/balances.go
jordipainan:
```suggestion
	if dest.Amount+amount < dest.Amount {
		return ErrBalanceOverflow
	}
	dest.Amount += amount
	if b.Amount < amount {
		return ErrBalanceOverflow
	}
	b.Amount -= amount
	if b.Nonce+1 <= b.Nonce {
		return ErrBalanceNonceInvalid
	}
```

vochain/balances.go
jordipainan:
```suggestion
	if balance.Amount+amount <= balance.Amount {
		return ErrBalanceOverflow
	}
	balance.Amount += amount
```

vochain/types.go
```diff
@@ -20,8 +20,10 @@ const (
 )
 
 var (
-	ErrVoteDoesNotExist = fmt.Errorf("vote does not exist")
-	ErrProcessNotFound  = fmt.Errorf("process not found")
+	ErrVoteDoesNotExist    = fmt.Errorf("vote does not exist")
+	ErrNotEnoughBalance    = fmt.Errorf("not enough balance to transfer")
+	ErrBalanceNonceInvalid = fmt.Errorf("invalid balance nonce")
+	ErrProcessNotFound     = fmt.Errorf("process not found")
 	// keys; not constants because of []byte
```
jordipainan, line 26:
```suggestion
	ErrBalanceOverflow = fmt.Errorf("balance overflow")
	ErrBalanceAmountZero = fmt.Errorf("zero amount not valid")
	// keys; not constants because of []byte
```

vochain/balances.go
ed255: nit: This check
```go
	if dest.Amount+amount <= dest.Amount {
		return ErrBalanceOverflow
	}
```
will give `ErrBalanceOverflow` when amount is 0.  Maybe do `if dest.Amount+amount < dest.Amount`.

For the `b.Amount` on a second thought, I think that by checking that there are enough funds already guarantees no overflow, so there's no need for the second overflow check.  So this should be enough (which appears already in the PR):
```go
	if b.Amount < amount {
		return ErrNotEnoughBalance
	}
```

vochain/balances.go
```diff
@@ -0,0 +1,123 @@
...
+		return fmt.Errorf("wrong balance data lenght: %d", len(data))
+	}
+	b.Amount = binary.LittleEndian.Uint64(data[:8])
+	b.Nonce = binary.LittleEndian.Uint32(data[8:])
+	return nil
+}
+
+// Transfer moves amount from the origin Balance to the dest Balance.
+func (b *Balance) Transfer(dest *Balance, amount uint64, nonce uint32) error {
+	if dest == nil {
```
jordipainan:
```suggestion
      if amount == 0 {
              return ErrBalanceAmountZero
      }
	if dest == nil {
```

vochain/balances.go
```diff
@@ -0,0 +1,123 @@
...
+		if err := v.Tx.DeepSet(to.Bytes(), balanceTo.Marshal(), BalancesCfg); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// MintBalance increments the existing balance of address by amount
+func (v *State) MintBalance(address common.Address, amount uint64) error {
+	var balance Balance
```
jordipainan:
```suggestion
    if amount == 0 {
        return ErrBalanceAmountZero
    }
	var balance Balance
```

vochain/balances.go
jordipainan: true that! thank u @ed255 , changed recheck if you want :)

vochain/balances.go
jordipainan: Prefer account too

jordipainan (approved):

ed255 (approved):

PR #325, opened 2021-10-19
Implement rolling census

ed255:
- census:
    - Add support for importing a Poseidon local census (from the vochain) via
      the censusdownloader.
    - Add a reverse mapping for Poseidon local census to resolve Public Key to
      Index, as the Poseidon census stores Public Keys in leaves by a
      sequential index in order to create a balanced tree thus reducing the
      number of levels.
    - In handler, for Poseidon census resolve the Public Key to Index before
      getting a proof.
- crypto/zk/artifacts:
    - Add `Parameters []int64` to set the parameters used for the circuit
      build, so that later we can figure out which circuit to use for a given
      process (based on the census size).
- service:
    - Move the code to download the ZK Verification Keys to BaseApplication
- vochain:
    - In BaseApplication, add LoadZkVKs function (moved from service)
    - In NewVochain, call LoadZkVKs.
    - In State, implement AddToRollingCensus, which adds a key to the rolling
      census of a process using a sequential index as tree key, and using the
      given key as value.
    - In State, add DumpRollingCensus so that the census manager can import a
      rolling census once the process starts (and the rolling census is
      closed).
    - In State, add OnProcessStart event, which is fired when there are
      processes that start at the current block.  This requires storing an
      index of ProcessIDs by StartBlock in the mainTree.NoState.
    - In State, in AddProcess, initialize a rolling census if the process
      requires it.
- statedb:
    - Replace BatchedTx by raw WriteTx (see comment in code)
    - Add Dump function to all variants of tree
 
---
## Anonymous Voting Status
- This PR doesn't add any test for a process with anonymous voting.  In order to do this we will need to generate SNARK proofs dynamically.  The simplest way to do it will be with the docker-compose test, and adding some auxiliar js code that generates the proofs using snarkjs.
- I haven't yet added any new API endpoint that the client can query in order to figure out the circuit to use for a given process with rolling census.  I would like to discuss possible options for this.


ed255:
This PR depends on https://github.com/vocdoni/dvote-protobuf/pull/30

census/auth.go
```diff
@@ -47,6 +48,9 @@ func (m *Manager) CheckAuth(reqOuter *api.RequestMessage, reqInner *api.MetaRequ
 
 	// Check signature with existing namespace keys
 	log.Debugf("namespace keys %s", ns.Keys)
+	// NOTE(Edu): If I understand this correctly, if a Census doesn't have
+	// any Key, it means that anyone can do authorized operations on this
+	// census?  What are those operations?  This feels dangerous to me!
```
p4u: The census that can be modified are identified by "<address_owner>/<arbitrary_id_provided_by_owner>". There is a check for the censusID (comming from the router), so when modifying a census only the owner can do it (address is extracted from signature).

go.mod
```diff
@@ -3,7 +3,7 @@ module go.vocdoni.io/dvote
 go 1.17
 
 // For testing purposes while dvote-protobuf becomes stable
-// replace go.vocdoni.io/proto => ../dvote-protobuf
+replace go.vocdoni.io/proto => ../../dvote-protobuf
```
p4u: this should not be here

ed255: yeah! I added a new type in the protobuf definitions.  Once this https://github.com/vocdoni/dvote-protobuf/pull/30 is merged I'll amend this.

census/auth.go
ed255: Thanks for the clarification.  After I wrote this note I realized that the `CheckAuth` is not called anywhere, so I just commented the whole file.  @p4u should I remove this file?

p4u: yes

go.mod
p4u: done

test/rollingcensus_test.go
```diff
@@ -0,0 +1,91 @@
...
+
+	_, err := app.State.Save()
+	qt.Assert(t, err, qt.IsNil)
+
+	// Block 2
+	app.State.Rollback()
+	app.State.SetHeight(2)
+
+	const numKeys = 128
+	keys := make([][]byte, numKeys)
```
p4u, line 57: Can we abstract this poseidon compatible key generation and put it somewhere (such as utils/) ? I feel its an easy mistake to do in the future (assume Poseidon works with the whole field).

test/testcommon/testutil/util.go
```diff
@@ -29,3 +30,22 @@ func B642byte(tb testing.TB, s string) []byte {
 	}
 	return b
 }
+
+type Random struct {
```
p4u, line 37: why not using this instead? https://github.com/vocdoni/vocdoni-node/blob/master/util/helpers.go#L18

vochain/app.go
```diff
@@ -312,19 +343,6 @@ func (app *BaseApplication) InitChain(req abcitypes.RequestInitChain) abcitypes.
 		}
 	}
 
-	var header models.TendermintHeader
```
p4u, line 315: I'm not against it, but could you explain why this is being removed?

vochain/censusdownloader/censusdownloader.go
```diff
@@ -48,7 +48,9 @@ func (c *CensusDownloader) importCensus(root, uri string) {
 func (c *CensusDownloader) Rollback() {
 	c.queueLock.Lock()
 	c.queue = make(map[string]string)
-	c.isFastSync = c.vochain.Node.ConsensusReactor().WaitSync()
+	if c.vochain.Node != nil {
+		c.isFastSync = c.vochain.Node.ConsensusReactor().WaitSync()
```
p4u: use `c.vochain.IsSynchronizing()` intead. In general we should avoid calling c.vochain.Node from anywhere out of the pure vochain package context

test/testcommon/testutil/util.go
ed255: Ah, because the `RandomBytes` function doesn't return a deterministic result, and I think unit tests should use deterministic inputs because you have better guarantees that if the test passes once, it will pass in later runs.  Also having deterministic inputs makes it easier to debug and follow whats going on, because each run gives the exact same values.

I think the function `RandomBytes` should never be used in a unit test.

vochain/app.go
ed255: I removed this because it was a reminiscence of the old StateDB which stored the tendermint block header.  We no longer rely on keeping the header in the StateDB, so I just deleted it, as it was never being queried.

p4u: cool

test/testcommon/testutil/util.go
p4u: @mvdan?

vochain/process.go
```diff
@@ -13,16 +13,56 @@ import (
...
+// case of Tx_NewProcess.  In that same function, before calling AddProcess,
+// NewProcessTxCheck is called, which returns error if the process exists, so
+// for a Tx_NewProcess the processID is expected to not exist.  If there are no
+// cases where AddProcess is supposed to overide an existing process, I think
+// we should update AddProcess and call DeepAdd instead of DeepSet (which will
+// return an error if the process already exists), and also allows us to make
+// other assumptions (like certain fields of a process won't be modified
+// later).
+// NOTE(Edu):  I have updated this funtion assuming that only new processes are
+// added and never modified via this funciton.
```
p4u: That's correct, AddProcess() should be only using for adding non-existing processes. Once a process is part of the state, the way to modify it is by transactions such as `SetProcessStatus`. Can you remove your comment and instead speficy that adding a process that already exist will return an error?

test/testcommon/testutil/util.go
mvdan: Yeah, I agree with Edu - tests should be deterministic where possible. Otherwise you can have a case where `go test` fails on CI, but if you try to reproduce locally you can't, because you got a different random seed.

The util package also uses crypto/rand, which is a bit overkill for the tests :) math/rand is plenty there.

vochain/process.go
```diff
@@ -13,16 +13,56 @@ import (
...
+		// store the mapping between index and key, use the NoState
+		// database in the censusTree.
+		if preRegister && anonymous {
+			census, err := v.Tx.DeepSubTree(ProcessesCfg, CensusPoseidonCfg.WithKey(p.ProcessId))
+			if err != nil {
+				return err
+			}
+			// We store census size as little endian 64 bits.  Set it to 0.
+			censusLenLE := [8]byte{}
+			if err := census.NoState().Set(keyCensusLen, censusLenLE[:]); err != nil {
```
p4u: this seems a possible future origin of problems (a developer not taking into account the little endianess. I think it would be better if we can make an abstraction of this... Maybe something like `census.NoState().SetUint64(keyCensusLen, 0)`

vochain/process.go
```diff
@@ -363,10 +403,13 @@ func NewProcessTxCheck(vtx *models.Tx, txBytes,
...
 
 	// check valid/implemented process types
-	switch {
-	case tx.Process.EnvelopeType.Anonymous:
-		return nil, fmt.Errorf("anonymous process not yet implemented")
-	case tx.Process.EnvelopeType.Serial:
+	// pre-regiser and anonymous must be either both enabled or disabled, as
+	// we only support a single scenario of pre-register + anonymous.
+	if tx.Process.Mode.PreRegister != tx.Process.Mode.PreRegister {
+		return nil, fmt.Errorf("pre-register mode only supported " +
```
p4u, line 400: Actually pre-register should also be supported by non anonymous envelope processes too. I can think of use cases where entities wants to perform a non-anonymous vote with pre-register.

If this means a lot of changing on your PR, you could let some TODO notes in the code. So we can takle this in the future.

vochain/start.go
```diff
@@ -39,6 +40,12 @@ func NewVochain(vochaincfg *config.VochainCfg, genesis []byte) *BaseApplication
 	if err := app.Node.Start(); err != nil {
 		log.Fatal(err)
 	}
+	// get the zk Circuits VerificationKey files
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute*1)
```
p4u: how big are these files? Is a minute enough for a low-band connection? Also I think we should put this into a constant, easy to find and modify.

vochain/state.go
```diff
@@ -53,6 +54,9 @@ func processGetCensusRoot(value []byte) ([]byte, error) {
 	if err := proto.Unmarshal(value, &sdbProc); err != nil {
 		return nil, fmt.Errorf("cannot unmarshal StateDBProcess: %w", err)
 	}
+	if len(sdbProc.Process.CensusRoot) != 32 {
```
p4u: this lenght should be equal to the hashing used by Arbo (instead of a magic number).

vochain/process.go
ed255: I'll add a TODO!  For this PR I only focused on the case with PreRegister and Anonymous; I haven't done a thorough analysis of what is required for PreRegister without Anonymous, and I prefer for now disabling that option to allow having a more controlled setting where we can make more assumptions and avoid untested configurations.

vochain/start.go
ed255: We are only downloading the verification keys here, which are pretty small.  The verification key is constant size regardless of the circuit size.  Here's an instance of the one we are downloading https://github.com/vocdoni/zk-circuits-artifacts/blob/master/zkcensusproof/dev/1024/verification_key.json which is 3.57 KiB
I'll move the value to a constant.

test/testcommon/testutil/util.go
p4u: Fine. So then we might file an issue to switch to the testutil/random method for all tests in the repository

mvdan: It would also be possible to make the util random funcs that use crypto/rand error if called directly from a `*_test.go` func, via https://pkg.go.dev/runtime#Callers.

p4u:
Is this ready for merging @arnaucube @mvdan ?

p4u:
I have rebased the branch with master and I'll merge it now since this commit touches many files and its easy to create conflict with other commits.

PR #329, opened 2021-10-25
Feature/processrollingapi

### LINKED ISSUE #327: feature: Report circuit artifacts to use for a started process with rolling census
Anonymous voting uses a process with a rolling census that is closed once the process starts.  We can have different circuit builds for different census sizes.  The client needs to figure out which circuit artifacts to use for a given process, and we can give that information via some API endpoint.
What the node needs to do is: get the census size, and pick the smaller circuit artifacts that supports that size from the `VochainGenesis.CircuitsConfig` in the vochain genesis.  That information should be supplied to the user via an API.

I've been looking at different endpoints and I think the best place to put this information is in the `getProcessInfo` endpoint in the scrutinizer.  This would would involve:
1. In `Scrutinizer.updateProcess` if the process is rolling, store the current census size (This one is optional, but it would allow querying the current rolling census size.  This would require that the Scrutinizer updates the census size on every Commit.  It could be done via a new event `OnGrowRollingCensus`, or by the Scrutinizer querying all the processes with rolling census with StartBlock < currentBlock and updating the census sizes from there).
2. In `Scrutinizer.OnProcessesStart` if the process is anonymous, figure out the circuit config and store the index in the process.
3. Extend `indexertypes.Process` with rolling census size, and circuit config index
4. In `Scrutinizer.ProcessInfo`, if the process is anonymous and started, query the Genesis.CircuitConfigs for the stored index, and return the data.

ed255:
See commit messages
```
Add MaxRollingCensusSize and RollingCensusSize to Scrutinizer

Add getCircuitConfig vote-api endpoint

    - Verify validity of models.Process.MaxRollingCensusSize value when
      adding a process
    - For each RegisterKey for a Rolling census, verify that we never
      superpass the process.MaxRollingCensusSize
    - When a process with rolling census starts, set the value of the census
      to process.RollingCensusSize
    - Add getCircuitConfig endpoint in vote-api which dynamically figures
      out the proper CircuitConfig to be used given the
      process.RollingCensusSize

State: Move OnProcessStart events to Save

    This removes the requirement for the event listeners to handle a
    possible Rollback of an OnProcessStart.

```

Depends on https://github.com/vocdoni/dvote-protobuf/pull/31
Resolve https://github.com/vocdoni/vocdoni-node/issues/327

### DIFF
--- a/vochain/census.go
+++ b/vochain/census.go
@@ -29,6 +29,10 @@ var keyCensusLen = []byte("len")
 func (v *State) AddToRollingCensus(pid []byte, key []byte, weight *big.Int) error {
 	v.Tx.Lock()
 	defer v.Tx.Unlock()
+	process, err := getProcess(v.mainTreeViewer(false), pid)
+	if err != nil {
+		return fmt.Errorf("cannot open process with pid %x: %w", pid, err)
+	}
 	census, err := v.Tx.DeepSubTree(ProcessesCfg, CensusPoseidonCfg.WithKey(pid))
 	if err != nil {
 		return fmt.Errorf("cannot open rolling census with pid %x: %w", pid, err)
@@ -39,6 +43,9 @@ func (v *State) AddToRollingCensus(pid []byte, key []byte, weight *big.Int) erro
 	censusLen, err := statedb.GetUint64(noState, keyCensusLen)
 	if err != nil {
 		return fmt.Errorf("cannot get ceneusLen: %w", err)
+	}
+	if censusLen >= *process.MaxRollingCensusSize {
+		return fmt.Errorf("maxRollingCensusSize already reached")
 	}
 	// Add key to census
 	index := [8]byte{}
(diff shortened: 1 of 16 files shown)


vochain/census.go
```diff
@@ -57,6 +64,27 @@ func (v *State) AddToRollingCensus(pid []byte, key []byte, weight *big.Int) erro
...
+	census, err := mainTreeView.DeepSubTree(ProcessesCfg, CensusPoseidonCfg.WithKey(pid))
+	if err != nil {
+		return 0, fmt.Errorf("cannot open rolling census with pid %x: %w", pid, err)
+	}
+	noState := census.NoState()
+	censusLen, err := statedb.GetUint64(noState, keyCensusLen)
+	if err != nil {
+		return 0, fmt.Errorf("cannot get ceneusLen: %w", err)
+	}
+	return censusLen, nil
```
p4u, line 77: This can be simplified:
`return censusLen, statedb.GetUint64(census.NoState(), keyCensusLen)`

vochain/census.go
```diff
@@ -57,6 +64,27 @@ func (v *State) AddToRollingCensus(pid []byte, key []byte, weight *big.Int) erro
...
+}
+
+func (v *State) GetRollingCensusSize(pid []byte, isQuery bool) (uint64, error) {
+	if !isQuery {
```
p4u, line 81: Is this lock required? Shouldn't `v.mainTreeViewer(isQuery)` be safe already?

p4u (approved):

vochain/census.go
ed255: The reason why I didn't simplify it was to wrap the error to add a meaningful message, because otherwise the message can be something like `key not found` which makes it hard to debug.

vochain/census.go
ed255: The function `mainTreeViewer` when `isQuery = false` returns the mainTree from the `v.Tx`, which is not thread safe because the `v.Tx` can be Discarded or Commited at any time.
The reason why the function `mainTreeViewer` doesn't lock is because the lock must persist after the function returns, because we are locking the access to `v.Tx`, which is returned when `isQuery = false`.
We could change the `mainTreeViewer` function to accept a closure that takes the TreeViewer as an argument and is executed inside a lock.  Let me know what you think.

arnaucube (approved):

PR #332, opened 2021-10-27
add a new http router

### LINKED ISSUE #334: Rest API compatible Router

p4u:
This is an http router which reuses some code of the multirpc but
following a simpler approach. Supports not only JSON-RPC handlers but
any which implements the RouterNamespace inteface. So it will allow
creating the new API based on URL requests.

The existing router and multirpc should be replaced with this one.

### DIFF
--- a/httprouter/jsonrpcapi/encoding.go
+++ b/httprouter/jsonrpcapi/encoding.go
@@ -1,0 +1,42 @@
+package jsonrpcapi
+
+import (
+	"encoding/hex"
+	"fmt"
+)
+
+// HexBytes is a []byte which encodes as hexadecimal in json, as opposed to the
+// base64 default.
+type HexBytes []byte
+
+func (b HexBytes) MarshalJSON() ([]byte, error) {
+	enc := make([]byte, hex.EncodedLen(len(b))+2)
+	enc[0] = '"'
+	hex.Encode(enc[1:], b)
+	enc[len(enc)-1] = '"'
+	return enc, nil
+}
+
+func (b *HexBytes) UnmarshalJSON(data []byte) error {
+	if len(data) == 0 {
+		return nil
+	}
+	if len(data) < 2 || data[0] != '"' || data[len(data)-1] != '"' {
+		return fmt.Errorf("invalid JSON string: %q", data)
+	}
+	data = data[1 : len(data)-1]
+
+	// Strip a leading "0x" prefix, for backwards compatibility.
+	if len(data) >= 2 && data[0] == '0' && (data[1] == 'x' || data[1] == 'X') {
+		data = data[2:]
+	}
+
+	decLen := hex.DecodedLen(len(data))
+	if cap(*b) < decLen {
+		*b = make([]byte, decLen)
+	}
+	if _, err := hex.Decode(*b, data); err != nil {
+		return err
+	}
+	return nil
+}
--- a/httprouter/jsonrpcapi/jsonrpcapi.go
+++ b/httprouter/jsonrpcapi/jsonrpcapi.go
@@ -1,0 +1,186 @@
+package jsonrpcapi
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+
+	ethcommon "github.com/ethereum/go-ethereum/common"
+	"go.vocdoni.io/dvote/crypto/ethereum"
+	"go.vocdoni.io/dvote/httprouter"
+	"go.vocdoni.io/dvote/log"
+)
+
+// SignedJRPC is a httprouter handler to easily build a JSON rpc endpoint.
+//
+// The basic JSON structure for the RPC is predefined and follows the format:
+//
+// {
+//   "request": {
+//     "id":"randomID",
+//     "method":"methodName",
+//     "timestamp":"currentTS",
+//     <customFields...>
+//     },
+//   "signature":"0x...",
+//   "id":"sameID"}
+// }
+//
+// The handler manages automatically signatures, ids and timestamps.
+// The developer using SignedJRPC only needs to take care of the information within the request field (method and customFields).
+// The handler is prepared to accept several methods within the same URL path, such as http://localhost/jsonapi, so there
+// is no need to declare multiple URL paths (but its also allowed).
+type SignedJRPC struct {
+	signer    *ethereum.SignKeys
+	reqType   func() MessageAPI
+	methods   map[string]*registeredMethod
+	adminAddr *ethcommon.Address
+}
+
+// SignedJRPCdata is the data type used by SignedJRPC
+type SignedJRPCdata struct {
+	Method             string
+	Address            ethcommon.Address
+	SignaturePublicKey []byte
+	Private            bool
+	Message            MessageAPI
+}
+
+type registeredMethod struct {
+	public        bool
+	skipSignature bool
+}
+
+// NewSignedJRPC returns an initialized instance of the SignedJRPC type.
+// A signer is required in order to create signatures and authenticate clients.
+// A function returning an empty type implementing the custom MessageAPI should also be provided.
+func NewSignedJRPC(signer *ethereum.SignKeys, reqType func() MessageAPI) *SignedJRPC {
+	return &SignedJRPC{
+		methods: make(map[string]*registeredMethod, 128),
+		signer:  signer,
+		reqType: reqType,
+	}
+}
+
+// AuthorizeRequest implements the httprouter interface
+func (s *SignedJRPC) AuthorizeRequest(data interface{}, isAdmin bool) bool {
+	request, ok := data.(*SignedJRPCdata)
+	if !ok {
+		log.Warnf("type is not SignedJRPCdata")
+		return false
+	}
+	// If admin method, compare with the admin address
+	if isAdmin {
+		if s.adminAddr == nil {
+			log.Warnf("admin address not defined")
+			return false
+		}
+		return *s.adminAddr == request.Address
+	}
+	// If private method, check authentication
+	if request.Private {
+		return s.signer.Authorized[request.Address]
+	}
+	// If the method is registered as public, return true
+	return true
+}
+
+// ProcessData implements the httprouter interface
+func (s *SignedJRPC) ProcessData(req *http.Request) (interface{}, error) {
+	reqBody, err := io.ReadAll(req.Body)
+	if err != nil {
+		return nil, fmt.Errorf("HTTP connection closed: (%v)", err)
+	}
+	return s.getRequest(reqBody)
+}
+
+// SendError builds a response message error and sends it using the provided http context.
+func (s *SignedJRPC) SendError(requestID string, errorMsg string, ctxt *httprouter.HTTPContext) error {
+	if ctxt == nil {
+		return fmt.Errorf("http context is nil")
+	}
+	msg := s.reqType()
+	msg.SetError(errorMsg)
+	data, err := BuildReply(s.signer, msg, requestID)
+	if err != nil {
+		return err
+	}
+	return ctxt.Send(data)
+}
+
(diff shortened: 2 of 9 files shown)


httpRouter/httprouter.go
```diff
@@ -0,0 +1,249 @@
+package httprouter
```
mvdan: if the package name is httprouter, the directory name should also be httprouter - without camelCase

httpRouter/bearerstdapi.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+	return &BearerStandardAPIdata{
+		Data:      respBody,
+		AuthToken: strings.ReplaceAll(req.Header.Get("Authorization"), bearerPrefix, ""),
+	}, nil
+}
+
+// AddAdminToken sets the bearer admin token capable to execute admin handlers
+func (b *BearerStandardAPI) AddAdminToken(bearerToken string) {
+	b.adminToken = bearerToken
```
mvdan: if authTokens is a sync.Map, shouldn't adminToken be an atomic.Value too? otherwise it could be racy if the user calls AddAdminToken too late.

Alternatively, clearly document that this must be called before any requests come in.

httpRouter/bearerstdapi.go
```diff
@@ -0,0 +1,86 @@
...
+// ProcessData is a function for the RouterNamespace interface.
+// The body of the http requests and the bearer auth token are readed.
+func (b *BearerStandardAPI) ProcessData(req *http.Request) (interface{}, error) {
+	respBody, err := io.ReadAll(req.Body)
+	if err != nil {
+		return nil, fmt.Errorf("HTTP connection closed: (%v)", err)
+	}
+	return &BearerStandardAPIdata{
+		Data:      respBody,
+		AuthToken: strings.ReplaceAll(req.Header.Get("Authorization"), bearerPrefix, ""),
```
mvdan: if bearerPrefix is indeed a prefix, ReplaceAll doesn't seem right - you want TrimPrefix

test/testcommon/testutil/util.go
```diff
@@ -53,6 +53,10 @@ func (r *Random) RandomBytes(n int) []byte {
 	return b
 }
 
+func (r *Random) RandomInt(n int) int {
+	return r.rand.Int() % n
+}
```
mvdan, line 58: Use rand's Intn, and probably call the method RandomIntn for consistency

httpRouter/httprouter.go
```diff
@@ -0,0 +1,249 @@
...
+
+	if len(r.TLSdomain) > 0 {
+		log.Infof("fetching letsencrypt TLS certificate for %s", r.TLSdomain)
+		s, m := r.generateTLScert(host, port)
+		s.ReadTimeout = 20 * time.Second
+		s.WriteTimeout = 15 * time.Second
+		s.IdleTimeout = 10 * time.Second
+		s.ReadHeaderTimeout = 5 * time.Second
+		s.Handler = r.Mux
+		if err := http2.ConfigureServer(s, nil); err != nil {
```
mvdan: why do we need these http2 calls? as far as I know, net/http already transparently supports http2.

httpRouter/bearerstdapi.go
```diff
@@ -0,0 +1,86 @@
...
+// NewBearerStandardAPI returns a BearerStandardAPI initialized type
+func NewBearerStandardAPI() *BearerStandardAPI {
+	return &BearerStandardAPI{}
+}
+
+// CheckAuth is a function for the RouterNamespace interface.
+// On private handlers checks if the supplied bearer token have still request credits
+func (b *BearerStandardAPI) CheckAuth(data interface{}, isAdmin bool) bool {
+
+	msg, ok := data.(*BearerStandardAPIdata)
```
mvdan: Why does the interface split ProcessData and CheckAuth? One is called right after the other, and they need to share data closely, as you can see by this type assertion here.

I think it would be much easier to have a single method on the interface that does everything; it could be like:

```
type RouterAuthorizer interface {
     AuthorizeRequest(req *http.Request, isAdmin bool) bool
}
```

I also think this name is slightly more intuitive :)

mvdan (approved):
> The existing router and multirpc should be replaced with this one.

Will you do this refactor shortly after? We're accumulating deprecated routing libraries at a fast pace :)

httpRouter/bearerstdapi.go
p4u: Yes it should. I just thought the admin token was not gonna be changed (just set at the beggining) but you are right, for consistency it should be also atomic.

httpRouter/bearerstdapi.go
p4u: ProcessData() is the interface function for digesting the data. The digested data is added to Message.Data (which is of type `interface{}` to allow any kind of type). This specific implementation just uses []byte as Message.Data but JSON or Protobuf structs could be used instead (and we will use it for sure). 

While ProcessData() needs to be called always (for each namespace request) CheckAuth() is only called when the method is declared private or admin.

So I think having two different methods makes sense.

httpRouter/httprouter.go
p4u: I don't remember, it is a legacy from the multirpc proxy package. There could be some reason related to the x509 certificates that we use here https://github.com/vocdoni/blind-ca

As I'm not sure I let the code as it was...

p4u:
> > The existing router and multirpc should be replaced with this one.
> 
> Will you do this refactor shortly after? We're accumulating deprecated routing libraries at a fast pace :)

Agree. I'll do it ASAP.

httpRouter/bearerstdapi.go
mvdan: I see, that makes sense. Perhaps clarify that the second method only gets called for private or admin endpoints.

p4u:
@mvdan can you review again? there is a new pacckage "jsonrpcapi"

jordipainan (approved):

httprouter/jsonrpcapi/jsonrpcapi.go
```diff
@@ -0,0 +1,186 @@
...
+//
+// {
+//   "request": {
+//     "id":"randomID",
+//     "method":"methodName",
+//     "timestamp":"currentTS",
+//     <customFields...>
+//     },
+//   "signature":"0x...",
+//   "id":"sameID"}
```
mvdan, line 27: that `}` seems like a typo

httprouter/jsonrpcapi/jsonrpcapi.go
```diff
@@ -0,0 +1,186 @@
...
+		signer:  signer,
+		reqType: reqType,
+	}
+}
+
+// AuthorizeRequest implements the httprouter interface
+func (s *SignedJRPC) AuthorizeRequest(data interface{}, isAdmin bool) bool {
+	request, ok := data.(*SignedJRPCdata)
+	if !ok {
+		log.Warnf("type is not SignedJRPCdata")
```
mvdan, line 70: IMO this kind of thing should be a panic, as it should never happen

httprouter/jsonrpcapi/jsonrpcapi.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+
+	// The method must be registered
+	method, ok := s.methods[request.Method]
+	if !ok {
+		return request, fmt.Errorf("method not valid: (%s)", request.Method)
+	}
+
+	// Extract signature and address
+	if !method.skipSignature {
```
mvdan, line 163: if you invert and early return, you can unindent the rest

httprouter/jsonrpcapi/encoding.go
```diff
@@ -0,0 +1,42 @@
+package jsonrpcapi
+
+import (
+	"encoding/hex"
+	"fmt"
+)
+
+// HexBytes is a []byte which encodes as hexadecimal in json, as opposed to the
+// base64 default.
+type HexBytes []byte
```
mvdan, line 10: do we not already have this somewhere?

httprouter/jsonrpcapi/message.go
```diff
@@ -0,0 +1,62 @@
...
+	GetID() string
+	SetID(string)
+	SetTimestamp(int32)
+	SetError(string)
+	GetMethod() string
+}
+
+// RequestMessage is the envelope for the SignedJRPC message request
+type RequestMessage struct {
+	MessageAPI json.RawMessage `json:"request"`
```
mvdan, line 23: IMO these should be named after the JSON fields, e.g. Request or Response.

httprouter/jsonrpcapi/message.go
```diff
@@ -0,0 +1,62 @@
...
+	Signature HexBytes `json:"signature"`
+}
+
+// BuildReply builds a response message (ID, Timestamp and Signature) and marshals it using JSON.
+// The output message is ready to send as reply to the client using reqMsg.Context.Send(data)
+func BuildReply(signer *ethereum.SignKeys, msg MessageAPI, requestID string) ([]byte, error) {
+	var err error
+	respRequest := &ResponseMessage{ID: requestID}
+	msg.SetID(requestID)
+	msg.SetTimestamp(int32(time.Now().Unix()))
```
mvdan, line 43: forcing int32 for timestamps will make our software break by 2038, which is not very far :) why not just use int64, as that's what time.Now().Unix() already returns?

mvdan (commented):
The code looks fine, but doesn't this implement exactly what we have already, anyway? I see the same request/response structure with the embedded signature, and the same Send/SendError methods as before. I guess you're using the new router interface, but the overall API seems to work exactly the same way as before for the user.

httprouter/jsonrpcapi/message.go
p4u: This was done this way to avoid precision problems on the javascript API side.

httprouter/jsonrpcapi/encoding.go
p4u: Yes, but import loops are annoying... Its just few lines that in some case is worth repeat it IMHO

p4u:
> The code looks fine, but doesn't this implement exactly what we have already, anyway? I see the same request/response structure with the embedded signature, and the same Send/SendError methods as before. I guess you're using the new router interface, but the overall API seems to work exactly the same way as before for the user.

This is a WIP PR. See feature/use_httprouter branch

httprouter/jsonrpcapi/message.go
mvdan: Then use integers stored in strings - it's just two more characters, and you entirely sidestep the issue of float precision in json :) In Go, they are natively supported like:

```
type T struct {
    IntAsString int64 `json:",string"`
}
```

mvdan: I should mention - the decoder side (e.g. JS) also needs to take care to decode the integer string as a non-float, such as using a real int64 or a "big int". But I imagine that's doable.

httprouter/jsonrpcapi/encoding.go
mvdan: Ah. What about just moving the code to jsonrpcapi, and having the other packages pull HexBytes from here?

Another option is to move HexBytes to the types package, since that should have no vocdoni-node imports.

p4u:
superseded by  #344

httprouter/jsonrpcapi/message.go
p4u: Sure, but we are not gonna do breaking changes now. I'll open an issue to change this in the future.

p4u: https://github.com/vocdoni/vocdoni-node/issues/345

httprouter/jsonrpcapi/encoding.go
p4u: done

PR #337, opened 2021-10-29
add vochainPrivValidator flag to support a remote private validator

altergui:
with `tmkms` this enables the use of Ledger or Yubikey for privkey storage and signing


altergui:
oops, i branched off `stage` instead of `master`, i'll rebase asap.
but feedback on the proposal already welcome.

altergui:
here's the output, using `--vochainPrivValidator tcp://0.0.0.0:26658`
(while running at the same time `tmkms start -c kms/tmkms.toml -v`)
```
dvotenode_1  | 2021-10-29T11:21:30Z	INFO	vochain/start.go:264	PrivValidatorListenAddr defined, Tendermint will wait for a remote private validator connection
dvotenode_1  | 2021-10-29T11:21:30Z	INFO	vochain/start.go:297	tendermint p2p node ID: 145b2a2df939f4ccc306db0dfa4fb66360b9f7de
dvotenode_1  | 2021-10-29T11:21:30Z	INFO	vochain/start.go:301	found genesis file /app/run/stage/vochain/config/genesis.json
```

p4u:
LGTM! @jordipainan?

jordipainan (approved):

jordipainan:
> LGTM! @jordipainan?

LGTM too! Thank you @altergui

jordipainan:
@altergui can we change the base to master ? We use it as dev branch :)

altergui:
definitely, i spotted that last friday but thought there would be conflicts in a rebase and would take me some time... in fact it was trivial. here you go

altergui:
a minor 'regression' in log output is that when this flag is enabled, the log doesn't print the pubkey anymore (since it's not yet known).
i could make it print the pubkey after node initialization, since i guess it's convenient for debugging.

jordipainan:
perf! yup would be nice to print the pubkey, I do not think is a big deal to print it after initialization :)

vochain/start.go
```diff
@@ -361,3 +379,9 @@ func AminoKeys(key crypto25519.PrivKey) (private, public []byte, err error) {
 
 	return aminoKey, aminoPubKey, nil
 }
+
+// AminoPubKey is a helper function that transforms a standard EDDSA pubkey into
+// Tendermint like amino format useful for creating genesis files.
+func AminoPubKey(pubkey []byte) (public []byte, err error) {
```
jordipainan: As public and err are not assigned here you can just have `func AminoPubKey(pubKey []byte) ([]byte, error)`

p4u:
What's missing here @altergui @jordipainan ? can we merge it?

vochain/start.go
altergui: nice, thanks! amended

altergui:
i think it's safe enough to merge, shouldn't introduce new bugs unless the flag is used :)

PR #346, opened 2021-11-05
Deprecate dumpPlain census handler

arnaucube:
Reasoning:
dumpPlain was used to return a json with an array of keys&values in
string hex encoding each one. While dump returns a byte array with the
keys&values in a much more efficient way, using arbo's Dump method.  So,
dumpPlain was doing the same that dump but in a much more inefficient
way, which affects the workload done by the backend for each request,
but also the size of data sent over the network.
With this change, the client can still download the keys&values from the
dump handler, and then parse them following the specified encoding.

This is the description of the dump byte array encoding:
```
dump exports all the Tree leafs in a byte array of length:
[ N * (2+len(k+v)) ]. Where N is the number of key-values, and for each k+v:
[ 1 byte | 1 byte | S bytes | len(v) bytes ]
[ len(k) | len(v) |   key   |     value    ]
Where S is the size of the output of the hash function used for the Tree.
```

### DIFF
--- a/benchmark/api_benchmark_test.go
+++ b/benchmark/api_benchmark_test.go
@@ -139,16 +139,6 @@ func censusBench(b *testing.B, cl *client.Client, size int) {
 		return
 	}
 
-	// dumpPlain
-	/*	log.Infof("[%d] dump claims", rint)
-		req.CensusKey = []byte{}
-		req.CensusKeys = [][]byte{}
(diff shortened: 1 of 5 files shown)


brickpop:
Also worth pointing out that `dumpPlain` was returning an empty list for non-empty censuses. Also deprecating on the client side.

PR #351, opened 2021-11-06
re-enable hashing on census keys

p4u:
We need to hash the census keys in order to ensure the size
of the key fits into the hash function field.

Enable all hashings on censusManager if Digested=False

### DIFF
--- a/census/handler.go
+++ b/census/handler.go
@@ -101,7 +101,19 @@ func (m *Manager) Handler(ctx context.Context, r *api.APIrequest,
 
 	case "addClaimBulk":
 		if validAuthPrefix {
-			invalidClaims, err := tr.AddBatch(r.CensusKeys, r.CensusValues)
+			batchKeys := r.CensusKeys
+			if !r.Digested {
+				batchKeys = [][]byte{}
+				for _, k := range r.CensusKeys {
+					hk, err := tr.Hash(k)
+					if err != nil {
+						log.Warnf("error hashing claim: %v", err)
+						return nil, err
+					}
+					batchKeys = append(batchKeys, hk)
+				}
+			}
+			invalidClaims, err := tr.AddBatch(batchKeys, r.CensusValues)
 			if err != nil {
 				return nil, err
 			}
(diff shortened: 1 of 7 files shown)


p4u:
> Maybe a 'short-term' solution is to not allow to work with this kind of census (not-digested ones) if the kind of tree used is with Poseidon (as it's a kind of tree to be used inside a zk-circuit)

The digested flag indicates wheter the hashing of the pubKey is done by the client or by the server. The hash needs to be done since we need to ensure the Key fits into the finite field of the merkle-tree. So its a requirement we cannot avoid.

p4u:
> In the case of trees that work with Poseidon, how this will be handled to be proved inside a zk circuit?

What you mean? the zkProof is valid for a censusRoot which contains the key Poseidon(pubKey) of the user.

p4u:
> Same than the comment of the hashFunc, I think that would be better to have this method in the tree package instead of the censustree.

The handler has no access to tree, that's why we need it on censusTree (to use the correct hash function depending on the merkle-tree type).

arnaucube:
> > Same than the comment of the hashFunc, I think that would be better to have this method in the tree package instead of the censustree.
> 
> The handler has no access to tree, that's why we need it on censusTree (to use the correct hash function depending on the merkle-tree type).

Make sense, anyway maybe we can still move the `hashFunc` & `Hash()` to the `tree` package and add a wrapper on the `censustree` package. I say this because from the concepts pov, the 'hash function' is not from the censustree itself, but from the tree itself.

arnaucube:
> > In the case of trees that work with Poseidon, how this will be handled to be proved inside a zk circuit?
> 
> What you mean? the zkProof is valid for a censusRoot which contains the key Poseidon(pubKey) of the 

What happens if somebody is using a 'key' in the tree of more than 496 bytes, and uses the digested flag to indicate that the hash needs to be done by the server?
Poseidon hash on server side will fail due not being able to hash more than 16 elements of the finite field (each one of 31bytes).

If the approach is the one from this PR, we should limit the max-length of the keys to be digested when using Poseidon hash (to avoid the described problem).


Apart from that, the described user flow means that the leaf contains the `Poseidon(pubKey)` where the `pubKey` (eth) is a byte-array of 65 bytes? If that's the case, the current circuits do not handle this case (hash the pubkey inside the circuit), and if needed, it's not 'constraint-efficient' to do the bytes cropping logic inside the circuit.

PR #359, opened 2021-11-08
first stone for building the new URL based API

p4u:
extend bearerstdapi package to support path variables
add the package urlApi with the initial code and some example methods


httprouter/bearerstdapi/bearerstdapi.go
```diff
@@ -66,6 +104,64 @@ func (b *BearerStandardAPI) ProcessData(req *http.Request) (interface{}, error)
...
+				log.Warn(err)
+				return
+			}
+			if err := msg.Context.Send(data); err != nil {
+				log.Warn(err)
+			}
+		}
+	}
+
+	path := strings.Join(routerPattern, "/")
```
mvdan: path.Join perhaps

httprouter/bearerstdapi/bearerstdapi.go
```diff
@@ -66,6 +104,64 @@ func (b *BearerStandardAPI) ProcessData(req *http.Request) (interface{}, error)
...
+
+	routerHandler := func(msg httprouter.Message) {
+		bsaMsg := msg.Data.(*BearerStandardAPIdata)
+		bsaMsg.PathVars = make(map[string]string, elementsCount)
+		for i, e := range msg.Path {
+			varName, ok := elementsMap[i+1]
+			if ok {
+				bsaMsg.PathVars[varName] = e
+			}
+		}
```
mvdan: what is this doing? probably worth documenting it

httprouter/bearerstdapi/bearerstdapi.go
```diff
@@ -66,6 +104,64 @@ func (b *BearerStandardAPI) ProcessData(req *http.Request) (interface{}, error)
 	}, nil
 }
 
+// RegisterMethod adds a new method under the URL pattern.
+// The pattern can contain variable names (starting with #), such as /send/#name/hello
+// The pattern can also contain wildcard at the end of the path, such as /send/#name/*
```
mvdan: Did you try reusing chi? It already supports these kind of variables and path matches out of the box; see https://github.com/go-chi/chi. And it's the HTTP URL router we've used elsewhere already.

I know this is not a lot of code, but I'd rather not reinvent the wheel. Some of the code you wrote down here is hard to follow without any docs, for example, and it's hard to say how many edge case bugs it has.

service/api.go
```diff
@@ -55,12 +52,13 @@ func API(apiconfig *config.API, rpc *rpcapi.RPCAPI, storage data.Storage, cm *ce
...
-		}
-	}()
+	/*	go func() {
+			for {
+				time.Sleep(120 * time.Second)
+				log.Infof("[router info] privateReqs:%d publicReqs:%d",
+					atomic.LoadUint64(&rpc.PrivateCalls), atomic.LoadUint64(&rpc.PublicCalls))
+			}
+		}()
+	*/
```
mvdan: just delete the code? not a big fan of accumulating commented out code :)

httprouter/bearerstdapi/bearerstdapi.go
```diff
@@ -66,6 +104,64 @@ func (b *BearerStandardAPI) ProcessData(req *http.Request) (interface{}, error)
...
+// The accessType can be of type private, public or admin.
+func (b *BearerStandardAPI) RegisterMethod(pattern, HTTPmethod string, accessType string, handler BearerStdAPIhandler) error {
+	if pattern[0] != '/' {
+		return fmt.Errorf("pattern must start with /")
```
mvdan: in my opinion these errors can only happen with mistakes like typos in constant strings, and RegisterMethod is only called when the service starts - so IMO they should be panics rather than errors. Either way, the application will fail to start. The bonus of panics is that they clarify "you called me with invalid input", and we don't have to do as much error checking that simply bubbles up to os.Exit.

urlapi/helpers.go
```diff
@@ -0,0 +1,47 @@
...
+			Status:    models.ProcessStatus_name[procInfo.Status],
+			StartDate: procInfo.CreationTime,
+			EndDate:   u.vocinfo.HeightTime(int64(procInfo.EndBlock)),
+		})
+	}
+	return processes, nil
+}
+
+func (u *URLAPI) formatProcessType(et *models.EnvelopeType) string {
+	ptype := ""
```
mvdan: nit: use a strings.Builder to avoid alloc+copy at every addition

urlapi/helpers.go
```diff
@@ -0,0 +1,47 @@
...
+
+	"go.vocdoni.io/proto/build/go/models"
+)
+
+func (u *URLAPI) getProcessSummaryList(pids ...[]byte) ([]*ProcessSummary, error) {
+	processes := []*ProcessSummary{}
+	for _, p := range pids {
+		procInfo, err := u.scrutinizer.ProcessInfo(p)
+		if err != nil {
+			return nil, fmt.Errorf("cannot fetch process info (%v)", err.Error())
```
mvdan: just use `fmt.Errorf("message: %w", err)` with errors: simpler, and also supports error wrapping

urlapi/handlers.go
```diff
@@ -0,0 +1,126 @@
...
+	jProcess.Status = models.ProcessStatus_name[proc.Status]
+	jProcess.Type = u.formatProcessType(proc.Envelope)
+	jProcess.VoteCount, err = u.scrutinizer.GetEnvelopeHeight(processID)
+	if err != nil {
+		return fmt.Errorf("cannot get envelope height (%w)", err)
+	}
+	if proc.HaveResults {
+		results, err := u.scrutinizer.GetResults(processID)
+		if err != nil {
+			return fmt.Errorf("cannot get envelope height (%w)", err)
```
mvdan: The standard format for these is `message: %w`, and I even think fmt.Errorf will only wrap if the format string has that `: %w` suffix.

httprouter/bearerstdapi/bearerstdapi.go
mvdan: Wait, we are *already* using chi under the hood, just two levels down, through httprouter. Why not just use the chi route path syntax directly? A new syntax on top that gets translated (e.g. `/#foo` to `/{foo}`) seems really confusing. It definitely confused me to think we had implemented our own router from scratch.

p4u: I was waiting for this comment, haha. I got the same thinking as you did while I was implementing the custom URL variable parser.

I need to think about how to use the go-chi Mux for this while preserving the abstraction layer of the httpRouter clean (so we can switch from go-chi to something else). Do you have a proposal?

Btw, the /{foo}/ is just added cause go-chi does not allow to use /*/ if its not the end of the path (just a workaround, but I'm not using the {foo} value obtained by go-chi.

httprouter/bearerstdapi/bearerstdapi.go
p4u: makes sense

urlapi/handlers.go
p4u: right

httprouter/bearerstdapi/bearerstdapi.go
mvdan: I see, I wasn't aware you were adding an abstraction layer on purpose. I can't help but think it's overkill here. Abstraction layers for transports, encryption mechanisms, or even message encodings make sense - it's likely we'll support at least a couple of implementations for each sooner than later. I personally don't think that will be the case with chi, as it's a pretty complete and performant http router. I don't see why we need to pay the price of the abstraction.

mvdan: > Btw, the /{foo}/ is just added cause go-chi does not allow to use /*/ if its not the end of the path (just a workaround, but I'm not using the {foo} value obtained by go-chi.

that seems fine - I'd perhaps consider `{discard}` just to clarify we discard the value

p4u: > I see, I wasn't aware you were adding an abstraction layer on purpose. I can't help but think it's overkill here. Abstraction layers for transports, encryption mechanisms, or even message encodings make sense - it's likely we'll support at least a couple of implementations for each sooner than later. I personally don't think that will be the case with chi, as it's a pretty complete and performant http router. I don't see why we need to pay the price of the abstraction.

I think the abstraction layer makes sense because now we have to very different API approaches:

- signed json RPC: the abstraction layer handles the JSON message (which has a specific, predefined format), check signature and resources access, handle error messaging, etc (close to what multirpc does)
- bearer URL API: the abstraction layer handles the bearer Auth, error messages (which have different format) and also the URL variable parsing (which is not required by the JSON RPC)

The httpRouter makes makes that these two implementations (and possible future ones) could coexist within the same http server (go-chi) in a unified and standarized way.

I really think it is quite consistent, the only question is if we could reuse the go-chi variable parsing instead of our own. But as you already said, its just a few lines of code, so not that painful. I'll study changing it (in order to not reinvent the wheel) but for now the current implementation works. We can change it in the future without a huge refactor.

p4u: > > Btw, the /{foo}/ is just added cause go-chi does not allow to use /*/ if its not the end of the path (just a workaround, but I'm not using the {foo} value obtained by go-chi.
> 
> that seems fine - I'd perhaps consider `{discard}` just to clarify we discard the value

does not work cause var names cannot be repeated in go-chi. So I'd need to use {discard1} {discard2} which is even more ugly (I would say).

mvdan: incremental changes sound good to me!

mvdan (approved):
I imagine you'll push the changes for the other bits of the review later? SGTM in any case.

httprouter/bearerstdapi/bearerstdapi.go
p4u: I found a quite elegant and clean solution (easy to apply) that allows using go-chi url parser instead of our own: https://github.com/vocdoni/vocdoni-node/pull/359/files#diff-9eb7c1f117ca67d37752f846cc98bfeaf31a980c9eedf181d547fae87a95d8faR30

p4u: So now use the same go-chi syntax: `/entities/{entity}/processes/{status}`, handlers can get the vars using the HTTP context (https://github.com/vocdoni/vocdoni-node/pull/359/files#diff-2cff19f634c3d726591bf8ffc4119faac8b72b3457b77bb4a18bdbf0b2d814a9R48) and we still preserve the abstraction layer (changing go-chi by another http-server would only require changing the httpRouter layer).

p4u:
> I imagine you'll push the changes for the other bits of the review later? SGTM in any case.

Your comments are addressed and applied. But you can make a final review if you want.

mvdan:
Ah, you're right, I had misread the diff.

httprouter/bearerstdapi/bearerstdapi.go
mvdan: great!

PR #360, opened 2021-11-08
fix web3 connection sync routine if nethermind:

jordipainan:
Adapt the sync progress routine to work properly
when the connected web3 endpoint runs the Nethermind
codebase. The eth_syncing JSON RPC call returns 0x0
as block number.

Add an extra call (eth_blockNumber) in the Connect
method to check that the client can obtain at least the
block number from the connected web3 endpoint.

### DIFF
--- a/ethereum/handler/ethereumHandler.go
+++ b/ethereum/handler/ethereumHandler.go
@@ -103,7 +103,15 @@ func (eh *EthereumHandler) Connect(dialEndpoint string) error {
 		// if RPC connection established, create an ethereum client using the RPC client
 		// TODO: @jordipainan this is racy
 		eh.EthereumClient = ethclient.NewClient(eh.EthereumRPC)
-		log.Infof("connected to %s web3 endpoint", dialEndpoint)
+		ctx, cancel := context.WithTimeout(context.Background(), time.Second*30)
+		defer cancel()
+		blkNum, err := eh.EthereumClient.BlockNumber(ctx)
+		if err != nil {
+			log.Debugf("cannot connect: %s", err)
+			maxtries--
+			continue
+		}
+		log.Infof("connected to %s web3 endpoint, latest known block number is: %d", dialEndpoint, blkNum)
 		return nil
 	}
 	return fmt.Errorf("could not connect to the web3 endpoint %s", dialEndpoint)
@@ -133,10 +141,7 @@ func (eh *EthereumHandler) SyncInfo(ctx context.Context) (*EthSyncInfo, error) {
 	if err != nil {
 		return nil, err
 	}
-	if sp != nil {
-		info.MaxHeight = sp.HighestBlock
-		info.Height = sp.CurrentBlock
-	} else {
+	if sp == nil {
 		header, err := eh.EthereumClient.HeaderByNumber(ctx, nil)
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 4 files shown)


ethereum/handler/ethereumHandler.go
```diff
@@ -134,8 +142,22 @@ func (eh *EthereumHandler) SyncInfo(ctx context.Context) (*EthSyncInfo, error) {
...
 	if sp != nil {
-		info.MaxHeight = sp.HighestBlock
-		info.Height = sp.CurrentBlock
+		// some clients return 0x0 on block number if are synced
+		// do an extra check getting the latest block number via eth_blockNumber
+		ctx, cancel := context.WithTimeout(context.Background(), time.Second*30)
+		defer cancel()
+		blkNum, err := eh.EthereumClient.BlockNumber(ctx)
+		if err != nil {
+			log.Debugf("cannot connect: %s", err)
```
p4u: if the error is returned it should not need to be printed

ethereum/handler/ethereumHandler.go
```diff
@@ -134,8 +142,22 @@ func (eh *EthereumHandler) SyncInfo(ctx context.Context) (*EthSyncInfo, error) {
 		return nil, err
 	}
 	if sp != nil {
-		info.MaxHeight = sp.HighestBlock
-		info.Height = sp.CurrentBlock
+		// some clients return 0x0 on block number if are synced
+		// do an extra check getting the latest block number via eth_blockNumber
+		ctx, cancel := context.WithTimeout(context.Background(), time.Second*30)
+		defer cancel()
+		blkNum, err := eh.EthereumClient.BlockNumber(ctx)
```
p4u: use info.Height directly

ethereum/handler/ethereumHandler.go
```diff
@@ -134,8 +142,22 @@ func (eh *EthereumHandler) SyncInfo(ctx context.Context) (*EthSyncInfo, error) {
 		return nil, err
 	}
 	if sp != nil {
```
p4u: lets reverse this
```
if sp == nil {
	header, err := eh.EthereumClient.HeaderByNumber(ctx, nil)
		if err != nil {
			return nil, err
		}
		info.Height = uint64(header.Number.Int64())
		info.MaxHeight = info.Height
		info.Synced = info.Height > 0
		return info,nil
		}
```

p4u (approved):

p4u:
We should cherry-pick this commit to production

ethereum/handler/ethereumHandler.go
```diff
@@ -133,17 +141,28 @@ func (eh *EthereumHandler) SyncInfo(ctx context.Context) (*EthSyncInfo, error) {
...
-	} else {
+	if sp == nil {
 		header, err := eh.EthereumClient.HeaderByNumber(ctx, nil)
 		if err != nil {
 			return nil, err
 		}
 		info.Height = uint64(header.Number.Int64())
 		info.MaxHeight = info.Height
 		info.Synced = info.Height > 0
+	} else {
```
p4u, line 163: The objective of the if reverse was to remove the "else" and add a return inside the if

jordipainan: done

jordipainan:
@mvdan it is kind of strange that the staticcheck returns an error in this case:
`rpcapi/votehandlers.go:287:18: func (*RPCAPI).getProcessRollingCensusSize is unused (U1000)` while this commit does not introduce such change.
Do you have any ideas ?

mvdan:
Assuming master is green, have you tried rebasing on the latest master?

jordipainan:
@mvdan  force pushed the rebase but b68e96d58c571a2006c6ab85c0003f02d0033545 was the latest commit on this branch and is also green in master, now compose-test error :)

p4u:
> @mvdan it is kind of strange that the staticcheck returns an error in this case: `rpcapi/votehandlers.go:287:18: func (*RPCAPI).getProcessRollingCensusSize is unused (U1000)` while this commit does not introduce such change. Do you have any ideas ?

I also saw this error in the past. It appears randomly. This comes from a lint ignore line added by @ed255 (but to be honest I thought I removed that line...)

EDIT: I misunderstood the error... nothing to do with what I found in the past

jordipainan:
> > @mvdan it is kind of strange that the staticcheck returns an error in this case: `rpcapi/votehandlers.go:287:18: func (*RPCAPI).getProcessRollingCensusSize is unused (U1000)` while this commit does not introduce such change. Do you have any ideas ?
> 
> I also saw this error in the past. It appears randomly. This comes from a lint ignore line added by @ed255 (but to be honest I thought I removed that line...)

thank you for the input @p4u lets see what is happening with the tests now that I just did the rebase with the latest master commit

PR #361, opened 2021-11-08
Add js script to generate snarkjs proof

arnaucube:
Add js script to generate snarkjs proof, to be used in the full-flow
testing.

gen-vote-snark.js takes as arguments: url path inputs. And outputs the
generated zkProof.


p4u:
How do we integrate this with the testsuit in order to run docker-compose e2e tests?

arnaucube:
The thing was that we need a way to generate a zkSNARK proof inside the full-flow-test, and currently we don't have a direct way to compute the Groth16 (compatible with circom ecosystem) from Go. Eventually [vocdoni/go-snark](https://github.com/vocdoni/go-snark) will be updated and we will be able to generate proofs with it (currently we use it to verify them), but meanwhile we can use this js script that uses snarkjs to generate the proof.
The go code of the full-flow-test will call this current js script, passing by arguments the url&path of the circuit-artifacts to use, and the inputs used to generate the zkSNARK (Groth16) proof.

This PR is to the branch `feature/testrolling4`, where @ed255 is doing the go full-flow-test where go code calls the js script with the test-generated inputs, to obtain a valid zk proof and then continue with the test flow.

Issue #354, opened 2021-11-07
error on zkProof format


p4u:
Some mismatch between dvote-js and vocdoni-node makes the vote transaction with zkSnarks fail
```
[...]

Issue #353, opened 2021-11-07
dvote-js registerKeyTx fails


p4u:
dvote-js snark proof fails with "inputs values not inside Finite Field"

```
2021-11-07T00:03:10Z	DEBUG	vochain/app.go:415	rejected tx: inputs values not inside Finite Field
2021-11-07T00:03:10Z	DEBUG	vochain/app.go:413	deliver tx: {"registerKey":{"nonce":"c5uPj9JwoBs9xEUj0z101UpkwDsPmMwvtA65A5E4xyo=", "processId":"Q3E/LJ1SsW3tXPAQseQv1U1VgmmFF6tI86qtTwcI5x0=", "proof":{"arbo":{"type":"BLAKE2B", "siblings":"pQABAB/+bhecw2Hg04koqjp1ODs6aW6AtA/RnK0Ym2x36UVXMMmZ59GcmLuw4ppLlQcOO6xgtEr7Vi/358kX7cHzQIw5Pu5npYMHbBEZCUNO+BEHHjGQvMb8Tvg5fdyKkgFW3e7gMLwcNhT7bkBlDgYiqD6xOotGQt48HZ9O5wMSBP1WrPs6zbRnU7uDVA63RhEJTEqZY5QZqiW1qGcJFI1FUrrF"}}, "newKey":"EikIPmrwSiJXWMDauSwbSJFlkHetsglaDBlMj1UQqeM=", "weight":"AQ=="}}
```

brickpop:
Sending the following transaction:

```json
{
  "method":"submitRawTx",
[...]

PR #364, opened 2021-11-09
Complete Anonymous Testing integration test

### LINKED ISSUE #331: Implement Anonymous Voting Integration test
Implement it in `cmd/vochaintest/vochaintest.go`.

The test should be similar to the `mkTreeVoteTest` with `encryptedVotes = false`, but once the census is built, there's a registration key phase, and in the voting step, users generate a SNARK proof using some js code.

### LINKED ISSUE #358: bug: Use the anonymous vote nullifier to avoid double voting
We are not marking the nullifier (the one from the circuit inputs) as used when registering an anonymous vote!  We need to index each vote data by it's nullifier to avoid the double vote.

Currently in anonymous voting we have nullifier in two places: publicInputs for the circuit and Vote.Nullifier.  And there are APIs that work with the "old" nullifier.  Maybe we could unify them.

ed255:
- Add node.js in the Docker test image
- Introduce a js script `gen-vote-snark.js` (by @arnaucube) in the Docker test
  image used to generate snarkjs proofs for the anonymous voting integration
  test.
- CensusManager:
    - When returing a proof of a Poseidon Merkle Tree (from a rolling census),
      prefix the proof key to the siblings in the return value.
- In client api, add the testGenSNARKProof function whic creates a SNARK proof
  for anonymous voting using the `gen-vote-snark.js` script.
- Update genesis CircuitsConfig for dev
- Complete missing parts in the anonymous testing integration test functions in
  `api/client.go`

Resolve #331
Resolve #358
#354 #353 should be resolved as well, as the integration test shows a full flow of anonymous voting working.

### DIFF
--- a/dockerfiles/testsuite/js/gen-vote-snark.js
+++ b/dockerfiles/testsuite/js/gen-vote-snark.js
@@ -7,40 +7,36 @@ const wc  = require("./witness_calculator.js");
 const snarkjs = require("snarkjs");
 const ffutils = require("ffjavascript").utils;
 
-async function generateProof(inputs, path) {
+function sleep(ms) {
+  return new Promise((resolve) => {
+    setTimeout(resolve, ms);
+  });
+}
+
+async function generateProof(inputs, artifactsPath, outPath) {
   // witness
-  let wasmBuff = fs.readFileSync(`${path}/circuit.wasm`);
+  let wasmBuff = fs.readFileSync(`${artifactsPath}/circuit.wasm`);
   let witnessCalculator = await wc(wasmBuff)
-  let wtnsBuff = await witnessCalculator.calculateWTNSBin(inputs, 0);
-  await fs.writeFile(`${path}/out.wtns`, wtnsBuff, function(err) {
+  let wtnsBuff = await witnessCalculator.calculateWTNSBin(inputs, 1);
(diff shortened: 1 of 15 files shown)


dockerfiles/testsuite/js/gen-vote-snark.js
```diff
@@ -31,16 +31,24 @@ async function generateProof(inputs, path) {
   globalThis.curve_bn128.terminate();
 }
 
-async function downloadFile(id, url, path, filename) {
+async function downloadFile(id, url, path, hashB64, filename) {
   // console.log("download file:", filename);
-  const file = fs.createWriteStream(`${path}/${id}/${filename}`);
+  const hashHex = Buffer.from(hashB64, "base64").toString("hex");
```
arnaucube: Maybe it's better to do the downloading & hash verification from go using 'zk artifacts' package: https://github.com/vocdoni/vocdoni-node/blob/master/crypto/zk/artifacts/artifacts.go#L76
As eventually in the future the proof verification will also be computed from go code, so the js script will no be longer used. So I think that makes sense to currently download & verify hashes from go.

p4u:
@ed255 can you rebase with master and attend Arnau's comment? Then I think we are fine for merging!

arnaucube (approved):
:partying_face:

PR #368, opened 2021-11-12
Detect too many pre-registers in rolling census

### LINKED ISSUE #333: bug: PreRegister design doesn't check for double preRegister
The current design and implementation of the PreRegister doesn't check that the same user from the pre-census registers more than one zkCensusKey in the rolling-census.

I propose that we add a new subTree for the process with PreRegister that sets which users in the pre-census have registered a key in the rolling-census.  This hew subTree would use the pre-census userID (public key) as the tree key.

ed255:
Keep track of used weight per user in the pre-register phase.  For now
we only accept weight 1 for each RegisterKeyTx; but in theory it should
be already possible to pre-register several keys (with weight 1 each) if
the pre-register census is weighted, for example with an ERC20 census.

Resolve https://github.com/vocdoni/vocdoni-node/issues/333


vochain/census.go
```diff
@@ -212,12 +215,56 @@ func (v *State) RegisterKeyTxCheck(vtx *models.Tx, txBytes, signature []byte, st
...
+	usedWeight, err := v.getNullifierUsedWeight(process.ProcessId, nullifier)
+	if err != nil {
+		return fmt.Errorf("cannot get nullifeir used weight: %w", err)
+	}
+	txWeight := new(big.Int).SetBytes(tx.Weight)
+	// TODO: In order to support tx.Weight != 1 for anonymous voting, we
+	// need to add the weight to the leaf in the CensusPoseidon Tree, and
+	// also add the weight as a public input in the circuit to verify it anonymously.
+	if txWeight.Cmp(bigOne) != 0 {
+		return fmt.Errorf("RegisterKey weight != 1 not supported, got: %v", txWeight)
```
p4u, line 236: s/RegisterKey/registerKey

p4u (approved):

p4u:
superseded by #369 (there were conflicting commits so I based 369 PR on this branch)

PR #369, opened 2021-11-14
Feature/enable weighted voting

p4u:


client/api.go
```diff
@@ -837,7 +844,7 @@ func (c *Client) TestSendAnonVotes(
 	if err != nil {
 		return 0, err
 	}
-	log.Infof("CircuitIndex: %v, CircuitConfit: %+v", *circuitIndex, *circuitConfig)
+	log.Infof("CircuitIndex: %v, CircuitPath: %+v", *circuitIndex, *&circuitConfig.CircuitPath)
```
arnaucube: Why `*&circuitConfig.CircuitPath` and not `circuitConfig.CircuitPath`? as `circuit.Config.CircuitPath is a `string`.

```suggestion
	log.Infof("CircuitIndex: %v, CircuitPath: %+v", *circuitIndex, circuitConfig.CircuitPath)
```

vochain/proof.go
```diff
@@ -89,8 +89,12 @@ func VerifyProofOffChainTree(process *models.Process, proof *models.Proof,
...
 			return false, nil, fmt.Errorf("cannot hash proof key: %w", err)
 		}
-		valid, err := tree.VerifyProof(hashFunc, hashedKey, []byte{}, p.Siblings, censusRoot)
-		return valid, bigOne, err
+		valid, err := tree.VerifyProof(hashFunc, hashedKey, p.Value, p.Siblings, censusRoot)
+		// Legacy: support p.Value == nil, assume then value=1
+		if p.Value == nil {
+			return valid, bigOne, err
+		}
+		return valid, new(big.Int).SetBytes(p.Value), err
```
arnaucube: This is converting bytes to bigint interpreting bytes as big-endian, when those bytes come from inside the merkletree where those same bytes are treated as little-endian.
Maybe would make sense to update this line to return the bigint interepreting the bytes as little-endian, as are some bytes coming from inside the tree (which works in little-endian).

In fact, arbo already has a method that abstracts from 'endianness' to convert from bytes to bigint and viceversa, using that, this line could be:
```suggestion
		return valid, arbo.BytesToBigInt(p.Value), err
```

client/api.go
p4u: Thas vs-code autocomplete, I will change it, thanks

vochain/proof.go
p4u: So, there is something I don't understand.

In order to add the value, we do `err := tr.Add(data, r.CensusValue)`. The r.CensusValue is a slice of bytes. Are you saying that Arbo will switch the order of that slice internally on tr.Add() but when returning the value it will just return the bytes (without reversing them again)?

If that's the case I don't understand the logic behind it, thb.

arnaucube: The thing is that if arbo is using a blake2b or sha256 hash internally, the endianness is not a problem. But, if the instance of arbo used is with Poseidon hash, arbo inputs are still `[]byte`, but later internally will need to be converted into `*big.Int` in order to compute the Poseidon hashes. So arbo interprets those byte arrays as little-endian in order to convert them to bigints (only in the Poseidon hash case): https://github.com/vocdoni/arbo/blob/master/tree.go#L278

Now, if we use the `Add` method without taking in account the endianness, we could input a byte-array that encodes a bigint in big-endian, and when internally is converted to bigint to compute the Poseidon hash it gives an error due the value being outside of the finite field.
As an example, the value `21888242871839275222246405745257275088548364400416034343698204186575808495616` (which is inside the finite field).
If we encode it as big-endian it is `0x30644e72e131a029b85045b68181585d2833e84879b9709143e1f593effff449` (in big-endian hex). If we give those bytes, internally when being interpreted as little-endian will give the value `33451715034712113185045486448192360878076553814222909500690766407339250443312`, which was not the original one, but that's not the only problem: when trying to compute the Poseidon hash of this value, it will give the error `inputs values not inside Finite Field`.

As an iteration on the previous example, if somebody sends a vote with `weight` equal to `11888242871839275222246405745257275088548364400416034343698204186575808499697` encoded as a big-endian byte array, when the Vochain tries to add it into a Poseidon Tree, it will not be able and it will return the mentioned error.

client/api.go
```diff
@@ -222,7 +222,7 @@ func (c *Client) TestResults(pid []byte, totalVotes int) ([][]string, error) {
 		}
 		log.Infof("no results yet at block %d", block+2)
 	}
-	total := fmt.Sprintf("%d", totalVotes)
+	total := fmt.Sprintf("%d", uint64(totalVotes)*withWeight)
```
mvdan, line 226: can this not overflow?

urlapi/urlapi.go
```diff
@@ -17,10 +17,11 @@ type URLAPI struct {
...
 
-	router       *httprouter.HTTProuter
-	api          *bearerstdapi.BearerStandardAPI
-	scrutinizer  *scrutinizer.Scrutinizer
-	vocapp       *vochain.BaseApplication
+	router      *httprouter.HTTProuter
+	api         *bearerstdapi.BearerStandardAPI
+	scrutinizer *scrutinizer.Scrutinizer
+	vocapp      *vochain.BaseApplication
+	//lint:ignore U1000 unused
```
mvdan, line 24: or just delete it? :)

p4u:
@arnaucube @emmdim @brickpop @marcvelmer 

In order to fix the endianess issue in a transparent way (for the voter) I made the following changes on the census API:

**addClaim & addClaimBulk:** Instead of `censusValue` and `censusValues` use `weight` and `weights`.
Weight is a string representation of a big number, such as "2" and weights is an array of weight.
This way the client do not need to care about the endianess nor the byte encoding used (before censusValue was a byte array representing a big number).

If weight is not specified the census service will use `1` as default.

example:
```json
{
  "id": "req-12345678",
  "request": {
    "method": "addClaim",
    "censusId": "0x12345678/0x23456789",
    "censusKey": "base64-string", 
    "weight": "2",
    "timestamp": 1556110671
  },
  "signature": "hexString"
}
```

**genProof**: Only needs "censusKey" and returns "siblings", "censusValue" and "weight".
For voting "siblings" and "censusValue" (byte arrays) must be used. The weight here is just to let the user know his voting power in a friendly way.

**checkProof**: Needs censusKey, siblings and censusValue to check the validity of the proof

I'll update the documentation once this is merged.

arnaucube:
> If weight is not specified the census service will use `1` as default.
> example:
> [...]
> "weight": "2.5",

One question, in the example of the comment, there are float values, I've checked the docs page and did not found any reference to it, and in the code the string of `weight` is converted into a `*big.Int`. Is a problem of the example, or the client will send float values in string format?

Just as a side comment (not requesting any change), I see this 'endianness problem' as an 'API level specification problem', as it's a problem on how the client needs to send the data to the API, but then, once the data enters the node, it should be in a the way that works better for the node (and this can be independent on how the client needs to send it). For instance, we could have the old API (previous to this last commit) and internally swap the endianness to have the `weight` ready for being added as `value` bytes in the tree:
- old (assuming that client sends big-endian bytes):
`base64/hex->bytes->swapendianness` --> 2 type conversions
- old (assuming that client sends little-endian bytes):
`base64/hex->bytes` --> 1 type conversion
- with this last commit:
`string->bigint->bytes->swapendianness` --> 3 type conversions

I can see that sending string of the bigint can be less misleading as there can be no confusions about the endianness of the representation, and I'm fine with the last commit change to use string, but just added this comment as I think that is important to separate the two worlds (client interface & node internals).

brickpop:
I agree that the float value doesn't look like playing well with floats. If the census needs a `0.5` resolution, I would rather scale everything by a factor of `2`, so that everything is a natural number

p4u:
Right, the "2.5" value is not correct, however it is just the example.

brickpop:
> genProof: Only needs "censusKey" and returns "siblings", "censusValue" and "weight".
For voting "siblings" and "censusValue" (byte arrays) must be used. The weight here is just to let the user know his voting power in a friendly way.

@p4u why would we have `censusValue` and `weight` together, if they represent the same thing? Shouldn't we simply stick to `weight` everywhere?

> checkProof: Needs censusKey, siblings and censusValue to check the validity of the proof

Same as before, shouldn't we use only one or the other, everywhere?

p4u:
@brickpop  the weight returned by "genProof" is just to inform the user about its weight in a friendly way without requiring encoding/decoding raw bytes

the censusValue is the raw bytes that are part of the proof and must be used as it is (without modifying)

PR #372, opened 2021-11-19
fix signaling oracle acl:

jordipainan:
- call add() when the process is actually created
- increase number of proposals that an address can create (3 per day)


p4u (approved):
This changes should be also cherry-pick to stage

PR #383, opened 2021-12-01
BearerStdApi: enable Quota auth method

NateWilliams2:


httprouter/jsonrpcapi/jsonrpcapi.go
```diff
@@ -76,33 +76,49 @@ func (s *SignedJRPC) AllowPrivate(allow bool) {
...
 			return false, s.returnError("admin address do not match", request.ID)
 		}
 		return true, nil
-	}
-	// If private method, check authentication.
-	// If no authorized addresses configured, allows any address.
-	if s.allowPrivate && request.Private {
-		if len(s.signer.Authorized) == 0 {
-			return true, nil
+	case httprouter.AccessTypeQuota:
```
p4u, line 95: This case should throw a panic() IMO because it means the developer is not aware of this restriction and panic is the best way to identify were exactly the error comes from.

p4u (approved):

PR #389, opened 2021-12-07
vochain/scrutinizer: start using sqlite and sqlc

mvdan:
(see commit message)


mvdan:
This *should* be safe to merge, but it's possible that it may cause panics on the staging/dev environment where it gets deployed due to the `go-cmp.Diff` check on methods like `ProcessInfo`.

If those panics come up they are most likely bugs in my SQL code, which can and will be fixed, but we should also avoid breaking the environments. So if any of you can help me test this before we deploy it, that would be helpful. All I can say is that all of our tests run without a problem. We'll see about compose-test.

mvdan:
compose-test does fail, though CI doesn't seem to be showing its logs, because I don't see the dvotenode failure. @vdo thoughts?

I'll try to repro locally and pull the logs from there, anyway.

mvdan:
I spotted and fixed the first failure, and now it runs into a different failure. Debugging that one now. I'll update the thread when done.

mvdan:
Ah, the second failure was just a mistake on my part. All green now and ready for review :)

Note that the files under `db/` are generated and should not be reviewed.

The main interesting bits are `migrations/` for the SQL schema, `queries/` for the SQL queries, and then the existing Go files being modified to use those via sqlc and sqlite.

vochain/scrutinizer/process.go
```diff
@@ -6,21 +6,65 @@ import (
...
+	sqlStartTime := time.Now()
+
+	queries, ctx, cancel := s.timeoutQueries()
+	defer cancel()
+	sqlProcInner, err := queries.GetProcess(ctx, pid)
+	if err != nil {
+		return nil, err
+	}
+	log.Debugf("ProcessInfo sqlite took %s", time.Since(sqlStartTime))
+	sqlProc := indexertypes.ProcessFromDB(&sqlProcInner)
```
p4u, line 58: Making two calls for each query is a bit annoying. Why not doing this in a single call?

vochain/scrutinizer/process.go
```diff
@@ -358,6 +426,38 @@ func (s *Scrutinizer) newEmptyProcess(pid []byte) error {
 		RollingCensusSize: p.GetRollingCensusSize(),
 	}
 	log.Debugf("new indexer process %s", proc.String())
+
+	queries, ctx, cancel := s.timeoutQueries()
+	defer cancel()
+	if _, err := queries.CreateProcess(ctx, scrutinizerdb.CreateProcessParams{
+		ID:                pid,
```
p4u, line 433: queries.CreateProcess could get `(ctx, p)` and build scrutinizerdb.CreateProcessParams{} itself. I think that would remove noise from process.go (where you don't care about database types).

vochain/scrutinizer/process.go
```diff
@@ -412,15 +512,56 @@ func (s *Scrutinizer) updateProcess(pid []byte) error {
...
+	}
+	// TODO: remove from results table
+	// TODO: hold a sql db transaction for multiple queries
+	queries, ctx, cancel := s.timeoutQueries()
+	defer cancel()
+	previousStatus, err := queries.GetProcessStatus(ctx, pid)
+	if err != nil {
+		return err
+	}
+	if _, err := queries.UpdateProcessFromState(ctx, scrutinizerdb.UpdateProcessFromStateParams{
```
p4u, line 529: same

vochain/scrutinizer/process.go
mvdan: not sure I follow; I need to convert from the sql type to our custom indexertypes type. that requires extra code to run, which I put in ProcessFromDB. I can't make GetProcess do it directly because it's code generated by sqlc.

vochain/scrutinizer/process.go
mvdan: Same as above; CreateProcess is generated code. I don't dictate how it works or what its function type is :)

vochain/scrutinizer/process.go
p4u: Ah, I see. So, would it make sense to build an intermediate method to do both queries at once so abstract the developer from the SQL specific stuff?

mvdan: It's an option, in terms of organizing the code. My intuition is that we should not try to abstract the SQL queries away, because the SQL queries are written *specifically* for the needs of this code. So adding an abstraction layer just hides how the code actually works, but doesn't really help the developer in any significant way.

It's true that translating between the DB types and the API types is a bit cumbersome, but the backend will have to cross that line at some stage. I would rather have that be explicit than implicit - in practice, I expect that some of the internal logic that doesn't touch the API or blockchain state could simply use the DB types directly, avoiding the cost/complexity of translating back and forth with the API types. We only _really_ need to convert to the API type when responding to the user or when storing to the blockchain state.

It's also true that a lot of the backend internals currently use the API types from `indexertypes` directly :) My assumption is that we could transition some of those code paths to using the internal type. I just haven't done that here, to keep the refactor small. The change is leaving the doors open to any future approach with the types, though.

p4u: Ok. I'm merging that as a first stone for this refactor. Then let's see how it evolves.

mvdan:
Thanks for the review and merge :) assuming this gets deployed, we need to monitor for any panics that might have been introduced. If the panics happen, please ping me and I'll try to fix quickly. If they aren't fixed quickly enough, we can temporarily comment out the panic line until I get a chance to fix it. FYI @vdo

vdo:
I will keep an eye @mvdan

PR #391, opened 2021-12-08
update ethless branch

jordipainan:


emmdim:
@jordipainan Let's merge it temporarily to the api testing branch

PR #398, opened 2021-12-13
VPROT-125 TransactionCosts{} represents the tx cost settings in the genesis file

randomshinichi:
Plus a (not yet used) helper function `TransactionCostsFieldFromStateKey` which should be useful when coding the state updating methods.

### DIFF
--- a/vochain/types.go
+++ b/vochain/types.go
@@ -67,26 +67,27 @@ type QueryData struct {
 // ________________________ TRANSACTION COSTS __________________________
 // TransactionCosts describes how much each operation should cost
 type TransactionCosts struct {
-	SetProcess            uint64 `json:"Tx_SetProcess" state:"c_setProcess"`
-	RegisterKey           uint64 `json:"Tx_RegisterKey" state:"c_registerKey"`
-	NewProcess            uint64 `json:"Tx_NewProcess" state:"c_newProcess"`
-	SendTokens            uint64 `json:"Tx_SendTokens" state:"c_sendTokens"`
-	SetAccountInfo        uint64 `json:"Tx_SetAccountInfo" state:"c_setAccountInfo"`
-	AddDelegateForAccount uint64 `json:"Tx_AddDelegateForAccount" state:"c_addDelegateForAccount"`
-	DelDelegateForAccount uint64 `json:"Tx_DelDelegateForAccount" state:"c_delDelegateForAccount"`
-	CollectFaucet         uint64 `json:"Tx_CollectFaucet" state:"c_collectFaucet"`
+	SetProcess            uint64 `json:"Tx_SetProcess"`
+	RegisterKey           uint64 `json:"Tx_RegisterKey"`
+	NewProcess            uint64 `json:"Tx_NewProcess"`
+	SendTokens            uint64 `json:"Tx_SendTokens"`
+	SetAccountInfo        uint64 `json:"Tx_SetAccountInfo"`
+	AddDelegateForAccount uint64 `json:"Tx_AddDelegateForAccount"`
+	DelDelegateForAccount uint64 `json:"Tx_DelDelegateForAccount"`
+	CollectFaucet         uint64 `json:"Tx_CollectFaucet"`
 }
 
 // StructAsBytes returns the contents of TransactionCosts as a map. Its purpose
 // is to keep knowledge of TransactionCosts' fields within itself, so the
 // function using it only needs to iterate over the key-values.
-func (t *TransactionCosts) StructAsBytes() (b map[string][]byte, err error) {
-	b = make(map[string][]byte)
+func (t *TransactionCosts) StructAsBytes() (map[string][]byte, error) {
+	b := make(map[string][]byte)
 
 	tType := reflect.TypeOf(*t)
 	tValue := reflect.ValueOf(*t)
 	for i := 0; i < tType.NumField(); i++ {
-		key := tType.Field(i).Tag.Get("state")
+		key := TransactionCostsFieldToStateKey(tType.Field(i).Name)
+
 		value := tValue.Field(i).Uint()
 		valueBytes, err := util.Uint64ToBytes(value)
 		if err != nil {
@@ -94,7 +95,15 @@ func (t *TransactionCosts) StructAsBytes() (b map[string][]byte, err error) {
 		}
 		b[key] = valueBytes
 	}
-	return
+	return b, nil
+}
+
+// TransactionCostsFieldToStateKey transforms "SetProcess" to "c_setProcess" for all of
+// TransactionCosts' fields
+func TransactionCostsFieldToStateKey(key string) string {
+	a := []rune(key)
+	a[0] = unicode.ToLower(a[0])
+	return strings.Join([]string{costPrefix, string(a)}, "")
 }
 
 // TransactionCostsFieldFromStateKey transforms "c_setProcess" to "SetProcess" for all of
--- a/vochain/types.go
+++ b/vochain/types.go
@@ -7,6 +7,7 @@ import (
 	"regexp"
 	"strings"
 	"time"
+	"unicode"
 
 	"go.vocdoni.io/dvote/types"
 	"go.vocdoni.io/dvote/util"
@@ -30,6 +31,8 @@ var (
 	ErrAccountBalanceZero  = fmt.Errorf("zero balance account not valid")
 	// keys; not constants because of []byte
 	voteCountKey = []byte("voteCount")
+
+	rxAlphabetical = regexp.MustCompile(`^[A-Za-z]+$`)
 )
 
 // PrefixDBCacheSize is the size of the cache for the MutableTree IAVL databases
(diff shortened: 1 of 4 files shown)


vochain/types.go
```diff
@@ -59,13 +63,74 @@ type QueryData struct {
...
+
+		b[key] = valueBytes
+	}
+	return
+}
+
+// TransactionCostsFieldFromStateKey transforms "c_setProcess" to "SetProcess" for all of
+// TransactionCosts' fields
+func TransactionCostsFieldFromStateKey(key string) (string, error) {
+	if key[0:2] != "c_" {
```
jordipainan: Can you add `c_` as a constant ?

jordipainan:
LGTM :) Now we should add the costs to the state in the Extra tree creating the get/set corresponding functions (Same did for the Treasurer). After this we are going to implement the APP handlers for setting the costs via Vochain transaction.

vochain/types.go
```diff
@@ -59,13 +64,74 @@ type QueryData struct {
...
+// function using it only needs to iterate over the key-values.
+func (t *TransactionCosts) StructAsBytes() (b map[string][]byte, err error) {
+	b = make(map[string][]byte)
+
+	tType := reflect.TypeOf(*t)
+	tValue := reflect.ValueOf(*t)
+	for i := 0; i < tType.NumField(); i++ {
+		key := tType.Field(i).Tag.Get("state")
+		value := tValue.Field(i).Uint()
+
```
p4u, line 93: No need for these blank lines here

vochain/types.go
```diff
@@ -59,13 +64,74 @@ type QueryData struct {
...
+	name := strings.TrimLeft(key, costPrefix)
+
+	// strings.Title will misbehave when there are punctuation marks in the
+	// string. To clean the input up, we ensure there are only alphabetical
+	// characters left in the string
+	isAlpha := regexp.MustCompile(`^[A-Za-z]+$`).MatchString
+	if !isAlpha(name) {
+		return "", fmt.Errorf("%s needs to be alphabetical only", name)
+	}
+	Name := strings.Title(name)
```
p4u: Not go friendly, please don't use capital letter for function scope variables.

vochain/types.go
```diff
@@ -59,13 +64,74 @@ type QueryData struct {
...
+	if key[0:2] != costPrefix {
+		return "", fmt.Errorf("state keys must start with '%s', got %s", costPrefix, key)
+	}
+	name := strings.TrimLeft(key, costPrefix)
+
+	// strings.Title will misbehave when there are punctuation marks in the
+	// string. To clean the input up, we ensure there are only alphabetical
+	// characters left in the string
+	isAlpha := regexp.MustCompile(`^[A-Za-z]+$`).MatchString
+	if !isAlpha(name) {
```
p4u: No need to declare isAlpha, better `if!isAlpha(regexp....)`

vochain/types.go
```diff
@@ -59,13 +64,74 @@ type QueryData struct {
...
+		return "", fmt.Errorf("%s needs to be alphabetical only", name)
+	}
+	Name := strings.Title(name)
+
+	// check if TransactionCosts actually has such a field
+	_, found := reflect.TypeOf(TransactionCosts{}).FieldByName(Name)
```
p4u: same as before: `if _, found := reflext.TypeOf...; !found {`

vochain/types.go
```diff
@@ -59,13 +64,74 @@ type QueryData struct {
...
+
+		b[key] = valueBytes
+	}
+	return
+}
+
+// TransactionCostsFieldFromStateKey transforms "c_setProcess" to "SetProcess" for all of
+// TransactionCosts' fields
+func TransactionCostsFieldFromStateKey(key string) (string, error) {
+	if key[0:2] != costPrefix {
```
p4u: before accessing `key[:2]` check the size and return an error

p4u (changes requested):

vochain/types.go
randomshinichi: Aww i thought it was an elegant way of expressing what that variable was supposed to contain

jordipainan:
@mvdan would be nice if you can recheck it :) Thank you

mvdan:
@randomshinichi could you please fix the conflicts first? otherwise we might be reviewing code that needs to be changed anyway :)

randomshinichi:
@mvdan whoops didn't see that, re-pushed!

vochain/types.go
```diff
@@ -59,13 +64,73 @@ type QueryData struct {
...
+
+// TransactionCostsFieldFromStateKey transforms "c_setProcess" to "SetProcess" for all of
+// TransactionCosts' fields
+func TransactionCostsFieldFromStateKey(key string) (string, error) {
+	if len(key) < 3 {
+		return "", fmt.Errorf("state key must have a length greater than 3, because it should include the costPrefix: got %s", key)
+	}
+	if key[0:2] != costPrefix {
+		return "", fmt.Errorf("state keys must start with '%s', got %s", costPrefix, key)
+	}
```
mvdan, line 120: you can replace these 6 lines with just `strings.HasPrefix`.

you can even reuse the result of TrimLeft, to avoid checking the string twice:

```
name := strings.TrimLeft(key, costPrefix)
if name == key { // missing prefix
    return "", fmt.Errorf("state keys must start with %q, got %q", costPrefix, key)
}
```

also note you should use %q when logging/erroring with string values.

vochain/types.go
```diff
@@ -59,13 +64,73 @@ type QueryData struct {
...
+	name := strings.TrimLeft(key, costPrefix)
+
+	// strings.Title will misbehave when there are punctuation marks in the
+	// string. To clean the input up, we ensure there are only alphabetical
+	// characters left in the string
+	if !regexp.MustCompile(`^[A-Za-z]+$`).MatchString(name) {
```
mvdan: don't compile the regexp every time we call this function; that's wasteful. use a global, like:

```
var rxAlphabetical = regexp.MustCompile(...)
```

vochain/types.go
```diff
@@ -59,13 +64,73 @@ type QueryData struct {
...
+// StructAsBytes returns the contents of TransactionCosts as a map. Its purpose
+// is to keep knowledge of TransactionCosts' fields within itself, so the
+// function using it only needs to iterate over the key-values.
+func (t *TransactionCosts) StructAsBytes() (b map[string][]byte, err error) {
+	b = make(map[string][]byte)
+
+	tType := reflect.TypeOf(*t)
+	tValue := reflect.ValueOf(*t)
+	for i := 0; i < tType.NumField(); i++ {
+		key := tType.Field(i).Tag.Get("state")
```
mvdan: you could do the reverse transformation of TransactionCostsFieldFromStateKey, meaning then that you can entirely get rid of the field tags and you make the code more consistent too.

you just need to lowercase the first letter (we assume ASCII letters so that's easy), and add the `c_` prefix.

vochain/types.go
```diff
@@ -59,13 +64,73 @@ type QueryData struct {
...
+	SetAccountInfo        uint64 `json:"Tx_SetAccountInfo" state:"c_setAccountInfo"`
+	AddDelegateForAccount uint64 `json:"Tx_AddDelegateForAccount" state:"c_addDelegateForAccount"`
+	DelDelegateForAccount uint64 `json:"Tx_DelDelegateForAccount" state:"c_delDelegateForAccount"`
+	CollectFaucet         uint64 `json:"Tx_CollectFaucet" state:"c_collectFaucet"`
+}
+
+// StructAsBytes returns the contents of TransactionCosts as a map. Its purpose
+// is to keep knowledge of TransactionCosts' fields within itself, so the
+// function using it only needs to iterate over the key-values.
+func (t *TransactionCosts) StructAsBytes() (b map[string][]byte, err error) {
```
mvdan: nit: avoid named returns in general. see https://github.com/golang/go/wiki/CodeReviewComments#naked-returns and https://github.com/golang/go/wiki/CodeReviewComments#named-result-parameters

vochain/types.go
randomshinichi: that's convenient

randomshinichi: > you can even reuse the result of TrimLeft, to avoid checking the string twice:

No, error checking should be pedantic and repetitive, otherwise input that slips through the first few tests will hit a later error which isn't the actual error. We have all encountered error messages which don't make sense, but happen because of the way things are ordered within the code.
Consider `TransactionCostsFieldFromStateKey("c_")` which will result in the error `"" needs to be alphabetical only`, because `name=""` after `strings.TrimPrefix()`.

Better to just check a string several times, what's the hurry? My dad calls this "trying to be smart".

vochain/types.go
randomshinichi: fortunately none of this involves returning an `error` so this was simple!

p4u (approved):

vochain/types.go
jordipainan: is this solved ?

PR #401, opened 2021-12-13
Add support for `MintTokensTx` and `SetAccountDelegateTx`

jordipainan:


jordipainan:
@mvdan can the failing test because of a false positive?, I think I need to import the `common` go-eth pkg

`vochain/state_test.go:8:2: package "github.com/ethereum/go-ethereum/common" is being imported more than once (ST1019)`

p4u:
run `go mod tidy`

vochain/account.go
```diff
@@ -282,3 +284,124 @@ func SetAccountInfoTxCheck(vtx *models.Tx, txBytes, signature []byte, state *Sta
...
+				return common.Address{}, common.Address{}, fmt.Errorf("delegate already added")
+			}
+		}
+		return sigAddress, delAcc, nil
+	case models.TxType_DEL_DELEGATE_FOR_ACCOUNT:
+		f := false
+		for i := 0; i < len(acc.DelegateAddrs); i++ {
+			delegateToCmp := common.BytesToAddress(acc.DelegateAddrs[i])
+			if delegateToCmp == delAcc {
+				f = true
```
p4u: just put the return here `return sigAddress, delAcc, nil` remove `f`and return by default `return common.Address{}, common.Address{}, fmt.Errorf("cannot remove a non existent delegate")`

vochain/state.go
```diff
@@ -503,38 +503,63 @@ func (v *State) IsOracle(addr common.Address) (bool, error) {
 
 // setTreasurer saves the Treasurer address to the state
 func (v *State) setTreasurer(address common.Address) error {
+	t := &models.Treasurer{
+		Address: address.Bytes(),
+		Nonce:   0,
+	}
+	tBytes, err := proto.Marshal(t)
```
p4u:
```
tBytes, err := proto.Marshal(&models.Treasurer{
		Address: address.Bytes(),
		Nonce:   0,
	}
```

p4u (approved):

p4u:
This is a good piece of code, congrats @jordipainan

PR #402, opened 2021-12-13
Don't require processID for getEnvelopeStatus

NateWilliams2:
There's no need for the processID in the getEnvelopeStatus call


p4u (approved):

p4u:
Can you update the docs?

PR #404, opened 2021-12-16
Decode resp message on submitRawTx errors

NateWilliams2:
Fixes an issue where messages returned by submitRawTx API methods are encoded in hex

### DIFF
--- a/rpcapi/votehandlers.go
+++ b/rpcapi/votehandlers.go
@@ -25,11 +25,11 @@ func (r *RPCAPI) submitRawTx(request *api.APIrequest) (*api.APIresponse, error) 
 		return nil, fmt.Errorf("no reply from vochain")
 	}
 	if res.Code != 0 {
-		return nil, fmt.Errorf("%s", res.Data)
+		return nil, fmt.Errorf("%s", string(res.Data))
 	}
 	log.Debugf("broadcasting tx hash:%s", res.Hash)
 	// return nullifier or other info
-	return &api.APIresponse{Payload: fmt.Sprintf("%x", res.Data)}, nil
+	return &api.APIresponse{Payload: fmt.Sprintf("%s", string(res.Data))}, nil
 }
 
 func (a *RPCAPI) submitEnvelope(request *api.APIrequest) (*api.APIresponse, error) {
--- a/rpcapi/votehandlers.go
+++ b/rpcapi/votehandlers.go
@@ -29,7 +29,7 @@ func (r *RPCAPI) submitRawTx(request *api.APIrequest) (*api.APIresponse, error) 
 	}
 	log.Debugf("broadcasting tx hash:%s", res.Hash)
 	// return nullifier or other info
-	return &api.APIresponse{Payload: fmt.Sprintf("%s", string(res.Data))}, nil
+	return &api.APIresponse{Payload: fmt.Sprintf("%x", res.Data)}, nil
 }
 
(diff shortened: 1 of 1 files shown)


mvdan:
please don't merge this yet; want to dig into it a little :)

rpcapi/votehandlers.go
```diff
@@ -25,11 +25,11 @@ func (r *RPCAPI) submitRawTx(request *api.APIrequest) (*api.APIresponse, error)
 		return nil, fmt.Errorf("no reply from vochain")
 	}
 	if res.Code != 0 {
-		return nil, fmt.Errorf("%s", res.Data)
+		return nil, fmt.Errorf("%s", string(res.Data))
 	}
 	log.Debugf("broadcasting tx hash:%s", res.Hash)
 	// return nullifier or other info
-	return &api.APIresponse{Payload: fmt.Sprintf("%x", res.Data)}, nil
+	return &api.APIresponse{Payload: fmt.Sprintf("%s", string(res.Data))}, nil
```
mvdan: This change is invalid; the non-error response will be binary and thus needs to be encoded as hex, so you can't simply return the non-encoded form as a JSON string. You can see how the tests broke because of this.

I'd say just revert this line change.

rpcapi/votehandlers.go
```diff
@@ -25,11 +25,11 @@ func (r *RPCAPI) submitRawTx(request *api.APIrequest) (*api.APIresponse, error)
 		return nil, fmt.Errorf("no reply from vochain")
 	}
 	if res.Code != 0 {
-		return nil, fmt.Errorf("%s", res.Data)
+		return nil, fmt.Errorf("%s", string(res.Data))
```
mvdan, line 28: this is the correct change. the problem is that tendermint's HexBytes has a String method that does an uppercase hex encode. we don't want that - error strings are always plaintext and not binary, so we can return them directly.

*however*, note that this will mean that the client side must only do hex decode on `resp.Data` if *and only if* `resp.Code==0`. Otherwise, the response error message will be in `resp.Data` as a plain string, without hex encoding. Is that how the API is documented today? If not, and we merge this, we should also fix the API docs.

cc @brickpop @marcvelmer

mvdan (approved):

rpcapi/votehandlers.go
p4u: I'm not sure I'm following your comment @mvdan 

I think in case of res.Code != 0 we should `return nil, fmt.Errorf(hex.DecodeString(util.trimHex(res.Data)))`
The RPCAPI will take care of setting the string as a JSON error message for the client.

mvdan: res.Data holds bytes, not a hex string, so I'm pretty sure your code wouldn't work?

the problem is that res.Data is tendermint's `type HexBytes []byte`, whose String method encodes as hex. our previous code was calling that String method via `%s`. the new code prevents that by instead converting to `string`, which then gives us the non-encoded bytes as a string directly.

p4u: I see, thanks.

p4u: So lets merge this @mvdan @NateWilliams2 ?

mvdan: Yeah, I think so. What I said earlier about `resp.Data` being hex only some of the time is wrong, it's always hex - and it comes from tendermint. It's us who have to be careful to not encode as hex when returning errors. That's the fix Nate is doing here. LGTM :)

mvdan:
Not sure why this merge broke compose-test on master; it passed here.

PR #408, opened 2021-12-16
vochain/scrutinizer: tweak ProcessList sql code

mvdan:
(see commit message)


p4u (approved):

mvdan:
Hmm, since I'm going to be out for two weeks over christmas, maybe we should leave the panics disabled. I won't have a way to tell when the mismatches happen, but at least I could pull the logs every week or so and `grep` for them.

p4u:
maybe @vdo can help with that

PR #413, opened 2021-12-17
State TxCost Setter/Getter and Genesis State setting

randomshinichi:
Tx Costs can now be set directly on the State, and the State is populated with TxCosts from the genesis.

### DIFF
--- a/vochain/app.go
+++ b/vochain/app.go
@@ -359,6 +359,15 @@ func (app *BaseApplication) InitChain(req abcitypes.RequestInitChain) abcitypes.
 		log.Fatalf("could not set State.Treasurer from genesis file: %s", err)
 	}
 
+	// add tx costs
+	txCosts := genesisAppState.TxCost.AsMap()
+	for k, v := range txCosts {
+		err = app.State.SetTxCost(k, v)
+		if err != nil {
+			log.Fatalf("could not set tx cost %q to value %q from genesis file to the State", k, v)
+		}
+	}
+
 	// Is this save needed?
 	if _, err := app.State.Save(); err != nil {
 		log.Fatalf("cannot save state: %s", err)
--- a/vochain/app.go
+++ b/vochain/app.go
@@ -360,8 +360,7 @@ func (app *BaseApplication) InitChain(req abcitypes.RequestInitChain) abcitypes.
 	}
 
 	// add tx costs
-	txCosts := genesisAppState.TxCost.AsMap()
-	for k, v := range txCosts {
+	for k, v := range genesisAppState.TxCost.AsMap() {
 		err = app.State.SetTxCost(k, v)
 		if err != nil {
 			log.Fatalf("could not set tx cost %q to value %q from genesis file to the State", k, v)
--- a/vochain/state.go
+++ b/vochain/state.go
@@ -591,12 +591,15 @@ func (v *State) incrementTreasurerNonce() error {
 	return v.Tx.DeepSet([]byte(TreasurerKey), tBytes, ExtraCfg)
 }
 
-func (v *State) SetTxCost(txType models.TxType, cost uint64) error {
+func (v *State) SetTxCostByTxType(txType models.TxType, cost uint64) error {
 	key, ok := TxTypeToKey[txType]
 	if !ok {
 		return fmt.Errorf("txType %v shouldn't cost anything", txType)
 	}
+	return v.SetTxCost(key, cost)
+}
 
+func (v *State) SetTxCost(key string, cost uint64) error {
 	v.Tx.Lock()
 	defer v.Tx.Unlock()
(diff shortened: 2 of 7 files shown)


vochain/state.go
```diff
@@ -25,14 +25,44 @@ import (
...
-const defaultHashLen = 32
+const (
+	// defaultHashLen is the most common default hash length for Arbo hashes.  This
+	// is the value of arbo.HashFunctionSha256.Len(), arbo.HashFunctionPoseidon.Len() and
+	// arbo.HashFunctionBlake2b.Len()
+	defaultHashLen = 32
+
+	// k* are the keys under which the State stores how much each
+	// transaction should cost
+	kSetProcess            = "c_setProcess"
```
jordipainan: Can we have a more self-describing name ? For `kSetProcess` I understand that this is some key for set process but nothing related with tx costs. I think `kCostSetProcess` or `kSetProcessCost` might be more self-describing.

vochain/state.go
```diff
@@ -25,14 +25,44 @@ import (
...
+	kSendTokens            = "c_sendTokens"
+	kSetAccountInfo        = "c_setAccountInfo"
+	kAddDelegateForAccount = "c_addDelegateForAccount"
+	kDelDelegateForAccount = "c_delDelegateForAccount"
+	kCollectFaucet         = "c_collectFaucet"
+)
+
+// TxTypeToKey translates models.TxType to a string which the State uses
+// as a key internally under the Extra tree
+var TxTypeToKey = map[models.TxType]string{
```
jordipainan: Maybe `TxTypeToCostKey` ?

vochain/state.go
```diff
@@ -561,6 +591,36 @@ func (v *State) incrementTreasurerNonce() error {
...
+func (v *State) SetTxCost(key string, cost uint64) error {
+	v.Tx.Lock()
+	defer v.Tx.Unlock()
+
+	costBytes := [8]byte{}
+	binary.LittleEndian.PutUint64(costBytes[:], cost)
+	return v.Tx.DeepSet([]byte(key), costBytes[:], ExtraCfg)
+}
+
+func (v *State) TxCostByTxType(txType models.TxType) (uint64, error) {
```
jordipainan: Can we have the `isQuery` approach here ? Like we have in `Oracles()` or `Treasurer()`

vochain/state.go
```diff
@@ -561,6 +591,36 @@ func (v *State) incrementTreasurerNonce() error {
 	return v.Tx.DeepSet([]byte(TreasurerKey), tBytes, ExtraCfg)
 }
 
+func (v *State) SetTxCostByTxType(txType models.TxType, cost uint64) error {
```
jordipainan: Given that this function calls internally `SetTxCost()` can we have `SetTxCost()` unexported ?
Is there any other way to set a tx cost that does not require to have a `TxType` ? Asking this because I cannot clearly see the fact that there are two ways of setting the tx cost. I there should be just one way to set this and is using the `txType`.

jordipainan: Same for the get methods.

randomshinichi: OK, this will mean that everything will have to be translated to the models.TxType int32 when one wants to specify what type of cost one wants to update/get.

randomshinichi: Err, perhaps I should re-clarify.
As you may remember in our zeal we made `TransactionCosts{}` translate itself into a key-value map where the keys were "c_*". This is useful when initializing the State from genesis. That's what `SetTxCost(key string)` is for.

However, for the rest of the programs' lifetime, it will receive requests by `models.TxType`. So, maybe we should make `TransactionCosts{}.AsMap()` also use `models.TxType`, even though using "c_*" as keys is more... intuitive.

vochain/app.go
```diff
@@ -359,6 +359,15 @@ func (app *BaseApplication) InitChain(req abcitypes.RequestInitChain) abcitypes.
 		log.Fatalf("could not set State.Treasurer from genesis file: %s", err)
 	}
 
+	// add tx costs
+	txCosts := genesisAppState.TxCost.AsMap()
```
p4u: No need to declare txCosts :)

vochain/state.go
p4u: Actually I think its easier to read if we don't declare the const and add the strings directly to the map (unless we need the consts somewhere else).

```
	models.TxType_SET_PROCESS_STATUS:     "c_setProcess",    
```

p4u (changes requested):

p4u:
Please, rebase with master

jordipainan:
Reminder: Squash commits

jordipainan (approved):

p4u (approved):
LGTM

PR #416, opened 2021-12-20
ipfssync: hotfix regression which made libp2p listen only on ipv6

altergui:
This regression was introduced in "dualstack: discover both IPv4 and IPv6..."

with the regression, if the node is ipv4-only, it will not connect to other nodes.
if it's dualstack, this bug goes unnoticed since it fallbacks to ipv6 and still works OK.


p4u (approved):

p4u:
I think this hotfix is not needed as the refactor is already merged. Please close the PR

PR #419, opened 2021-12-21
WIP: ethless vochain 0.1

jordipainan:


vochain/account.go
```diff
@@ -32,15 +34,18 @@ func (a *Account) Unmarshal(data []byte) error {
 }
 
 // Transfer moves amount from the origin Account to the dest Account.
-func (a *Account) Transfer(dest *Account, amount uint64, nonce uint32) error {
+func (a *Account) Transfer(dest *Account, amount uint64, nonce uint32, faucet bool) error {
```
p4u: Umm, I don't like this approach (personally). I'd prefer to let Transfer() as a method for transfering balance among accounts and create a new method under `Account` named something like `CollectFaucet()`.

The reason is that Transfer is a very important method that will be used by several parts of the code. Not all developers will understand what the "faucet" boolean means and its easy to make errors (and important ones).

vochain/account.go
```diff
@@ -89,13 +93,10 @@ func (a *Account) DelDelegate(addr common.Address) error {
 // and updates the state with the new values (including nonce).
 // If origin address acc is not enough, ErrNotEnoughBalance is returned.
 // If provided nonce does not match origin address nonce+1, ErrAccountNonceInvalid is returned.
-// If isQuery is set to true, this method will only check against the current state (no changes will be stored)
-func (v *State) TransferBalance(from, to common.Address, amount uint64, nonce uint32, isQuery bool) error {
+func (v *State) TransferBalance(from, to common.Address, amount uint64, nonce uint64, faucet bool) error {
```
p4u: Same as above. I don't like using the same Method for collect faucet funds and send tokens between accounts. I think in this case its better to make things very explicit.

vochain/account.go
```diff
@@ -392,6 +399,98 @@ func (v *State) setDelegate(accountAddr, delegateAddr common.Address, txType mod
 	}
 }
 
+func SendTokensTxCheck(vtx *models.Tx, txBytes, signature []byte, state *State) (common.Address, common.Address, uint64, uint32, error) {
```
p4u: I know returning many individual parameters is a quite common behavior here. However I think we should slowly change it. So IMO it would be better if this method returns a struct+error instead of 5 unamed parameters.

p4u (changes requested):

p4u:
ping @jordipainan

jordipainan:
@p4u

- Rebased with master
- Addressed comments regarding transfer and collect from faucet, both function have been decoupled
- Added support for transaction costs implemented by randomshinichi
- Some other minor changes

For supporting these latest changes some testing has been refactored.
Also adds unit tests as well as integration tests for the newly added functionalities.

Commits before 95550fc are already reviewed, added them to this commit for maintaining the sync with the master branch. 

Also, I am going to rebase and make the proper changes when #437 is merged in this PR if you are fine with it ^^
Thank you

jordipainan:
Added WIP until branch stabilization cc: @emmdim

jordipainan:
Ping @mvdan https://github.com/vocdoni/vocdoni-node/runs/4929394156?check_suite_focus=true#step:4:413

mvdan:
Thanks, looking.

mvdan:
@jordipainan if this is blocking you, you can temorarily comment out the `log.Fatal` line in `sqliteWarnf` in your branch.

p4u:
@jordipainan this PR is quite impossible to properly review :( Its too long and there is a mix of commits from other PRs which are shown as new.

I'm not sure what to do here @mvdan ideas? I do see two options:

1. Just merge it and cross fingers (after the merge we need to run a lot of integration testing)
2. Rebuild the whole PR from scratch starting from the current Master branch once the Tendermint WIP PR is also merged.

jordipainan:
No problem, I will rebuild the entire PR starting from the master branch when the other PR (Tendermint 0.35 & New signing schema) are merged.

PR #420, opened 2021-12-21
vochain: if startBlock equals 0 or 1, start on next block

p4u:

### DIFF
--- a/client/api.go
+++ b/client/api.go
@@ -1037,15 +1037,21 @@ func (c *Client) CreateProcess(oracle *ethereum.SignKeys,
 	envelopeType *models.EnvelopeType,
 	mode *models.ProcessMode,
 	censusOrigin models.CensusOrigin,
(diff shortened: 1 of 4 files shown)


jordipainan:
As the assignation of the start block is set on CheckTx what happens if the tx is not included in the next 2 blocks ?

p4u:
> As the assignation of the start block is set on CheckTx what happens if the tx is not included in the next 2 blocks ?

CheckTx does not modify the state. DeliverTx will also execute it and set correctly the startBlock.

jordipainan:
> > As the assignation of the start block is set on CheckTx what happens if the tx is not included in the next 2 blocks ?
> 
> CheckTx does not modify the state. DeliverTx will also execute it and set correctly the startBlock.

Yup, was thinking about the checkTx (I know that it does not change the state)  and deliverTx not executing it. Thank you

PR #423, opened 2022-01-04
Add entropy to router prometheus middleware

NateWilliams2:
for the purpose of integration tests, I'm trying to create one router on localhost:portA and another on localhost:portB, one of which serves a local gateway api and the other serves the VaaS API to query locally from the tests. The error I'm getting is 
```
panic: duplicate metrics collector registration attempted
```
The problem is solved with this fix, which adds a source of entropy so the middleware is not identical across multiple router instances.


p4u:
@NateWilliams2 I think it would be a better approach to be able to set "manually" the prometheus metrics name. So you can just add a new exported string variable here https://github.com/vocdoni/vocdoni-node/pull/423/files#diff-720838454c767cfa980d2867c161fb37a513d19ac9c32721ed4f8ed0e73dc045R41 

If no value, the default value must be used. So only if the caller sets it `httpRouter.PrometheusID = "foobar"` this value will be used.

NateWilliams2:
@p4u how's this?

vocone/vocone.go
```diff
@@ -296,7 +296,7 @@ func startAPI(host string, port int, path string) (*rpcapi.RPCAPI, error) {
 	if err := signer.Generate(); err != nil {
 		return nil, err
 	}
-	httpRouter := httprouter.HTTProuter{}
+	httpRouter := httprouter.HTTProuter{PrometheusID: "vocone-chi"}
```
p4u: Why "vocone-chi"? I'd not set a custom ID if not necessary (when Grafana is configured, this ID is relevant).

p4u:
Great, thanks.

Btw, why go.mod is modified?

NateWilliams2:
When I  try to run vocone with the go.sum from master, I get 
```
go: github.com/vocdoni/arbo@v0.0.0-20211217085703-d56ab859f109 requires
	go.vocdoni.io/dvote@v1.0.4-0.20211025120558-83c64f440044 requires
	go.vocdoni.io/proto@v1.13.3-0.20211019215004-0ca7d4436553: missing go.sum entry; to add it:
	go mod download go.vocdoni.io/proto
```
After running go mod tidy, this is fixed, but as you can see the ci test is failing because of the go.mod. Could this be an issue with me using macOS again? @mvdan

p4u:
I'm not sure what happened but if `go mod tidy` fixes it, I'd say its fine to run it and push it. Add a message to the commit `run go mod tidy`

mvdan:
I can reproduce that, on master, running `go mod tidy` changes a few things. It removes recws and updates the go.sum entries for dvote/proto.

But why does this PR remove so many lines from go.mod? `go mod tidy` doesn't do that for me. Did you do that manually?

mvdan:
For some context: removing those indirect requirements has the effect of downgrading indirect dependencies, which we most likely don't want. We want to keep all dependencies relatively up to date, which we're doing over time as we can. A mass-downgrade of indirect deps like this is quite risky :)

NateWilliams2:
@mvdan no, didn’t remove anything manually. This is just the result of running go mod tidy.

mvdan:
Hmm, could you please share repro steps along with what Go version you ran it with? Perhaps on Discord to avoid friction :)

go.mod
```diff
@@ -322,6 +319,8 @@ require (
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 )
 
+require github.com/libp2p/go-libp2p-pubsub v0.4.2 // direct
```
mvdan: kinda weird that this got moved out of the first block :) I also don't get this with 1.17.5. perhaps undo it manually? it might be a leftover from the earlier mess with 1.16.

NateWilliams2:
@p4u can I merge?

p4u (approved):

p4u:
sure

PR #426, opened 2022-01-09
hotfix: compose-test workaround for network issue

### LINKED ISSUE #425: bug: CI github runner sometimes fails to connect to deb.debian.org:80
since last week, the `compose-test` sometimes fails:
```
Could not connect to deb.debian.org:80 (199.232.150.132), connection timed out
```

examples:
* https://github.com/vocdoni/vocdoni-node/runs/4750624991?check_suite_focus=true#step:3:166
* https://github.com/vocdoni/vocdoni-node/runs/4748289606?check_suite_focus=true#step:3:591
* https://github.com/vocdoni/vocdoni-node/runs/4728901369?check_suite_focus=true#step:3:175
* https://github.com/vocdoni/vocdoni-node/runs/4728640708?check_suite_focus=true#step:3:216
(couldn't find any earlier failures in the last couple of months)

i found a similar issue affecting gitlab users
* https://forum.gitlab.com/t/gitlab-ci-get-timeout-from-debian-org/59040/7
* https://gitlab.com/gitlab-com/gl-infra/production/-/issues/5590

which suggests passing `docker build --network host` to workaround (AFAIU) a MTU issue of docker-in-docker. I'll try this, but since it happens only randomly, only time will tell if it's "fixed", so I'd keep this report open for a week or so.

altergui:
fixes #425

### DIFF
--- a/dockerfiles/testsuite/docker-compose.yml
+++ b/dockerfiles/testsuite/docker-compose.yml
@@ -5,6 +5,7 @@ services:
     image: "dvotenode:${TESTSUITE_BUILD_TAG:-latest}"
     build: # reused by the other services
       context: ../../
+      network: host
     env_file: "${COMPOSE_HOST_PATH:-.}/env.seed"
     volumes:
       - data-seed:/app/run/
@@ -98,6 +99,7 @@ services:
     build:
       context: ../../
       target: test
+      network: host
     links:
       - gateway
     command: "true"


p4u (commented):
@vdo can you review?

vdo:
As long as its only the build and the test images, LGTM. The ports can't conflict in that case.

Although I think it would be preferable to fix it in the Docker daemon of the CI servers instead, so we don't have the same issue with other builds, adding to daemon.json a lower MTU like:
```
{
  "mtu": 1454
}
```
@altergui what do you think?

vdo:
I actually think it's a bad idea to use the host network feature in the compose, since it could conflict with other running instances in the CI server

I applied the MTU setting to both ci servers, let me know if you still have the same issue

altergui:
sounds great, i definitely agree is better to fix it system-wide, thanks!! :heart:
i'm closing this workaround then, and will keep #425 open to report on whether it happens again after the mtu fix

PR #430, opened 2022-01-12
CI: don't use -race and -coverprofile at once

mvdan:
(see commit message)


mvdan:
`-covermode=atomic` is also unnecessary I think; it's more expensive, and I don't think it gains us anything. It is good when run with `-race`, but we're not doing that anymore.

p4u (approved):

PR #432, opened 2022-01-14
vochain/scrutinizer: sort process lists consistently

mvdan:
(see commit message)

### DIFF
--- a/vochain/app.go
+++ b/vochain/app.go
@@ -16,6 +16,7 @@ import (
 	abcitypes "github.com/tendermint/tendermint/abci/types"
 	mempl "github.com/tendermint/tendermint/mempool"
 	nm "github.com/tendermint/tendermint/node"
+	tmprototypes "github.com/tendermint/tendermint/proto/tendermint/types"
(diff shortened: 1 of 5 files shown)


mvdan:
FYI @p4u @vdo I *think* this fixes the recent sqlite "diff" errors we've been seeing.

@p4u @jordipainan I'd love a review of what I've done here to emulate tendermint's "end block, commit, start block" flow. I assume we don't want to run the real tendermint for these tests, but we also want to be somewhat realistic when it comes to timestamps and heights.

I guess we could also use a more realistic version of the app, such as vocone, but I'm not sure how easy that would be. Perhaps @NateWilliams2 has thoughts on all this, given his recent experience with vocone in integration tests.

p4u:
Yes, vocone could be used for the scrutinizier testing since it should reproduce almost the same behavior of Tendermint.

But your mock method `AdvanceTestBlock` looks good to me for an unit test.

p4u (approved):

mvdan:
Thanks for double checking! I agree what I've done here is good enough for now. If we find that we need closer emulation of Tendermint, we can look at using vocone instead.

PR #434, opened 2022-01-17
hotfix disable failing ipfssync test due to upstream issue

altergui:
ExternalIP() is taking too long due to ident.me timing out

### DIFF
--- a/ipfssync/ipfssync_test.go
+++ b/ipfssync/ipfssync_test.go
@@ -1,78 +1,83 @@
 package ipfssync
 
-import (
-	"bytes"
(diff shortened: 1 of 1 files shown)


p4u:
Umm, why the tests are failing? it does not look as something related to this PR. @mvdan ?

mvdan:
The tests just seem to be running into the 1m timeout, still.

mvdan:
We could increase the timeout to 2m or 3m, but it still seems like the ipfssync/subpub tests still take too long

p4u:
@altergui ?

PR #436, opened 2022-01-18
util: prevent PublicIP from taking tens of seconds

mvdan:
(see commit message)

Closes #434.

### DIFF
--- a/util/net.go
+++ b/util/net.go
@@ -2,7 +2,6 @@ package util
 
 import (
 	"context"
-	"fmt"
 	"math/rand"
 	"net"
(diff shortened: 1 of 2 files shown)


mvdan:
I think keeping the 1m test timeout per package is useful, precisely because it will help us keep our tests reasonably fast and catch long timeouts or sleeps like this one :)

altergui (approved):
looks great to me

NateWilliams2:
looks great 👍

NateWilliams2 (approved):

PR #439, opened 2022-01-24
ipfssync: don't use util.PublicIP() anymore, IPFS knows its own address

altergui:
IPFS node already keeps track of its own addresses, thanks to libp2p PEX
these can be scraped from PeerHost.Addrs() (which is populated passively)
so use that, instead of our own util.PublicIP()

also, remove unused MultiAddrIPv{4,6} from subpub,
(which again called util.PublicIP(), totally unnecessarily)

vochain.go still needs util.PublicIP() because tendermint doesn't support
autodiscovering its own IP (it doesn't even support dualstack)

### DIFF
--- a/ipfssync/ipfssync.go
+++ b/ipfssync/ipfssync.go
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"net"
 	"os"
 	"path"
 	"strings"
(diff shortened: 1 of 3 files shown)


altergui:
also, this makes the startup much simpler and shaves a couple of seconds in tests
https://github.com/vocdoni/vocdoni-node/runs/4925686296?check_suite_focus=true#step:4:322
```
ok  	go.vocdoni.io/dvote/ipfssync	2.917s	coverage: 25.9% of statements
ok  	go.vocdoni.io/dvote/ipfssync/subpub	0.541s	coverage: 45.1% of statements
```

in contrast to waiting for PublicIP() (even now that it's not timing out)
https://github.com/vocdoni/vocdoni-node/runs/4922253489?check_suite_focus=true
```
ok  	go.vocdoni.io/dvote/ipfssync	4.386s
ok  	go.vocdoni.io/dvote/ipfssync/subpub	2.827s
```

p4u (approved):
great

PR #443, opened 2022-01-26
run staticcheck on CI in a more obvious way

mvdan:
(see commit message)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -16,16 +16,19 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v2
-      - name: Go test
+      - name: gofmt
+        # Run gofmt first, as it's quick and issues are common.
+        run: diff -u <(echo -n) <(gofmt -s -d $(git ls-files '*.go'))
+      - name: go test
         run: |
           # we run vet in another step
           go test -vet=off -timeout=1m -coverprofile=covprofile ./...
           # -race can easily make the crypto stuff 10x slower
           go test -vet=off -timeout=15m -race ./...
-      - name: Go analyze
+      - name: go vet
+        run: go vet ./...
+      - name: staticcheck
         run: |
-          diff -u <(echo -n) <(gofmt -s -d $(git ls-files '*.go'))
-          go vet ./...
           curl -L https://github.com/dominikh/go-tools/releases/download/2021.1.2/staticcheck_linux_amd64.tar.gz | tar -xzf -
           ./staticcheck/staticcheck ./...
       - name: Go build for Mac


mvdan:
It happened! Staticcheck isn't running. Please do not merge.

altergui (approved):
wonderful! thanks for the fixes

altergui:
mmmh, i've noticed that in https://github.com/vocdoni/vocdoni-node/pull/429
the checks are being run twice, once for "pull_request" and once for "push"
the one from `pull_request` runs staticcheck correctly (and fails, as expected)
the one from `push` staticcheck fails silently (with `warning: "./..." matched no packages`)

here in https://github.com/vocdoni/vocdoni-node/pull/443
only `pull_request` is run, i believe because the branch is under @mvdan fork (instead of being a branch of github.com/vocdoni)

indeed, i just checked a couple of github actions from `push`, and in them staticcheck always fails silently with `warning: "./..." matched no packages`

vdo:
> here in #443 only `pull_request` is run, i believe because the branch is under @mvdan fork (instead of being a branch of github.com/vocdoni)
> 
That would make sense, thanks @altergui `

mvdan:
That's an interesting observation, but I don't think the kind of event is the cause. We have the warning happening on both kinds of CI jobs: push and pull_request. It seems somewhat probabilistic, as far as I can tell.

mvdan:
Got an answer from Dominik, the author of staticcheck; he says staticcheck should indeed fail if it has nothing to do, as that's certainly wrong. That would have made us catch the bug much sooner.

He also mentions we can use `GOPACKAGESDEBUG=1` as a way to get a bit more debugging info on what staticcheck is doing to load packages. I'll try that tomorrow.

altergui:
with GOPACKAGESDEBUG=1 here's what staticcheck is doing
https://github.com/vocdoni/vocdoni-node/runs/4969013309?check_suite_focus=true#step:7:448
```
2022/01/27 16:09:09 5.769271ms for GOROOT= GOPATH=/go GO111MODULE=off GOPROXY= *** go list -e -f {{context.ReleaseTags}}
2022/01/27 16:09:09 7.492213ms for GOROOT= GOPATH=/go GO111MODULE= GOPROXY= *** go list -f "{{context.GOARCH}} {{context.Compiler}}" -- unsafe
2022/01/27 16:09:09 397.389771ms for GOROOT= GOPATH=/go GO111MODULE= GOPROXY= *** go list -e -json -compiled=true -test=true -export=true -deps=true -find=false -- ./...
warning: "./..." matched no packages
```

i tried running those same commands, the third one fails
https://github.com/vocdoni/vocdoni-node/runs/4969013309?check_suite_focus=true#step:7:452
```
raw cmds: go list -e -f {{context.ReleaseTags}}
[go1.1 go1.2 go1.3 go1.4 go1.5 go1.6 go1.7 go1.8 go1.9 go1.10 go1.11 go1.12 go1.13 go1.14 go1.15 go1.16 go1.17]
raw cmds: go list -f {{context.GOARCH}} {{context.Compiler}} -- unsafe
amd64 gc
raw cmds: go list -e -json -compiled=true -test=true -export=true -deps=true -find=false -- ./...
open /github/home/.cache/go-build/fc/fc0c834f7738ff86f05ec60adbd5736130df7c72423ddd801e9349680096154d-d: permission denied
```

a permissions issue?
```
drwxr-xr-x 4 root root 4096 Jan 27 16:11 /github/
drwxr-xr-x 6 1000 sudo 4096 Sep  8 14:10 /github/home/
drwxr-xr-x 6 root root 4096 Sep  8 14:10 /github/home/.cache/
drwxr-xr-x 258 root root 4096 Aug 31 13:01 /github/home/.cache/go-build/
drwxr-xr-x 2 root root 12288 Aug 31 13:02 /github/home/.cache/go-build/fc/
```
```
ls -l /github/home/.cache/go-build/fc/
total 12
-rw-r--r-- 1 root root 175 Aug 31 14:08 fc35a45292933ea25adeb5aaa850e25c734ba17e33fff0c13ff04c7207bfdbf2-a
-rw-r--r-- 1 root root  12 Aug 31 14:08 fc40abd0343f9a80c5b3dab549574858d22f9d70262ed2d3a847fdb54e3e303e-d
-rw-r--r-- 1 root root 175 Aug 31 14:08 fc8d850f33dc102eba02ef9faa9825bc312bca463d9d5752ac4f497e7588c67d-a
```
`whoami` reports being user `1000`

since this SOMETIMES works, i'm absolutely confused. but does any of this make any sense at all to you @vdo?

i could only understand this if there are many github runners, and some of them have `/github/home/.cache/` chown'd root (and fail), and others are chown'd 1000 (and work for the lucky PR)

altergui:
fixed it! using a PR run https://github.com/vocdoni/vocdoni-node/pull/449/commits/9489baa452b8c8378ff3f62d4db59a501ce058ee
i `mv`'d the root-chowned .cache directory out of the way (tried to `rm -rf` first but didn't have permission in the inner directories. but since i own the home, i could rename it :sunglasses:)
https://github.com/vocdoni/vocdoni-node/runs/4969405231?check_suite_focus=true

mvdan:
Aaaah, so it *is* related to the CI setup after all. @vdo FYI

altergui:
and i believe i understood why it SOMETIMES works, even in the same host: the  `/github/home/.cache/go-build/` has many subdirs (named ab, fc, ek, ...) and i guess SOME of them are owned by 1000, so when it hit them, staticcheck worked?

mvdan:
Yeah that sounds right. I think the GOCACHE somehow got corrupted with files that UID 1000 can't read or write. Perhaps a run happened with the root UID.

mvdan:
In terms of fixes to prevent this from happening ever again: I guess we could make CI use a root command to apply a recursive chmod on GOCACHE and GOMODCACHE so that it's all owned by UID 1000, but that seems kinda expensive. Walking the entire cached directory tree could take multiple seconds.

Is there a way to force docker to mount host directories in the container so that they are readable and writable by everyone, as if it was all chmod 0777 all the time? That would probably be a lot cheaper, and then we don't have to worry about UID problems ever again.

altergui:
i was thinking along the lines of
`time if [ -n "$(find ~/.cache/go-build/ -not -user 1000)" ] ; then mv ~/.cache ~/.cache-broken-$(date +%s) ; fi`

quite drastic but cheap

mvdan:
That still walks the entire cache, so I don't think it would be much cheaper than chmod. We want this to not add any cost to CI runs when the caches are in a good state, which will hopefully be the common scenario.

altergui:
true, i meant cheap in the sense that in case it's broken, a single mv is cheaper than chowning a bunch of files
but pretty shortsighted since rebuilding the whole cache afterwards is definitely more expensive than the sudo chown :facepalm:

mvdan:
Also that :)

mvdan:
Went full circle with Dominik: he confirms that the reason staticcheck wasn't failing is because cmd/go does not properly surface the "permission denied" error, and simply prints it as a warning. So the two causes that led us to not notice this permission problem (staticcheck willing to do nothing, cmd/go hiding GOCACHE errors) should be fixed in the next couple of months.

mvdan:
Upstream cmd/go issue is now live: https://github.com/golang/go/issues/50869

For now will wait on @vdo to advise how best to prevent permission issues in the future.

mvdan:
@vdo this should be ready to review. The way to check that staticcheck really is working is that:

* the "no packages" warning isn't shown
* the "staticcheck" step takes at least 5-10 seconds; if it takes <1s, it's quickly running into the permission error and doing nothing

p4u:
should we merge that? @vdo

mvdan:
For full context:

* Upstream Go will fix the underlying bug that caused us to not see the permission errors.
* With vdo we agreed to not start using root for the tests just yet, with the assumption that the permission error was a one time error that shouldn't happen again.

PR #445, opened 2022-01-26
ipfssync: lint double import (reported by staticcheck)

altergui:
this was introduced in
a03e171f "ipfssync: don't use util.PublicIP() anymore, IPFS knows its own address"

sorry about that :disappointed:


mvdan:
Please see https://github.com/vocdoni/vocdoni-node/pull/443 :) No need for this PR as well, I think.

altergui:
:partying_face:

PR #446, opened 2022-01-26
dvotecli: fix unset treasurer

### LINKED ISSUE #424: bug: `dvotecli genesis` generates a genesis with empty treasurer
since this commit
https://github.com/vocdoni/vocdoni-node/commit/187c471a33d4f22237e0a9d23dfca9ae51862523#diff-088e50f432c8393ca86f7bb7c1054fbd1867eef3d5e25e8773d0dbeabf7bc8adR354
dvotenode will not start if the `genesis.json` doesn't have a valid address in `treasurer`.

the hardcoded genesis.json in `vochain/genesis.go` does specify a hardcoded value (`0x309Bd6959bf4289CDf9c7198cF9f4494e0244b7d`)
but the `dvotecli` produces a genesis with `"treasurer": ""`
and if this `genesis.json` is used, `dvotenode` fails to start

**To Reproduce**

- Steps to reproduce the behavior:
  1. generate a genesis.json using
  ``` 
  dvotecli genesis \
          --chainId {{chain_id}} \
          --seeds {{seeds}} \
          --oracles {{oracles}} \
          --miners {{miners}}
  ```
  2. copy-paste the relevant output and save it as `genesis.json`  
  3. run `dvotenode` with that `genesis.json`
  4. `dvotenode` dies with `treasurer should be a valid Ethereum address: %!s(<nil>)`

i imagine two solutions:
* either set a default value for `treasurer` (maybe `0x0000000000000000000000000000000000000000`?) in `cmd/dvotecli/commands/genesis.go` (current default is "")

* or make `dvotenode` (`vochain/app.go`) continue gracefully in case `treasurer` equals the empty string

but i can't decide which one is the reasonable, since i'm totally lost on what a treasurer is.

jordipainan:
Fixes #424

### DIFF
--- a/cmd/dvotecli/commands/genesis.go
+++ b/cmd/dvotecli/commands/genesis.go
@@ -84,10 +84,33 @@ func genesisGen(cmd *cobra.Command, args []string) error {
 		Block:     vochain.BlockParams(tmConsensusParams.Block),
 		Validator: vochain.ValidatorParams(tmConsensusParams.Validator),
 	}
-	treasurer, _ := cmd.Flags().GetString("treasurer")
+
+	// Get treasurer
+	treasurer, err := cmd.Flags().GetString("treasurer")
+	if err != nil {
+		return err
+	}
+	t := &ethereum.SignKeys{}
+	if treasurer == "" {
+		// generate new treasurer
+		t = ethereum.NewSignKeys()
+		if err := t.Generate(); err != nil {
+			return err
+		}
+	} else {
+		if err := t.AddHexKey(treasurer); err != nil {
+			return err
+		}
+	}
+	prettyHeader("Treasurer")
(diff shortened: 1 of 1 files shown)


altergui (approved):
LGTM! i guess ideally (instead of failing) it would generate a priv/pub keypair and print it, like it does with the rest of the nodes. but that can be implemented later, along with incorporating this new type of node in testsuite (i'd happy to do that), and so on.

jordipainan:
> LGTM! i guess ideally (instead of failing) it would generate a priv/pub keypair and print it, like it does with the rest of the nodes. but that can be implemented later, along with incorporating this new type of node in testsuite (i'd happy to do that), and so on.

Yup, right! Changed for generate a key and to not fail

jordipainan:
Btw, remains another PR for adding transaction costs :) Going to do it in a while

PR #447, opened 2022-01-26
all: staticcheck fixes while we figure out CI issues

mvdan:
(see commit message)


mvdan:
I've spent hours on https://github.com/vocdoni/vocdoni-node/pull/443 and still haven't found why staticcheck isn't running, so for now, I've split the fixes here.

altergui (approved):
:+1:

Issue #399, opened 2021-12-13
module dependency cycle with arbo prevents downgrading bazil.org/fuse


maektwain:
**Describe the bug**
Running on mac specifically build leads to fuse error of undefined values

**To Reproduce (please complete the following information)**

run on mac os X  

`` go build ./cmd/dvotecli   ``

**Current behavior**
```
# bazil.org/fuse
../../go/pkg/mod/bazil.org/fuse@v0.0.0-20200524192727-fb710f7dfd05/error_std.go:26:20: undefined: errNoXattr
../../go/pkg/mod/bazil.org/fuse@v0.0.0-20200524192727-fb710f7dfd05/fuse.go:171:12: undefined: mount
../../go/pkg/mod/bazil.org/fuse@v0.0.0-20200524192727-fb710f7dfd05/fuse.go:236:24: undefined: maxWrite
../../go/pkg/mod/bazil.org/fuse@v0.0.0-20200524192727-fb710f7dfd05/fuse.go:425:32: undefined: maxWrite
../../go/pkg/mod/bazil.org/fuse@v0.0.0-20200524192727-fb710f7dfd05/fuse.go:810:12: undefined: openFlags
../../go/pkg/mod/bazil.org/fuse@v0.0.0-20200524192727-fb710f7dfd05/fuse.go:828:18: undefined: openFlags
../../go/pkg/mod/bazil.org/fuse@v0.0.0-20200524192727-fb710f7dfd05/fuse.go:845:18: undefined: openFlags
../../go/pkg/mod/bazil.org/fuse@v0.0.0-20200524192727-fb710f7dfd05/fuse.go:868:18: undefined: openFlags
../../go/pkg/mod/bazil.org/fuse@v0.0.0-20200524192727-fb710f7dfd05/fuse.go:991:12: undefined: openFlags
../../go/pkg/mod/bazil.org/fuse@v0.0.0-20200524192727-fb710f7dfd05/fuse.go:1413:20: undefined: maxWrite
../../go/pkg/mod/bazil.org/fuse@v0.0.0-20200524192727-fb710f7dfd05/fuse.go:1413:20: too many errors

```
**Expected behavior**
It should build 

**System (please complete the following information):**

- OS: Mac OS X 11.5.2
- Software version
- Commit hash 

**Additional context**

mvdan:
Unfortunately, this is a bug caused by upstreams - ipfs and fuse. See https://github.com/ipfs/go-ipfs/issues/7889.

[...]

PR #448, opened 2022-01-26
fix build with Mac yet again, lock it in via CI

mvdan:
(see commit message)

Updates #399.


mvdan:
FYI @NateWilliams2, this is for Mac builds once again. Hopefully we won't be here ever again, as I've added a CI check.

NateWilliams2 (approved):
looks great, thanks!

PR #455, opened 2022-01-31
Fix/tendermint0.35

p4u:
This PR adds some fixes and improvements done during the benchmarking session of the tendermint-0.35 based vochain

### DIFF
--- a/vochain/start.go
+++ b/vochain/start.go
@@ -306,7 +306,7 @@ func newTendermint(app *BaseApplication,
 
 	// We need to fetch chain_id in order to make Replay work,
 	// since signatures depend on it.
-	log.Warnf("genesis file at %s", tconfig.GenesisFile())
+	log.Infof("genesis file at %s", tconfig.GenesisFile())
 	type genesisChainID struct {
 		ChainID string `json:"chain_id"`
 	}
--- a/vochain/start.go
+++ b/vochain/start.go
@@ -182,8 +182,8 @@ func newTendermint(app *BaseApplication,
 	tconfig.BlockSync.Enable = true
 	tconfig.StateSync.Enable = false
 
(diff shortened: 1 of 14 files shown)


vochain/start.go
```diff
@@ -309,6 +304,23 @@ func newTendermint(app *BaseApplication,
 		}
 	}
 
+	// We need to fetch chain_id in order to make Replay work,
+	// since signatures depend on it.
+	log.Warnf("genesis file at %s", tconfig.GenesisFile())
```
altergui: shouldn't this be Info or Debug? (i imagine this being a leftover from debugging)

vochain/vochaininfo/vochaininfo.go
```diff
@@ -128,21 +128,21 @@ func (vi *VochainInfo) VoteCacheSize() int {
...
 	if vi.vnode.Node != nil {
 		return vi.vnode.Node.NetInfo(context.Background())
 	}
 	return nil, fmt.Errorf("vochain node not ready")
 }
 
 // Peers returns the whole list of peers, including ID and URL
-func (vi *VochainInfo) Peers() []coretypes.Peer {
-	ni, err := vi.netInfo()
+func (vi *VochainInfo) Peers() int {
```
altergui: maybe call it NPeers then? to make it consistent with what's return'ed

altergui (commented):
i didn't review the oracle changes since they are out of my current scope of understanding but i added a couple comments

vochain/start.go
p4u: right

vochain/vochaininfo/vochaininfo.go
p4u: yep

PR #458, opened 2022-02-01
new tool to generate a custom-sized testnet (using ansible-playbook)

altergui:
> while trying to understand all the different components of the testsuite, i quickly wrote a small ansible playbook to flexibly recreate the files used by testsuite (docker-compose.yml, genesis.json and env files).

this ansible playbook turned out pretty useful to simulate different networks and helped me greatly into debugging `tendermint-0.35` branch.

definitely not the most traditional use of ansible, but i feel at home with jinja2 templating so for me it was the quickest way to solve this


p4u:
Pretty cool! Can you review @vdo ?

cmd/dvotecli/commands/genesis.go
```diff
@@ -125,12 +140,48 @@ func genesisGen(cmd *cobra.Command, args []string) error {
...
+	}
+
+	fmt.Println()
+
+	if file, _ := cmd.Flags().GetString("file"); file != "" {
+		if err := os.WriteFile(file, data.Bytes(), 0o600); err != nil {
+			return err
+		}
+	} else {
+		prettyHeader("Genesis JSON")
```
vdo: Maybe we can avoid this header so the raw output JSON can be piped to tools like jq or ijq

vdo:
Nice combination of ansible and docker compose :+1:

p4u:
Can we automatize this testing somehow? Maybe a server running the test on each commit?

p4u (approved):

p4u:
@altergui please address the comments and click "resolved" whenever is done :)

p4u:
conflicts should be fixed :)

cmd/dvotecli/commands/genesis.go
altergui: nice one! done

altergui:
> Can we automatize this testing somehow? Maybe a server running the test on each commit?

i thought about this but i want to avoid making the tests run even longer. I'll try a couple of alternatives and come up with a solution in another PR

PR #459, opened 2022-02-01
WIP: Enables the ethereum-less approach

jordipainan:
See: https://www.notion.so/Ethereum-less-vochain-description-67780a73a3e54be5892ed2a2a145b7c7

Changes:
- Adds support for: `MintTokensTx`, `SetAccountDelegate`, `SendTokens`, `CollectFaucetTx` & `SetTransactionCostsTx`.
- Adds support for transaction costs on current and new defined transactions as well as on the genesis.json and `InitChain`.
- Delegates can change account infoURI, create processes and set the process status on behalf of an account, make the proper changes to support this behavior on `NewProcessTx` and `SetProcessStatus`.
- Adds unit and integration tests for the newly defined transactions.
- Refactors current unit and integration tests for making them work with the ethereum-less approach.
- Adapt the remaining code to the new signing schema.
- Adds API Client methods for the new transactions.

### DIFF
--- a/client/api.go
+++ b/client/api.go
@@ -1,27 +1,17 @@
 package client
 
 import (
-	"context"
-	"crypto/sha256"
-	"encoding/binary"
 	"encoding/hex"
-	"encoding/json"
 	"fmt"
-	"math"
 	"math/big"
 	"math/rand"
-	"os/exec"
-	"strings"
-	"sync"
 	"time"
 
 	"github.com/ethereum/go-ethereum/common"
-	"github.com/vocdoni/arbo"
 	"go.vocdoni.io/dvote/api"
 	"go.vocdoni.io/dvote/crypto/ethereum"
 	"go.vocdoni.io/dvote/crypto/zk/artifacts"
 	"go.vocdoni.io/dvote/log"
-	"go.vocdoni.io/dvote/types"
 	"go.vocdoni.io/dvote/util"
 	"go.vocdoni.io/dvote/vochain"
 	"go.vocdoni.io/dvote/vochain/scrutinizer/indexertypes"
@@ -34,6 +24,7 @@ type pkeys struct {
 	priv []api.Key
 }
 
+// GetChainID returns the chainID of the Vochain
 func (c *Client) GetChainID() (string, error) {
 	var req api.APIrequest
 	req.Method = "getInfo"
@@ -47,6 +38,7 @@ func (c *Client) GetChainID() (string, error) {
 	return resp.ChainID, nil
 }
 
+// GetProcessInfo returns useful process information and parameters
 func (c *Client) GetProcessInfo(pid []byte) (*indexertypes.Process, error) {
 	var req api.APIrequest
 	req.Method = "getProcessInfo"
@@ -61,6 +53,7 @@ func (c *Client) GetProcessInfo(pid []byte) (*indexertypes.Process, error) {
 	return resp.Process, nil
 }
 
+// GetEnvelopeStatus returns true if the vote envelope is registered
 func (c *Client) GetEnvelopeStatus(nullifier, pid []byte) (bool, error) {
 	var req api.APIrequest
 	req.Method = "getEnvelopeStatus"
@@ -75,6 +68,7 @@ func (c *Client) GetEnvelopeStatus(nullifier, pid []byte) (bool, error) {
 	return *resp.Registered, nil
 }
 
+// GetProof tries to generate a return a census proof
 func (c *Client) GetProof(pubkey, root []byte, digested bool) ([]byte, []byte, error) {
 	var req api.APIrequest
 	req.Method = "genProof"
@@ -93,6 +87,7 @@ func (c *Client) GetProof(pubkey, root []byte, digested bool) ([]byte, []byte, e
 	return resp.Siblings, resp.CensusValue, nil
 }
 
+// GetResults gets the results of a given process from the scrutinizer
 func (c *Client) GetResults(pid []byte) ([][]string, string, bool, error) {
 	var req api.APIrequest
 	req.Method = "getResults"
@@ -110,6 +105,7 @@ func (c *Client) GetResults(pid []byte) ([][]string, string, bool, error) {
 	return resp.Results, resp.State, *resp.Final, nil
 }
 
+// GetEnvelopeHeight returns the height at which the envelope was registered
 func (c *Client) GetEnvelopeHeight(pid []byte) (uint32, error) {
 	var req api.APIrequest
 	req.Method = "getEnvelopeHeight"
@@ -124,6 +120,7 @@ func (c *Client) GetEnvelopeHeight(pid []byte) (uint32, error) {
 	return *resp.Height, nil
 }
 
+// CensusSize returns the size of a census as int64
 func (c *Client) CensusSize(cid []byte) (int64, error) {
 	var req api.APIrequest
 	req.Method = "getSize"
@@ -138,6 +135,7 @@ func (c *Client) CensusSize(cid []byte) (int64, error) {
 	return *resp.Size, nil
 }
 
+// ImportCensus adds, imports and publish a given census
 func (c *Client) ImportCensus(signer *ethereum.SignKeys, uri string) ([]byte, error) {
 	var req api.APIrequest
 	req.Method = "addCensus"
@@ -170,6 +168,7 @@ func (c *Client) ImportCensus(signer *ethereum.SignKeys, uri string) ([]byte, er
 	return resp.Root, nil
 }
 
+// GetKeys gets the process encryption keys
 func (c *Client) GetKeys(pid, eid []byte) (*pkeys, error) {
 	var req api.APIrequest
 	req.Method = "getProcessKeys"
@@ -188,6 +187,7 @@ func (c *Client) GetKeys(pid, eid []byte) (*pkeys, error) {
 	}, nil
 }
 
+// GetCircuitConfig gets the ZK circuit config used for anonymous voting
 func (c *Client) GetCircuitConfig(pid []byte) (*int, *artifacts.CircuitConfig, error) {
 	var req api.APIrequest
 	req.Method = "getProcessCircuitConfig"
@@ -202,6 +202,7 @@ func (c *Client) GetCircuitConfig(pid []byte) (*int, *artifacts.CircuitConfig, e
 	return resp.CircuitIndex, resp.CircuitConfig, nil
 }
 
+// GetRollingCensusSize returns the size of a given rolling census
 func (c *Client) GetRollingCensusSize(pid []byte) (int64, error) {
 	var req api.APIrequest
 	req.Method = "getProcessRollingCensusSize"
@@ -216,6 +217,7 @@ func (c *Client) GetRollingCensusSize(pid []byte) (int64, error) {
 	return *resp.Size, nil
 }
 
+// GetRollingCensusVoterWeight returns the weight in the census of a specific address
 func (c *Client) GetRollingCensusVoterWeight(pid []byte, address common.Address) (*big.Int, error) {
 	var req api.APIrequest
 	req.Method = "getPreregisterVoterWeight"
@@ -231,6 +233,7 @@ func (c *Client) GetRollingCensusVoterWeight(pid []byte, address common.Address)
 	return resp.Weight.ToInt(), nil
 }
 
+// TestResults checks the correctness of a given process results
 func (c *Client) TestResults(pid []byte, totalVotes int, withWeight uint64) ([][]string, error) {
 	log.Infof("waiting for results...")
 	var err error
@@ -262,11 +265,13 @@ func (c *Client) TestResults(pid []byte, totalVotes int, withWeight uint64) ([][
 	return results, nil
 }
 
+// Proof wraps a merkle proof siblings and value
 type Proof struct {
 	Siblings []byte
 	Value    []byte
 }
 
+// GetMerkleProofBatch returns an array of merkle proofs for a set of given addresses
 func (c *Client) GetMerkleProofBatch(signers []*ethereum.SignKeys,
 	root []byte,
 	tolerateError bool) ([]*Proof, error) {
@@ -290,6 +295,7 @@ func (c *Client) GetMerkleProofBatch(signers []*ethereum.SignKeys,
 	return proofs, nil
 }
 
+// GetMerkleProofPoseidonBatch returns an array of poseidon merkle proofs for a set of given addresses
 func (c *Client) GetMerkleProofPoseidonBatch(signers []*ethereum.SignKeys,
 	root []byte,
 	tolerateError bool) ([]*Proof, error) {
@@ -314,6 +320,7 @@ func (c *Client) GetMerkleProofPoseidonBatch(signers []*ethereum.SignKeys,
 	return proofs, nil
 }
 
+// GetCSPproofBatch returns an array of CSP like merkle proofs for a set of given addresses
 func (c *Client) GetCSPproofBatch(signers []*ethereum.SignKeys,
 	ca *ethereum.SignKeys,
 	pid []byte) ([]*Proof, error) {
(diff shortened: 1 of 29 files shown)


client/api.go
```diff
@@ -1050,7 +1051,277 @@ func (c *Client) TestSendAnonVotes(
 	return votingElapsedTime, nil
 }
 
-func (c *Client) CreateProcess(oracle *ethereum.SignKeys,
+func (c *Client) CreateAccount(signer *ethereum.SignKeys, infoURI string, nonce uint32) error {
```
p4u: Please document at least all exported Methods and Types :)

client/api.go
```diff
@@ -1050,7 +1051,277 @@ func (c *Client) TestSendAnonVotes(
...
+		return fmt.Errorf("%s failed: %s", req.Method, resp.Message)
+	}
+	return nil
+}
+
+func (c *Client) SendTokens(from *ethereum.SignKeys, to common.Address, value uint64, nonce uint32) error {
+	var req api.APIrequest
+	var err error
+	req.Method = "submitRawTx"
+
```
p4u, line 489: There is some repeated code here. You could move most of the code to a new method such as `c.SignAndSendTx(tx models.Tx)`

client/api.go
jordipainan: yep I did this for all methods in this branch, I can add it on top of this PR, also I simplified the CreateAccount / SetAccountInfoURI and I merged both functions. I just wanted to not add more stuff. Here is the branch with other changes I found useful https://github.com/vocdoni/vocdoni-node/tree/feature/improve_client

NateWilliams2:
As @p4u and I discussed, it would be very helpful to the vaas-api if this could We could merge the `setAccountInfo` and `collectFaucet` transactions into a new transaction (or add the faucet functionality to `setAccountInfo` so that an account can be created from scratch with a single transaction. Otherwise we need to first request the tokens, wait to confirm this transaction has been mined, and then create the account itself. This is difficult to implement cleanly on the API side and degrades the usability of the integrator API. @jordipainan would this be possible?

jordipainan:
IDK why compose-test is failing here, I assume is a CI problem, I executed the tests several times locally and they did not throw any error. 

Going to add the collect faucet functionality upon account creation and we will figure out this later.

jordipainan:
@NateWilliams2 added here: d7bee008f3af0dd75999212f3567548095507f31
Unfortunately tests not running atm, will continue asap

vochain/process.go
```diff
@@ -335,7 +360,7 @@ func (v *State) GetProcessResults(pid []byte) ([][]string, error) {
 }
 
 // SetProcessCensus sets the census for a given process, only if that process enables dynamic census
-func (v *State) SetProcessCensus(pid, censusRoot []byte, censusURI string, commit bool) error {
+func (v *State) SetProcessCensus(pid, censusRoot []byte, censusURI string, txSender common.Address, commit bool) error {
```
p4u, line 363: The state methods which are not "tx" related (such as SetProcessCensus, SetProcessStatus, AddVote, etc...) should be agnostic to the transaction based checks.

The only checks that these methods shall do are State related, i.e the process exist? But not the cost or other transaction related stuff. These checks go to specific-scope check transaction methods such as `SetProcessCensusTxCheck()`

p4u: This comment applies to all other methods.

vochain/account.go
```diff
@@ -42,15 +44,15 @@ func (a *Account) Transfer(dest *Account, amount uint64, nonce uint32) error {
...
+	if (a.Balance + cost) < amount {
 		return ErrNotEnoughBalance
 	}
 	if dest.Balance+amount <= dest.Balance {
 		return ErrBalanceOverflow
 	}
 	dest.Balance += amount
-	a.Balance -= amount
-	a.Nonce++
+	a.Balance -= amount + cost
```
p4u, line 55: The spec design said the cost of a transaction should be sent to 0xFFF... burn address.

That will be useful to have within the state always the number of tokens that has been used. Then we can use this number to mint more tokens and keep always the same liquid.

So I'd not add the cost here, I'd just let this method to transfer among amounts. The caller function (tx related) should be the one checking the balance+cost and making to transfers, one to 0xFFF and the other to the destination account.

This way in some moment we can also just switch 0xFFF by the validator address if we want to reward it.

vochain/account.go
```diff
@@ -74,64 +76,101 @@ func (a *Account) AddDelegate(addr common.Address) error {
...
-		}
+	return nil
+}
+
+// CollectFaucet transfers balance from faucet generated package address to collector address,
+// and updates the state with the new values (including nonce).
+func (v *State) CollectFaucet(from, to common.Address, amount uint64, nonce uint64) error {
+	var accFrom, accTo *Account
+	var err error
+	if accFrom, err = v.GetAccount(from, false); err != nil || accFrom == nil {
```
p4u, line 130: the accFrom nil check is redundant here

vochain/account.go
```diff
@@ -74,64 +76,101 @@ func (a *Account) AddDelegate(addr common.Address) error {
...
+	if accFrom == nil {
+		return ErrAccountNotFound
+	}
+	if accTo, err = v.GetAccount(to, false); err != nil {
+		return err
+	}
+	if accTo == nil {
+		return ErrAccountNotFound
+	}
+	if amount == 0 {
```
p4u, line 142: The basic, non expensive checks, should be placed on the begining of the function

vochain/account.go
```diff
@@ -175,108 +214,502 @@ func (v *State) GetAccount(address common.Address, isQuery bool) (*Account, erro
 	return &acc, acc.Unmarshal(raw)
 }
 
-// VerifyAccountBalance extracts an account address from a signed message, and verifies if
-// there is enough balance to cover an amount expense
-func (v *State) VerifyAccountBalance(message, signature []byte, amount uint64) (bool, common.Address, error) {
-	address, err := ethereum.AddrFromSignature(message, signature)
+// AccountFromSignature extracts an address from a signed message and returns an account if exists
+func (v *State) AccountFromSignature(message, signature []byte) (common.Address, *Account, error) {
```
p4u, line 218: I think common.Address should be a pointer here since *Account is a pointer too

vochain/account.go
```diff
@@ -175,108 +214,502 @@ func (v *State) GetAccount(address common.Address, isQuery bool) (*Account, erro
...
+	pubKey, err := ethereum.PubKeyFromSignature(message, signature)
+	if err != nil {
+		return common.Address{}, nil, fmt.Errorf("cannot extract public key from signature: %w", err)
+	}
+	address, err := ethereum.AddrFromPublicKey(pubKey)
 	if err != nil {
-		return false, address, err
+		return common.Address{}, nil, fmt.Errorf("cannot extract address from public key: %w", err)
+	}
+	if address == types.EthereumZeroAddressBytes {
```
p4u, line 227: Can you explain this check?

vochain/account.go
```diff
@@ -175,108 +214,502 @@ func (v *State) GetAccount(address common.Address, isQuery bool) (*Account, erro
...
+		return common.Address{}, nil, ErrAccountNotFound
 	}
-	return acc.Balance >= amount, address, nil
+	return address, acc, nil
 }
 
-func (v *State) SetAccountInfoURI(address common.Address, infoURI string) error {
-	if infoURI == "" {
-		return fmt.Errorf("cannot set an empty URI")
+func (v *State) SetAccountInfoURI(accountAddress, txSender common.Address, infoURI string) error {
```
p4u, line 241: Document me please

vochain/account.go
```diff
@@ -175,108 +214,502 @@ func (v *State) GetAccount(address common.Address, isQuery bool) (*Account, erro
...
 	}
-	var acc Account
-	v.Tx.Lock()
-	defer v.Tx.Unlock()
-	raw, err := v.Tx.DeepGet(address.Bytes(), AccountsCfg)
-	if err != nil && !errors.Is(err, arbo.ErrKeyNotFound) {
+	if acc == nil {
+		return ErrAccountNotFound
+	}
+	SetAccountInfoCost, err := v.TxCost(models.TxType_SET_ACCOUNT_INFO, false)
```
p4u, line 249: This check should not be here

vochain/account.go
```diff
@@ -175,108 +214,502 @@ func (v *State) GetAccount(address common.Address, isQuery bool) (*Account, erro
...
+	if err != nil {
 		return err
-	} else if err == nil {
-		if err := acc.Unmarshal(raw); err != nil {
+	}
+	if accountAddress == txSender {
```
p4u, line 253: I don't understand this logic. If txSender is not accountAddress, what is happening?

vochain/account.go
```diff
@@ -175,108 +214,502 @@ func (v *State) GetAccount(address common.Address, isQuery bool) (*Account, erro
...
+		return v.Tx.DeepSet(accountAddress.Bytes(), accBytes, AccountsCfg)
+	}
+	sender, err := v.GetAccount(txSender, false)
+	if err != nil {
+		return err
+	}
+	if sender == nil {
+		return ErrAccountNotFound
+	}
+	sender.Balance -= SetAccountInfoCost
```
p4u, line 272: I think all these balance/account operations should be perform with TransferBalance() and send it to 0xFFFF

Also I'm not sure the balance should be handled at this level... it feels to me that the transfers should stay in some other TX related layer

vochain/account.go
```diff
@@ -175,108 +214,502 @@ func (v *State) GetAccount(address common.Address, isQuery bool) (*Account, erro
...
-	newAccount.InfoURI = infoURI
+	acc.InfoURI = infoURI
 	if len(delegates) > 0 {
-		newAccount.DelegateAddrs = make([][]byte, len(delegates))
-		for i := 0; i < len(delegates); i++ {
-			if delegates[i].String() != types.EthereumZeroAddress {
-				newAccount.DelegateAddrs = append(newAccount.DelegateAddrs, delegates[i].Bytes())
+		acc.DelegateAddrs = make([][]byte, len(delegates))
+		for _, v := range delegates {
+			if v.String() != types.EthereumZeroAddressString {
```
p4u, line 312: Why these checks with EthereumZeroAddressStrong ?

vochain/account.go
```diff
@@ -175,108 +214,502 @@ func (v *State) GetAccount(address common.Address, isQuery bool) (*Account, erro
...
-	newAccount.Balance = initBalance
-	v.Tx.Lock()
-	defer v.Tx.Unlock()
-	accBytes, err := newAccount.Marshal()
-	if err != nil {
-		return err
+	acc.Balance = initBalance
+
+	// handle faucet claim
+	if !bytes.Equal(faucetSender.Bytes(), types.EthereumZeroAddressBytes[:]) {
```
p4u, line 320: I think faucet address should be a pointer and check if its nil or not. But in any case I feel like this faucet stuff should not be here. Once thing is create the Account and the other is fund it with faucet payload.

p4u (changes requested):

p4u:
This is a big and important PR, the more eyes the better. Please @randomshinichi @mvdan @NateWilliams2 @altergui @emmdim review those parts of the code where you feel comfortable, if you have the time.

PS: if you can suggest instead of just coment, that will help Jordi :)

vochain/transaction.go
```diff
@@ -548,6 +629,46 @@ func AdminTxCheck(vtx *models.Tx, txBytes, signature []byte, state *State) error
...
+		return 0, err
+	}
+	// check signature recovered address
+	tAddr := common.BytesToAddress(treasurer.Address)
+	if tAddr != sigAddress {
+		return 0, fmt.Errorf("address recovered not treasurer: expected %s got %s", treasurer.String(), sigAddress.String())
+	}
+	// check nonce
+	if tx.Nonce != treasurer.Nonce {
+		return 0, fmt.Errorf("invalid nonce %d, expected: %d", tx.Nonce, treasurer.Nonce+1)
```
altergui: why is the error expecting `treasurer.Nonce+1` if the check is comparing `treasurer.Nonce`?
maybe it's intentional/correct, but it looks mismatched so in that case maybe add a comment to clarify?

vochain/state.go
```diff
@@ -504,38 +542,144 @@ func (v *State) IsOracle(addr common.Address) (bool, error) {
...
+// SetFaucetNonce stores an already used faucet nonce in the
+// FaucetNonce subtree
+func (v *State) SetFaucetNonce(key []byte) error {
+	v.Tx.Lock()
+	defer v.Tx.Unlock()
+	return v.Tx.DeepSet(key, nil, FaucetNonceCfg)
+}
+
+func (v *State) substractTxCostIncrementNonce(accoutAddress common.Address, value uint64) error {
+	acc, err := v.GetAccount(accoutAddress, false)
```
altergui: typo here (and previous line, and in return line), `accoutAddress` is missing `n`

rpcapi/accounthandlers.go
```diff
@@ -31,10 +31,29 @@ func (r *RPCAPI) getAccount(request *api.APIrequest) (*api.APIresponse, error) {
...
+			response.Delegates = append(response.Delegates, common.BytesToAddress(v).String())
+		}
 	}
-	response.Delegates = delegates
+	return &response, nil
+}
+
+func (r *RPCAPI) getTreasurer(request *api.APIrequest) (*api.APIresponse, error) {
+
+	// try get treasyrer and send reply
```
altergui: treasurer

vochain/process.go
```diff
@@ -484,56 +527,69 @@ func (app *BaseApplication) NewProcessTxCheck(vtx *models.Tx, txBytes,
...
-		// If owner authorized, check entityId matches with owner address
-		if !bytes.Equal(process.EntityId, addr.Bytes()) {
-			return fmt.Errorf("process entityID and transaction owner do not match")
-		}
-	} else {
-		// Check if the transaction comes from an oracle
-		// Oracles can create processes with any entityID
-		if isOracle, err = state.IsOracle(addr); err != nil {
-			return err
+		// If address authorized, check entityId matches with address address or a delegate
```
altergui, line 554: this logic looks very similar to line 455 (`// check tx signer can create process`)
maybe make a function out of them?

altergui: i see you need isOracle further down, so i don't know. maybe it's not worth it.

altergui (commented):
just skimmed over the diff and made a couple trivial comments, i'm afraid i can't contribute much more right now

vochain/transaction.go
jordipainan: yup thats wrong, changed

vochain/state.go
```diff
@@ -504,38 +542,150 @@ func (v *State) IsOracle(addr common.Address) (bool, error) {
 
 // setTreasurer saves the Treasurer address to the state
 func (v *State) setTreasurer(address common.Address) error {
```
randomshinichi, line 544: Would be great if you could cherry-pick this commit to reexport SetTreasurer
https://github.com/randomshinichi/vocdoni-node/commit/2b50175fc5458dfb530794a12d373ce4c5387d6d

NateWilliams2: Second this! This would make integration testing with vocone tokens possible

vochain/process.go
```diff
@@ -61,6 +62,16 @@ func (v *State) AddProcess(p *models.Process) error {
 	if err != nil {
 		return err
 	}
+	processCreationCost, err := v.TxCost(models.TxType_NEW_PROCESS, false)
+	if err != nil {
+		return err
```
NateWilliams2, line 67: imo wrap this error so it's more descriptive

vochain/process.go
```diff
@@ -250,12 +261,19 @@ func (v *State) SetProcessStatus(pid []byte, newstatus models.ProcessStatus, com
 		for _, l := range v.eventListeners {
 			l.OnProcessStatusChange(process.ProcessId, process.Status, v.TxCounter())
 		}
+		processCreationCost, err := v.TxCost(models.TxType_SET_PROCESS_STATUS, false)
+		if err != nil {
+			return err
```
NateWilliams2, line 266: same as above, wrap with more info

vochain/process.go
```diff
@@ -316,6 +334,13 @@ func (v *State) SetProcessResults(pid []byte, result *models.ProcessResult, comm
 				log.Warnf("onProcessResults callback error: %v", err)
 			}
 		}
+		processCreationCost, err := v.TxCost(models.TxType_SET_PROCESS_RESULTS, false)
+		if err != nil {
+			return err
```
NateWilliams2, line 339: same, etc

vochain/account.go
```diff
@@ -175,108 +214,510 @@ func (v *State) GetAccount(address common.Address, isQuery bool) (*Account, erro
...
+				if used {
+					return nil, fmt.Errorf("nonce %d already used", faucetPkgPayload.Identifier)
+				}
+				// check issuer have enough funds
+				issuerAcc, err := state.GetAccount(issuerAddress, false)
+				if err != nil {
+					return nil, fmt.Errorf("cannot get faucet account: %v", err)
+				}
+				if issuerAcc == nil {
+					return nil, ErrAccountNotFound
```
NateWilliams2, line 426: wrap this to clarify the faucet account, not the recipient account, was not found

PR #473, opened 2022-02-07
ethless: support state TxCosts

jordipainan:
THIS COMMIT BREAKS GENESIS on DEV and STAGE ENVIRONMENTS

### DIFF
--- a/cmd/vochaintest/vochaintest.go
+++ b/cmd/vochaintest/vochaintest.go
@@ -11,6 +11,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/ethereum/go-ethereum/common"
 	flag "github.com/spf13/pflag"
 
 	"go.vocdoni.io/dvote/client"
@@ -146,6 +147,11 @@ func main() {
 		censusImport(*host, entityKey)
 	case "censusGenerate":
 		censusGenerate(*host, entityKey, *electionSize, *keysfile, 1)
+	case "testtxs":
+		// end-user voting is not tested here
+		testAllTransactions(*host,
+			*oraclePrivKey,
+		)
 	default:
 		log.Fatal("no valid operation mode specified")
 	}
@@ -946,4 +952,105 @@ func cspVoteTest(
 		log.Infof("results: %+v", r)
 	}
 	log.Infof("all done!")
+}
+
+// enduser voting is not tested here
+func testAllTransactions(
+	host,
+	oraclePrivKey string,
+) {
+	// oracle is also the treasurer
+	var err error
+	oracleSigner := ethereum.NewSignKeys()
+	if err := oracleSigner.AddHexKey(oraclePrivKey); err != nil {
+		log.Fatal(err)
+	}
+
+	log.Infof("connecting to main gateway %s", host)
+	// add the first connection, this will be the main connection
+	var mainClient *client.Client
+
+	for tries := 10; tries > 0; tries-- {
+		mainClient, err = client.New(host)
+		if err == nil {
+			break
+		}
+		time.Sleep(1 * time.Second)
+	}
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer mainClient.Close()
+
+	chainId, err := mainClient.GetChainID()
+	if err != nil {
+		log.Fatal(err)
+	}
+	oracleSigner.VocdoniChainID = chainId
+
+	// check set transaction cost
+	if err := testSetTxCost(mainClient, oracleSigner); err != nil {
+		log.Fatal(err)
+	}
+
+}
+
+func testSetTxCost(mainClient *client.Client, oracleSigner *ethereum.SignKeys) error {
+	// get treasurer
+	treasurer, err := mainClient.GetTreasurer(oracleSigner)
+	if err != nil {
+		return err
+	}
+	log.Infof("treasurer fetched %s with nonce %d", common.BytesToAddress(treasurer.Address), treasurer.Nonce)
+
+	// get current tx cost
+	txCost, err := mainClient.GetTransactionCost(oracleSigner, models.TxType_SET_ACCOUNT_INFO)
+	if err != nil {
+		return err
+	}
+	log.Infof("tx cost of %s fetched successfully (%d)", models.TxType_SET_ACCOUNT_INFO, txCost)
+
+	h, err := mainClient.GetCurrentBlock()
+	if err != nil {
+		return fmt.Errorf("cannot get current height")
+	}
+	// set tx cost
+	if err := mainClient.SetTransactionCost(oracleSigner,
+		models.TxType_SET_ACCOUNT_INFO,
+		1000,
+		treasurer.Nonce); err != nil {
+		return fmt.Errorf("cannot set transaction cost: %v", err)
+	}
+
+	log.Infof("waiting for new block ...")
+	for {
+		time.Sleep(time.Millisecond * 500)
+		if h2, err := mainClient.GetCurrentBlock(); err != nil {
+			log.Warnf("error getting current height: %v", err)
+			continue
+		} else {
+			if h2 > h {
+				break
+			}
+		}
+	}
+
+	// check tx cost changed and treasurer nonce incremented
+	treasurer2, err := mainClient.GetTreasurer(oracleSigner)
+	if err != nil {
+		return err
+	}
+	newTxCost, err := mainClient.GetTransactionCost(oracleSigner, models.TxType_SET_ACCOUNT_INFO)
+	if err != nil {
+		return err
+	}
+	if treasurer2.Nonce != treasurer.Nonce+1 {
+		return fmt.Errorf("treasurer nonce expected to be %d got %d", treasurer.Nonce+1, treasurer2.Nonce)
+	}
+	if newTxCost != 1000 {
+		return fmt.Errorf("newProcessTx cost expected to be %d got %d", newTxCost, txCost)
+	}
+	log.Infof("tx cost of %s changed successfully from %d to %d", models.TxType_SET_ACCOUNT_INFO, txCost, newTxCost)
+	log.Infof("treasurer nonce changed successfully from %d to %d", treasurer.Nonce, treasurer2.Nonce)
+	return nil
 }
--- a/cmd/vochaintest/vochaintest.go
+++ b/cmd/vochaintest/vochaintest.go
@@ -22,11 +22,12 @@ import (
 )
 
 var ops = map[string]bool{
-	"vtest":          true,
-	"cspvoting":      true,
-	"anonvoting":     true,
-	"censusImport":   true,
-	"censusGenerate": true,
+	"vtest":             true,
+	"cspvoting":         true,
+	"anonvoting":        true,
+	"censusImport":      true,
+	"censusGenerate":    true,
+	"tokentransactions": true,
 }
 
 func opsAvailable() (opsav []string) {
@@ -42,6 +43,7 @@ func main() {
 	loglevel := flag.String("logLevel", "info", "log level")
 	opmode := flag.String("operation", "vtest", fmt.Sprintf("set operation mode: %v", opsAvailable()))
 	oraclePrivKey := flag.String("oracleKey", "", "hexadecimal oracle private key")
+	treasurerPrivKey := flag.String("treasurerKey", "", "hexadecimal treasurer private key")
 	entityPrivKey := flag.String("entityKey", "", "hexadecimal entity private key")
 	host := flag.String("gwHost", "http://127.0.0.1:9090/dvote", "gateway websockets endpoint")
 	electionType := flag.String("electionType", "encrypted-poll", "encrypted-poll or poll-vote")
@@ -78,6 +80,12 @@ func main() {
 		fmt.Fprintf(os.Stderr,
 			"\t./test --operation=censusGenerate --gwHost "+
 				"wss://gw1test.vocdoni.net/dvote --electionSize=10000 --keysFile=keys.json\n")
+		fmt.Fprintf(os.Stderr,
+			"=> tokentransactions\n\tTests all token "+
+				"related transactions\n")
+		fmt.Fprintf(os.Stderr,
+			"\t./test --operation=tokentransactions --gwHost "+
+				"wss://gw1test.vocdoni.net/dvote\n")
 	}
 	flag.Parse()
 
@@ -147,11 +155,9 @@ func main() {
 		censusImport(*host, entityKey)
 	case "censusGenerate":
 		censusGenerate(*host, entityKey, *electionSize, *keysfile, 1)
-	case "testtxs":
+	case "tokentransactions":
 		// end-user voting is not tested here
-		testAllTransactions(*host,
-			*oraclePrivKey,
-		)
+		testTokenTransactions(*host, *treasurerPrivKey)
 	default:
 		log.Fatal("no valid operation mode specified")
 	}
@@ -955,14 +961,13 @@ func cspVoteTest(
 }
 
 // enduser voting is not tested here
-func testAllTransactions(
+func testTokenTransactions(
 	host,
-	oraclePrivKey string,
+	treasurerPrivKey string,
 ) {
-	// oracle is also the treasurer
 	var err error
-	oracleSigner := ethereum.NewSignKeys()
-	if err := oracleSigner.AddHexKey(oraclePrivKey); err != nil {
+	treasurerSigner := ethereum.NewSignKeys()
+	if err := treasurerSigner.AddHexKey(treasurerPrivKey); err != nil {
 		log.Fatal(err)
 	}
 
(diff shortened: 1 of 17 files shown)


NateWilliams2 (approved):
LGTM, I have one comment though-

Maybe this could be added in a different PR, but could we include `SetTreasurer`, `MintTokens`, and `SetTxCost` functions exposed in Vocone.go? This would be really helpful in testing, i.e. in the vaas-api I can use these to allocate tokens to a faucet account.

I implemented these methods locally to test with:
```
// SetTreasurer configures the vocone treasurer account address
func (vc *Vocone) SetTreasurer(treasurer common.Address) error {
	if err := vc.app.State.SetTreasurer(treasurer); err != nil {
		return err
	}
	vc.app.Commit()
	return nil
}

// MintTokens mints tokens to the given account address
func (vc *Vocone) MintTokens(to common.Address, amount uint64) error {
	if err := vc.app.State.MintBalance(to, amount); err != nil {
		return err
	}
	vc.app.Commit()
	return nil
}

// SetTxCost configures the transaction cost for the given tx type
func (vc *Vocone) SetTxCost(txType models.TxType, cost uint64) error {
	if err := vc.app.State.SetTxCost(txType, cost); err != nil {
		return err
	}
	vc.app.Commit()
	return nil
}
```

I included vc.App.Commit() so that this code can be run synchronously without waiting for the next block to be mined, etc. I understand if this is improper and shouldn't be used though. @jordipainan  and maybe @p4u what do u think?

p4u:
> I included vc.App.Commit() so that this code can be run synchronously without waiting for the next block to be mined, etc. I understand if this is improper and shouldn't be used though. @jordipainan and maybe @p4u what do u think?

Yes, we can export that methods, I think that makes sense.

Regarding the app.Commit() no way :/ haha

p4u:
> > I included vc.App.Commit() so that this code can be run synchronously without waiting for the next block to be mined, etc. I understand if this is improper and shouldn't be used though. @jordipainan and maybe @p4u what do u think?
> 
> Yes, we can export that methods, I think that makes sense.
> 
> Regarding the app.Commit() no way :/ haha

Can you send a suggestion so JP can just apply it?

cmd/vochaintest/vochaintest.go
```diff
@@ -146,6 +147,11 @@ func main() {
 		censusImport(*host, entityKey)
 	case "censusGenerate":
 		censusGenerate(*host, entityKey, *electionSize, *keysfile, 1)
+	case "testtxs":
```
p4u: I think we can improve this naming to at least "testTxs", but I'd suggest "testTokenTransactions"

testAllTransactions should take the treasurer key not the oracle (even if its the same, for consistency the function signature should have the correct naming for input variables).

cmd/vochaintest/vochaintest.go
```diff
@@ -947,3 +953,104 @@ func cspVoteTest(
 	}
 	log.Infof("all done!")
 }
+
+// enduser voting is not tested here
+func testAllTransactions(
+	host,
+	oraclePrivKey string,
```
p4u: that should be treasurerKey

cmd/vochaintest/vochaintest.go
```diff
@@ -947,3 +953,104 @@ func cspVoteTest(
...
+) {
+	// oracle is also the treasurer
```
p4u: Why this assumption? You really don't know who will call this function

cmd/vochaintest/vochaintest.go
```diff
@@ -947,3 +953,104 @@ func cspVoteTest(
...
+	oracleSigner.VocdoniChainID = chainId
+
+	// check set transaction cost
+	if err := testSetTxCost(mainClient, oracleSigner); err != nil {
+		log.Fatal(err)
+	}
+
+}
+
+func testSetTxCost(mainClient *client.Client, oracleSigner *ethereum.SignKeys) error {
```
p4u: oracleSigner should be treasurerSigner maybe?

NateWilliams2:
> Regarding the app.Commit() no way :/ haha

Thought so but I had to try 😛

dockerfiles/testsuite/start_test.sh
```diff
@@ -67,10 +74,18 @@ testid="/tmp/.vochaintest$RANDOM"
...
+[ $TEST -eq 5 -o $TEST -eq 0 ] && {
+	echo "### Running test 5 ###"
+	test_txs ${testid}5 &
+} || echo 0 >${testid}5
+
+echo "### Waiting for all state txs tests ###"
 wait
 
-[ "$(cat ${testid}1)" == "0" -a "$(cat ${testid}2)" == "0" -a "$(cat ${testid}3)" == "0" -a "$(cat ${testid}4)" == "0" ] && {
+[ "$(cat ${testid}1)" == "0" -a "$(cat ${testid}2)" == "0" -a "$(cat ${testid}3)" == "0" -a "$(cat ${testid}4)" == "0" -a "$(cat ${testid}5)" == "0" ] && {
```
p4u: I think its time to break this line

rpcapi/accounthandlers.go
```diff
@@ -31,10 +33,44 @@ func (r *RPCAPI) getAccount(request *api.APIrequest) (*api.APIresponse, error) {
 	response.Nonce = new(uint32)
```
p4u, line 32: Regarding the previous lines (not part of this PR but related):

```
	// account does not exist
	if acc == nil {
		// response ok meaning the account does not exist
		return &response, nil
	}
```
I think we should return an error if the account does not exists.

rpcapi/accounthandlers.go
```diff
@@ -31,10 +33,44 @@ func (r *RPCAPI) getAccount(request *api.APIrequest) (*api.APIresponse, error) {
...
 	*response.Nonce = acc.Nonce
 	response.InfoURI = acc.InfoURI
-	delegates := make([]string, len(acc.DelegateAddrs))
-	for i := 0; i < len(acc.DelegateAddrs); i++ {
-		delegates = append(delegates, common.BytesToAddress(acc.DelegateAddrs[i]).String())
+	if len(acc.DelegateAddrs) > 0 {
+		for _, v := range acc.DelegateAddrs {
+			response.Delegates = append(response.Delegates, common.BytesToAddress(v).String())
```
p4u, line 36: Is not the same as doing `fmt.Sprintf("0x%x", v)` ?

rpcapi/accounthandlers.go
```diff
@@ -31,10 +33,44 @@ func (r *RPCAPI) getAccount(request *api.APIrequest) (*api.APIresponse, error) {
...
+func (r *RPCAPI) getTreasurer(request *api.APIrequest) (*api.APIresponse, error) {
+	// try get treasurer and send reply
+	var response api.APIresponse
+	t, err := r.vocapp.State.Treasurer(true)
+	if err != nil {
+		return nil, fmt.Errorf("cannot get treasurer: %w", err)
+	}
+	// treasurer does not exist
+	if t == nil {
+		// response ok meaning the treasurer does not exist
```
p4u: Same, I think this should be an error

NateWilliams2:
Interesting, It seems not possible to suggest changes to a file (vocone.go) that's not included in the original PR. I'll push a commit and ask Jordi to review it

rpcapi/accounthandlers.go
```diff
@@ -31,10 +33,44 @@ func (r *RPCAPI) getAccount(request *api.APIrequest) (*api.APIresponse, error) {
...
+		return &response, nil
+	}
+	response.EntityID = common.BytesToAddress(t.Address).String()
```
p4u: You can build response here

`response := api.APIresponse { EntityID : ... `

Also I think you don't need to call common.BytesToAddress

vocone/vocone.go
```diff
@@ -195,6 +196,21 @@ func (vc *Vocone) AddOracle(oracleKey *ethereum.SignKeys) error {
 	return nil
 }
 
+// SetTreasurer configures the vocone treasurer account address
```
NateWilliams2, line 199: @jordipainan what do you think of these changes?

vochain/genesis.go
```diff
@@ -288,12 +288,16 @@ var Genesis = map[string]VochainGenesis{
       ],
       "treasurer": "0x309Bd6959bf4289CDf9c7198cF9f4494e0244b7d",
       "tx_cost": {
-         "Tx_SetProcess": 100,
+         "Tx_SetProcessStatus": 100,
```
p4u, line 291: This is breaking for the Genesis.

If a commit modigy the Genesis it should update the genesis version and time and also add a comment to the commit message. Also I think stage genesis should be updated.

vochain/state.go
```diff
@@ -504,39 +520,103 @@ func (v *State) IsOracle(addr common.Address) (bool, error) {
...
+// IsTreasurer returns true if the given address matches the Treasurer address
 func (v *State) IsTreasurer(addr common.Address) (bool, error) {
-	tAddr, err := v.Treasurer(false)
+	t, err := v.Treasurer(false)
 	if err != nil {
 		return false, err
 	}
-
+	tAddr := common.BytesToAddress(t.Address)
 	return addr == tAddr, nil
```
p4u: `return addr == common.BytesToAddress(t.Address), nil` or `bytes.Equal(addr, t.Address)`

vochain/state.go
```diff
@@ -504,39 +520,103 @@ func (v *State) IsOracle(addr common.Address) (bool, error) {
...
 }
 
+// incrementTreasurerNonce increments the treasurer nonce
+func (v *State) incrementTreasurerNonce() error {
+	t, err := v.Treasurer(false)
+	if err != nil {
+		return err
```
p4u: Ideally all these errors should be wrapped. Else its very hard to identify where it comes from

vochain/state.go
```diff
@@ -504,39 +520,103 @@ func (v *State) IsOracle(addr common.Address) (bool, error) {
...
+// SetTxCost sets the given transaction cost
+func (v *State) SetTxCost(txType models.TxType, cost uint64) error {
+	key, ok := TxTypeCostToStateKey[txType]
+	if !ok {
+		return fmt.Errorf("txType %v shouldn't cost anything", txType)
+	}
+	v.Tx.Lock()
+	defer v.Tx.Unlock()
+
+	costBytes := [8]byte{}
```
p4u: make is more idiomatic here

vochain/transaction.go
```diff
@@ -585,3 +597,43 @@ func checkRevealProcessKeys(tx *models.AdminTx, process *models.Process) error {
 	}
 	return nil
 }
+
+func SetTransactionCostsTxCheck(vtx *models.Tx, txBytes, signature []byte, state *State) (uint64, error) {
+	tx := vtx.GetSetTransactionCosts()
```
p4u, line 618: vtx should be checked before accessing fields since its a pointer

vochain/transaction.go
```diff
@@ -585,3 +597,43 @@ func checkRevealProcessKeys(tx *models.AdminTx, process *models.Process) error {
...
+	// check signature available
+	if signature == nil || tx == nil || txBytes == nil {
+		return 0, fmt.Errorf("missing signature and/or transaction")
+	}
+	// check value
+	if tx.Value <= 0 {
```
p4u: This check removes the possibility to declare a tx with cost 0, I think we should allow that

vochain/transaction.go
```diff
@@ -585,3 +597,43 @@ func checkRevealProcessKeys(tx *models.AdminTx, process *models.Process) error {
...
+		return 0, fmt.Errorf("cannot extract address from public key: %w", err)
+	}
+	// get treasurer
+	treasurer, err := state.Treasurer(false)
+	if err != nil {
+		return 0, err
+	}
+	// check signature recovered address
+	tAddr := common.BytesToAddress(treasurer.Address)
+	if tAddr != sigAddress {
```
p4u: nit: no need to declare tAddr

vochain/transaction.go
```diff
@@ -585,3 +597,43 @@ func checkRevealProcessKeys(tx *models.AdminTx, process *models.Process) error {
...
+		return 0, fmt.Errorf("address recovered not treasurer: expected %s got %s", treasurer.String(), sigAddress.String())
+	}
+	// check nonce
+	if tx.Nonce != treasurer.Nonce {
```
p4u, line 629: I think these two last checks should go before extracting the pubkey from the signature (which is the expensive check)

vochain/transaction.go
```diff
@@ -585,3 +597,43 @@ func checkRevealProcessKeys(tx *models.AdminTx, process *models.Process) error {
 	}
 	return nil
 }
+
+func SetTransactionCostsTxCheck(vtx *models.Tx, txBytes, signature []byte, state *State) (uint64, error) {
```
p4u, line 614: Please document all exported methods :)

vochain/types.go
```diff
@@ -59,13 +60,87 @@ type QueryData struct {
...
+	"SendTokens":              models.TxType_SEND_TOKENS,
+	"SetAccountInfo":          models.TxType_SET_ACCOUNT_INFO,
+	"RegisterKey":             models.TxType_REGISTER_VOTER_KEY,
+	"NewProcess":              models.TxType_NEW_PROCESS,
+	"AddDelegateForAccount":   models.TxType_ADD_DELEGATE_FOR_ACCOUNT,
+	"DelDelegateForAccount":   models.TxType_DEL_DELEGATE_FOR_ACCOUNT,
+	"CollectFaucet":           models.TxType_COLLECT_FAUCET,
+}
+
+func TxCostNameToTxType(key string) models.TxType {
```
p4u, line 112: document me!

p4u (changes requested):

jordipainan:
@p4u I cannot comment on your suggestions, regarding the one left (`Is not the same as doing fmt.Sprintf("0x%x", v) ?`) is not the same cause this does not convert to the checksum addr

jordipainan:
seems the last integration test is not executed, do not merge until solved

p4u (approved):

p4u:
great job

NateWilliams2:
@p4u @jordipainan I realized a problem with the `SetTreasurer`, `MintTokens`, and `SetTxCost` functions I proposed. These methods all rely on making some changes to the state, so if they are called directly in the code from Vocone.go (without a hacky solution like manually calling `app.State.Save()`), they will have no effect. The next time a block is mined, the state will roll back, and then transactions from the mempool will be applied. The effects of calling these methods from `Vocone.go` will be discarded altogether. It could still be helpful to have these methods in the vocone, but maybe they can add an actual tx to the mempool rather than calling the methods directly. 

I'm actually confused because there are tests in vocdoni-node that seem to rely on calling methods like `app.State.SetTxCost` directly, without committing the state.

jordipainan:
@NateWilliams2  
You're right, I thought you are going to call Save() later.
I also thought that the vocone is just for testing, so you can call Save() later or do not call it if you do not need it. We can talk more about this because maybe I miss some information.

The testing thing you mention is done for setting up the testing easily and does not impact on the actual test itself.

NateWilliams2:
> I thought you are going to call Save() later

In this case, we need to expose `app.State.Save` to the vocone as well

vocone/vocone.go
```diff
@@ -195,6 +196,40 @@ func (vc *Vocone) AddOracle(oracleKey *ethereum.SignKeys) error {
...
+		log.Debugf("setting tx cost for txtype %s", models.TxType_name[int32(k)])
+		err := vc.app.State.SetTxCost(k, txCosts)
+		if err != nil {
+			return err
+		}
+
+	}
+	return vc.app.State.IncrementTreasurerNonce()
+}
+
```
NateWilliams2, line 254:
```suggestion
func (vc *Vocone) SaveState() error {
	_, err := vc.app.State.Save()
	return err
}
```

vochain/state.go
```diff
@@ -504,37 +520,100 @@ func (v *State) IsOracle(addr common.Address) (bool, error) {
...
+	t, err := v.Treasurer(false)
 	if err != nil {
 		return false, err
 	}
+	return addr == common.BytesToAddress(t.Address), nil
+}
 
-	return addr == tAddr, nil
+// IncrementTreasurerNonce increments the treasurer nonce
+func (v *State) IncrementTreasurerNonce() error {
```
p4u, line 569: does this method needs to be exported?

p4u (approved):

randomshinichi:
> I'm actually confused because there are tests in vocdoni-node that seem to rely on calling methods like app.State.SetTxCost directly, without committing the state.

In my original functions, I was calling Save(), I just thought that since the functions you added here didn't have them, you had some solution of your own.

jordipainan:
@p4u can I merge it ? I think we can squash everything in one commit

p4u:
> @p4u can I merge it ? I think we can squash everything in one commit

Go ahead, ok to squash

Issue #414, opened 2021-12-20
sqlite vs badgerhold inconsistency errors


p4u:
```
2021-12-20T10:24:45Z	DEBUG	scrutinizer/process.go:199	ProcessList badgerhold took 1.674072ms
2021-12-20T10:24:45Z	DEBUG	scrutinizer/process.go:213	ProcessList sqlite took 1.336863ms
  []string{
  	"b1fb3c077cd8e6b97f1fd9da50455c3ce5097a47d99cda870f9221844ef43af9",
- 	"4b23b18d326ac8d7f8ca1a603cc17094c7ecd83c005827c40345284158a82ff8",
- 	"981f43f7ccb528d82eef9ba1d9ff6c6670a460e65ba15961ff25beee057294d5",
  }

params: []interface {}{[]uint8(nil), 0, 64, "", 0x0, "", "", false}
panic: ping mvdan to fix this bug

goroutine 6383142 [running]:
go.vocdoni.io/dvote/vochain/scrutinizer.(*Scrutinizer).ProcessList(0xc0031488c0, {0x0, 0x0, 0x0}, 0x0, 0x40, {0x0, 0x0}, 0x0, {0x0, ...}, ...)
	go.vocdoni.io/dvote/vochain/scrutinizer/process.go:225 +0x2e91
go.vocdoni.io/dvote/rpcapi.(*RPCAPI).getProcessList(0xc0003e2a20, 0xc00c121900)
	go.vocdoni.io/dvote/rpcapi/votehandlers.go:133 +0x118
go.vocdoni.io/dvote/rpcapi.(*RPCAPI).route(0xc0003e2a20, {{0x26f0aa0, 0xc00bef79d0}, {0xc068345b68281de3, 0xd983d3b969, 0x43f5de0}, {0xc0134f2830, 0x1, 0x1}, 0xc0134f27e0})
	go.vocdoni.io/dvote/rpcapi/api.go:109 +0xca
created by go.vocdoni.io/dvote/httprouter.(*HTTProuter).routerHandler.func1
	go.vocdoni.io/dvote/httprouter/httprouter.go:237 +0x3f9
```

p4u:
```
2021-12-20T16:08:46Z	DEBUG	scrutinizer/process.go:199	ProcessList badgerhold took 7.583132ms
2021-12-20T16:08:46Z	DEBUG	scrutinizer/process.go:213	ProcessList sqlite took 1.205918ms
  []string{
[...]

PR #474, opened 2022-02-07
vochain/scrutinizer: fix 0.4% of entity IDs with sqlite

mvdan:
(see commit message)

Updates #414.


mvdan:
I'm not closing #414 yet, because some of the recent panics do not include an entity_id starting with a null byte. So there might be another bug to fix.

mvdan:
Also, sqlc is actively working on native sqlite support, so hopefully we'll be able to remove these workarounds soon. For the time being, they're not too painful.

mvdan:
I should note that I'm not sure whether our real environments produce random entity IDs, but even if they don't, it's still better to plug this trap.

NateWilliams2 (approved):
LGTM, I think there is actually a chance of this occurring in real behavior.

mvdan:
We'll see if dev or staging still see warnings after this.

PR #476, opened 2022-02-07
vochain/scrutinizer: fix encrypted processes with sqlite

mvdan:
(see commit message)

Updates #414.


NateWilliams2 (approved):

NateWilliams2:
I've confirmed that this change fixes the bug I was getting in vaas-api testing

PR #477, opened 2022-02-08
Run go-tests as root

vdo:


mvdan (approved):
I imagine we could just remove the option entirely, given root is the default and that's very, very unlikely to change.

PR #479, opened 2022-02-09
Vocone.SetBulkTxCosts()

randomshinichi:
Maybe it's better to instead increment the nonce once, after all the setting is done?

EDIT: okay, let's not waste nonces.


jordipainan:
I think is fine to increment the nonce, just once at the end

jordipainan (approved):

PR #482, opened 2022-02-11
refactor account related methods & add support for MintTokensTx

jordipainan:


jordipainan:
@altergui Can you help me a bit on the `start_test.sh` script. When I try to run just the last test with `./start_test.sh 5` it runs the first one. I do not know a lot about bash and seems that the script does some kind of magic.

Thank you!

altergui:
> @altergui Can you help me a bit on the `start_test.sh` script. When I try to run just the last test with `./start_test.sh 5` it runs the first one. I do not know a lot about bash and seems that the script does some kind of magic.
> 
> Thank you!

bash is horrible, yes.
but i just checked and it works as expected. try removing the second line of the script `set -x`, which makes the output really confusing (it prints out every line that is being executed, prefixed with `+ `)

```
./start_test.sh 5
### Starting test suite ###
### Waiting for test suite to be ready ###
[blah]
### Running test 5 ###
### Waiting for tests to finish ###
```

jordipainan:
@altergui so I do not know what I am doing wrong of if I have something locally not well configured, but when I run `start_test.sh 5` (also removing `set -x`) the only test executed is `test()` and not `test_token_transactions()` :S

cmd/vochaintest/vochaintest.go
```diff
@@ -1059,3 +1077,136 @@ func testSetTxCost(mainClient *client.Client, treasurerSigner *ethereum.SignKeys
...
+	}
+	// wait for new block
+	log.Infof("waiting for new block ...")
+	for {
+		time.Sleep(time.Millisecond * 500)
+		if h2, err := mainClient.GetCurrentBlock(); err != nil {
+			log.Warnf("error getting current height: %v", err)
+			continue
+		} else {
+			if h2 > h {
```
mvdan: nit: join else and if

cmd/vochaintest/vochaintest.go
```diff
@@ -1059,3 +1077,136 @@ func testSetTxCost(mainClient *client.Client, treasurerSigner *ethereum.SignKeys
...
```
mvdan: nit: same

vochain/account.go
```diff
@@ -192,94 +171,406 @@ func (v *State) VerifyAccountBalance(message, signature []byte, amount uint64) (
...
-func (v *State) CreateAccount(address common.Address, infoURI string, delegates []common.Address, initBalance uint64) error {
+// Create account creates an account
+func (v *State) CreateAccount(accountAddress common.Address,
+	infoURI string,
+	delegates []common.Address,
+	initBalance uint64,
+) error {
 	// check valid address
-	if address.String() == types.EthereumZeroAddress {
+	if bytes.Equal(accountAddress.Bytes(), types.EthereumZeroAddressBytes[:]) {
```
mvdan: FYI, you can replace all instances of `bytes.Equal(foo.Bytes(), types.EthereumZeroAddressBytes[:])` with just `foo == types.EthereumZeroAddressBytes` - you can compare array values directly.

types/consts.go
```diff
@@ -7,8 +7,9 @@ func Bool(b bool) *bool { return &b }
 // These exported variables should be treated as constants, to be used in API
 // responses which require *bool fields.
 var (
-	False = Bool(false)
-	True  = Bool(true)
+	False                    = Bool(false)
+	True                     = Bool(true)
+	EthereumZeroAddressBytes = [20]byte{}
```
mvdan: nit: I'd do `EthereumZeroAddress ethcommon.Address`.

This drops the `Bytes`, because it's a byte array, not a byte slice, and "bytes" tends to imply a slice.

you also give it the proper type of Address, which helps avoid misuse, and also allows you to not hard-code the length of 20 here.

vochain/account.go
```diff
@@ -192,94 +171,406 @@ func (v *State) VerifyAccountBalance(message, signature []byte, amount uint64) (
...
+		return common.Address{}, 0, fmt.Errorf("invalid To address")
+	}
+	// get treasurer
+	treasurer, err := state.Treasurer(false)
+	if err != nil {
+		return common.Address{}, 0, err
+	}
+	// check nonce
+	if tx.Nonce != treasurer.Nonce {
+		return common.Address{}, 0, fmt.Errorf("invalid nonce %d, expected: %d", tx.Nonce, treasurer.Nonce+1)
```
mvdan: you check against Nonce but report Nonce+1?

vochain/account.go
```diff
@@ -192,94 +171,406 @@ func (v *State) VerifyAccountBalance(message, signature []byte, amount uint64) (
...
+					if err != nil {
+						return nil, fmt.Errorf("cannot get %s tx cost: %w", models.TxType_COLLECT_FAUCET, err)
+					}
+					if (issuerAcc.Balance) < faucetPkgPayload.Amount+cost {
+						return nil, fmt.Errorf("faucet does not have enough balance %d < %d", issuerAcc.Balance, faucetPkgPayload.Amount+cost)
+					}
+					returnValues.FaucetPayloadSigner = issuerAddress
+					returnValues.FaucetPayload = tx.GetFaucetPackage().Payload
+				} else {
+					return nil, fmt.Errorf("payload to and tx sender missmatch")
```
mvdan: tip: if you reverse this if and return early, then you can unindent the bulk of the code. same for some other "ifs" that have this pattern.

vochain/account.go
```diff
@@ -136,26 +123,18 @@ func (v *State) MintBalance(address common.Address, amount uint64) error {
...
-	raw, err := v.Tx.DeepGet(address.Bytes(), AccountsCfg)
-	if err != nil && !errors.Is(err, arbo.ErrKeyNotFound) {
-		return err
-	} else if err == nil {
-		if err := acc.Unmarshal(raw); err != nil {
-			return err
-		}
+	acc, err := v.GetAccount(address, false)
+	if err != nil {
+		return fmt.Errorf("mintBalance (%w)", err)
```
mvdan: nit: the format is `foo: %w` for %w to work.

vochain/account.go
```diff
@@ -87,46 +87,33 @@ func (a *Account) DelDelegate(addr common.Address) {
...
-	if !isQuery {
-		v.Tx.Lock()
-		defer v.Tx.Unlock()
-	}
-	accFromRaw, err := v.Tx.DeepGet(from.Bytes(), AccountsCfg)
-	if err != nil {
+func (v *State) TransferBalance(from, to common.Address, amount uint64, nonce uint32) error {
+	var accFrom, accTo *Account
+	var err error
+	if accFrom, err = v.GetAccount(from, false); err != nil {
```
mvdan: You don't need these `var`s:

```suggestion
	accFrom, err := v.GetAccount(from, false)
	if err != nil {
```

mvdan: joining the error check with the `if` can help reduce lines in some situations, but in this case it just doesn't help

mvdan (commented):
quick review, happy to be more thorough when it's no longer WIP

altergui:
> @altergui so I do not know what I am doing wrong of if I have something locally not well configured, but when I run `start_test.sh 5` (also removing `set -x`) the only test executed is `test()` and not `test_token_transactions()` :S

sooooooo.... you're right. i was just looking at the `if`s and didn't notice that the `test` lines are wrong, tests "3", "4" (and your "5") are not working, @NateWilliams2 also got this wrong a couple of weeks ago in 4f95b7dcab3a927fa7a93757b4042e90a73cb981 because, well, bash is horrible. the correct syntax is to call `test_token_transactions ${testid}5 &` (without prefixing `test`, which makes it execute `test()` instead)
(nate fixed this yesterday in 97ad3e3a7da88fe5efd2d24d1413668e378149b9 <3)

p4u:
@jordipainan this needs to be rebased

randomshinichi (approved):
LGTM

PR #469, opened 2022-02-05
dvotenode: new flag --ipfsSyncLogLevel

altergui:
This addresses issue #460

it's not the most flexible or extendable solution, but i believe it's the most elegant (in terms of lines changed)

please review, especially if the approach is too hacky or non-idiomatic :grin: 

in terms of behaviour, what happens now is:
* running `dvotenode --logLevel=debug --ipfsSyncLogLevel=info` achieves what @p4u wanted, getting debug messages for everything except ipfssync code
* note that due to how zap is implemented, the opposite is not possible, running `dvotenode --logLevel=info --ipfsSyncLogLevel=debug` will start with a warning and just print info messages for everything. but i believe this is not a problem.
* not specifying `--ipfsSyncLogLevel` behaves like before, ipfssync inherits global `--logLevel`

in any case, ipfssync messages are also now tagged with `[ipfssync]` (using zap mechanisms), which i believe makes it very easy to filter in or out with `| grep` or `| grep -v`, FWIW

Issue #460, opened 2022-02-02
Reduce ipfsSync log messages


p4u:
In DEBUG mode currently ipfsSync and its packages are flooding the log output so its very hard to find other (probably more relevant) messages.

I'd suggest to somehow add a flag to the vocdoni-node in order to configure the ipfsSync log output independently.

I'm not sure how to do that with the current log implementation, probably we need to extend the log package somehow and allow creating origin tags by log type and then be able to configure different log levels per origin. Something like:

main.go
```
log.SetOriginLogLevel("ipfsSync", "info")
```

ipfsSync.go
```
logger := log.WithOrigin("ipfsSync")
logger.Debugf("...")
```

@mvdan could probably help us here to find the best solution

altergui:
since f83a4b2648616df9f1a4e2df009ce7d091704e74, ipfsconnect (ex-ipfsSync) is much less verbose. closing this

PR #488, opened 2022-02-16
WIP: log: allow more granular logging with --logLevel=component:level

altergui:
note that this commit doesn't change any of the current logging behaviour,
keeping backwards compatibility for code using log.Init()

to use granular logging, call log.InitWithLevels()
    
* fix Logger() since it was returning an unwrapped logger with CallerSkip(1)
* if using log.Named("somename"), prefix [somename] in logs
    
this addresses issue #460 "Reduce ipfsSync log messages", in an extensible way

supersedes #469, implementing @mvdan suggestion https://github.com/vocdoni/vocdoni-node/pull/469#issuecomment-1033037869

### DIFF
--- a/cmd/dvotenode/dvotenode.go
+++ b/cmd/dvotenode/dvotenode.go
@@ -71,8 +71,9 @@ func newConfig() (*config.DvoteCfg, config.Error) {
 		"use developer mode (less security)")
 	globalCfg.PprofPort = *flag.Int("pprof", 0,
 		"pprof port for runtime profiling data (zero is disabled)")
-	globalCfg.LogLevel = *flag.StringP("logLevel", "l", "info",
-		"log level (debug, info, warn, error, fatal)")
+	globalCfg.LogLevel = *flag.StringSliceP("logLevel", "l", []string{"info"},
+		"log level for each subsystem, in format [<subsystem|*>:]<debug|info|warn|error|fatal>\n"+
+			"subsystems: "+strings.Join(log.GetSubsystems(), ","))
 	globalCfg.LogOutput = *flag.String("logOutput", "stdout",
 		"log output (stdout, stderr or filepath)")
 	globalCfg.LogErrorFile = *flag.String("logErrorFile", "",
@@ -361,7 +362,7 @@ func main() {
 	if globalCfg == nil {
 		log.Fatal("cannot read configuration")
 	}
-	log.Init(globalCfg.LogLevel, globalCfg.LogOutput)
+	log.Init(strings.Join(globalCfg.LogLevel, ","), globalCfg.LogOutput)
 	if path := globalCfg.LogErrorFile; path != "" {
 		if err := log.SetFileErrorLog(path); err != nil {
 			log.Fatal(err)
--- a/data/ipfs.go
+++ b/data/ipfs.go
@@ -27,9 +27,14 @@ import (
 	crypto "go.vocdoni.io/dvote/crypto/ethereum"
 	"go.vocdoni.io/dvote/db/lru"
 	"go.vocdoni.io/dvote/ipfs"
-	"go.vocdoni.io/dvote/log"
+	logger "go.vocdoni.io/dvote/log"
 	"go.vocdoni.io/dvote/types"
 )
+
+var log = logger.Named("ipfs")
+
+// register subsystem name to allow setting its logLevel
+var _ = logger.Named("ipfs.daemon")
 
 const (
 	MaxFileSizeBytes      = 1024 * 1024 * 50 // 50 MB
@@ -40,7 +45,6 @@ type IPFSHandle struct {
 	Node         *ipfscore.IpfsNode
 	CoreAPI      coreiface.CoreAPI
 	DataDir      string
-	LogLevel     string
 	retriveCache *lru.Cache
 
 	// cancel helps us stop extra goroutines and listeners which complement
@@ -50,10 +54,11 @@ type IPFSHandle struct {
 }
 
 func (i *IPFSHandle) Init(d *types.DataStore) error {
-	if i.LogLevel == "" {
-		i.LogLevel = "ERROR"
+	l, found := logger.GetLogLevel("ipfs.daemon")
+	if !found || l == "" { // default to ERROR LogLevel
+		l = "ERROR"
 	}
-	ipfslog.SetLogLevel("*", i.LogLevel)
+	ipfslog.SetLogLevel("*", l)
 	ipfs.InstallDatabasePlugins()
 	ipfs.ConfigRoot = d.Datadir
 
--- a/ipfssync/ipfssync.go
+++ b/ipfssync/ipfssync.go
@@ -20,9 +20,11 @@ import (
 	"google.golang.org/protobuf/proto"
 
 	"go.vocdoni.io/dvote/data"
-	"go.vocdoni.io/dvote/log"
+	logger "go.vocdoni.io/dvote/log"
 	"go.vocdoni.io/dvote/statedblegacy/gravitonstate"
 )
+
+var log = logger.Named("ipfssync")
 
 const (
 	MaxKeySize = 64
--- a/ipfssync/ipfssync.go
+++ b/ipfssync/ipfssync.go
@@ -353,8 +353,6 @@ func (is *IPFSsync) unicastMsg(address string, imsg *models.IpfsSync) error {
 
 // Start initializes and start an IPFSsync instance
 func (is *IPFSsync) Start() {
-	var err error
-
 	// Init pin storage
 	log.Infof("initializing new pin storage")
 	dbDir := path.Join(is.DataDir, "db")
@@ -362,10 +360,10 @@ func (is *IPFSsync) Start() {
 		log.Fatal(err)
 	}
 	is.state = &gravitonstate.GravitonState{}
-	if err = is.state.Init(dbDir, "disk"); err != nil {
+	if err := is.state.Init(dbDir, "disk"); err != nil {
 		log.Fatal(err)
 	}
-	if err = is.state.AddTree("ipfsSync"); err != nil {
+	if err := is.state.AddTree("ipfsSync"); err != nil {
 		log.Fatal(err)
 	}
 	is.hashTree = is.state.Tree("ipfsSync")
--- a/log/log.go
+++ b/log/log.go
@@ -3,12 +3,11 @@ package log
 import (
 	"fmt"
 	"os"
+	"strings"
 	"time"
 
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
-	"google.golang.org/protobuf/encoding/protojson"
-	"google.golang.org/protobuf/reflect/protoreflect"
 )
 
 var (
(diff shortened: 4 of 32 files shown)


altergui:
(i did include a trivial nitpick to ipfssync.go) :eyes:

p4u:
LGTM

cmd/dvotenode/dvotenode.go
```diff
@@ -71,8 +71,8 @@ func newConfig() (*config.DvoteCfg, config.Error) {
 		"use developer mode (less security)")
 	globalCfg.PprofPort = *flag.Int("pprof", 0,
 		"pprof port for runtime profiling data (zero is disabled)")
-	globalCfg.LogLevel = *flag.StringP("logLevel", "l", "info",
-		"log level (debug, info, warn, error, fatal)")
+	globalCfg.LogLevel = *flag.StringSliceP("logLevel", "l", []string{"*:info"},
+		"log level for each subsystem, in format <subsystem|*>:<debug|info|warn|error|fatal>")
```
mvdan: I would say make the default just `info` rather than `*:info`, then we don't break existing commands.

ipfssync/ipfssync.go
```diff
@@ -20,10 +20,15 @@ import (
...
 	"go.vocdoni.io/dvote/data"
-	"go.vocdoni.io/dvote/log"
+	logger "go.vocdoni.io/dvote/log"
 	"go.vocdoni.io/dvote/statedblegacy/gravitonstate"
 )
 
+// global log is set to a logger.Named(loggerName) on Start
+var log = logger.Logger()
+
+const loggerName = "ipfssync"
```
mvdan: why not just `var log = logger.Named("ipfssync")`?

cmd/dvotenode/dvotenode.go
mvdan: Ah, I see you still support it, but I didn't get that from your docs.

vochain/start.go
```diff
@@ -23,9 +23,14 @@ import (
 	tmjson "github.com/tendermint/tendermint/libs/json"
 	tmlog "github.com/tendermint/tendermint/libs/log"
 	tmos "github.com/tendermint/tendermint/libs/os"
-	"go.vocdoni.io/dvote/log"
+	logger "go.vocdoni.io/dvote/log"
```
mvdan, line 26: if you call `Named` directly in the global var decl, then we don't need these renames.

log/log.go
```diff
@@ -14,34 +16,98 @@ import (
...
+// logLevels are strings like "somename:debug" or "*:info" (equivalent to simply "info").
+// out can be either "stdout", "stderr" or "/path/to/file" and it's a global setting.
+//
+// It's safe to use concurrently.
+func InitWithLevels(logLevels []string, out string) {
+	for _, ll := range logLevels {
+		split := strings.Split(ll, ":")
+		if len(split) > 1 {
+			// split[0] is subsystem name, split[1] is level
+			logLevel.Store(split[0], split[1])
```
mvdan: if you set up all named loggers as static globals, then in here you can also check if a log level component is unknown; otherwise we'll silently ignore e.g. `foobar:info`.

data/ipfs.go
```diff
@@ -50,10 +54,13 @@ type IPFSHandle struct {
...
 
 func (i *IPFSHandle) Init(d *types.DataStore) error {
-	if i.LogLevel == "" {
-		i.LogLevel = "ERROR"
+	log = logger.Named(loggerName)
+
+	l := logger.LogLevelFor(loggerName)
+	if l == "" { // default to ERROR LogLevel
+		l = "ERROR"
 	}
```
mvdan, line 60: API-wise, it would be nicer if creating the logger also set the default log level for that component, like:

```
var log = log.Named(data.ipfs, log.Error)
```

Then it's shorter and you don't need to expose API like LogLevelFor.

log/log.go
```diff
@@ -14,34 +16,98 @@ import (
...
 	}
-	Init(level, "stderr")
+
+	InitWithLevels(level, defaultOutput)
+}
+
+// Init initializes the logger, with a single logLevel (debug|info|warn|error|fatal).
+// out can be either "stdout", "stderr" or "/path/to/file" and it's a global setting.
+//
+// kept for backwards compatibility, it just wraps InitWithLevels()
```
mvdan: just modify Init to take a list? I don't think we need to worry about backwards compatibility in the log package, if the changes are pretty easy to adapt to.

log/log.go
```diff
@@ -14,34 +16,98 @@ import (
...
+	level := LogLevelFor(name)
+	if level == "" {
+		level = LogLevelFor("*")
+	}
+	defer log.Infof("logger construction succeeded for subsystem %s at level %s", name, level)
+	return newLogger(level).Named(name)
+}
+
+// Logger returns the unwrapped logger
+func Logger() *zap.SugaredLogger {
```
mvdan: How about calling this `Unnamed` now, so then we end up with `log.Named(...)` and `log.Unnamed()`?

altergui:
thanks for the review! i did look for ways to avoid renaming and cleaner init'ing, but couldn't wrap my head around it. will try again :) and rework this PR.

also, while hacking on this i noticed that both ipfs and tendermint have their own `log` packages which (unsurprisingly) wrap zap and have the same features that i'm trying to develop, so i want to evaluate using them directly.

mvdan:
my 2c: neither tendermint nor go-ipfs guarantee any form of stability for their Go APIs, especially not for packages that aren't really intended for direct use, like logging. So I'd probably not want to use them.

vochain/start.go
altergui: mmmh, i tried that first, as in 
`var log = log.Logger()`
or, well, the ideal `var log = log.Named("ipfssync")`
but go complains that `log already declared through import of package log ("go.vocdoni.io/dvote/log")`

mvdan: Ah, I guess because of the global scope. then maybe `var log = logpkg.Named("ipfssync")`

ipfssync/ipfssync.go
altergui: Named() clones the global logger, which at the time of package init is uninitialized. Afterwards, during dvotenode init() , log.Init() is called and sets the global logger up, but it's too late.
i'll see how i can refactor Named() so that the when log.Init() is later run, the named loggers are also init'd

mvdan: I think the right solution is to instead initialize the log package's global logger at init time as well. then the settings applied via flags would modify the logger in-place.

data/ipfs.go
altergui: mmh here i'm trying to do the opposite, snoop into what LogLevel was set upstream for this component (`ipfs`) and pass that same level to the (underlying) ipfs

right now my approach is that:
* each component chooses their own name (ipfssync defines itself as "ipfssync")
* the caller chooses which logLevel to use for each component (so, dvotenode chooses "error" or "debug" for component "ipfssync")

but i'm thinking it'd probably be wiser and more consistent that the caller chooses BOTH the component name and loglevel. i'll rework the approach like this, i think the result will make much more sense

log/log.go
altergui: dropped this func

log/log.go
altergui: good point, done

data/ipfs.go
altergui: in the end i kept the original approach, which is also the same implemented by `ipfs/go-log`

ipfssync/ipfssync.go
altergui: right, i did this, adapting code from `ipfs/go-log` which had exactly this approach.
it was an interesting adventure :)

cmd/dvotenode/dvotenode.go
altergui: i updated the docs to be clearer

log/log.go
```diff
@@ -22,26 +21,60 @@ func init() {
...
-	cfg := newConfig(logLevel, output)
+// Init initializes the logger.
+//
+// logLevels is a comma-separated string like "somename:debug,*:info"
+// (for backwards-compatibility, the "*:" is optional, so "warn" is interpreted as "*:warn")
+//
+// output can be either "stdout", "stderr" or "/path/to/file" and it's a global setting.
+//
+// It's safe to use concurrently.
+func Init(logLevels string, output string) {
```
mvdan, line 39: why not just `logLevels []string`? some of the callers use Join, and you use Split here.

mvdan: also, for the sake of clarity, I would call this func `GlobalInit`

log/setup.go
```diff
@@ -0,0 +1,235 @@
...
+var primaryCore zapcore.Core
+
+// loggerCore is the base for all loggers created by this package
+var loggerCore = &lockedMultiCore{}
+
+// SetupLogging will initialize the logger backend and set the flags.
+// TODO calling this in `init` pushes all configuration to env variables
+// - move it out of `init`? then we need to change all the code (js-ipfs, go-ipfs) to call this explicitly
+// - have it look for a config file? need to define what that is
+func SetupLogging(cfg Config) {
```
mvdan, line 60: why is this exported?

log/setup.go
```diff
@@ -0,0 +1,235 @@
...
+
+type LogFormat int
+
+const (
+	ColorizedOutput LogFormat = iota
+	PlaintextOutput
+	JSONOutput
+)
+
+type Config struct {
```
mvdan, line 24: why is this exported?

vochain/start.go
```diff
@@ -23,9 +23,11 @@ import (
...
 )
 
+var log = logger.Named("vochain")
```
mvdan, line 29: nit: I would call the imported package `logpkg` rather than `logger`, because then it's more obvious that the lines wanting to print a log line shouldn't use `logger`.

log/setup.go
```diff
@@ -0,0 +1,235 @@
...
+	sort.Strings(subs)
+	return subs
+}
+
+// Named returns a named logger.
+func Named(name string) *zap.SugaredLogger {
+	loggerMutex.Lock()
+	defer loggerMutex.Unlock()
+	log, ok := loggers[name]
+	if !ok {
```
mvdan, line 208: if you invert this condition, you can just `return log` and unindent the rest.

log/setup.go
```diff
@@ -0,0 +1,235 @@
...
+		if rem.MatchString(name) {
+			levels[name].SetLevel(zapcore.Level(lvl))
+		}
+	}
+	return nil
+}
+
+// GetSubsystems returns a sorted slice containing the
+// names of the current loggers
+func GetSubsystems() []string {
```
mvdan, line 191: avoid GetFoo names: https://go.dev/doc/effective_go#Getters

but also, you could use a slightly better name, like SubsystemNames

log/setup.go
```diff
@@ -0,0 +1,235 @@
...
+	if primaryCore != nil {
+		loggerCore.ReplaceCore(primaryCore, core)
+	} else {
+		loggerCore.AddCore(core)
+	}
+	primaryCore = core
+}
+
+// SetAllLoggers changes the logging level of all loggers to lvl
+func SetAllLoggers(lvl zapcore.Level) {
```
mvdan, line 108: why is this exported?

log/setup.go
```diff
@@ -0,0 +1,235 @@
...
+			Sugar()
+
+		loggers[name] = log
+	}
+
+	return log
+}
+
+// GetLogLevel returns the configured log level of a specific subsystem
+func GetLogLevel(name string) (level string, found bool) {
```
mvdan, line 229: like before, avoid the Get prefix; just call this LogLevel

log/setup.go
```diff
@@ -0,0 +1,235 @@
...
+	case "fatal":
+		return zapcore.FatalLevel, nil
+	default:
+		return zapcore.InfoLevel, fmt.Errorf("couldn't parse log level '%s'", logLevel)
+	}
+}
+
+// SetLogLevel changes the log level of a specific subsystem
+// name=="*" changes all subsystems
+func SetLogLevel(name, level string) error {
```
mvdan, line 142: as far as I can tell, SetLogLevel and SetLogLevelRegex are unused. also, why are they exported?

mvdan (commented):
I would recommend doing `go doc ./log`, to see what API you're actually exposing. Right now, you're exposing way too much stuff :)

mvdan:
Also: bear in mind that `init` functions get run sequentially, so anything that they do doesn't require a mutex. You only need a mutex if any of the loggers are _modified_ after they start being used asynchronously by different goroutines.

jordipainan:
@altergui there is something blocking this PR ?

altergui:
yeah, i need to address all the points of the review and push a better version

jordipainan:
> yeah, i need to address all the points of the review and push a better version

perfect! no rush just asked in the case there is something external that is blocking you. Thank you <3

altergui:
i'm closing this PR since it would need a big rebase or rewrite from scratch,
and i believe #746 might have solved the usecase for this feature, at least partially.

PR #490, opened 2022-02-21
Rename isQuery to committed in vochain

G10h4ck:
Added also some documentation about it's usage on all methods that sports this param

### DIFF
--- a/vochain/state.go
+++ b/vochain/state.go
@@ -719,7 +719,7 @@ func (v *State) RemoveValidator(address []byte) error {
 	return validators.Set(address, nil)
 }
 
-// Validators returns a list of the validators saved on persistent storage
+// Validators returns a list of the chain validators
 // TODO: do 'saved on persistent storage' make sense also when committed == false?
 // When committed is false, the operation is executed also on not yet commited
 // data from the currently open StateDB transaction.
--- a/vochain/state.go
+++ b/vochain/state.go
@@ -454,12 +454,12 @@ func (v *State) setMainTreeView(treeView *statedb.TreeView) {
 	v.mainTreeViewValue.Store(treeView)
 }
 
-// mainTreeViewer returns the mainTree as a treeViewer.  When isQuery is false,
-// the mainTree returned is the not yet commited one from the currently open
-// StateDB transaction.  When isQuery is true, the mainTree returned is the
-// last commited version.
-func (v *State) mainTreeViewer(isQuery bool) statedb.TreeViewer {
-	if isQuery {
+// mainTreeViewer returns the mainTree as a treeViewer.
+// When committed is false, the mainTree returned is the not yet commited one
+// from the currently open StateDB transaction.
+// When committed is true, the mainTree returned is the last commited version.
+func (v *State) mainTreeViewer(committed bool) statedb.TreeViewer {
+	if committed {
 		return v.MainTreeView()
 	}
 	return v.Tx.AsTreeView()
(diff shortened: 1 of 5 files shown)


G10h4ck:
I have got a few doubts about the documentation I have written, flagged those with TODO, may someone with more insight than me take a look?

vochain/state.go
```diff
@@ -658,13 +667,17 @@ func (v *State) RemoveValidator(address []byte) error {
 }
 
 // Validators returns a list of the validators saved on persistent storage
-func (v *State) Validators(isQuery bool) ([]*models.Validator, error) {
-	if !isQuery {
+// TODO: do 'saved on persistent storage' make sense also when committed == false?
```
jordipainan: Nope, in this case and following the other methods I think we should rewrite the comment.
I propose to change it to: `Validators returns a list of the chain validators`

vochain/state.go
```diff
@@ -881,14 +903,18 @@ func (v *State) EnvelopeExists(processID, nullifier []byte, isQuery bool) (bool,
 }
 
 // iterateVotes iterates fn over state tree entries with the processID prefix.
-// if isQuery, the IAVL tree is used, otherwise the AVL tree is used.
+// if committed is true, the IAVL tree is used, otherwise the AVL tree is used.
+// TODO: does previous line is more descriptive then the followings 3 ?
```
jordipainan: If you're not aware that `I` stands for Immutable then sadly not ... So the 3 following lines are better and maintain consistency with the other methods.

vochain/state.go
```diff
@@ -658,13 +667,17 @@ func (v *State) RemoveValidator(address []byte) error {
 }
 
 // Validators returns a list of the validators saved on persistent storage
```
jordipainan:
```suggestion
// Validators returns a list of the chain validators
```

vochain/state.go
```diff
@@ -881,14 +903,18 @@ func (v *State) EnvelopeExists(processID, nullifier []byte, isQuery bool) (bool,
 }
 
 // iterateVotes iterates fn over state tree entries with the processID prefix.
-// if isQuery, the IAVL tree is used, otherwise the AVL tree is used.
+// if committed is true, the IAVL tree is used, otherwise the AVL tree is used.
```
jordipainan, line 906:
```suggestion
```

vochain/state.go
randomshinichi: And what does AVL stand for?

jordipainan: https://pkg.go.dev/github.com/tendermint/iavl

jordipainan:
Missing changes on https://github.com/G10h4ck/vocdoni-node/blob/mv_isquery_committed/vochain/process.go#L116 & https://github.com/G10h4ck/vocdoni-node/blob/mv_isquery_committed/vochain/process.go#L125

With these remaining lines I think we can say we got rid of IsQuery naming :)

G10h4ck:
> Missing changes on https://github.com/G10h4ck/vocdoni-node/blob/mv_isquery_committed/vochain/process.go#L116 & https://github.com/G10h4ck/vocdoni-node/blob/mv_isquery_committed/vochain/process.go#L125
> 
> With these remaining lines I think we can say we got rid of IsQuery naming :)

I forgot to save process.go before committing ;-)

jordipainan (approved):

randomshinichi:
Thanks for taking on this issue, code feels better now!

jordipainan:
self-reminder: Squash

PR #491, opened 2022-02-21
add support for set delegate tx

jordipainan:
Merge after #489

### DIFF
--- a/client/util.go
+++ b/client/util.go
@@ -15,19 +15,21 @@ import (
 	"go.vocdoni.io/dvote/vochain"
 )
 
+const TimeBetweenBlocks = 6 * time.Second
+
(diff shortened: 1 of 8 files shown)


client/util.go
```diff
@@ -21,13 +21,13 @@ func (c *Client) WaitUntilBlock(block uint32) {
...
 			log.Error(err)
-			time.Sleep(5 * time.Second)
+			time.Sleep(6 * time.Second)
 			continue
 		}
 		if cb >= block {
 			break
 		}
-		time.Sleep(5 * time.Second)
+		time.Sleep(6 * time.Second)
```
G10h4ck: If it need to be the same in more then one place better to extrapolate that value to a constant

G10h4ck (changes requested):

PR #494, opened 2022-02-24
mod: update all patch versions via go get -u=patch

mvdan:
(see commit message)


jordipainan (approved):

jordipainan:
Already created #497 for the CI failure

mvdan:
If master is green, it could be that upgrading one of the libraries broke the tests.

jordipainan:
> If master is green, it could be that upgrading one of the libraries broke the tests.

master actually have the same problem

Issue #454, opened 2022-01-30
Refactor vochainInfo


p4u:
Refactor in a more advanced, smart and idiomatic way the vochaininfo package [1]. Also extend it to support more kind of statistics, on log but also for prometheus.

Open to creativity of the developer taking the issue :)

[1] https://github.com/vocdoni/vocdoni-node/tree/master/vochain/vochaininfo

altergui:
i'm delighted!

i even hacked a bit on this while working on tendermint-0.35, for example i made a flag for the (currently hardcoded) stats log interval. i didn't push this commit since it needed more work (which was out of scope for tendermint-0.35), and also because i felt it wasn't worth it to add yet another flag for this. but just to say, i'm absolutely eager to work on this :smile: 

<details>
  <summary>hacking on stats interval</summary>
  
```patch
    dvotenode: new flag --vochainLogStatsInterval useful for debugging

-------------------------- cmd/dvotenode/dvotenode.go --------------------------
index 55c35d52..d4332395 100644
@@ -133,6 +133,8 @@ func newConfig() (*config.DvoteCfg, config.Error) {
 		"tendermint node log level (error, info, debug, none)")
 	globalCfg.VochainConfig.LogLevelMemPool = *flag.String("vochainLogLevelMemPool", "error",
 		"tendermint mempool log level")
+	globalCfg.VochainConfig.LogStatsInterval = *flag.Duration("vochainLogStatsInterval", time.Duration(20*time.Second),
+		"time between periodic logging of several status info")
 	globalCfg.VochainConfig.Peers = *flag.StringArray("vochainPeers", []string{},
 		"comma-separated list of p2p peers")
 	globalCfg.VochainConfig.Seeds = *flag.StringArray("vochainSeeds", []string{},
@@ -240,6 +242,7 @@ func newConfig() (*config.DvoteCfg, config.Error) {
 	viper.BindPFlag("vochainConfig.RPCListen", flag.Lookup("vochainRPCListen"))
 	viper.BindPFlag("vochainConfig.LogLevel", flag.Lookup("vochainLogLevel"))
 	viper.BindPFlag("vochainConfig.LogLevelMemPool", flag.Lookup("vochainLogLevelMemPool"))
+	viper.BindPFlag("vochainConfig.LogStatsInterval", flag.Lookup("vochainLogStatsInterval"))
 	viper.BindPFlag("vochainConfig.Peers", flag.Lookup("vochainPeers"))
[...]

PR #495, opened 2022-02-24
refactor vochaininfo

altergui:
here's some progress for #454

### DIFF
--- a/vochain/vochaininfo/vochaininfo.go
+++ b/vochain/vochaininfo/vochaininfo.go
@@ -9,26 +9,26 @@ import (
 )
 
 // VochainInfo stores some metrics and information regarding the Vochain Blockchain
-// Avg1/10/60/360 are the block time average for 1 minute, 10 minutes, 1 hour and 6 hours
 type VochainInfo struct {
-	sync   bool
-	height int64
-	// NOTE(Edu): After the integration of the arbo-based StateDB, there's
-	// no single voteTree, but we can still count total number of votes.  A
-	// more appropiate name for this variable would be voteCount
-	voteTreeSize    uint64
-	processTreeSize uint64
-	mempoolSize     int
-	voteCacheSize   int
-	votesPerMinute  int
-	avg1            int32
-	avg10           int32
-	avg60           int32
-	avg360          int32
-	avg1440         int32
-	vnode           *vochain.BaseApplication
-	close           chan bool
-	lock            sync.RWMutex
+	height         int64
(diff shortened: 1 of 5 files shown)


vochain/vochaininfo/vochaininfo.go
```diff
@@ -156,99 +164,100 @@ func (vi *VochainInfo) NPeers() int {
...
-	var h1, h10, h60, h360, h1440 int64
-	var n1, n10, n60, n360, n1440, vm int64
-	var a1, a10, a60, a360, a1440 int32
-	var sync bool
-	var oldVoteTreeSize uint64
-	duration = time.Second * time.Duration(sleepSecs)
+// statsInterval defines how often the (cached) calculations are updated.
+func (vi *VochainInfo) Start(statsInterval time.Duration) {
+	log.Infof("starting vochain info service every %s", statsInterval)
+	var err error
```
mvdan, line 170: I think you can move this var further down

vochain/vochaininfo/vochaininfo.go
```diff
@@ -156,99 +164,100 @@ func (vi *VochainInfo) NPeers() int {
...
+func (vi *VochainInfo) updateAvgBlockTimes(interval time.Duration) {
+	// during fastsync, stats are totally skewed: don't even try.
+	if !vi.Sync() {
+		return
+	}
+	oldHeight := vi.vnode.Height()
+	select {
+	case <-time.After(interval):
+		vi.lock.Lock()
+		vi.avgBlockTimes[interval] = time.Duration(int64(interval) / int64(vi.vnode.Height()-oldHeight))
```
mvdan, line 238: I assume you're fine with integer division, because the interval is a number of nanoseconds, so it will be large enough that integer divisions will be precise enough.

You should use the Nanoseconds method, which makes that clearer: https://cs.opensource.google/go/go/+/refs/tags/go1.17.7:src/time/time.go;l=736

mvdan (commented):
The code seems fine to me. My only worry is the amount of goroutines you have involved here; the more goroutines, the harder the code is to understand and ensure that we don't have races or deadlocks.

I imagine it's so that you can have multiple tickers? FYI, you could have just one goroutine with multiple tickers, and then a big select statement that gets triggered on any one of them. Potentially that could be better anyway, because then a lot of the locking would be simpler or even unnecessary.

jordipainan:
@altergui there is something blocking this PR ?

PR #498, opened 2022-02-25
Add getCensusWeight handler

### LINKED ISSUE #463: feature: Add `getCensusWeight`
**Describe the feature**
Currently, the census service has the `getSize` call to retrieve the number of voters in a census. However, the addition of weighted censuses means that we would also need a call to know the total weight of that census.

**Motivation**
Displaying the turnout of a weighted voting election.

**Proposal**
Add a `getCensusWeight` or `getWeight` call (the second name could be ambiguous).

**Components affected**
Census Service

**Constraints & Assumptions**

**Additional information (if required)**

NateWilliams2:
Adds getCensusWeight handler, tests as well

This handler uses the arbo tree `IterateLeaves()` method. @arnaucube performance-wise, is this method realistic to use with large census sizes? 

Closes #463

### DIFF
--- a/census/handler.go
+++ b/census/handler.go
@@ -323,6 +323,19 @@ func (m *Manager) Handler(ctx context.Context, r *api.APIrequest,
 		resp.Size = &sizeInt64
 		return resp, nil
 
+	case "getCensusWeight":
+		resp.Weight = &types.BigInt{}
+		getWeightCallback := func(key, value []byte) bool {
+			weight := tr.BytesToBigInt(value)
+			resp.Weight = resp.Weight.Add((*types.BigInt)(weight), resp.Weight)
+			return true
+		}
+		err := tr.IterateLeaves(getWeightCallback)
+		if err != nil {
+			return nil, err
+		}
+		return resp, nil
+
 	case "dumpPlain":
 		return nil, fmt.Errorf("dumpPlain is deprecated, dump should be used instead")
 
--- a/census/handler.go
+++ b/census/handler.go
@@ -324,16 +324,11 @@ func (m *Manager) Handler(ctx context.Context, r *api.APIrequest,
 		return resp, nil
 
 	case "getCensusWeight":
-		resp.Weight = &types.BigInt{}
-		getWeightCallback := func(key, value []byte) bool {
-			weight := tr.BytesToBigInt(value)
-			resp.Weight = resp.Weight.Add((*types.BigInt)(weight), resp.Weight)
-			return true
-		}
-		err := tr.IterateLeaves(getWeightCallback)
+		weight, err := tr.GetCensusWeight()
 		if err != nil {
 			return nil, err
 		}
+		resp.Weight = (*types.BigInt)(weight)
 		return resp, nil
 
 	case "dumpPlain":
--- a/censustree/censustree.go
+++ b/censustree/censustree.go
@@ -1,6 +1,7 @@
 package censustree
 
 import (
+	"errors"
 	"fmt"
 	"math/big"
 	"sync"
@@ -12,6 +13,8 @@ import (
 	"go.vocdoni.io/dvote/tree"
 	"go.vocdoni.io/proto/build/go/models"
 )
+
+var censusWeightKey = []byte("censusWeight")
 
 // Tree implements the Merkle Tree used for census
 // Concurrent updates to the tree.Tree can lead to losing some of the updates,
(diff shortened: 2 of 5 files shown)


jordipainan (approved):

jordipainan:
Waiting 'til @arnaucube answer :) cc: @brickpop

arnaucube (commented):
Yes, as you comment @NateWilliams2 , this can be a bit intensive from performance point of view. It will walk over all the nodes & leaves of the tree, so if the tree is big enough, it may take more than 10 seconds for example, even more than 1 minute.
Maybe we can go with this for now, if the `getCensusWeight` is not going to be used too much, having all this in mind.
Maybe another option would be to add the functionality of storing the total census weight in the `censustree` package at `Add` & `AddBatch` methods (https://github.com/vocdoni/vocdoni-node/blob/master/censustree/censustree.go#L153).
In this way, the total weight is updated in the key-value database each time that a leaf is added to the census, and when somebody calls `getCensusWeight` internally it just needs to get the 'total census weight' from the keyvalue db, instead of iterating over all the nodes and leaves of the tree.

jordipainan:
How about, as @arnaucube said, storing the total census weight that is being updated on each leaf addition ? 
@NateWilliams2  do you see it feasible in a timely manner ? Given that the iteration approach can be very consuming, maybe we can kill two birds in one shot here using Arnau's suggestion.

Thank you :)

NateWilliams2:
Absolutely, i think that's a good call.

NateWilliams2:
@arnaucube I made a commit that should store the censusWeight in the tree's prefixed database. When testing, the censusWeight key is not actually getting set (therefore the weight is always returned as 0). Is the database used by the tree unable to be used to store other key/values?

jordipainan:
@arnaucube  can you help nate on this ? thank you

arnaucube:
Sorry for the delay, and thanks for the pings @NateWilliams2 & @jordipainan ^^

The `test/census.go` was getting 0 at the value of the census weight because internally the node was not using `censustree.Add` method, and it was using `censustree.AddBatch`, which didn't had the code to store the census total weight.
But additionally, on the `test/census.go`, prior to getting the weight, the test dumps & imports the leafs of the tree, meaning that imports the leafs into an empty new tree, and the `ImportDump` method didn't had the logic to also store the total census weight while importing the leafs.

I've added a commit with some updates to fix that (more details in the [commit message](https://github.com/vocdoni/vocdoni-node/commit/9182da6fe769ec860c34f87bedfd0e7cdba56790)).

But still I think that we should add one more thing:
The CensusWeight update should be done only for the censustrees that have weight. Other kind of censuses (eg. AnonymousVoting censustree) do not use the `value` parameter of the leaves as weight, as it's used for other stuff (like the pubK), so the CensusWeight in those cases would get wrong values.
Maybe for the moment it's fine as it is, as in a non-weighted census, probably is not interesting to call the getCensusWeight, so it's not a problem that it returns a wrong census weight. But for the future would be worth to have some way to know if the censustree is for a weighted census or not, an if it's not then the census weight should not be computed neither stored. I've added a message in the three ([[1]](https://github.com/vocdoni/vocdoni-node/commit/9182da6fe769ec860c34f87bedfd0e7cdba56790#diff-49e35dc9c17882db4d5de205ceb0e4b9fbb96bff7d818ffa63cbcf36b5679facR182), [[2]](https://github.com/vocdoni/vocdoni-node/commit/9182da6fe769ec860c34f87bedfd0e7cdba56790#diff-49e35dc9c17882db4d5de205ceb0e4b9fbb96bff7d818ffa63cbcf36b5679facR216), [[3]](https://github.com/vocdoni/vocdoni-node/commit/9182da6fe769ec860c34f87bedfd0e7cdba56790#diff-49e35dc9c17882db4d5de205ceb0e4b9fbb96bff7d818ffa63cbcf36b5679facR241)) affected places of the censustree code to have this in mind for the future.

jordipainan:
I was thinking about limiting to API for returning 0 in the case where the census tree is not weighted but seems that we cannot correlate the actual tree with the process attributes.
So maybe the best way to do so is clearly documenting the API call explaining that the weight returned for non weighted censuses must be ignored because does not make sense and is not correct.
Thoughts on this ? @NateWilliams2 @brickpop

jordipainan:
@arnaucube how much overhead adds to compute and update the census weight for non weighted censuses ? Is something we can assume in terms of performance ? Cause at first glance sounds strange for me, but as miracles cannot be done I would like to know if this can affect too much and the tradeoff is acceptable, cause for sure is better than the old solution based on iterate all the tree for getting the total weight. Thank you :)

arnaucube:
Regarding the overhead, it's a bit, specially in the `ImportDump` method, but I guess that for the moment can be assumed, as you said, it's less overhead than iterating over all the leaves each time.
But for a next iteration there could be a parameter in the [`censustree.Tree` struct](https://github.com/vocdoni/vocdoni-node/blob/master/censustree/censustree.go#L25), like `CensusWeight` and if it's true, all these methods do the total census weight computation logic, and they skip it if it's set to false.
(I say a 'new parameter' like `CensusWeight`, as the `CensusType` is already used for another [kind of `type`](https://github.com/vocdoni/dvote-protobuf/blob/master/src/vochain/vochain.proto#L23)).

jordipainan:
Perfect thank you so much @arnaucube, I am going to add a feature request for the future to keep this in mind.

PR #499, opened 2022-02-28
vocone: create burn account on constructor

jordipainan:


NateWilliams2 (approved):
Thanks, this should fix my issue!

Issue #345, opened 2021-11-04
feature: use timestamp as string for the rpc API


p4u:
**Describe the feature**
Use string timestamps instead of int32

**Motivation**
Support timestamps  > year 2038

**Proposal**
Use integers stored in strings - it's just two more characters, and you entirely sidestep the issue of float precision in json :) In Go, they are natively supported like:

```golang
type T struct {
    IntAsString int64 `json:",string"`
}
```

**Components affected**
vocdoni-node and dvote-js

**Additional information (if required)**
This change should be perform in paralel for golang and javascript libraries

G10h4ck:
Do we need to keep some kind of retro-compatibility with old RPC clients?

G10h4ck:
Also the problem about 64bit integer is not withing JSON itself but in the languages on which are written the RPC clients, Dart and JavaScript are affected but Go/C++ etc. do not, I guess the RPC clients are written in JavaScript, am I wrong? In that case we have to change the representation of all 64bit integers and not just the timestamps in the RPC protocol.

jordipainan:
> Do we need to keep some kind of retro-compatibility with old RPC clients?

Just saw this @G10h4ck I do not think we need to maintain this retro-compatibility given that the clients we have are updated or in process to be updated to the new specs.

jordipainan:
> Also the problem about 64bit integer is not withing JSON itself but in the languages on which are written the RPC clients, Dart and JavaScript are affected but Go/C++ etc. do not, I guess the RPC clients are written in JavaScript, am I wrong? In that case we have to change the representation of all 64bit integers and not just the timestamps in the RPC protocol.

Yep you're right, we also have RPC client written in JS, but I do not know if some work around is already implemented, ping @brickpop or @marcvelmer for having better clarity on this. Thank you

PR #500, opened 2022-03-02
Use int64 for API timestamp

G10h4ck:
Address https://github.com/vocdoni/vocdoni-node/issues/345

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 # Binaries for programs and plugins
 /dvotenode
 /dvotecli
+/ipfsSync
 *.exe
 *.exe~
 *.dll
--- a/api/types.go
+++ b/api/types.go
@@ -75,7 +75,7 @@ type APIrequest struct {
 	Signature    types.HexBytes                 `json:"signature,omitempty"`
 	SrcNetId     string                         `json:"sourceNetworkId,omitempty"`
 	Status       string                         `json:"status,omitempty"`
-	Timestamp    int32                          `json:"timestamp"`
+	Timestamp    int64                          `json:"timestamp,string"`
 	TxIndex      int32                          `json:"txIndex,omitempty"`
 	Type         string                         `json:"type,omitempty"`
 	URI          string                         `json:"uri,omitempty"`
@@ -98,7 +98,7 @@ func (a *APIrequest) SetAddress(addr *common.Address) {
 func (a *APIrequest) SetID(id string) {
 }
 
-func (a *APIrequest) SetTimestamp(ts int32) {
+func (a *APIrequest) SetTimestamp(ts int64) {
 	a.Timestamp = ts
 }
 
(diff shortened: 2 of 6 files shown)


G10h4ck:
Seems we have similar but unrelated problem in ipfsync too https://github.com/vocdoni/vocdoni-node/blob/master/ipfssync/ipfssync.go#L40 @altergui I guess that should be treated separately

altergui:
> Seems we have similar but unrelated problem in ipfsync too https://github.com/vocdoni/vocdoni-node/blob/master/ipfssync/ipfssync.go#L40 @altergui I guess that should be treated separately

right, but in ipfssync, AFAIU, there's no api involved, the messages are only passed between ipfssync instances. so maybe it's a non-issue?

G10h4ck:
> right, but in ipfssync, AFAIU, there's no api involved, the messages are only passed between ipfssync instances. so maybe it's a non-issue?

If you plan to make it obsolete before 2037 no problem ;)

jordipainan:
LGTM 

@marcvelmer @brickpop @emmdim I think this also affects other repos can we sync on this ?

G10h4ck:
I have made a PR for dvote-js too https://github.com/vocdoni/dvote-js/pull/99

NateWilliams2:
@G10h4ck Why are you using the json `string` tag for these fields? Is it needed?
I'm getting an error: `getInfo: json: invalid use of ,string struct tag, trying to unmarshal unquoted value into int64`

NateWilliams2:
@G10h4ck Why are you using the json `string` tag for these fields? Is it needed?
I'm getting an error: `getInfo: json: invalid use of ,string struct tag, trying to unmarshal unquoted value into int64`

G10h4ck:
> @G10h4ck Why are you using the json `string` tag for these fields? Is it needed? I'm getting an error: `getInfo: json: invalid use of ,string struct tag, trying to unmarshal unquoted value into int64`

Are you getting that error attempting to parse a JSON generated by older version with the new one?
Using string representation for 64 bit integers is necessary to make it properly parse-able by clients that doesn't natively supports 64 bit integers natively such as Javascript and Dart

p4u:
Jumping in without full context. As @G10h4ck states in the issue, there could be a precission problem if using int64 when interacting with JS libraries, please do not merge this until we are sure we won't have interoperability problems: @mvdan @brickpop @marcvelmer ?

In the worst case we can let it unmerged since the RPC API will require a full refactor in some moment.

mvdan:
I thought we were fine with this small API breaking change, given that we can update the clients too at the same time. If we want the transition to be smoother, the server could accept both forms when decoding by implementing `json.Unmarshaler`. I can help with that if it sounds useful. We would still always encode as a string in the new version, so this would only help with older requests that didn't use strings, it won't help with clients that expect non-string responses.

Issue #497, opened 2022-02-25
bug: sometimes compose-test fails with "could not get pre-register key"


jordipainan:
With 

`2022-02-25T11:15:48Z	FATAL	vochaintest/vochaintest.go:654	[http://gateway0:9090/dvote] could not get pre-register key main.mkTreeAnonVoteTest.func1`

altergui:
i edited this issue  since the first line was actually reported half a year ago in #299 and i'd track it's progress there instead

i'd assume the "could not get pre-register key" is unrelated, and track progress here.

altergui:
PR #608 merged

Issue #299, opened 2021-10-05
bug: Sometimes compose-test fails with "process keys is empty"


ed255:
**Describe the bug**
For the past month or so, from time to time (I'd say between 1/5 and 1/10) the compose-test fails with this error:
```
2021-10-04T14:51:54Z	FATAL	vochaintest/vochaintest.go:361	[http://gateway:9090/dvote] process keys is empty
main.mkTreeVoteTest.func1
	go.vocdoni.io/dvote/cmd/vochaintest/vochaintest.go:361
```
[logs_2215.zip](https://github.com/vocdoni/vocdoni-node/files/7284186/logs_2215.zip)

mvdan:
This sounds like some form of race; perhaps vochaintest fetches the process keys before they are ready. If it were a true backend bug, I imagine the backend would be returning or logging some form of error.

altergui:
i bumped into this several times while working on the CI, so i'll dig into it

altergui:
i'm chasing this one, since on the branch i'm working, it's happening 7 times out of 10
(all of the failures are the same error, process keys never appear)
![image](https://user-images.githubusercontent.com/2815267/192602248-0b738286-9b7e-436e-ae6e-187e6d30635e.png)

altergui:
catched it! the process ends up with empty keys when `rejected tx: adminTxCheck: cannot add process keys to a process that has started or finished status (READY)`

i.e. startBlock is too near in time, and by the time process keys are to be added, it already started.

https://github.com/vocdoni/vocdoni-node/actions/runs/3138070184/jobs/5103477505#step:3:8215
```
miner1_1    | 2022-09-27T18:45:38Z	DEBUG	vochain/transaction.go:599	checking admin signed tx {"txtype":"ADD_PROCESS_KEYS","processId":"VzThEAPFzOwsLWWCYfvcQLBP7AbUkFckLTkBBAAAAAE=","encryptionPublicKey":"Bbu9as36F8x+P7nUDpEjqdcEezOz6NpuIlaQime3CSk=","keyIndex":4,"nonce":128863522} by addr f7fb77ee1f309d9468fb6dcb71add0f934a33c6b
miner1_1    | 2022-09-27T18:45:41Z	INFO	service/vochain.go:230	[vochain info] height:30 mempool:3 peers:9 processes:4 votes:600 vxm:0 voteCache:0 blockTime:{1m:7.50}
[...]

PR #502, opened 2022-03-02
testsuite: create accounts for entity and oracle and mint tokens

altergui:
here's a proposal based on jordi's code, to make the test less brittle. Reviews welcome


cmd/vochaintest/vochaintest.go
```diff
@@ -237,6 +264,78 @@ func censusImport(host string, signer *ethereum.SignKeys) {
...
+func retryMintTokensUntilSuccess(mainClient *client.Client, treasurer, account *ethereum.SignKeys, amount uint64, retries int) error {
+	// instead of waiting until a certain prerequired transaction is mined,
+	// retry the desired tx until success
+	for i := 0; i < retries; i++ {
+		treasurerAccount, err := mainClient.GetTreasurer(treasurer)
+		if err != nil {
+			return fmt.Errorf("cannot get treasurer: %w", err)
+		}
+
+		err = mainClient.MintTokens(treasurer, account.Address(), treasurerAccount.GetNonce(), amount)
```
jordipainan: I think passing the account that will receive the tokens is enough and we do not need to pass the whole signer.

cmd/vochaintest/vochaintest.go
```diff
@@ -237,6 +264,78 @@ func censusImport(host string, signer *ethereum.SignKeys) {
...
+	}
+
+	if err := retryMintTokensUntilSuccess(mainClient, treasurer, mainSigner, 10000, 5); err != nil {
+		log.Fatalf("cannot mint tokens for account %s with treasurer %s: %w", mainSigner.Address(), treasurer.Address(), err)
+	}
+
+	return nil
+}
+
+func retryMintTokensUntilSuccess(mainClient *client.Client, treasurer, account *ethereum.SignKeys, amount uint64, retries int) error {
```
jordipainan: I think the best approach for knowing if tokens were actually minted is to use the `GetAccount()` method, so we are going to be sure that the balance is incremented.

cmd/vochaintest/vochaintest.go
jordipainan: Also I think this can be more idiomatic with `if err := mainClient.MintTokens ...; err == nil { ... }`

cmd/vochaintest/vochaintest.go
```diff
@@ -105,6 +105,33 @@ func main() {
...
+		}
+	}
+
+	treasurerKey := ethereum.NewSignKeys()
+	if len(*treasurerPrivKey) > 0 {
+		if err := treasurerKey.AddHexKey(*treasurerPrivKey); err != nil {
+			log.Fatal(err)
+		}
+	} else {
+		if err := treasurerKey.Generate(); err != nil {
```
jordipainan: I made this wrong, if the treasurer address is not provided in advance (genesis) the randomly generated key is not going to work as the genesis will take into account the hardcoded one

jordipainan (changes requested):

cmd/vochaintest/vochaintest.go
altergui: right, i reimplemented this way, thanks for the pointer

cmd/vochaintest/vochaintest.go
altergui: fixed in https://github.com/vocdoni/vocdoni-node/pull/502/files#diff-6bff2e932680154a119e047cbfcaa19a9d11335ff65a7eccf163ac8a588ca1a4R173

cmd/vochaintest/vochaintest.go
altergui: i ended up totally ignoring errors from MintTokens since they are pretty useless, what matters in the end is the balance

altergui: i didn't understand this comment? i'm indeed passing just the account, not the whole signer, AFAIU

altergui:
i reimplemented the whole test in a much more robust way. feedback welcome

* treasurerKey is now a required flag for tokentransactions
* txs for create account and for top up (mint tokens) are retried 5 times until success, waiting 1 block between retries

cmd/vochaintest/vochaintest.go
jordipainan: Yup sorry I did not make the proper comment. On the old code we were passing account as *ethereum.SignKeys and the only thing we required is the account.Address() so there was no need to pass the whole signer. With the modified code is required to pass the signer because of https://github.com/vocdoni/vocdoni-node/pull/502/files#diff-6bff2e932680154a119e047cbfcaa19a9d11335ff65a7eccf163ac8a588ca1a4R330 so this comment no longer applies here :)

cmd/vochaintest/vochaintest.go
```diff
@@ -237,6 +263,108 @@ func censusImport(host string, signer *ethereum.SignKeys) {
...
+				return fmt.Errorf("cannot create account %s: %s", account.Address(), err)
+			}
+		}
+
+		waitUntilNextBlock(mainClient)
+	}
+	return fmt.Errorf("cannot create account %s after %d retries", account.Address(), retries)
+}
+
+func ensureAccountHasTokens(mainClient *client.Client, treasurer, account *ethereum.SignKeys, amount uint64, retries int) error {
```
jordipainan: So yeah, following the comment I did on the unresolved thread ... You can see in this function that we are passing account as *ethereum.SignKeys but only account.Address() is used. So instead of passing the whole account signer is enough to pass the address of the account.

From `func ensureAccountHasTokens(mainClient *client.Client, treasurer, account *ethereum.SignKeys, amount uint64, retries int) error` to `func ensureAccountHasTokens(mainClient *client.Client, treasurer *ethereum.SignKeys, account common.Address, amount uint64, retries int) error`

altergui: right, so true. pushed this change just now, thanks

altergui:
anyway the compose-test is failing (and taking very long) so i need to review what's happening :disappointed:

jordipainan:
> anyway the compose-test is failing (and taking very long) so i need to review what's happening disappointed

lmk if I can help :)

altergui:
so, the "taking very long" for the tests was unrelated to my code and was apparently solved (or got better) today

on the other hand, compose-test is failing due to #497 and #299 so i'm debugging them as well, ideally i will solve all of those in one go, and finally make the compose-test robust (never fail randomly)

p4u:
Is there a blocker for this PR @altergui @jordipainan ?

jordipainan:
> Is there a blocker for this PR @altergui @jordipainan ?

this was an effort for integrating process txs with ethless making the tests work, there is a new pr, closing this

PR #506, opened 2022-03-10
voconed --treasurer, --txCosts flag

randomshinichi:
Having these flags should let one easily substitute vocone for dvotenode in integration test suites.

### DIFF
--- a/cmd/voconed/voconed.go
+++ b/cmd/voconed/voconed.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/ethereum/go-ethereum/common"
 	flag "github.com/spf13/pflag"
 	"github.com/spf13/viper"
 	"go.vocdoni.io/dvote/crypto/ethereum"
@@ -15,8 +16,9 @@ import (
 
 // VoconeConfig contains the basic configuration for the voconed
 type VoconeConfig struct {
-	logLevel, dir, oracle, path   string
-	port, blockSeconds, blockSize int
+	logLevel, dir, oracle, path, treasurer string
+	port, blockSeconds, blockSize          int
+	txCosts                                uint64
 }
 
 func main() {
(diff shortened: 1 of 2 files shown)


cmd/voconed/voconed.go
```diff
@@ -97,6 +103,20 @@ func main() {
 		log.Fatal(err)
 	}
 
+	if len(config.treasurer) > 0 {
+		log.Info("setting treasurer ", config.treasurer)
+		err = vc.SetTreasurer(common.HexToAddress(config.treasurer))
```
jordipainan: we tend to use `if err := X` as it is more idiomatic

cmd/voconed/voconed.go
```diff
@@ -97,6 +103,20 @@ func main() {
...
+		if err != nil {
+			log.Fatal(err)
+		}
+	}
+
+	log.Info("setting tx costs to ", config.txCosts)
+	err = vc.SetBulkTxCosts(config.txCosts)
```
jordipainan: same as commented regarding `if err :=`

cmd/voconed/voconed.go
```diff
@@ -97,6 +103,20 @@ func main() {
 		log.Fatal(err)
 	}
 
+	if len(config.treasurer) > 0 {
+		log.Info("setting treasurer ", config.treasurer)
```
jordipainan: `log.Infof("setting treasurer %s", config.treasurer)` ? 

Or maybe adding some lines below `log.Infof("treasurer set to %s", config.treasurer)`

cmd/voconed/voconed.go
```diff
@@ -97,6 +103,20 @@ func main() {
...
+		err = vc.SetTreasurer(common.HexToAddress(config.treasurer))
+		if err != nil {
+			log.Fatal(err)
+		}
+	}
+
+	log.Info("setting tx costs to ", config.txCosts)
```
jordipainan: same as commented for the log.Info above :)

PR #512, opened 2022-03-16
ethless: add support for collectFaucetTx

jordipainan:
context: https://www.notion.so/aragonorg/Ethereum-less-vochain-description-67780a73a3e54be5892ed2a2a145b7c7

### DIFF
--- a/client/api.go
+++ b/client/api.go
@@ -1511,3 +1511,47 @@ func (c *Client) SetAccountDelegate(signer *ethereum.SignKeys, delegate common.A
 	}
 	return nil
 }
+
+// SetDelegate sends a set delegate transaction, if op == true adds a delegate, deletes a delegate otherwise
+func (c *Client) CollectFaucetTx(from, to *ethereum.SignKeys, amount, identifier uint64, nonce uint32) error {
+	var req api.APIrequest
+	var err error
+	req.Method = "submitRawTx"
+
+	faucetPayload := &models.FaucetPayload{
+		Identifier: identifier,
+		To:         to.Address().Bytes(),
+		Amount:     amount,
+	}
+	faucetPayloadBytes, err := proto.Marshal(faucetPayload)
+	if err != nil {
+		return err
+	}
+	faucetPayloadSignature, err := from.SignEthereum(faucetPayloadBytes)
+	if err != nil {
+		return err
+	}
+	faucetPkg := &models.FaucetPackage{
+		Payload:   faucetPayload,
+		Signature: faucetPayloadSignature,
+	}
+	tx := &models.CollectFaucetTx{
+		FaucetPackage: faucetPkg,
+		Nonce:         nonce,
+		TxType:        models.TxType_COLLECT_FAUCET,
+	}
+
+	stx := models.SignedTx{}
+	stx.Tx, err = proto.Marshal(&models.Tx{Payload: &models.Tx_CollectFaucet{CollectFaucet: tx}})
+	if err != nil {
+		return err
+	}
+	resp, err := c.SubmitRawTx(to, &stx)
+	if err != nil {
+		return err
+	}
+	if !resp.Ok {
+		return fmt.Errorf("submitRawTx failed: %s", resp.Message)
+	}
+	return nil
+}
--- a/client/api.go
+++ b/client/api.go
@@ -1512,8 +1512,8 @@ func (c *Client) SetAccountDelegate(signer *ethereum.SignKeys, delegate common.A
 	return nil
 }
 
-// SetDelegate sends a set delegate transaction, if op == true adds a delegate, deletes a delegate otherwise
-func (c *Client) CollectFaucetTx(from, to *ethereum.SignKeys, amount, identifier uint64, nonce uint32) error {
+// CollectFaucetTx sends a collect faucet transaction
+func (c *Client) CollectFaucet(from, to *ethereum.SignKeys, amount, identifier uint64, nonce uint32) error {
 	var req api.APIrequest
 	var err error
 	req.Method = "submitRawTx"
--- a/vochain/account.go
+++ b/vochain/account.go
@@ -768,7 +768,7 @@ func CollectFaucetTxCheck(vtx *models.Tx, txBytes, signature []byte, state *Stat
 	}
 	// check tx sender is the same as the one contained in the payload
 	if !bytes.Equal(toAddr.Bytes(), faucetPkg.Payload.To) {
-		return nil, fmt.Errorf("txSender %s and faucet payload To %s missmatch",
+		return nil, fmt.Errorf("txSender %x and faucet payload To %x mismatch",
 			toAddr,
 			common.BytesToAddress(faucetPkg.Payload.To),
 		)
--- a/vochain/account.go
+++ b/vochain/account.go
@@ -282,10 +282,9 @@ func (v *State) ConsumeFaucetPayload(from common.Address, faucetPayload *models.
 		return fmt.Errorf("cannot transfer balance: %w", err)
 	}
 
-	// transfer tx cost to burn address
-	// if tx is set account info (create fresh account) burn tokens here
-	// other case is collect faucet Tx where tokens are burned in a different function
-	// if tx is not set account (no new account) increment nonce of the tx sender
+	// burn the tx fee (by sending to burn address)
+	// SetAccountInfo: burn tokens
+	// CollectFaucetTx: tokens are burned elsewhere, so just increment the sender's nonce
 	if txType == models.TxType_SET_ACCOUNT_INFO {
 		collectFaucetCost, err := v.TxCost(models.TxType_COLLECT_FAUCET, false)
 		if err != nil {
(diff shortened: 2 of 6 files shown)


vochain/account.go
```diff
@@ -459,7 +463,9 @@ func (v *State) SubstractCostIncrementNonce(accountAddress common.Address, txTyp
 		return fmt.Errorf("substractCostIncrementNonce: %w", err)
```
randomshinichi: Spelling is wrong: should be `SubtractCostIncrementNonce`

vochain/account.go
```diff
@@ -283,12 +283,19 @@ func (v *State) ConsumeFaucetPayload(from common.Address, faucetPayload *models.
 	}
 
 	// transfer tx cost to burn address
-	collectFaucetCost, err := v.TxCost(models.TxType_COLLECT_FAUCET, false)
-	if err != nil {
-		return err
-	}
-	if err := accFrom.Transfer(burnAcc, collectFaucetCost); err != nil {
-		return fmt.Errorf("cannot transfer balance: %w", err)
+	// if tx is set account info (create fresh account) burn tokens here
```
randomshinichi: How about this comment which takes less brainpower to parse:

> burn the tx fee (by sending to burn address)
> SetAccountInfo: burn tokens
> CollectFaucetTx: tokens are burned elsewhere, so just increment the sender's nonce

vochain/transaction.go
```diff
@@ -276,9 +276,29 @@ func (app *BaseApplication) AddTx(vtx *VochainTx, commit bool) ([]byte, error) {
 			default:
 				return []byte{}, fmt.Errorf("setAccountDelegate: invalid transaction type")
 			}
-			// substract tx costs and increment nonce
 			return vtx.TxID[:], err
 		}
+	case *models.Tx_CollectFaucet:
+		fromAcc, err := CollectFaucetTxCheck(vtx.Tx, vtx.SignedBody, vtx.Signature, app.State)
+		if err != nil {
+			return []byte{}, fmt.Errorf("collectFaucetTxCheck: %w", err)
```
mvdan: nit: just use `nil` when returning a `[]byte` with an error, the caller shouldn't care and it's shorter and more consistent with other APIs

vochain/account.go
```diff
@@ -692,6 +698,97 @@ func (v *State) SetAccountDelegate(accountAddr, delegateAddr common.Address, txT
...
+		return nil, fmt.Errorf("cannot get faucet account: %v", err)
+	}
+	if issuerAcc == nil {
+		return nil, fmt.Errorf("the account signing the faucet payload does not exist")
+	}
+	cost, err := state.TxCost(models.TxType_COLLECT_FAUCET, false)
+	if err != nil {
+		return nil, fmt.Errorf("cannot get %s tx cost: %w", models.TxType_COLLECT_FAUCET, err)
+	}
+	if (issuerAcc.Balance) < faucetPkg.Payload.Amount+cost {
```
mvdan:
```suggestion
	if issuerAcc.Balance < faucetPkg.Payload.Amount+cost {
```

vochain/account.go
```diff
@@ -692,6 +698,97 @@ func (v *State) SetAccountDelegate(accountAddr, delegateAddr common.Address, txT
...
+		return nil, fmt.Errorf("faucet does not have enough balance %d < %d", issuerAcc.Balance, faucetPkg.Payload.Amount+cost)
+	}
+	// check tx sender is the same as the one contained in the payload
+	if !bytes.Equal(toAddr.Bytes(), faucetPkg.Payload.To) {
+		return nil, fmt.Errorf("txSender %s and faucet payload To %s missmatch",
```
mvdan: is it safe to use `%s` on these bytes? presumably they may contain non-utf8, so perhaps use %x for hex or %q for quoted Go strings, which will escape non-printable chars

mvdan (approved):
I can't really understand how ethless works, but the code looks fine :)

vochain/account.go
altergui: typo: s/missmatch/mismatch/

client/api.go
```diff
@@ -1511,3 +1511,47 @@ func (c *Client) SetAccountDelegate(signer *ethereum.SignKeys, delegate common.A
 	}
 	return nil
 }
+
+// SetDelegate sends a set delegate transaction, if op == true adds a delegate, deletes a delegate otherwise
```
altergui: this comment is misplaced

altergui: (and CollectFaucetTx needs doc)

vochain/account.go
```diff
@@ -692,6 +698,97 @@ func (v *State) SetAccountDelegate(accountAddr, delegateAddr common.Address, txT
...
+		return nil, fmt.Errorf("cannot extract public key from signature: %w", err)
+	}
+	toAddr, err := ethereum.AddrFromPublicKey(pubKey)
+	if err != nil {
+		return nil, fmt.Errorf("cannot extract address from public key: %w", err)
+	}
+	// extract the account that generated the payload
+	faucetPackageBytes, err := proto.Marshal(faucetPkg.Payload)
+	if err != nil {
+		return nil, fmt.Errorf("cannot extract faucet package payload: %v", err)
```
altergui: why %v and not %w?

(maybe it's correct, there's also another %v a couple of returns later, in line 750)

vochain/account.go
altergui: should the edge case `issuerAcc.Balance <= faucetPkg.Payload.Amount+cost` be valid as well? i.e. faucet balance would end up being exactly zero

vochain/account.go
altergui: mmh, i have no idea what are the conventions but i do notice that all of the errors in this PR are consistently lowercase, so either this misspelling applies to all of them or all of them are correct :)

vochain/account.go
```diff
@@ -283,12 +283,19 @@ func (v *State) ConsumeFaucetPayload(from common.Address, faucetPayload *models.
...
+	// if tx is set account info (create fresh account) burn tokens here
+	// other case is collect faucet Tx where tokens are burned in a different function
+	// if tx is not set account (no new account) increment nonce of the tx sender
+	if txType == models.TxType_SET_ACCOUNT_INFO {
+		collectFaucetCost, err := v.TxCost(models.TxType_COLLECT_FAUCET, false)
+		if err != nil {
+			return err
+		}
+		if err := accFrom.Transfer(burnAcc, collectFaucetCost); err != nil {
+			return fmt.Errorf("cannot transfer balance: %w", err)
```
altergui: wouldn't this be "cannot burn fee"? transfer balance sounds confusing to me, but maybe i'm just lacking context.

vochain/account.go
altergui: :+1: much clearer

altergui (commented):
i'm also not fully immersed in this feature yet, but here are some contributions :hugs:

vochain/account.go
jordipainan: changed all of them :) thank you

vochain/account.go
jordipainan: changed thank you, it is more clear now ^^

vochain/account.go
jordipainan: yes the edge case is valid as well, 0 balance is fine

client/api.go
jordipainan: ups, true that, thank you

vochain/account.go
jordipainan: is correct, but we wrap all errors at this level, changed all of them

vochain/account.go
jordipainan: mm not sure, because you are transferring to the burn address but .. thinking a bit more i think is more useful to talk about burn fee in order to have more self explanatory error.

jordipainan:
all comments addressed, thank you @randomshinichi @altergui @mvdan  <3

vochain/account.go
jordipainan: yup, added a better error :) thank you

client/api.go
```diff
@@ -1512,8 +1512,8 @@ func (c *Client) SetAccountDelegate(signer *ethereum.SignKeys, delegate common.A
 	return nil
 }
 
-// SetDelegate sends a set delegate transaction, if op == true adds a delegate, deletes a delegate otherwise
-func (c *Client) CollectFaucetTx(from, to *ethereum.SignKeys, amount, identifier uint64, nonce uint32) error {
+// CollectFaucetTx sends a collect faucet transaction
+func (c *Client) CollectFaucet(from, to *ethereum.SignKeys, amount, identifier uint64, nonce uint32) error {
```
altergui, line 1516: just a minor nit, this `CollectFaucet` name got out of sync with the comment in the previous line. a tip for next time, in VSCode, you can rename the function with F2 and it will also change the comment (along other references in code, of course)

altergui:
woohoo for the merge!

client/api.go
jordipainan: yup! https://github.com/vocdoni/vocdoni-node/pull/523

PR #520, opened 2022-03-21
vochain/scrutinizer: start porting votes to sqlite

mvdan:
(see commit message)


mvdan:
Not sure who else would be a good reviewer here, since Pau is currently out :) @jordipainan feel free to add other reviewers as you see fit

jordipainan:
As far as I can tell LGTM

vochain/scrutinizer/queries/vote_reference.sql
```diff
@@ -0,0 +1,13 @@
...
+INSERT INTO vote_references (
+	nullifier, process_id, height, weight,
+	tx_index, creation_time
+) VALUES (
+	?, ?, ?, ?,
+	?, ?
+);
+
+-- name: GetVoteReference :one
+SELECT * FROM vote_references
```
jordipainan, line 11: Just a question: vote_references contains all vote references of all processes or all vote references for a given process? I understand it contains all vote references of all processes.
If that is the case could this query be faster filtering first by processId ?

Thank you

altergui:
phew... no way. i read it, but i can't make a single comment or review  :no_mouth: other than "i trust you both" :wink:

vochain/scrutinizer/queries/vote_reference.sql
mvdan: the tables are all global, so it contains entries for all processes.

I haven't spent any time optimizing the queries or tables, beyond adding column indexes where it makes sense for sure :) it's entirely likely that we'll want indexes on vote_references by process ID as well, but I'd rather do that in the future when we can prove with some benchmarks that it actually helps.

mvdan:
@jordipainan can you please approve? Merging is blocked until then. Or if you don't want to approve, I'm not sure who else can :) Usually it would be Pau. Not sure if @emmdim has the bandwidth.

G10h4ck (approved):
I have reviewed the changes and looks fine to me, but I am the one who has read that code less times, so take my review with a grain of salt, and use it for merging just if necessary if no one else has different opinion ;)

PR #523, opened 2022-03-23
fix comment on api/client CollectFaucet

jordipainan:


altergui:
possibly as well:
vochain/account_test.go:448
vochain/account_test.go:540
vochain/transaction.go:298

(i understand the "TxCheck" references shouldn't lose the "Tx" part)

jordipainan:
> possibly as well: vochain/account_test.go:448 vochain/account_test.go:540 vochain/transaction.go:298
> 
> (i understand the "TxCheck" references shouldn't lose the "Tx" part)

are fine there, what we can do is to normalize to use Tx as suffix or not, but this is out of the scope, as a lot more things must keep advancing lets freeze this until we have enough resources

PR #528, opened 2022-03-23
CI: avoid duplicate runs when doing a PR from our own branches

altergui:
this cuts in half the amount of running jobs for most of our PRs (which are done using branches in this same repo, so tests are run both for the "push" and for the "pull_request" unnecessarily)

with this change, only one job is run per commit:
either on push (and skipped on pull_request) for branches of this repo,
or on pull_request (and not on push) for branches of forks.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -2,6 +2,15 @@
 name: Main
 
 on: [push, pull_request]
+# Actually we don't want to run unconditionally on every push and pull_request:
+# 1. We want this to run on any push to any branch in *our* repo, but not in personal forks
+#   since they don't run on our significantly faster nodes, so they fail and cause confusion.
+#   i.e. github.event_name == 'push' && github.event.repository.fork == false
+# 2. We also want to run on external PRs, but not on our own internal PRs as they'll be run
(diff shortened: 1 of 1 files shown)


altergui:
this also means that the checks ends earlier, since the run starts when you *push* the changes over git, and by the time you create the PR, the job is already running. The job for "pull_request" is skipped (before, you had to wait for the dup job to finish, and this dup job started only after creating the PR)

altergui:
you can see this in action in this very PR: `go-tests` and `compose-test` are only run once (for push, even before this PR was created) and skipped for pull_request

mvdan (approved):
SGTM, but note that a PR from a branch in our main repo isn't really a situation where you have duplicate jobs. The push job runs on the branch as-is, but the pull_request job runs on the _merge_ of the source branch on the target/base branch. This can matter if the PR's source branch is behind by a few commits.

In general I think this is a good idea, but still FYI that the duplication isn't entirely pointless.

altergui:
:facepalm: now i'm ashamed, i didn't realise this crucial difference. makes total sense to run both, (or, if i had to choose, actually the pull_request one). nothing to see here, i'm going back to my cave :grimacing:

PR #529, opened 2022-03-24
vocli 1.0

randomshinichi:
vocli is a user-friendly CLI to send transactions to the vochain.

It can manage keys, send SetAccountInfoTx for an account, mint, send tokens, generate and claim faucet payloads.

Voting functionality is not included yet.

An integration test is included that does not test all possible options - instead it tests most functionality and requires that certain things be said in the output to help new users along.

### DIFF
--- a/cmd/vocli/commands/account.go
+++ b/cmd/vocli/commands/account.go
@@ -1,0 +1,117 @@
+package commands
+
+import (
+	"fmt"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/spf13/cobra"
+	"go.vocdoni.io/dvote/client"
+)
+
+var accInfoCmd = &cobra.Command{
+	Use:   "info",
+	Short: "Get information about an account from the vochain.",
+	Args:  cobra.ExactArgs(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		c, err := client.New(gatewayRpc)
+		if err != nil {
+			return err
+		}
+		resp, err := c.GetAccount(common.HexToAddress(args[0]))
+		if err != nil {
+			return err
+		}
+		fmt.Fprintln(Stdout, resp.String())
+		return err
+	},
+}
+
+var accSetInfoCmd = &cobra.Command{
+	Use:   "set <keystore, ipfs://info-URI>",
+	Short: "Create/Update an account on the vochain",
+	Args:  cobra.ExactArgs(2),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		key, _, err := openKeyfile(args[0], "Please unlock your key: ")
+		if err != nil {
+			return err
+		}
+		infoUri := args[1]
+
+		if err := createAccount(key, gatewayRpc, infoUri, faucetHex); err != nil {
+			return err
+		}
+		return nil
+	},
+}
+
+var accTreasurerCmd = &cobra.Command{
+	Use:   "treasurer",
+	Short: "Get information about the treasurer account (a special account) on the vochain.",
+	Args:  cobra.ExactArgs(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		c, err := client.New(gatewayRpc)
+		if err != nil {
+			return err
+		}
+
+		resp, err := c.GetTreasurer()
+		if err != nil {
+			return err
+		}
+		fmt.Fprintln(Stdout, "address", common.BytesToAddress(resp.Address))
+		fmt.Fprintln(Stdout, "nonce", resp.Nonce)
+		return err
+	},
+}
+
+var accAddDelegateCmd = &cobra.Command{
+	Use:   "add-delegate <keystore> <address>",
+	Short: "Add a delegate address for an account",
+	Args:  cobra.ExactArgs(2),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		c, err := client.New(gatewayRpc)
+		if err != nil {
+			return err
+		}
+
+		_, signer, err := openKeyfile(args[0], "Please unlock your key: ")
+		if err != nil {
+			return err
+		}
+
+		nonce, err := getNonce(c, signer.Address().String())
+		if err != nil {
+			return err
+		}
+
+		return c.SetAccountDelegate(signer, common.HexToAddress(args[1]), true, nonce)
+	},
+}
+
+var accDelDelegateCmd = &cobra.Command{
+	Use:   "rm-delegate <keystore> <address>",
+	Short: "Remove a delegate address for an account",
+	Args:  cobra.ExactArgs(2),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		c, err := client.New(gatewayRpc)
+		if err != nil {
+			return err
+		}
+
+		_, signer, err := openKeyfile(args[0], "Please unlock your key: ")
+		if err != nil {
+			return err
+		}
+
+		nonce, err := getNonce(c, signer.Address().String())
+		if err != nil {
+			return err
+		}
+
+		return c.SetAccountDelegate(signer, common.HexToAddress(args[1]), false, nonce)
+	},
+}
+var accCmd = &cobra.Command{
+	Use:   "account",
+	Short: "Manage a key's account. Accounts are stored on the vochain and are controlled by keys.",
+}
--- a/cmd/vocli/commands/helpers.go
+++ b/cmd/vocli/commands/helpers.go
@@ -1,0 +1,61 @@
+package commands
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path"
+	"path/filepath"
+	"time"
+
+	"github.com/ethereum/go-ethereum/common"
+)
+
+func writeKeyFile(filename string, k []byte) error {
+	// Create the keystore directory with appropriate permissions
+	// in case it is not present yet.
+	const dirPerm = 0700
+	if err := os.MkdirAll(filepath.Dir(filename), dirPerm); err != nil {
+		return err
+	}
+
+	// Here we diverge from geth. No need for atomic write? just write it out
+	// already
+	if err := ioutil.WriteFile(filename, k, 0600); err != nil {
+		return err
+	}
+	return nil
+}
+
+// getDataDir returns ~/.dvote when passed a "" as an argument, and <path>/~.dvote otherwise
+func getDataDir(root string) (string, error) {
+	var err error
+	if root == "" {
+		root, err = os.UserHomeDir()
+	}
+	if err != nil {
+		return "", err
+	}
+
+	return path.Join(root, ".dvote"), nil
+}
+
+func getKeysDir(root string) (string, error) {
+	dataDir, err := getDataDir(root)
+	if err != nil {
+		return "", err
+	}
+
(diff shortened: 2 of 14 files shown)


altergui:
go-tests is failing because voconed spawns an ipfs daemon in default port 5001 and it should use a random port (in tests)

i'm not sure how to do this for voconed, but here's the snippet from a test dvote server
https://github.com/vocdoni/vocdoni-node/blob/468207e337d3379c126f0fc01dd0cd2ee67a383f/test/testcommon/apis.go#L47-L64

actually i think you'll need to write a function in that same spirit but for "test voconed server", and call that instead of NewVocone() since there's no way to pass a custom (random) IPFS port to NewVocone()

cmd/vocli/commands/account.go
```diff
@@ -0,0 +1,54 @@
...
+	"strings"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/spf13/cobra"
+	"go.vocdoni.io/dvote/client"
+)
+
+var accInfoCmd = &cobra.Command{
+	Use:   "info",
+	Short: "Get information about an account from the vochain. The address may or may not include the 0x prefix",
```
jordipainan: Maybe is not really needed to say that the account can be passed with 0x or without. If we want to point this we should also point that the address can be checksumed or not.

cmd/vocli/commands/account.go
```diff
@@ -0,0 +1,54 @@
...
+		if err != nil {
+			return err
+		}
+		fmt.Fprintln(Stdout, "resp", resp.String(), "err", err)
+		return err
+	},
+}
+
+var accSetInfoCmd = &cobra.Command{
+	Use:   "set <keystore, ipfs://info-URI>",
```
jordipainan, line 30: `ipfs://` prefix is not required, but optional, you can see data origins on the docs

cmd/vocli/commands/account.go
```diff
@@ -0,0 +1,54 @@
...
+	Short: "Create/Update an Account on the vochain",
+	Args:  cobra.ExactArgs(2),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		key, _, err := openKeyfile(args[0], "Please unlock your key: ")
+		if err != nil {
+			return err
+		}
+		infoUri := args[1]
+		if !strings.HasPrefix(infoUri, "ipfs://") {
```
jordipainan: Following the last comment, this check is not required

cmd/vocli/commands/account.go
```diff
@@ -0,0 +1,54 @@
...
+		if err := createAccount(key, gatewayRpc, infoUri, faucetHex); err != nil {
+			return err
+		}
+		return nil
+	},
+}
+
+var accCmd = &cobra.Command{
+	Use:   "account",
+	Short: "Manage the Account information stored in the vochain for a particular key",
```
jordipainan: If you are going to create an account the account is not yet stored in the Vochain, so I think we should change this command description

cmd/vocli/commands/keys-eth.go
```diff
@@ -0,0 +1,52 @@
+package commands
```
jordipainan, line 1: What is the reasoning for having different files for keys?. I think given that this file have 3 functions, it can be merged into keys.go file.
But I do not strongly oppose to this approach :)

cmd/vocli/commands/keys.go
```diff
@@ -0,0 +1,267 @@
...
+			if err != nil {
+				return "", err
+			}
+			return strings.TrimSpace(p), nil // newPassword
+		}
+	}
+	return password, nil
+}
+
+func createAccount(key *ethkeystore.Key, gatewayRpc, infoUri, faucetPkg string) error {
```
jordipainan, line 224: Should this function be included in the account.go file as it has to do with accounts ?

cmd/vocli/commands/keys.go
```diff
@@ -0,0 +1,267 @@
...
+		if err != nil {
+			return err
+		}
+	}
+
+	fmt.Fprintf(Stdout, "Account created/updated on chain %s\n", gatewayRpc)
+	return nil
+}
+
+func parseFaucetPayloadHex(faucetPayload string) (*models.FaucetPackage, error) {
```
jordipainan, line 260: Same as before

cmd/vocli/commands/helpers.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+func generateKeyFilename(address common.Address) (string, error) {
+	keysDir, err := getKeysDir(home)
+	if err != nil {
+		return "", fmt.Errorf("couldn't get a suitable datadir (normally ~/.dvote/keys) - %v", err)
+	}
+
+	t := time.Now()
+	keyFilename := fmt.Sprintf("%s-%d-%d-%d%s", address.Hex(), t.Year(), t.Month(), t.Day(), keyExt) // 0xDad23752bB8F80Bb82E6A2422A762fdeAf8dAb74-2022-1-13.vokey
```
jordipainan, line 59: This line is too long, can you add the comment in top of the code line?

cmd/vocli/commands/keys.go
```diff
@@ -0,0 +1,267 @@
...
+const (
+	scryptN = ethkeystore.StandardScryptN
+	scryptP = ethkeystore.StandardScryptP
+	keyExt  = ".vokey"
+)
+
+var keysCmd = &cobra.Command{
+	Use:   "keys",
+	Short: "Do things related to keys",
+	Long:  "In Vochain, we make the distinction between a key and an account. A Key controls an Account which is stored in the blockchain's state. The Key itself is stored in encrypted form on the user's disk.",
```
jordipainan: This line is too long, maybe you can split using `` instead of ""

cmd/vocli/commands/keys.go
```diff
@@ -0,0 +1,267 @@
...
+}
+
+var keysNewCmd = &cobra.Command{
+	Use:   "new [pre-existing keyfile]",
+	Short: "Generate a new key and create its corresponding Account on the chain. Or just create the Account onchain.",
+	Long:  "Generate a new key and send a SetAccountInfo Tx in one step. Or, if you already have a keyfile, send a SetAccountInfo Tx.",
```
jordipainan: This line is too long

cmd/vocli/commands/keys.go
```diff
@@ -0,0 +1,267 @@
...
+	Args:  cobra.MaximumNArgs(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		var key *ethkeystore.Key
+		var keyPath string
+
+		fmt.Fprintf(Stdout, "Step 1/2: First we will generate the key and save it on your disk, encrypted.\nUnlike other blockchains, an Account for this key must first be created on the blockchain for others to send you funds (or any other operations).\nOnce the key is generated and saved, we will send a SetAccountInfo transaction to the blockchain in order to create an Account for this key. \nFear not, if you have only generated a key, you may skip the key generation process and directly send the SetAccountInfo transaction by re-running 'keys new <full path to keyfile>'\n\n")
```
jordipainan: This line is too long

cmd/vocli/commands/root.go
```diff
@@ -0,0 +1,212 @@
...
+func init() {
+	Stdout = os.Stdout
+	Stderr = os.Stderr
+	Stdin = os.Stdin
+	RootCmd.CompletionOptions.DisableDefaultCmd = true
+	RootCmd.PersistentFlags().StringVarP(&gatewayRpc, "url", "u", "https://gw1.dev.vocdoni.net/dvote", "Gateway RPC URL")
+	RootCmd.PersistentFlags().StringVar(&home, "home", "", "root directory where all vochain files are stored (normally ~/.dvote)")
+	RootCmd.PersistentFlags().StringVar(&password, "password", "", "supply the password as an argument instead of prompting")
+	RootCmd.PersistentFlags().BoolVarP(&debug, "debug", "d", false, "prints additional information")
+	RootCmd.PersistentFlags().Uint32VarP(&nonce, "nonce", "n", 0, "account nonce to use when sending transaction (useful when it cannot be queried ahead of time, e.g. offline transaction signing)")
```
jordipainan: This line is too long

cmd/vocli/commands/root.go
```diff
@@ -0,0 +1,212 @@
...
+	RootCmd.AddCommand(keysCmd)
+	accCmd.AddCommand(accInfoCmd)
+	accCmd.AddCommand(accSetInfoCmd)
+	keysCmd.AddCommand(keysNewCmd)
+	keysCmd.AddCommand(keysImportCmd)
+	keysCmd.AddCommand(keysListCmd)
+	keysCmd.AddCommand(keysChangePasswordCmd)
+	keysCmd.AddCommand(keysShowPrivKeyCmd)
+
+	keysNewCmd.Flags().StringVar(&faucetHex, "faucet", "", "specify an optional hex-encoded faucet payload to immediately top up the new account with tokens")
```
jordipainan: This line is too long

jordipainan (changes requested):

jordipainan:
> go-tests is failing because voconed spawns an ipfs daemon in default port 5001 and it should use a random port (in tests)
> 
> i'm not sure how to do this for voconed, but here's the snippet from a test dvote server
> 
> https://github.com/vocdoni/vocdoni-node/blob/468207e337d3379c126f0fc01dd0cd2ee67a383f/test/testcommon/apis.go#L47-L64
> 
> actually i think you'll need to write a function in that same spirit but for "test voconed server", and call that instead of NewVocone() since there's no way to pass a custom (random) IPFS port to NewVocone()

Agree with this, there is also another error with the census `panic: duplicate metrics collector registration attempted [recovered]`

jordipainan:
@randomshinichi There are 3 missing methods:
- AddDelegate for an account
- DelDelegate for an account
- SetTransactionCosts made by the treasurer

Adding them to this PR also fine :)

jordipainan:
Re-request review when comments are addressed, thank you!

cmd/vocli/commands/keys.go
randomshinichi: Do you mean i should use `` in Go comments or i should rephrase the help message?

jordipainan: Use `` :)

cmd/vocli/commands/account.go
randomshinichi: `The URI must be http(s):// or ipfs:// prefixed and should have a maximum length (256 bytes?)` ?
from Ethereum-less-vochain-description

cmd/vocli/commands/account.go
randomshinichi: `Manage a key's Account. Accounts are stored on the vochain.`
How about this?

cmd/vocli/commands/keys-eth.go
randomshinichi: Functions in this file are the shim between us and the Ethereum libraries. Keeps things neat.

p4u:
Is there a blocker for this PR @randomshinichi?

randomshinichi:
> Is there a blocker for this PR @randomshinichi?

No, just that the integration tests fail most of the time but it is not related to this code.

PR #532, opened 2022-03-28
Add lock to vocone blocks

NateWilliams2:
as suggested in https://github.com/vocdoni/vaas-api/pull/42#discussion_r836505952
adds a lock to vocone block creation so helpers like `mintTokens` can be used without race conditions


jordipainan (commented):
lgtm

G10h4ck:
looks good to me too, the CI seems to fail for an apparently unrelated thing

mvdan (approved):
SGTM. I'd just document what the lock actually protects - I assume it's the `app` field, but best not to guess.

PR #545, opened 2022-04-26
Integrate ethless process logic

jordipainan:
- *[BREAKING CHANGES]* Adds the logic for NewProcessTx and SetProcessTx to work with the ethless approach.
- Adapts the testing both unit and e2e.
- E2E tests now can be run serially if specified.
- Adds some other minor changes to pack all together.


jordipainan:
Done together with @altergui

client/api.go
```diff
@@ -1104,12 +1105,19 @@ func (c *Client) CreateProcess(oracle *ethereum.SignKeys,
 		VoteOptions:   &models.ProcessVoteOptions{MaxCount: 16, MaxValue: 8},
 		MaxCensusSize: &maxCensusSize,
```
p4u, line 1106: BESITOS :*

p4u (approved):

client/api.go
mvdan: :scream_cat:

cmd/vochaintest/vochaintest.go
```diff
@@ -123,9 +145,9 @@ func main() {
 			false)
 	case "vtest":
 		mkTreeVoteTest(*host,
-			*oraclePrivKey,
+			oracleKey,
 			*electionType == "encrypted-poll",
-			entityKey,
+			entityKeys[2],
```
p4u: This looks like a very static approach and hard to understand if the code is not read. Also it is a not consistent since opmode can only be one per execution so we are creating entityKeys that are not being used. Consider the following changes:

- I'd change the naming to `accountKeys`, seems more consistent with the rest of the namings.

- `initaccounts` would take `accountKeys` and `treasurerKey`. Initialize the whole list of account keys provided.

- A test operation such as vtest uses the `accountKeys[0]` only, the rest of test operation also use only `accountKeys[0]`. If `accountKeys` is empty then it requires `treasurerKey` in order to generate a new random key and initialize it.

cmd/vochaintest/vochaintest.go
```diff
@@ -94,23 +95,44 @@ func main() {
...
 			log.Fatal(err)
 		}
 	} else {
-		if err := entityKey.Generate(); err != nil {
+		if err := oracleKey.Generate(); err != nil {
 			log.Fatal(err)
 		}
 	}
+	// create treasurer key
+	if len(*treasurerPrivKey) == 0 {
```
p4u: We should not require always the treasurerPrivKey, just if its needed

cmd/vochaintest/vochaintest.go
```diff
@@ -142,9 +164,9 @@ func main() {
 			log.Fatal(err)
 		}
 		cspVoteTest(*host,
-			*oraclePrivKey,
+			oracleKey,
```
p4u: Is oracleKey needed? I don't think so since we are using fresh accounts

cmd/vochaintest/vochaintest.go
```diff
@@ -238,8 +260,177 @@ func censusImport(host string, signer *ethereum.SignKeys) {
...
+func waitUntilNextBlock(mainClient *client.Client) error {
+	h, err := mainClient.GetCurrentBlock()
+	if err != nil {
+		return fmt.Errorf("cannot get current height: %w", err)
+	}
+	mainClient.WaitUntilBlock(h + 1)
+	return nil
+}
+
+func ensureProcessCreated(
```
p4u: We don't use this kind of line split in general

cmd/vochaintest/vochaintest.go
```diff
@@ -238,8 +260,177 @@ func censusImport(host string, signer *ethereum.SignKeys) {
...
+		waitUntilNextBlock(mainClient)
+		p, err := mainClient.GetProcessInfo(pid)
+		if err != nil {
+			log.Warnf("ensureProcessCreated: cannot get process %x (%s)", pid, err.Error())
+		}
+		if p != nil {
+			log.Infof("ensureProcessCreated: got process %x info %+v", pid, p)
+			return start, nil
+		}
+		if i == retries { // that was the last chance, break and fail immediately
```
p4u: No need to break here, just set the `for` condition as `i<retries`

cmd/vochaintest/vochaintest.go
```diff
@@ -238,8 +260,177 @@ func censusImport(host string, signer *ethereum.SignKeys) {
...
+		waitUntilNextBlock(mainClient)
+		p, err := mainClient.GetProcessInfo(processID)
+		if err != nil {
+			log.Warnf("ensureProcessEnded: cannot force end process: cannot get process %x info: %s", processID, err.Error())
+		}
+		log.Infof("ensureProcessEnded: got process %x info %+v", processID, p)
+		if p != nil && p.Status == int32(models.ProcessStatus_ENDED) || p.Status == int32(models.ProcessStatus_RESULTS) {
+			return nil
+		}
+		if i == retries { // that was the last chance, break and fail immediately
```
p4u, line 317: same

cmd/vochaintest/vochaintest.go
```diff
@@ -238,8 +260,177 @@ func censusImport(host string, signer *ethereum.SignKeys) {
...
+	}
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer mainClient.Close()
+
+	retries := 10
+	// create and top up accounts
+	if err := ensureAccountExists(mainClient, oracle, retries); err != nil {
+		log.Fatal(err)
```
p4u, line 351: If the function returns an error, Fatal should not be used

cmd/vochaintest/vochaintest.go
```diff
@@ -238,8 +260,177 @@ func censusImport(host string, signer *ethereum.SignKeys) {
...
+}
+
+func ensureAccountExists(mainClient *client.Client, account *ethereum.SignKeys, retries int) error {
+	for i := 0; i <= retries; i++ {
+		// if account exists, we're done
+		if acct, _ := mainClient.GetAccount(account.Address()); acct != nil {
+			return nil
+		}
+		if i == retries { // that was the last chance, break and fail immediately
+			break
```
p4u: This is a unnecessary break

cmd/vochaintest/vochaintest.go
```diff
@@ -238,8 +260,177 @@ func censusImport(host string, signer *ethereum.SignKeys) {
...
+	treasurer *ethereum.SignKeys,
+	accountAddr common.Address,
+	amount uint64,
+	retries int) error {
+	for i := 0; i <= retries; i++ {
+		// if balance is enough, we're done
+		if acct, _ := mainClient.GetAccount(accountAddr); acct != nil && acct.Balance > 1000 {
+			return nil
+		}
+		if i == retries { // that was the last chance, break and fail immediately
```
p4u: same

p4u (changes requested):

cmd/vochaintest/vochaintest.go
jordipainan: As some accounts must be always initialized, the treasurer must be able to send those transactions, without the treasurer the accounts start just empty and no test can be run.  I think we should always require the treasurer addr

cmd/vochaintest/vochaintest.go
jordipainan: I used for setting the process status to end so is checked that the oracle can set arbitrary processes, but I'm doing it in other functions so we can remove the oracle key here.

cmd/vochaintest/vochaintest.go
p4u: Only if it is required. Consider the caller knows what is doing. He might been funded the accounts previously.

p4u (approved):

PR #546, opened 2022-04-26
adapt admintx to treasurer

jordipainan:
**BREAKING** Vochain state

This PR changes the oracle checks for admin transactions. The treasurer is now the only admin account that can make some admin transactions. 
Disables the event handling for genesis events fired from the genesis smart contract. Deleted unused code.
I've not totally cleaned the references to the genesis contract cause a more in depth cleaning must be done and for now i think is fine to just ignore the events.

Add/Remove oracle can only be done by the Treasurer.
Add/Reveal process keys can still be done by the oracle address acting as a Keykeeper, without any cost and without increasing the nonce for the oracle account.
Add/Remove validator logic deactivated cause is not yet supported in the current release.

### DIFF
--- a/ethereum/ethevents/handlers.go
+++ b/ethereum/ethevents/handlers.go
@@ -250,95 +250,96 @@ func HandleVochainOracle(ctx context.Context, event *ethtypes.Log, e *EthereumEv
 			return fmt.Errorf("cannot broadcast tx: %w, res: %+v", err, res)
 		}
 		log.Infof("oracle transaction sent, hash: %x", res.Hash)
-
-	case ethereumEventList["genesisOracleAdded"]:
-		log.Infof("executing GenesisOracleAdd event")
-		tctx, cancel := context.WithTimeout(ctx, time.Minute)
-		defer cancel()
-		addOracleTx, err := genesisOracleAddedMeta(tctx,
-			&e.ContractsInfo[ethereumhandler.ContractNameGenesis].ABI, event.Data, e.VotingHandle)
-		if err != nil {
-			return fmt.Errorf("cannot obtain addOracle data for creating the transaction: %w", err)
-		}
-		log.Infof("found add oracle %x genesis event on ethereum", addOracleTx.Address)
-		oracles, err := e.VochainApp.State.Oracles(true)
-		if err != nil {
-			return fmt.Errorf("cannot fetch the oracle list from the Vochain: %w", err)
-		}
-		// check oracle not already on the Vochain
-		for idx, o := range oracles {
-			if o == common.BytesToAddress(addOracleTx.Address) {
-				return fmt.Errorf("cannot add oracle, already exists on the Vochain at position: %d", idx)
+		/* DEPRECATED
+		case ethereumEventList["genesisOracleAdded"]:
+			log.Infof("executing GenesisOracleAdd event")
+			tctx, cancel := context.WithTimeout(ctx, time.Minute)
+			defer cancel()
+			addOracleTx, err := genesisOracleAddedMeta(tctx,
+				&e.ContractsInfo[ethereumhandler.ContractNameGenesis].ABI, event.Data, e.VotingHandle)
+			if err != nil {
+				return fmt.Errorf("cannot obtain addOracle data for creating the transaction: %w", err)
 			}
-		}
-		// create admin tx
-		stx := &models.SignedTx{}
-		stx.Tx, err = proto.Marshal(&models.Tx{Payload: &models.Tx_Admin{Admin: addOracleTx}})
-		if err != nil {
-			return fmt.Errorf("cannot marshal admin tx (addOracle): %w", err)
-		}
-		stx.Signature, err = e.Signer.SignVocdoniTx(stx.Tx, e.VochainApp.ChainID())
-		if err != nil {
-			return fmt.Errorf("cannot sign oracle tx: %w", err)
-		}
-		tx, err := proto.Marshal(stx)
-		if err != nil {
-			return fmt.Errorf("error marshaling admin tx (addOracle) tx: %w", err)
-		}
-		log.Debugf("broadcasting tx: %s", log.FormatProto(addOracleTx))
-
-		res, err := e.VochainApp.SendTx(tx)
-		if err != nil || res == nil {
-			return fmt.Errorf("cannot broadcast tx: %w, res: %+v", err, res)
-		}
-		log.Infof("oracle transaction sent, hash: %x", res.Hash)
-	case ethereumEventList["genesisOracleRemoved"]:
-		log.Infof("executing GenesisOracleRemove event")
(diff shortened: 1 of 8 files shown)


jordipainan:
closes #542

ethereum/ethevents/handlers.go
```diff
@@ -250,95 +250,96 @@ func HandleVochainOracle(ctx context.Context, event *ethtypes.Log, e *EthereumEv
...
-		log.Infof("found add oracle %x genesis event on ethereum", addOracleTx.Address)
-		oracles, err := e.VochainApp.State.Oracles(true)
-		if err != nil {
-			return fmt.Errorf("cannot fetch the oracle list from the Vochain: %w", err)
-		}
-		// check oracle not already on the Vochain
-		for idx, o := range oracles {
-			if o == common.BytesToAddress(addOracleTx.Address) {
-				return fmt.Errorf("cannot add oracle, already exists on the Vochain at position: %d", idx)
+		/* DEPRECATED
```
p4u: Just remove the code

vochain/transaction.go
```diff
@@ -560,19 +560,25 @@ func AdminTxCheck(vtx *models.Tx, txBytes, signature []byte, state *State) error
...
+		return fmt.Errorf("cannot extract public key from signature: %w", err)
+	}
+	addr, err := ethereum.AddrFromPublicKey(pubKey)
+	if err != nil {
+		return fmt.Errorf("cannot extract address from public key: %w", err)
+	}
+	log.Debugf("checking admin signed tx %+v by addr %s", tx, addr.String())
+	log.Debugf("got treasurer addr %s", common.BytesToAddress(treasurer.Address).String())
+	if !bytes.Equal(addr.Bytes(), treasurer.Address) {
+		return fmt.Errorf("signature extracted address does not match with treasurer address")
```
p4u: Since this is an error we are returning to the users I'd do something more high level such as `not authorized for executing admin transactions`

vochain/transaction.go
```diff
@@ -560,19 +560,25 @@ func AdminTxCheck(vtx *models.Tx, txBytes, signature []byte, state *State) error
...
+	pubKey, err := ethereum.PubKeyFromSignature(txBytes, signature)
+	if err != nil {
+		return fmt.Errorf("cannot extract public key from signature: %w", err)
+	}
+	addr, err := ethereum.AddrFromPublicKey(pubKey)
+	if err != nil {
+		return fmt.Errorf("cannot extract address from public key: %w", err)
+	}
+	log.Debugf("checking admin signed tx %+v by addr %s", tx, addr.String())
+	log.Debugf("got treasurer addr %s", common.BytesToAddress(treasurer.Address).String())
```
p4u: just `"%x", treasurer.Address` would work

p4u (changes requested):

p4u:
> Add/Remove validator logic deactivated cause is not yet supported in the current release.

I think we agreed on leaving this logic up to the treasurer in order to be able to add/reomve validators in case of some urgency.  What is not supported? I think its supported while the transaction is build and sent to the blockchain, isn't?

vochain/transaction.go
```diff
@@ -563,106 +560,165 @@ func (app *BaseApplication) VoteEnvelopeCheck(ve *models.VoteEnvelope, txBytes,
...
-
+		// get oracles
+		oracles, err := state.Oracles(false)
+		if err != nil || len(oracles) == 0 {
+			return common.Address{}, fmt.Errorf("cannot check authorization against a nil or empty oracle list")
+		}
+		if authorized, addr, err := verifySignatureAgainstOracles(
+			oracles, txBytes, signature); err != nil {
+			return common.Address{}, err
+		} else if !authorized {
```
p4u: The else is redundant

vochain/transaction.go
```diff
@@ -563,106 +560,165 @@ func (app *BaseApplication) VoteEnvelopeCheck(ve *models.VoteEnvelope, txBytes,
...
+
+	pubKey, err := ethereum.PubKeyFromSignature(txBytes, signature)
+	if err != nil {
+		return common.Address{}, fmt.Errorf("cannot extract public key from signature: %w", err)
+	}
+	addr, err := ethereum.AddrFromPublicKey(pubKey)
+	if err != nil {
+		return common.Address{}, fmt.Errorf("cannot extract address from public key: %w", err)
 	}
+	log.Debugf("checking admin signed tx %+v by addr %x", tx, addr)
```
p4u: When printing protobuf use `"%s", log.FormatProtobuf(tx)`

vochain/transaction.go
```diff
@@ -563,106 +560,165 @@ func (app *BaseApplication) VoteEnvelopeCheck(ve *models.VoteEnvelope, txBytes,
...
-			return fmt.Errorf("process does not require keys")
+			return common.Address{}, fmt.Errorf("process does not require keys")
 		}
-
+		// get oracles
+		oracles, err := state.Oracles(false)
+		if err != nil || len(oracles) == 0 {
+			return common.Address{}, fmt.Errorf("cannot check authorization against a nil or empty oracle list")
+		}
+		if authorized, addr, err := verifySignatureAgainstOracles(
```
p4u: This line break is weird

vochain/transaction.go
```diff
@@ -563,106 +560,165 @@ func (app *BaseApplication) VoteEnvelopeCheck(ve *models.VoteEnvelope, txBytes,
...
-	if authorized, addr, err := verifySignatureAgainstOracles(
-		oracles, txBytes, signature); err != nil {
-		return err
-	} else if !authorized {
-		return fmt.Errorf("unauthorized to perform an adminTx, address: %s", addr.Hex())
+	// only treasurer can do admin txs
+	// oracles (keykeepers) can still add and reveal process keys
+	if tx.Txtype != models.TxType_ADD_PROCESS_KEYS && tx.Txtype != models.TxType_REVEAL_PROCESS_KEYS {
+		// get treasurer
+		treasurer, err := state.Treasurer(false)
```
p4u: Better this code in a function to check if the address is treasurer, it might be reused

vochain/transaction.go
```diff
@@ -563,106 +560,165 @@ func (app *BaseApplication) VoteEnvelopeCheck(ve *models.VoteEnvelope, txBytes,
...
+		if !bytes.Equal(addr.Bytes(), treasurer.Address) {
+			return common.Address{}, fmt.Errorf("not authorized for executing admin transactions")
+		}
+		// check treasurer account
+		if treasurer.Nonce != tx.Nonce {
+			return common.Address{}, ErrAccountNonceInvalid
+		}
 	}
 
 	switch tx.Txtype {
```
p4u, line 583: This is weird, this switch appears twice and there is a previous IF doing the same query.
I'm pretty sure this can be packed in a more elegant and clear way.

p4u (changes requested):

p4u (approved):

PR #547, opened 2022-04-26
vochain: implements deterministic process ID type

p4u:
Introduce a new type `ProcessID` with Marshal and Unmarshal methods.

The new identifier holds information about the process
- chainID: a 6 bytes trunked hash of the blockchain identifier
- organization: hex address of the wallet creating the process
- census proof type: kind of census used for the process (from protobuf models)
- envelope type: flags for configuring the ballot counting mode
- nonce: an incremental uint32 number

So processID: 395b41cc9abcd8da6bf26964af9d7eed9e03e53415d37aa96045062100000001  represents
- chID: vocdoni-azeno
- addr: 0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045
- proof: arbo
- envelopetype: 0x21
- nonce: 1

This commit only introduces the new type, not the required changes on the blockchain logic.

Discussion: https://github.com/vocdoni/protocol/discussions/28

### DIFF
--- a/vochain/process_id.go
+++ b/vochain/process_id.go
@@ -1,0 +1,178 @@
+package vochain
+
+import (
+	"encoding/binary"
+	"fmt"
+
+	"github.com/ethereum/go-ethereum/common"
+	"go.vocdoni.io/dvote/crypto/ethereum"
+	"go.vocdoni.io/dvote/log"
+	"go.vocdoni.io/dvote/types"
+	"go.vocdoni.io/proto/build/go/models"
+)
+
+// ProcessID is a 32 bytes identifier that holds the following information about the voting process:
+//
+// - chainID: a 6 bytes trunked hash of the blockchain identifier
+//
+// - organization: hex address of the wallet creating the process
+//
+// - census origin: kind of census used for the process (from protobuf models)
+//
+// - envelope type: ballot properties configuration
+//
+// - nonce: an incremental uint32 number
+//
+// processID: 395b41cc9abcd8da6bf26964af9d7eed9e03e53415d37aa96045000600000001 represents
+// chID:vocdoni-azeno addr:0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045 pt:arbo nc:1
+type ProcessID struct {
+	chainID          string
+	organizationAddr common.Address
+	censusOrigin     uint8
+	envType          uint8
+	nonce            uint32
+}
+
+// Marshal encodes to bytes
+func (p *ProcessID) Marshal() []byte {
+	// processID =
+	//  chainID[6bytes] + organizationAddress[20bytes] + proofType[1byte] + envelopeType[1byte] + nonce[4bytes]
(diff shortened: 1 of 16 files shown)


jordipainan (approved):
LGTM!

vochain/process_id.go
```diff
@@ -0,0 +1,120 @@
...
+// chID:vocdoni-azeno addr:0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045 pt:arbo nc:1
+type ProcessID struct {
+	chainID          string
+	organizationAddr *common.Address
+	proofType        uint16
+	nonce            uint32
+}
+
+// Marshal encodes to bytes
+func (p *ProcessID) Marshal() []byte {
```
mvdan, line 37: might as well implement https://pkg.go.dev/encoding#BinaryMarshaler and its unmarshaler counterpart

vochain/process_id.go
```diff
@@ -0,0 +1,120 @@
...
+//
+// - census proof type: kind of census used for the process (from protobuf models)
+//
+// - nonce: an incremental uint32 number
+//
+// processID: 395b41cc9abcd8da6bf26964af9d7eed9e03e53415d37aa96045000600000001 represents
+// chID:vocdoni-azeno addr:0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045 pt:arbo nc:1
+type ProcessID struct {
+	chainID          string
+	organizationAddr *common.Address
```
mvdan: I don't think you should be using a pointer here. a pointer weighs 8 bytes usually; an address is just 20 bytes.

vochain/process_id.go
```diff
@@ -0,0 +1,120 @@
...
+	return p.chainID
+}
+
+// SetAddr sets the organizer address (20 bytes)
+func (p *ProcessID) SetAddr(addr common.Address) {
+	p.organizationAddr = &addr
+}
+
+// Addr returns the organizer address
+func (p *ProcessID) Addr() *common.Address {
```
mvdan: per the above, I don't think this should return a pointer (SetAddr is already good)

vochain/process_id.go
```diff
@@ -0,0 +1,120 @@
...
+		return fmt.Errorf("proof is nil")
+	}
+	// get the Protobuf oneOf descriptor index number and use it as proofType
+	protoDesc := proof.ProtoReflect().Descriptor().Oneofs().Get(0)
+	p.proofType = uint16(proof.ProtoReflect().WhichOneof(protoDesc).Number())
+	return nil
+}
+
+// Proof returns a string representing the census proof type
+func (p *ProcessID) Proof() string {
```
mvdan: nit: SetProof takes a pointer, but Proof takes a string. would be nice for the two to be consistent.

mvdan (approved):

p4u:
Reduced Proof to 1 byte and added envelopeType of 1 byte.

Up to 256 kind of proofs should be enough.

@mvdan can you review again? specially the binary operations which I tried to make as safe as possible.

PR #550, opened 2022-04-27
vochain/scrutinizer: use a tx when doing two sqlite queries

mvdan:
(see commit message)


mvdan:
@p4u ironic that I said I could not reproduce your ProcessInfo warning, and then there it just failed in CI.

PR #555, opened 2022-05-03
WIP: import state from snapshot

jordipainan:
add --snapshotURL and --vochainForceResync flags
and the underlying functionality

The former is used for passing an URL from which to
download a state snapshot compressed file.
If forceResync is set to true all existing
vochain data will be wiped and replaced by
the decompressed one extracted from the
downloaded snapshot, otherwise the downloaded
state snapshot compressed file will just be saved.

### DIFF
--- a/vochain/start.go
+++ b/vochain/start.go
@@ -64,9 +64,9 @@ func NewVochain(vochaincfg *config.VochainCfg, genesis []byte) *BaseApplication 
 		log.Infof("importing snapshot...")
 		snapshotPath, err := FetchFile(filepath.Join(vochaincfg.DataDir, "snapshot"), vochaincfg.SnapshotURL)
 		if err != nil {
-			log.Fatalf("cannot import snapshot: %s", err)
+			log.Fatalf("cannot download snapshot: %s", err)
 		}
-		log.Infof("snapshot successfully imported !")
+		log.Infof("snapshot successfully downloaded!")
 		if vochaincfg.ForceResync {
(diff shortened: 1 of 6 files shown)


vochain/start.go
```diff
@@ -644,3 +652,58 @@ func AminoKeys(key crypto25519.PrivKey) (private, public []byte, err error) {
 func AminoPubKey(pubkey []byte) ([]byte, error) {
 	return tmjson.Marshal(pubkey)
 }
+
+// ImportSnapshot imports a state snapshot given a URL and decompress it
+// if ForceResync==true overrides the node data, else download and ignore
+func ImportSnapshot(vochaincfg *config.VochainCfg) error {
```
p4u: I think its cleaner and more elegant to **not** make this function dependent on vochain config. Instead is better to write atomic/standalone functions, makes the code more composable.

So ImportSnapshot() should take just the required argunemts (dataDir, URL). The caller should make the proper checks (ForceResync) and decide if the function must be executed or not.

p4u (changes requested):

vochain/start.go
jordipainan: Just refactored the code, now I think is more modular and atomic

p4u:
I'll review again once the tests are green :)

p4u:
Can you rebase with master and see if the tests pass?

altergui:
this looks abandoned (or maybe merged but not marked as such?), in any case could we close this?

PR #556, opened 2022-05-03
import state from snapshot (release-1.3)

jordipainan:
add --snapshotURL and --vochainForceResync flags
and the underlying functionality

The former is used for passing an URL from which to
download a state snapshot compressed file.
If forceResync is set to true all existing
vochain data will be wiped and replaced by
the decompressed one extracted from the
downloaded snapshot, otherwise the downloaded
state snapshot compressed file will just be saved.

### DIFF
--- a/vochain/apputils.go
+++ b/vochain/apputils.go
@@ -337,13 +337,8 @@ func Untar(src string, destination string) error {
 
 // FetchFile downloads a file from a given URL
 func FetchFile(destDir, url string) (string, error) {
-	// check if dest dir exists
-	if _, err := os.Stat(destDir); err != nil && os.IsNotExist(err) {
-		if err := os.MkdirAll(destDir, 0o777); err != nil {
-			return "", fmt.Errorf("cannot create destination directory: %w", err)
-		}
-	} else if err != nil {
-		return "", fmt.Errorf("cannot check if destination directory exists")
+	if err := os.MkdirAll(destDir, 0o777); err != nil {
+		return "", fmt.Errorf("cannot create destination directory: %w", err)
 	}
 	client := grab.NewClient()
 	req, _ := grab.NewRequest(destDir, url)
@@ -369,5 +364,6 @@ loop:
 	if err := resp.Err(); err != nil {
 		return "", fmt.Errorf("download failed: %w", err)
 	}
+	log.Infof("%s saved at %s", filepath.Base(resp.Filename), destDir)
 	return resp.Filename, nil
 }
--- a/vochain/apputils.go
+++ b/vochain/apputils.go
@@ -1,14 +1,20 @@
 package vochain
 
 import (
+	"archive/tar"
 	"bytes"
+	"compress/gzip"
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
+	"io"
 	"math/big"
 	"math/rand"
 	"os"
+	"path/filepath"
 	"strconv"
+	"strings"
+	"time"
 
 	"go.vocdoni.io/dvote/censustree/gravitontree"
 	"go.vocdoni.io/dvote/config"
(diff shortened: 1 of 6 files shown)


vochain/start.go
```diff
@@ -335,3 +349,23 @@ func AminoKeys(key crypto25519.PrivKey) (private, public []byte, err error) {
...
+// restoreStateFromSnapshot overrides the node data with a given snapshot
+func restoreStateFromSnapshot(dataDir, src string) error {
+	log.Infof("deleting old datadir ... %s", dataDir)
+	if _, err := os.Stat(dataDir); err != nil && !os.IsNotExist(err) {
+		return fmt.Errorf("cannot check if destination directory exists")
+	} else if err != nil {
+		if err := os.RemoveAll(dataDir); err != nil {
+			return fmt.Errorf("cannot clean directory before applying the snapshot")
+		}
+	}
```
altergui: i don't think this previous checks and nested ifs are necessary.  `os.RemoveAll()` returns success if the path didn't exist in the first place.
(also, the `else if err != nil` looks wrong?)

> RemoveAll removes path and any children it contains. It removes everything it can but returns the first error it encounters. If the path does not exist, RemoveAll returns nil (no error). If there is an error, it will be of type *PathError.

anyway i'd replace those 8 lines with these 4
```
	log.Infof("cleaning up old datadir %s before applying the snapshot", dataDir)
	if err := os.RemoveAll(dataDir); err != nil {
		return fmt.Errorf("cannot clean old datadir: %w", err)
	}
```

vochain/apputils.go
```diff
@@ -268,3 +275,101 @@ func GetFriendlyResults(results []*models.QuestionResult) [][]string {
...
+// FetchFile downloads a file from a given URL
+func FetchFile(destDir, url string) (string, error) {
+	// check if dest dir exists
+	if _, err := os.Stat(destDir); err != nil && os.IsNotExist(err) {
+		if err := os.MkdirAll(destDir, 0o777); err != nil {
+			return "", fmt.Errorf("cannot create destination directory: %w", err)
+		}
+	} else if err != nil {
+		return "", fmt.Errorf("cannot check if destination directory exists")
+	}
```
altergui: same here: 

> MkdirAll creates a directory named path, along with any necessary parents, and returns nil, or else returns an error. The permission bits perm (before umask) are used for all directories that MkdirAll creates. If path is already a directory, MkdirAll does nothing and returns nil.

no need to check. just call os.MkdirAll
```
		if err := os.MkdirAll(destDir, 0o777); err != nil {
			return "", fmt.Errorf("cannot create destination directory: %w", err)
		}
```

vochain/start.go
```diff
@@ -25,8 +26,21 @@ import (
 
 // NewVochain starts a node with an ABCI application
 func NewVochain(vochaincfg *config.VochainCfg, genesis []byte) *BaseApplication {
+	if vochaincfg.SnapshotURL != "" {
+		log.Infof("importing snapshot...")
+		snapshotPath, err := FetchFile(filepath.Join(vochaincfg.DataDir, "snapshot"), vochaincfg.SnapshotURL)
+		if err != nil {
+			log.Fatalf("cannot import snapshot: %s", err)
```
altergui: FetchFile just downloads, so i guess this would be

"cannot download snapshot: %s"

vochain/start.go
```diff
@@ -25,8 +26,21 @@ import (
...
+		}
+		log.Infof("snapshot successfully imported !")
```
altergui: "snapshot successfully downloaded!"

vochain/apputils.go
```diff
@@ -268,3 +275,101 @@ func GetFriendlyResults(results []*models.QuestionResult) [][]string {
...
+		case <-resp.Done:
+			// download complete
+			break loop
+		}
+	}
+	if err := resp.Err(); err != nil {
+		return "", fmt.Errorf("download failed: %w", err)
+	}
+	filename := strings.Split(resp.Filename, "/")
+	log.Infof("%s saved at %s", filename[len(filename)], destDir)
```
altergui: better
```go
	log.Infof("%s saved at %s", filepath.Base(resp.Filename), destDir)
```

altergui (commented):
a couple places for simpler code

jordipainan:
working now on the scrutinizer part

jordipainan:
thank you for committing the suggestions @altergui

p4u (approved):

PR #564, opened 2022-05-16
fix/benchmark (depends on PR 563)

randomshinichi:
`BenchmarkVochainBatchProof`, `BenchmarkVochainSingleProof` benchmarks vote processing on a single node with a distributed census. In one, the proofs for the votes are fetched in a batch; the other, the proofs are requested when forming the vote envelope, one at a time.

`BenchmarkVochainCSP` benchmarks vote processing on a single node with a centralized census. The proofs that the voters are in the census are fetched as a batch in the beginning.

Core i7 8550U plugged into AC, ~2.8-2.9GHz
```
BenchmarkVochainSingleProof-8   	    3428	    307200 ns/op	  101673 B/op	     956 allocs/op
BenchmarkVochainBatchProof-8   	    4600	    287593 ns/op	   62049 B/op	     578 allocs/op`
BenchmarkVochainCSP-8   	    3936	    306749 ns/op	   48031 B/op	     515 allocs/op
```

### DIFF
--- a/benchmark/vochain_benchmark_csp_test.go
+++ b/benchmark/vochain_benchmark_csp_test.go
@@ -1,0 +1,143 @@
+package test
+
+import (
+	"encoding/json"
+	"fmt"
+	"sync/atomic"
+	"testing"
+
+	"go.vocdoni.io/dvote/api"
+	"go.vocdoni.io/dvote/client"
+	"go.vocdoni.io/dvote/crypto/ethereum"
+	"go.vocdoni.io/dvote/log"
+	"go.vocdoni.io/dvote/test/testcommon"
+	"go.vocdoni.io/dvote/util"
+	"go.vocdoni.io/dvote/vochain"
+	models "go.vocdoni.io/proto/build/go/models"
+	"google.golang.org/protobuf/proto"
+)
+
+func BenchmarkVochainCSP(b *testing.B) {
+	log.Init("info", "stdout")
+	b.ReportAllocs()
+	var dvoteServer testcommon.DvoteAPIServer
+	host := *hostFlag
+	if host == "" {
+		dvoteServer.Start(b, "file", "census", "vote")
+		host = dvoteServer.ListenAddr
+	}
+
+	cl, err := client.New(host)
+	if err != nil {
+		b.Fatal(err)
+	}
+
+	// create random key batch
+	keySet := testcommon.CreateEthRandomKeysBatch(b, b.N)
+	log.Infof("generated %d keys", len(keySet))
+
+	// setup the basic parties
+	cspKey := ethereum.NewSignKeys()
+	if err := cspKey.Generate(); err != nil {
+		b.Fatal(err)
+	}
+	entityID := dvoteServer.Signer.Address().Bytes()
+
+	log.Info("creating a new process")
+	start, processID, err := cl.CreateProcess(
+		dvoteServer.Signer,
+		entityID,
+		cspKey.PublicKey(),
+		"https://dummycsp.foo",
+		&models.EnvelopeType{EncryptedVotes: false},
+		nil,
+		models.CensusOrigin_OFF_CHAIN_CA,
+		0,
+		numberOfBlocks,
+		uint64(b.N),
+	)
+	if err != nil {
+		b.Fatalf("couldn't create process: %v", err)
+	}
+	log.Infof("created a new process, will start at block %d", start)
+
+	if err := waitProcessReady(b, cl, processID); err != nil {
+		b.Fatal(err)
+	}
+
+	// batch generate CSP proofs
+	proofs, err := cl.GetCSPproofBatch(keySet, cspKey, processID)
+	if err != nil {
+		b.Fatalf("could not generate batch CSP proofs: %v", err)
+	}
+
+	// send votes in parallel
+	log.Infof("sending %d votes in parallel", b.N)
+	count := int32(0)
+	b.ResetTimer()
+
+	b.RunParallel(func(pb *testing.PB) {
+		cl, err := client.New(host)
+		if err != nil {
+			b.Fatal(err)
+		}
+		for pb.Next() {
+			sendVoteCSP(b,
+				cl,
+				dvoteServer.VochainAPP.ChainID(),
+				keySet[count],
+				proofs[count],
+				processID)
+			atomic.AddInt32(&count, 1)
+		}
+	})
+}
+
+func sendVoteCSP(b *testing.B, cl *client.Client, chainID string, s *ethereum.SignKeys,
+	proof *client.Proof, processID []byte) {
+	req := &api.APIrequest{}
+	doRequest := cl.ForTest(b, req)
+
+	// Generate VotePackage and put it in VoteEnvelope
+	votePkg := &vochain.VotePackage{
+		Nonce: fmt.Sprintf("%x", util.RandomHex(32)),
+		Votes: []int{1},
+	}
+	voteBytes, err := json.Marshal(votePkg)
+	if err != nil {
+		b.Fatalf("cannot marshal vote: %s", err)
+	}
+	p := &models.ProofCA{}
+	if err := proto.Unmarshal(proof.Siblings, p); err != nil {
+		b.Fatal(err)
+	}
+	tx := &models.VoteEnvelope{
+		Nonce:     util.RandomBytes(32),
+		ProcessId: processID,
+		Proof: &models.Proof{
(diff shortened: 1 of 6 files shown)


benchmark/vochain_benchmark_test.go
```diff
@@ -27,214 +27,133 @@ const (
 	hexProcessID   = "0xe9d5e8d791f51179e218c606f83f5967ab272292a6dbda887853d81f7a1d5105"
 )
 
-func BenchmarkVochain(b *testing.B) {
+var zeroReq = &api.APIrequest{}
+
+func reset(r *api.APIrequest) {
```
jordipainan: What do you think about defining this in `api/types.go` as a helper in the following way?

```golang
func (r *APIrequest) reset() {
	r = &APIrequest{}
}
```

benchmark/vochain_benchmark_csp_test.go
```diff
@@ -0,0 +1,145 @@
...
+		b.Fatal(err)
+	}
+
+	// create random key batch
+	keySet := testcommon.CreateEthRandomKeysBatch(b, b.N)
+	log.Infof("generated %d keys", len(keySet))
+
+	// setup the basic parties
+	signerPubHex, _ := dvoteServer.Signer.HexString()
+	signerPub := testutil.Hex2byte(b, signerPubHex)
```
jordipainan: This two intermediate variables can be deleted and you can just set the entityID with the signer

benchmark/vochain_benchmark_csp_test.go
```diff
@@ -0,0 +1,145 @@
...
+	cspKey := ethereum.NewSignKeys()
+	if err := cspKey.Generate(); err != nil {
+		b.Fatal(err)
+	}
+	entityID := signerPub[:types.EntityIDsize]
```
jordipainan: `entityID := dvoteServer.Signer.Address().Bytes()`

jordipainan: With this you do not need to import `"go.vocdoni.io/dvote/test/testcommon/testutil"` & `"go.vocdoni.io/dvote/types"`

benchmark/vochain_benchmark_test.go
```diff
@@ -273,19 +194,66 @@ func voteBench(b *testing.B, cl *client.Client, s *ethereum.SignKeys,
...
 
-	// check vote added
-	/*	reset(req)
-		req.ProcessID = processID
-		req.Nullifier = vochain.GenerateNullifier(s.Address(), processID)
-		for {
-			resp = doRequest("getEnvelopeStatus", nil)
-			if *resp.Registered {
-				break
+func waitProcess(b *testing.B, cl *client.Client, processID []byte) {
```
jordipainan: I would rather name this function `waitProcessReady()` but maybe we can pass another argument `status` and make this function more flexible, in this case we can call it `waitProcessStatus()` and we pass the status we expect the process to have.

benchmark/vochain_benchmark_test.go
```diff
@@ -273,19 +194,66 @@ func voteBench(b *testing.B, cl *client.Client, s *ethereum.SignKeys,
...
+				if *height.Height >= procInfo.Process.StartBlock {
+					break
+				}
 			}
-			time.Sleep(time.Second * 2)
 		}
-	*/
+		failures--
+		if failures == 0 {
+			b.Fatalf("processID %s should be started by now", hexProcessID)
```
jordipainan: This Fatal can be substituted by an error and we can execute the fatal in an upper layer. So `waitProcess(...args) error`

benchmark/vochain_benchmark_test.go
```diff
@@ -273,19 +194,66 @@ func voteBench(b *testing.B, cl *client.Client, s *ethereum.SignKeys,
...
+		}
+		time.Sleep(time.Second)
+	}
+}
+
+func createDistributedCensusAndProcess(b *testing.B, cl *client.Client, dvoteServer testcommon.DvoteAPIServer, keySet []*ethereum.SignKeys, entityID []byte) (censusRoot []byte, censusURI string, processID []byte) {
```
jordipainan: The signature of this function is too long, please split it like: 

```golang
func createDistributedCensusAndProcess(b *testing.B,
    cl *client.Client,
    dvoteServer testcommon.DvoteAPIServer,
    keySet []*ethereum.SignKeys,
    entityID []byte) (censusRoot []byte, censusURI string, processID []byte) {
```
Also as commented before I think we can return an error and do the Fatal in the upper layer

jordipainan: Btw, I personally do not like a lot this function, `DistributedCensus` does not add much information, as the census are normally distributed under the Vocdoni approach. And as we already have a create process and create census function I think is fine to keep them separated. But if this duplicated a lot of code maybe is fine.
If you decide to keep it please change the name to `createCensusAndProcess`

benchmark/vochain_benchmark_test.go
```diff
@@ -27,214 +27,133 @@ const (
...
+			)
+			atomic.AddInt32(&count, 1)
 		}
 	})
 }
 
-func voteBench(b *testing.B, cl *client.Client, s *ethereum.SignKeys,
-	censusRoot, processID []byte) {
-	// API requests
+func sendVote(b *testing.B, cl *client.Client, chainID string, s *ethereum.SignKeys,
```
jordipainan, line 147: I assume you're not returning an error here and in the other functions because old code do the things that way, but  what do you think if we wrap the error and we call the fatals on the main function of the bench? I think is cleaner

jordipainan (commented):
I've made some minor comments.

Regarding the genesis changes, these are breaking changes and I would like them to not be included in this PR, even if using tmjson makes more sense this change have critical implications, like the reset of the state in the dev env because of the node not knowing that the genesis is the same.

Apart from that, the rest looks fine! What a pleasure to have the benchs working again :partying_face: ! 
Good job :+1:

jordipainan:
@jordipainan Self-reminder to squash.
@randomshinichi You can rebase with master and the txcost fix is going to be applied without conflicts :)

benchmark/vochain_benchmark_csp_test.go
randomshinichi: I don't understand - is the entityID derived differently nowadays?

benchmark/vochain_benchmark_test.go
randomshinichi: New name sounds good.
Let's wait until we actually need the flexibility elsewhere to change the function though. Because if we really need this, then it shouldn't be just in the benchmark/test package

benchmark/vochain_benchmark_test.go
randomshinichi: The name is like that so in the future one won't ask "why doesn't BenchmarkVochainCSP use this" - but yeah it is an ugly name.

I'll rename it, but the function saves a lot of code duplication as you can see. Setup code is already long enough as it is, this is an easy way to keep it somewhat legible.

benchmark/vochain_benchmark_test.go
randomshinichi: Yes, I was following the old code's way.

The code leading up to the vote submission fails with a Fatal, but vote submission fails with an Error. If we move this upwards, this distinction will be lost. Is it important? Perhaps everything should be fatal?

benchmark/vochain_benchmark_test.go
randomshinichi: I decided to keep it because the aim was always to keep the brain free while reading the higher level code, not code tidiness per se.

benchmark/vochain_benchmark_test.go
randomshinichi: Changed it to this:
```
for pb.Next() {
    err = sendVote(b,
        cl,
        dvoteServer.VochainAPP.ChainID(),
        keySet[count],
        proofs[count],
        processID)
    if err != nil && strings.HasPrefix(err.Error(), "error submitting vote") {
        b.Error(err)
    } else if err != nil {
        b.Fatal(err)
    }

    atomic.AddInt32(&count, 1)
}
```
OK so on the higher level for loop the code feels noisier. `sendVote()` did not get simplified either, and I feel like `b.Fatal()/b.Error()` is a great opportunity to simplify error handling, which we're wasting in this case.

randomshinichi: Basically, my strategy is: sacrifice code 'tidiness' and DRY-ness in favour of clarity of purpose when reading higher level code.

jordipainan (approved):

PR #566, opened 2022-05-16
vochain/scrutinizer: port most of vote results to sqlite

mvdan:
(see commit message)

### DIFF
--- a/vochain/scrutinizer/db/process.sql.go
+++ b/vochain/scrutinizer/db/process.sql.go
@@ -23,7 +23,10 @@ INSERT INTO processes (
 	envelope_pb, mode_pb, vote_opts_pb,
 	private_keys, public_keys,
 	question_index, creation_time,
-	source_block_height, source_network_id
+	source_block_height, source_network_id,
+
+	results_votes, results_weight, results_envelope_height,
+	results_signatures, results_block_height
 ) VALUES (
 	?, ?, ?, ?, ?,
 	?, ?, ?,
@@ -33,7 +36,10 @@ INSERT INTO processes (
 	?, ?, ?,
 	?, ?,
 	?, ?,
-	?, ?
+	?, ?,
+
+	?, "0", 0,
+	"", 0
 )
 `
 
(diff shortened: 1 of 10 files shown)


mvdan:
@p4u FYI this keeps the existing design when it comes to updating the existing processes with some computation, e.g. WalkEnvelopes, but it's very likely that we could make some of it more efficient with sqlite by writing SQL queries with logic in them. That way we would avoid some of the data encoding and decoding cost. We can talk about that once the sqlite implementation is complete though, but still FYI.

vochain/scrutinizer/db/process.sql.go
```diff
@@ -94,11 +101,12 @@ func (q *Queries) CreateProcess(ctx context.Context, arg CreateProcessParams) (s
 		arg.CreationTime,
 		arg.SourceBlockHeight,
 		arg.SourceNetworkID,
+		arg.ResultsVotes,
 	)
 }
 
 const getProcess = `-- name: GetProcess :one
-SELECT id, entity_id, entity_index, start_block, end_block, results_height, have_results, final_results, census_root, rolling_census_root, rolling_census_size, max_census_size, census_uri, metadata, census_origin, status, namespace, envelope_pb, mode_pb, vote_opts_pb, private_keys, public_keys, question_index, creation_time, source_block_height, source_network_id FROM processes
+SELECT id, entity_id, entity_index, start_block, end_block, results_height, have_results, final_results, census_root, rolling_census_root, rolling_census_size, max_census_size, census_uri, metadata, census_origin, status, namespace, envelope_pb, mode_pb, vote_opts_pb, private_keys, public_keys, question_index, creation_time, source_block_height, source_network_id, results_votes, results_weight, results_envelope_height, results_signatures, results_block_height FROM processes
```
jordipainan, line 109: Is this not the same as `SELECT * FROM` ?

jordipainan:
LGTM! Thank you @mvdan 

Fyi: I made a comment but the code is autogenerated so does not apply and I deleted it, in case some mail notification arises :)

p4u:
Great @mvdan, we are close to finally remove badgerhold.

p4u (approved):

PR #570, opened 2022-06-09
Revert "change to NOTTY is true if not set (previously it was false i…

randomshinichi:
…f not set)"

This reverts commit 22c5ed4a2d6a9ce9313f70bc107866986d68cd08.

It turns out we do want the previous behaviour after all. This means you will have to set `NOTTY=1` if you are running `start_test.sh` from the CLI.


p4u:
Yes, but it has a reason, the no TTY option might fail in some systems (such as remote CI, depending on the environment) and also if true by default it becomes a hidden option that blinds some input/output (a developer might get confused of this behavior). Also, in case we want to do NOTTY=true by default, the var logic should be switched (so declare TTY=1).

p4u (approved):

PR #572, opened 2022-06-16
minor build and CI version bumps

mvdan:
(see commit message)


mvdan:
Hm, not sure how this broke compose-test.

Issue #576, opened 2022-07-04
bug: compose-test can fail with "Cannot create container for service test"


altergui:
https://github.com/vocdoni/vocdoni-node/runs/7172269901?check_suite_focus=true#step:3:479
```
### Waiting for test suite to be ready ###
### Test suite ready ###
### Running test vocli concurrently with others ###
Starting 252074766_testsuite_seed_1 ... 
Starting 252074766_testsuite_seed_1 ... done
Starting 252074766_testsuite_gateway0_1 ... 
Starting 252074766_testsuite_gateway0_1 ... done
Cannot create container for service test: Conflict. The container name "/vocli" is already in use by container "1809fc88ac681e54de9c15e6a34ca493c175c5a26a00f93d7f837755c21f655b". You have to remove (or rename) that container to be able to reuse that name.
```

it sounds like two concurrent compose-test were running and the second one couldn't run because the first took some namespace. this shouldn't happen, two compose-test jobs should be able to run in parallel without interfering each other (besides competing for resources like cpu, ram, etc)

PR #567, opened 2022-05-20
Reliability and QoL improvements for integration tests

randomshinichi:

PR #577, opened 2022-07-05
start_test.sh: use unique container names to allow concurrent runs

altergui:
bug https://github.com/vocdoni/vocdoni-node/issues/576 was introduced in PR https://github.com/vocdoni/vocdoni-node/pull/567, here's a fix


jordipainan (approved):

randomshinichi:
Oops.

PR #593, opened 2022-08-31
return an empty list in getEntityList if no entities found

### LINKED ISSUE #592: bug: getEntityList return "ok": true although the response is empty when using searchTerm
**Describe the bug**

When  calling `getEntityList` using a random `searchTerm`, it return `"response": {"ok":true}` but without a result (no results).

**To Reproduce (please complete the following information)**

Executing:

```bash
curl -s https://gw1-azeno.vocdoni.net/dvote -X POST -H Content-Type:application/json --data '{"id": "req00'$RANDOM'", "request": {"method": "getEntityList", "from": 0,"listSize": 10, "searchTerm": "1234", "timestamp":'$(date +%s)'}}' | jq 
```

We got:

```bash
{
  "response": {
    "ok": true,
    "request": "req0016518",
    "timestamp": 1661781576
  },
  "id": "req0016518",
  "signature": "8b1c706afd51295ebbe3a823290f8129046b1e23623d22b4a02164cd1952c85a0f517b4c233490d1c69de23064347702f1faff86c84882bad4df5dbb1166d30401"
}
```

**Expected behavior**

It should return an empty `entityIds` list or a ` "ok": false` response.

jordipainan:
closes #592

### DIFF
--- a/rpcapi/votehandlers.go
+++ b/rpcapi/votehandlers.go
@@ -382,7 +382,9 @@ func (r *RPCAPI) getEntityList(request *api.APIrequest) (*api.APIresponse, error
 	if request.ListSize > MaxListSize || request.ListSize <= 0 {
 		request.ListSize = MaxListSize
 	}
-	response.EntityIDs = r.scrutinizer.EntityList(request.ListSize, request.From, request.SearchTerm)
+	response.EntityIDs = new([]string)
+	eids := r.scrutinizer.EntityList(request.ListSize, request.From, request.SearchTerm)
+	response.EntityIDs = &eids
 	return &response, nil
 }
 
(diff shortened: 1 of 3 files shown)


rpcapi/votehandlers.go
```diff
@@ -382,7 +382,15 @@ func (r *RPCAPI) getEntityList(request *api.APIrequest) (*api.APIresponse, error
 	if request.ListSize > MaxListSize || request.ListSize <= 0 {
 		request.ListSize = MaxListSize
 	}
-	response.EntityIDs = r.scrutinizer.EntityList(request.ListSize, request.From, request.SearchTerm)
+	response.EntityIDs = new([]string)
+	*response.EntityIDs = append(
```
mvdan: no need to use append; you can do `list := call(); result.Field = &list`

mvdan (approved):
I guess we don't want to remove omitempty because we don't want this field to appear in responses to other queries? In that case SGTM - I don't particularly like it, but we can't really change how responses are implemented in this PR :)

jordipainan:
> I guess we don't want to remove omitempty because we don't want this field to appear in responses to other queries? In that case SGTM - I don't particularly like it, but we can't really change how responses are implemented in this PR :)

yup, it is the case you're commenting and we cant really change this now :S

Issue #444, opened 2022-01-26
ci: staticcheck 'warning: "./..." matched no packages'


altergui:
a couple days ago i introduced a double import
which was caught today by this github action running staticcheck 
https://github.com/vocdoni/vocdoni-node/runs/4949616468?check_suite_focus=true#step:5:16

this should have been caught in the ci of the original commit, but it didn't:
https://github.com/vocdoni/vocdoni-node/runs/4925896910?check_suite_focus=true
the check just passed (albeit with a 'warning: "./..." matched no packages', maybe that's because?)

just reporting this to investigate further afterwards.

altergui:
this is being adressed in #443

altergui:
i've just noticed staticcheck gave `warning: "./..." matched no packages` in a run today
https://github.com/vocdoni/vocdoni-node/runs/7675558159?check_suite_focus=true#step:7:13

and the step passed successfully (which i believe it shouldn't, since any actual staticcheck errors would go unnoticed)

reopening this since:
1) we thought this error would not happen again (see https://github.com/vocdoni/vocdoni-node/pull/443#issuecomment-1040302935). maybe we just need to use latest staticcheck version? i didn't follow up with this, really
2) irrespective of the above, i believe the correct behaviour should be to FAIL if this warning appears

PR #595, opened 2022-09-06
ci: if staticcheck matched no packages, report failure and fix cache dir

altergui:
this ensures a failure is reported in case staticcheck does nothing due to permission issues

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -27,10 +27,26 @@ jobs:
           go test -vet=off -timeout=15m -race ./...
       - name: go vet
         run: go vet ./...
-      - name: staticcheck
+      - name: Download staticcheck
         run: |
           curl -L https://github.com/dominikh/go-tools/releases/download/2022.1.2/staticcheck_linux_amd64.tar.gz | tar -xzf -
-          ./staticcheck/staticcheck ./...
+      - name: staticcheck
+        run: |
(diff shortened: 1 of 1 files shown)


altergui:
i reopened issue https://github.com/vocdoni/vocdoni-node/issues/444 since it happened again,

@mvdan commented in https://github.com/vocdoni/vocdoni-node/pull/443#issuecomment-1040302935
> Upstream Go will fix the underlying bug that caused us to not see the permission errors.

that fix (https://github.com/golang/go/issues/50869) is not yet merged upstream (many months after) so i worked around this with bash :roll_eyes: 

here's an example run that (now) fails when staticcheck does nothing
https://github.com/vocdoni/vocdoni-node/runs/8202798940?check_suite_focus=true

and here another run that passes since staticcheck didn't return a warning
https://github.com/vocdoni/vocdoni-node/runs/8202802022?check_suite_focus=true

mvdan (approved):
SGTM. FYI you can use `staticcheck/stderr` rather than `./staticcheck/stderr`.

altergui:
now it's evident that staticcheck is failing quite often, so i added a snippet of code to `mv` away the ~/.cache directory in case broken permissions are found.

i'm not really happy with this since i have the feeling that the real cause of the broken permissions (which i still don't understand) will happen recurrently, which combined with this snippet will invalidate the whole cache frequently, just to make staticcheck work.
anyway, let's see how this evolves

PR #596, opened 2022-09-07
indexer: add voterID for getting, if any, the identifier for a voter

jordipainan:


jordipainan:
@mvdan added migrations :)

vochain/scrutinizer/migrations/0003_alter_table_processes_votes.sql
```diff
@@ -0,0 +1,21 @@
+-- +goose Up
+ALTER TABLE vote_references
+ADD voter_id BLOB NOT NULL;
```
mvdan: nit: indent like the rest, and same for the DROP COLUMN

vochain/scrutinizer/migrations/0003_alter_table_processes_votes.sql
```diff
@@ -0,0 +1,21 @@
...
+ALTER TABLE processes
+    ADD results_votes           TEXT NOT NULL,
+    ADD results_weight          TEXT NOT NULL,
+    ADD results_envelope_height INTEGER NOT NULL,
+    ADD results_signatures      TEXT NOT NULL,
+    ADD results_block_height    INTEGER NOT NULL;
+
+-- +goose Down
+ALTER TABLE vote_references
+DROP COLUMN voter_id;
```
mvdan: nit: avoid "COLUMN" for consistency? is either syntax OK?

mvdan (approved):

jordipainan:
cc: @p4u

vochain/transaction.go
```diff
@@ -332,37 +332,37 @@ func (app *BaseApplication) AddTx(vtx *VochainTx, commit bool) (*AddTxResponse,
 // VoteEnvelopeCheck is an abstraction of ABCI checkTx for submitting a vote
 // All hexadecimal strings should be already sanitized (without 0x)
 func (app *BaseApplication) VoteEnvelopeCheck(ve *models.VoteEnvelope, txBytes, signature []byte,
-	txID [32]byte, forCommit bool) (*models.Vote, error) {
-
+	txID [32]byte, forCommit bool) (*models.Vote, []byte, error) {
+	voterID := make([]byte, 0)
```
p4u: This is a bit weird. IMO this method should return just `nil` and voterID only when it has a valid value.

vochain/scrutinizer/scrutinizer_test.go
```diff
@@ -945,13 +945,13 @@ func TestCountVotes(t *testing.T) {
 	for i := 0; i < 100; i++ {
 		v := &models.Vote{ProcessId: pid, VotePackage: vp, Nullifier: util.RandomBytes(32)}
 		// Add votes to votePool with i as txIndex
-		sc.OnVote(v, int32(i))
+		sc.OnVote(v, []byte{}, int32(i))
```
p4u: make it just `nil`

vochain/scrutinizer/scrutinizer.go
```diff
@@ -488,11 +491,11 @@ func (s *Scrutinizer) OnProcess(pid, eid []byte, censusRoot, censusURI string, t
 
 // OnVote scrutinizer stores the votes if the processId is live results (on going)
 // and the blockchain is not synchronizing.
-func (s *Scrutinizer) OnVote(v *models.Vote, txIndex int32) {
+func (s *Scrutinizer) OnVote(v *models.Vote, voterID []byte, txIndex int32) {
```
p4u: We should document what voterID

vochain/scrutinizer/scrutinizer.go
```diff
@@ -106,6 +106,7 @@ type Scrutinizer struct {
 // VoteWithIndex holds a Vote and a txIndex. Model for the VotePool.
 type VoteWithIndex struct {
 	vote    *models.Vote
+	voterID []byte
```
p4u: Do we actually need to store voterID here?

vochain/scrutinizer/indexertypes/types.go
```diff
@@ -212,6 +212,7 @@ type VotePackage struct {
 type VoteReference struct {
 	Nullifier    types.HexBytes `badgerholdKey:"Nullifier"`
 	ProcessID    types.HexBytes `badgerholdIndex:"ProcessID"`
+	VoterID      types.HexBytes
```
p4u: Do we know why VoteReference and EnvelopeMetadata are almost duplicated types? Its quite weird.

VoterID should be stored only once in the correct place.

p4u (changes requested):

vochain/scrutinizer/indexertypes/types.go
jordipainan: VoteReference is for the database itself and is required for fetching db info, EnvelopeMetadata is used for the envelope list API so voterID should be included in both as long as we use the same approach of having different types for the db and the API. Also for generating sqlc code voterID must be added into the VoteReference.

vochain/scrutinizer/scrutinizer.go
jordipainan: Yes because this model is used on the VoteIndexPool (holds the list of votes that will be persisted on the database).

jordipainan:
@p4u I added the changes requested, plus the voterIDType on the first byte of the VoterID

types/types.go
```diff
@@ -12,6 +18,71 @@ type EntityID = []byte
...
+// The first byte of the slice indicates one of the supported identifiers
+// For example for an Ethereum public key the VoterID is [1, pubkb0, pubkb1, ...]
+// where pubkb0 is the first byte of the Ethereum public key
+type VoterID []byte
+
+// VoterIDType represents the type of a voterID
+type VoterIDType = uint8
+
+const (
+	VoterID_Type_Undefined VoterIDType = 0
```
p4u: Why use underscores instead of camel case here. Is there a reason?

types/types.go
```diff
@@ -12,6 +18,71 @@ type EntityID = []byte
 
 type CensusRoot = []byte
 
+// VoterID is the indentifier of a voter, it is represented as random bytes
```
p4u: I don't understand the "random bytes" thing. What you mean?

types/types.go
```diff
@@ -12,6 +18,71 @@ type EntityID = []byte
...
+		"UNDEFINED": VoterID_Type_Undefined,
+		"ECDSA":     VoterID_Type_ECDSA,
+	}
+)
+
+var (
+	ErrUnsupportedVoterIDType error = errors.New("voterID type not supported")
+)
+
+func (v VoterID) Type() VoterIDType {
```
p4u, line 44: Please document all exported methods and vars :)

types/types.go
```diff
@@ -12,6 +18,71 @@ type EntityID = []byte
...
+	VoterID_Type_ECDSA     VoterIDType = 1
+)
+
+// Enum value maps for VoterIDType.
+var (
+	VoterID_Type_name = map[VoterIDType]string{
+		VoterID_Type_Undefined: "UNDEFINED",
+		VoterID_Type_ECDSA:     "ECDSA",
+	}
+	VoterID_Type_value = map[string]VoterIDType{
```
p4u: This map duplication might be avoid. Is it really necessary?

types/types.go
```diff
@@ -12,6 +18,71 @@ type EntityID = []byte
...
+	}
+)
+
+var (
+	ErrUnsupportedVoterIDType error = errors.New("voterID type not supported")
+)
+
+func (v VoterID) Type() VoterIDType {
+	if _, ok := VoterID_Type_name[v[0]]; !ok {
+		return VoterID_Type_Undefined
```
p4u: Actually if you query a map key that does not exist, the value equals to the zero value. So you might not need this check and just return `VoterIDType(v[0])`

types/types.go
```diff
@@ -12,6 +18,71 @@ type EntityID = []byte
...
+type VoterIDType = uint8
+
+const (
+	VoterID_Type_Undefined VoterIDType = 0
+	VoterID_Type_ECDSA     VoterIDType = 1
+)
+
+// Enum value maps for VoterIDType.
+var (
+	VoterID_Type_name = map[VoterIDType]string{
```
p4u: I think we don't need to export this map. There are helper methods on the VoterID type that make the job.

types/types.go
```diff
@@ -12,6 +18,71 @@ type EntityID = []byte
...
+		return VoterID_Type_name[VoterID_Type_Undefined]
+	}
+	return VoterID_Type_name[v[0]]
+}
+
+func (v VoterID) Nil() []byte {
+	return []byte{}
+}
+
+func (v VoterID) isNil() bool {
```
p4u: Why this method is not exported?

vochain/scrutinizer/vote.go
```diff
@@ -87,7 +87,7 @@ func (s *Scrutinizer) GetEnvelope(nullifier []byte) (*indexertypes.EnvelopePacka
 		return nil, fmt.Errorf("transaction is not an Envelope")
 	}
 	log.Debugf("getEnvelope took %s", time.Since(t))
-	return &indexertypes.EnvelopePackage{
+	envPkg := &indexertypes.EnvelopePackage{
```
p4u: `envPkg` is not very self descriptive, I'd name it just envelope or envelopePkg (or envelopePackage)

vochain/scrutinizer/vote.go
```diff
@@ -188,13 +193,18 @@ func (s *Scrutinizer) GetEnvelopes(processId []byte, max, from int,
 					return fmt.Errorf("transaction is not an Envelope")
 				}
 				envelope.Nullifier = txRef.Nullifier
-				envelopes = append(envelopes, &indexertypes.EnvelopeMetadata{
+				envMeta := &indexertypes.EnvelopeMetadata{
```
p4u: envelopeMetadata?

vochain/transaction.go
```diff
@@ -332,40 +332,40 @@ func (app *BaseApplication) AddTx(vtx *VochainTx, commit bool) (*AddTxResponse,
...
-		return nil, fmt.Errorf("process %x not in READY state - current state: %s", ve.ProcessId, process.Status.String())
+		return nil, types.VoterID{}.Nil(), fmt.Errorf("process %x not in READY state - current state: %s", ve.ProcessId, process.Status.String())
 	}
 
 	// Check in case of keys required, they have been sent by some keykeeper
 	if process.EnvelopeType.EncryptedVotes &&
 		process.KeyIndex != nil &&
 		*process.KeyIndex < 1 {
-		return nil, fmt.Errorf("no keys available, voting is not possible")
+		return nil, types.VoterID{}.Nil(), fmt.Errorf("no keys available, voting is not possible")
```
p4u: since voterID is a return value, you might create the voterID var in the begging and return `voterID.Nil()` instead of `types.VoterID{}.Nil()`. I think that would be cleaner.

p4u (changes requested):

types/types.go
jordipainan: I was following the proto style, but already changed

types/types.go
jordipainan: I mean that can be an arbitrary slice of bytes, that for example in the ethereum case is an ecdsa pubkey bytes. Anyways I deleted this part as the first part of the sentence already describes the VoterID fine

types/types.go
jordipainan: If you have the string and you want the VoterIDType is necessary, but is not used atm. So i'll delete it and will add in the future if necessary

jordipainan:
@p4u all coments addressed

p4u (approved):
great :)

PR #601, opened 2022-09-14
F/state snapshot

p4u:
This PR introduces the state snapshot capability for the Vochain.

This is an initial step for enabling to big features:
- Auto prune of state (export the state, remove the current DB, import the state)
- Tendermint Snapshot sync for faster blockchain synchronization

The current PR adds the Snapshot creation every 50k blocks. But it does not yet perform any other action.

Does not break the state since it does not change anything from the current data sets.

A small refactor on how the StateDB merkle trees are accesses has been done. A bigger refactor to simplify the API on the StateDB package should be done in the near future (currently is too complex and hard to understand).

See commit messages for more info.


census/census.go
```diff
@@ -318,12 +318,20 @@ func (m *Manager) DelNamespace(name string) error {
...
 	nsConfig := filepath.Join(m.StorageDir, "namespaces.json")
 	data, err := json.Marshal(m.Census)
 	if err != nil {
-		return err
+		return fmt.Errorf("error marshaling census file: %w", err)
 	}
-	return os.WriteFile(nsConfig, data, 0o600)
+	log.Warnf("NSCONFIG: %s", nsConfig)
+	if err := os.WriteFile(nsConfig, data, 0o600); err != nil {
+		log.Warnf("ERROR: %v", err)
```
jordipainan: You can wrap the error as in line 327 and also unify the message (either on 327 with "ERROR" or changing "ERROR" on line 331 by "error writing file: %w")

vochain/state_trees.go
```diff
@@ -0,0 +1,163 @@
...
+			ParentLeafSetRoot: processSetVotesRoot,
+		}),
+	}
+)
+
+// StateTree returns the state merkle tree with name.
+func StateTreeCfg(name string) statedb.TreeConfig {
+	t, ok := MainTrees[name]
+	if !ok {
+		panic(fmt.Sprintf("state tree %s does not exist", name))
```
mvdan, line 144: nit: use %q for strings in errors, in case the strings contain weird chars

statedb/treeview.go
```diff
@@ -42,9 +44,11 @@ type TreeViewer interface {
 	// Iterate iterates over all leafs of this tree.  When callback returns true,
 	// the iteration is stopped and this function returns.
 	Iterate(callback func(key, value []byte) bool) error
-	// Dump exports all the tree leafs in a byte array.
-	Dump() ([]byte, error)
-	// Root returns the root of the tree, which cryptographically summarises the
+	// Dump exports all the tree leafs into a writer buffer.
+	Dump(w io.Writer) error
+	// Import writes a buffer containing all the tree leafs.
```
mvdan: Import writes? :)

vochain/app.go
```diff
@@ -629,7 +630,16 @@ func (app *BaseApplication) DeliverTx(req abcitypes.RequestDeliverTx) abcitypes.
...
 	data, err := app.State.Save()
 	if err != nil {
-		log.Fatalf("cannot save state: %s", err)
+		log.Fatalf("cannot save state: %v", err)
+	}
+	if app.Height()%50000 == 0 && !app.IsSynchronizing() {
+		startTime := time.Now()
+		log.Infof("performing a state snapshot on block %d", app.Height())
+		if _, err := app.State.snapshot(); err != nil {
+			log.Fatalf("cannot make state snapshot: %v", err)
```
mvdan, line 639: do we want to attempt to try again soon if it fails, without waiting another 50k blocks?

jordipainan:
How much the Snapshot creation affect the Commit() call? Did you make any bench? Maybe saving temporary the "latest state" and invoke the Snapshot func in a goroutine using this temporary "latest state" can avoid Commit() to wait until the snapshot is made.
For the rest LGTM

vochain/process.go
```diff
@@ -143,8 +146,30 @@ func (v *State) CountProcesses(committed bool) (uint64, error) {
 	return count, nil
 }
 
+// ListProcessIDs returns the full list of process identifiers (pid).
+func (v *State) ListProcessIDs(committed bool) ([][]byte, error) {
+	if !committed {
+		v.Tx.RLock()
+		defer v.Tx.RUnlock()
+	}
```
mvdan, line 154: instead of taking a boolean parameter to tell the method whether or not to grab the lock, what I would do is make it the caller's responsibility to hold the lock. this way it is more explicit, as you don't have to remember what the boolean does.

vochain/process.go
```diff
@@ -143,8 +146,30 @@ func (v *State) CountProcesses(committed bool) (uint64, error) {
...
+	processesTree, err := v.mainTreeViewer(committed).SubTree(StateTreeCfg(TreeProcess))
+	if err != nil {
+		return nil, err
+	}
+	var pids [][]byte
+	if err := processesTree.Iterate(func(key []byte, value []byte) bool {
+		p := make([]byte, len(key))
+		copy(p, key)
+		pids = append(pids, p)
+		return false
```
mvdan, line 164: doesn't this mean that we only iterate once and then stop? as in, `false` means we want to stop iterating.

p4u:
> How much the Snapshot creation affect the Commit() call? Did you make any bench? Maybe saving temporary the "latest state" and invoke the Snapshot func in a goroutine using this temporary "latest state" can avoid Commit() to wait until the snapshot is made. For the rest LGTM

The snapshot is quite fast, I don't have benchmark numbers but I did some tests. In a huge state with hundreds of thousands of tree leaves, it should take less than 15s.

The idea is not just to perform the snapshot but to do a prune and import from scratch in order to reduce the size of the storage (currently Arbo stores all nodes even if they are not used anymore). This operation requires to stop processing blocks since nothing else can be commited after the snapshot. Thus we cannot use a goroutine here.

p4u:
@mvdan I'd need you special attention on the file `vochain/state_snapshot.go`.
Specially in the IO operations regarding files.

The StateSnapshot type is made to fit the io.Writer and io.Reader interfaces so it can be passed to Arbo in order to perform "Dump" and "ImportDump" operations.

This method is probably the most complex and I'm not yet testing it since Arbo is missing the io.Reader interface (it will be added soon) https://github.com/vocdoni/vocdoni-node/pull/601/files#diff-2dc8bad68f36101bd28e7bf07823ecaad9cf8eaf2f43e7a8ba9d2bb4ec14df62R142

The `Save` methos is also quite complex and it would be nice if you can have a look on it (maybe there is something that can be optimized) https://github.com/vocdoni/vocdoni-node/pull/601/files#diff-2dc8bad68f36101bd28e7bf07823ecaad9cf8eaf2f43e7a8ba9d2bb4ec14df62R219

p4u:
Some extra context regarding the decisions taken on StateSnapshot.

The objective is to build a single binary file that contains all State the information (which includes several merkle trees).

Initially I wanted to use Protobuf but after making some research I realized:
1. Protobuf requires to store all data in memory before performing the Marshaling. That's something I want to avoid cause the snapshot can be quite big, so I want to use disk IO operations directly.
2. Protobuf is not made for large files (taken from the official documentation)

Gob is slow and also has the same issues. I did not find other encoders that accomplish my requirements so I decided to do my own encoding since its quite simple.

The first 32 bytes of the file is the size of the header manifest. In the header you find the information for finding the byte chunks regarding each merkle tree and its metadata information.

So I'm basically first fetching the header and then moving the file cursor when Read() or Write() are called. When the last byte position of the current tree is reached, Read returns io.EOF

vochain/process.go
p4u: usually yes but https://github.com/vocdoni/vocdoni-node/blob/master/statedb/treeview.go#L42

vochain/app.go
p4u: If it fails the execution should stop since there is no point for failing (only if the disk is full or there are some serious I/O issues). Thus should not fail.

vochain/process.go
p4u: I think the same but this snipped is used all over the State code implementation. We could fill and issue to change this behavior.

mvdan: SGTM

statedb/treeview.go
p4u: Can you rephrase it? I'm quite bad with words :/

vochain/process.go
mvdan: Huh, that's a bit weird. We should probably have a TODO to invert that.

vochain/state_snapshot.go
```diff
@@ -0,0 +1,422 @@
...
+	}
+	// write the header
+	hs, err := file.Write(buf.Bytes())
+	if err != nil {
+		return err
+	}
+	log.Debugf("snapshot header size is %d bytes", hs)
+
+	// write the trees (by copying the tmpFile)
+	if _, err := s.file.Seek(0, 0); err != nil {
```
mvdan: use `io.SeekStart` for readability

vochain/state_snapshot.go
```diff
@@ -0,0 +1,422 @@
...
+func (s *StateSnapshot) Save() error {
+	s.lock.Lock()
+	defer s.lock.Unlock()
+
+	file, err := os.Create(s.path)
+	defer func() {
+		if err := file.Close(); err != nil {
+			log.Warn(err)
+		}
+	}()
```
mvdan: this can be just a `defer file.Close()` since you already check the error at the end, in the happy path. we do not care about close errors if we stopped early because something went wrong already.

vochain/state_snapshot.go
```diff
@@ -0,0 +1,422 @@
...
+// - treeN is the raw bytes dump of all trees.
+type StateSnapshot struct {
+	path               string
+	file               *os.File
+	lock               sync.Mutex
+	header             SnapshotHeader
+	headerSize         uint32
+	currentTree        int // indicates the current tree
+	treeReadPointer    int // points to the last read byte
+	currentTreeReadEnd int // points to the last byte of the current tree
```
mvdan: I'd probably use sized integers here just in case, since this does encoding/decoding. you really don't want to run into issues on 32 bit if e.g. the files near gigabytes in size.

vochain/state_snapshot.go
```diff
@@ -0,0 +1,422 @@
...
+}
+
+// Save builds the snapshot started with `Create` and stores in disk its contents.
+// After calling this method the snapshot is finished.
+// `EndTree` must be called before saving.
+func (s *StateSnapshot) Save() error {
+	s.lock.Lock()
+	defer s.lock.Unlock()
+
+	file, err := os.Create(s.path)
```
mvdan: name this `finalFile` or something like that, otherwise it's confusing with the `file` field named the same

vochain/state_snapshot.go
```diff
@@ -0,0 +1,422 @@
...
+
+	// build the header
+	var buf bytes.Buffer
+	enc := gob.NewEncoder(&buf)
+	if err := enc.Encode(s.header); err != nil {
+		return err
+	}
+	// write the size of the header in the first 32 bytes
+	headerSize := make([]byte, 32)
+	binary.LittleEndian.PutUint32(headerSize, uint32(buf.Len()))
```
mvdan, line 236: FYI, starting in Go 1.19, you can use AppendUint32, meaning you can do it all via

    file.Write(binary.LittleEndian.AppendUint32(nil, uint32(buf.Len()))

statedb/treeview.go
```diff
@@ -42,9 +44,11 @@ type TreeViewer interface {
 	// Iterate iterates over all leafs of this tree.  When callback returns true,
 	// the iteration is stopped and this function returns.
 	Iterate(callback func(key, value []byte) bool) error
-	// Dump exports all the tree leafs in a byte array.
-	Dump() ([]byte, error)
-	// Root returns the root of the tree, which cryptographically summarises the
+	// Dump exports all the tree leafs into a writer buffer.
```
mvdan: nit: this should probably not mention a buffer, as it takes a writer. how about:

> Dump exports all the tree leaves into w. The leaves can be read again via Import.

While you are here, you could also rename Dump to Export, which is the opposite of Import, and will be easier to understand.

statedb/treeview.go
mvdan: I thought it was a typo: import takes a reader, so it feels wrong that it says it "writes". Perhaps something like this, but tweak it because I'm assuming how it works internally:

> Import reads exported tree leaves from r.

vochain/state_snapshot.go
```diff
@@ -0,0 +1,422 @@
...
+//
+// The structure of the snapshot encoded file is:
+//
+//  [headerLen][header][tree1][tree2][treeN]
+//
+// - headerlen is a fixed 32 bytes little endian number indicating the size of the header.
+//
+// - header is the Gob encoded structure containing the information of the trees (size, name, etc.).
+//
+// - treeN is the raw bytes dump of all trees.
```
mvdan, line 41: how about encoding the whole thing with gob? note that gob can be streamed; you can use a single encoder to encode multiple values (such as one header, tree1, and then tree2), and then the decoder can keep decoding until it hits EOF. You know what types to decode into in the second case because the first type is always a header, and the other types are always `[]byte` for trees.

I say this because the code will end up being much clearer - no need to deal with sizes and lengths in the encoding yourself, because gob will do that for you.

mvdan (commented):
Here's one review pass. I think you should just use gob for the entire dump, see my comment for details. Happy to do a quick demo if that's not clear. I can do a second review if/when you've changed the code.

vochain/state_snapshot.go
p4u: I'm not sure how to do it with Gob without having to load the tree bytes in memory. Note that I wrote a io.Writer compatible type so it can be directly passed to Arbo in order to perform the Dump of the tree (so the dump bytes go directly to the file).

Can you provide an example so I can understand?

Regarding pros/cons of using Gob I see:
- Pro: code cleaner
- Con: hard to implement in other languages (since Gob is goLang specific while the custom encoding can be easily implemented by copying the logic in any other language).

More?

vochain/state_snapshot.go
p4u: No, this vars are not encoded, only SnapshotHeader and SnapshotHeaderTree are and both use sized types.

p4u: These are counters and its more convenient to use INT since they are used for indexing arrays

vochain/state_snapshot.go
p4u: Nice, for the future :)

statedb/treeview.go
p4u: For consistency with the underlying package (Arbo) I think Dump is good. But thanks for the nit.

p4u:
@mvdan two of your suggestions make staticcheck cry https://github.com/vocdoni/vocdoni-node/actions/runs/3055785945/jobs/4929234376#step:8:6

mvdan:
The first warning is actually right. Note that you're checking the close error on the _temporary_ file, but you care about the close error on the _final_ file.

Why do we need two files, in fact? If you want atomic writes, what you do is write to a temporary file in the same directory, then _rename_ to the final place. You're writing to a temporary file and then doing an entire copy to the final place, which is not atomic, as it can still fail halfway through. If we don't need atomicity, you can always avoid the temporary file and write directly to the final file.

The second warning is because you put the constant in the wrong place :) it's the second parameter.

vochain/state_snapshot.go
mvdan: Ah, I had missed that you want to read/decode the trees directly with Arbo without having to scan the bytes twice. That makes sense. Yes, you can't really do that if you make everything be gob.

I wonder if you could remove headerLen, because gob should know when the value (the header struct) is finished without needing to rely on EOF. But I can't see that on the docs. So I think what you're doing is fine.

vochain/state_snapshot.go
```diff
@@ -0,0 +1,419 @@
...
+	var err error
+	s.path = filePath
+	s.file, err = os.Open(filePath)
+	if err != nil {
+		return err
+	}
+	headerSizeBytes := make([]byte, 32)
+	if _, err = s.file.Read(headerSizeBytes); err != nil {
+		return err
+	}
```
mvdan: note that `Read` may read fewer bytes. You need `io.ReadFull` to actually ensure you read 32 bytes. this is a fairly common gotcha as in practice short reads do read fully, but you can still be caught by weird bugs where you silently ignore bytes in edge cases.

vochain/state_snapshot.go
```diff
@@ -0,0 +1,419 @@
...
+	s.headerSize = binary.LittleEndian.Uint32(headerSizeBytes)
+	headerBytes := make([]byte, int(s.headerSize))
+	if _, err = s.file.Read(headerBytes); err != nil {
+		return err
+	}
+	decoder := gob.NewDecoder(bytes.NewBuffer(headerBytes))
```
mvdan: no need to read the bytes into a buffer to build an io.Reader for gob. you can instead use a limit reader via https://pkg.go.dev/io#LimitReader:

```
decoder := gob.NewDecoder(io.NewLimitReader(s.file, s.headerSize))
```

That way, gob will see an EOF when it reaches the end of the header size, rather than the end of the whole file.

vochain/state_snapshot.go
```diff
@@ -0,0 +1,419 @@
...
+func (s *StateSnapshot) Open(filePath string) error {
+	s.lock.Lock()
+	defer s.lock.Unlock()
+	var err error
+	s.path = filePath
+	s.file, err = os.Open(filePath)
+	if err != nil {
+		return err
+	}
+	headerSizeBytes := make([]byte, 32)
```
mvdan: perhaps make 32 a named constant for clarity

vochain/state_snapshot.go
```diff
@@ -0,0 +1,419 @@
...
+	if s.treeReadPointer < s.currentTreeReadEnd {
+		return ErrTreeNotYetFullyRead
+	}
+	s.currentTree++
+	if len(s.header.Trees) == s.currentTree {
+		return io.EOF
+	}
+	s.currentTreeReadEnd = s.treeReadPointer +
+		int(s.header.Trees[s.currentTree].Size)
+	_, err := s.file.Seek(int64(s.treeReadPointer), 0)
```
mvdan: all these second `0` parameters to Seek can be SeekStart too :)

vochain/state_snapshot.go
```diff
@@ -0,0 +1,419 @@
...
+	}
+	s.treeReadPointer = snapshotHeaderLenSize + int(s.headerSize)
+	s.currentTree--
+	return s.FetchNextTree()
+}
+
+// FetchNextTree prepares everything for reading the next tree.
+// It returns ErrTreeNotYetFullyRead if the current tree has still
+// pending bytes for read. Returns io.EOF when there are no more trees.
+func (s *StateSnapshot) FetchNextTree() error {
```
mvdan, line 120: I think you should consider a LimitReader here too. As long as the file's read offset is at the start of the next tree, which it should be, then you just need a LimitReader with the tree's size. Then you no longer need the Read methods - you can simply return the LimitReader and let users use e.g. io.ReadAll on it to get a byte slice.

A huge bonus is that you no longer have to keep track of file offsets, append bytes, make sure you don't have panics, etc. A lot of that complexity is taken care of for you.

mvdan:
> Why do we need two files, in fact? If you want atomic writes, what you do is write to a temporary file in the same directory, then rename to the final place. You're writing to a temporary file and then doing an entire copy to the final place, which is not atomic, as it can still fail halfway through. If we don't need atomicity, you can always avoid the temporary file and write directly to the final file.

I just realised why: you write the trees first, and the header later, since you don't know how large the header will be at the beginning. But you need the header to go at the start of the final file.

You could work around this by having one singular header before each tree, and only one overall header at the start with basic fixed-size metadata like the dump version. Each single tree header would contain its own information like its size. Then you would have:

    [dump header] [tree1 header] [tree1 data] [tree2 header] [tree2 data] ...

To get to tree 10 you will have to scan nine headers to skip the nine trees before it, but I think that won't be a common thing, since reading a dump will be a sequential operation in general. And this way you don't have to write all the trees twice when saving a snapshot, which could really matter in terms of i/o and cpu usage.

vochain/state_snapshot.go
p4u: Wow, that's and amazing improvement. I did not know about `io.LimitReader`, so nice :)

p4u: Implemented! Please take a look again.

p4u:
> > Why do we need two files, in fact? If you want atomic writes, what you do is write to a temporary file in the same directory, then rename to the final place. You're writing to a temporary file and then doing an entire copy to the final place, which is not atomic, as it can still fail halfway through. If we don't need atomicity, you can always avoid the temporary file and write directly to the final file.
> 
> I just realised why: you write the trees first, and the header later, since you don't know how large the header will be at the beginning. But you need the header to go at the start of the final file.
> 
> You could work around this by having one singular header before each tree, and only one overall header at the start with basic fixed-size metadata like the dump version. Each single tree header would contain its own information like its size. Then you would have:
> 
> ```
> [dump header] [tree1 header] [tree1 data] [tree2 header] [tree2 data] ...
> ```
> 
> To get to tree 10 you will have to scan nine headers to skip the nine trees before it, but I think that won't be a common thing, since reading a dump will be a sequential operation in general. And this way you don't have to write all the trees twice when saving a snapshot, which could really matter in terms of i/o and cpu usage.

That would work but IMHO it does not add much value. The current approach is simpler and already implemented. The only constrain in comparison with your proposal is that the data needs to be copied. If someday we have issues with that (cause we talking about hundreds of GiB, then it will be worth it.

PS: ok, save this text and one day you might say "I told you" :) We might make NFTs when something like this happens and there is an actual proof on github XD

mvdan:
I already made "I told you we should have used SQL for the scrutinizer" into an NFT :stuck_out_tongue_winking_eye:

vochain/state_snapshot.go
```diff
@@ -0,0 +1,416 @@
...
+	defer s.lock.Unlock()
+	b, err := io.ReadAll(s.currentTreeReadBuffer)
+	if err != nil {
+		return nil, err
+	}
+	if len(b) == 0 {
+		return nil, io.EOF
+	}
+	return b, nil
+}
```
mvdan, line 166: if you make FetchNextTree return the LimitReader, these two can go, because the user can read from the returned value directly

vochain/state_snapshot.go
```diff
@@ -0,0 +1,416 @@
...
+	if err != nil {
+		return err
+	}
+	headerSizeBytes := make([]byte, snapshotHeaderLenSize)
+	if _, err = io.ReadFull(s.file, headerSizeBytes); err != nil {
+		return err
+	}
+	s.headerSize = binary.LittleEndian.Uint32(headerSizeBytes)
+	headerBytes := make([]byte, int(s.headerSize))
+	s.file.Seek(snapshotHeaderLenSize, 0)
```
mvdan: seeking can fail; check the error

vochain/state_snapshot.go
```diff
@@ -0,0 +1,416 @@
...
+	if _, err = io.ReadFull(s.file, headerBytes); err != nil {
+		return err
+	}
+	decoder := gob.NewDecoder(bytes.NewBuffer(headerBytes))
```
mvdan: this can still be a LimitReader, no need for a bytes.Buffer. I think I even showed an example in my previous review? :)

vochain/state_snapshot.go
```diff
@@ -0,0 +1,416 @@
...
+	s.lock.Lock()
+	defer s.lock.Unlock()
+
+	// create the final file
+	finalFile, err := os.Create(s.path)
+	defer func() {
+		finalFile.Close()
+		if err != nil {
+			log.Error(err)
+		}
```
mvdan: this is weird. you're not actually checking the close error? also, you should return the error, not log it. so this should go at the end of the function instead, and here you should do a `defer finalFile.Close()`.

vochain/state_snapshot.go
```diff
@@ -0,0 +1,416 @@
...
+	}
+	decoder := gob.NewDecoder(bytes.NewBuffer(headerBytes))
+	if err := decoder.Decode(&s.header); err != nil {
+		return fmt.Errorf("cannot decode header: %w", err)
+	}
+	if s.header.Version != snapshotHeaderVersion {
+		return fmt.Errorf("snapshot version not compatible")
+	}
+	s.currentTree--
+	return s.FetchNextTree()
```
mvdan, line 115: this looks confusing, perhaps it needs docs.

vochain/state_snapshot.go
p4u: I tried and something failed (gob could not decode it). I'm not sure why, just reverted to the buffer (which works and is also OK). 

EDIT: I found the bug, changed to LimitReader

vochain/state_snapshot.go
p4u: Yes I know but I prefer to just make the type compatible with io.Reader and io.Writer. I think its more convenient.

vochain/state_snapshot.go
p4u: right

PR #603, opened 2022-09-16
F/get envelope/add creation time

jordipainan:


jordipainan:
Note that creation time is declared as omitempty on the JSON annotation because the struct is used in other parts of the code that do not have access to the time information

PR #607, opened 2022-09-21
bump go-ipfs v0.13.1 -> kubo v0.15.0

altergui:
upgrade go-ipfs to latest version (kubo v0.15.0), which pulls go-libp2p v0.21.0


altergui:
latest kubo v0.15.0 pulls go-libp2p v0.21.0

latest go-libp2p is v0.23.1 , i did try to upgrade to that version but:
1) i'm not sure it's a good idea, i guess it's safer to stay with v0.21.0 and upgrade only when a newer kubo requires a newer go-libp2p?
2) go-libp2p v0.23.1 only compiles with Go 1.18 ... and the CI servers are using Go 1.17
https://github.com/vocdoni/vocdoni-node/actions/runs/3111746493/jobs/5044384324#step:5:356
```
# github.com/lucas-clemente/quic-go/internal/qtls
/go/pkg/mod/github.com/lucas-clemente/quic-go@v0.29.0/internal/qtls/go_oldversion.go:5:5: cannot use "The version of quic-go you're using can't be built using outdated Go... (type untyped string) as type int in assignment
note: module requires Go 1.18
```

altergui:
so i just rolled back HEAD to drop the `go1.18.4 get github.com/libp2p/go-libp2p`
should be ready to merge

mvdan:
When upgrading ipfs, use the libp2p version that it wants. Any newer version is likely to be incompatible.

You're right that we need to upgrade our Go version at the same time. Unfortunately, libp2p does some crypto shenanigans and it usually only supports one or two Go versions at a time.

mvdan:
The changes here SGTM, though I assume you need to upgrade to Go 1.18.x as well in this PR.

p4u:
So let's upgrade to 1.18.X @altergui

altergui:
:thinking: not really *needed* , i mean, everything compiles with go 1.17

go 1.18 was only required if upgrading to libp2p v0.23.1, which we're not doing

mvdan:
Ah, I see. We still want to upgrade Go as 1.17.x is no longer maintained, but we should probably do that in a separate PR then.

altergui:
sounds good, i can prepare that PR.

please approve this one and i'll squash-merge it

mvdan (approved):

Issue #129, opened 2021-03-16
CI: Cannot get merkle proof


vdo:
A possible race condition causes the vochaintest to fail sometimes during the `compose-test` job:
```
2021-03-15T11:29:11Z    INFO    vochaintest/vochaintest.go:137    census created and published
Root: dd9268afeaf99fb790cef9f5279d41c05aea3323afa6504c1a1885cf2ac0b884
URI: ipfs://Qmc4vo1nGiKmraAqdafGSHMRbphwrAdbc7XV1DWnuZxfG3
2021-03-15T11:29:11Z    INFO    client/api.go:193    generating proofs...
2021-03-15T11:29:11Z    INFO    client/api.go:629    get root
2021-03-15T11:29:11Z    FATAL    vochaintest/vochaintest.go:140    cannot get merkle proof: ()
main.censusGenerate
    /src/cmd/vochaintest/vochaintest.go:140
main.mkTreeVoteTest
    /src/cmd/vochaintest/vochaintest.go:220
main.main
    /src/cmd/vochaintest/vochaintest.go:88
runtime.main
    /usr/local/go/src/runtime/proc.go:204
2021-03-15T11:29:11Z    INFO    vochaintest/vochaintest.go:137    census created and published
Root: c79afe3bf9483676cbdffdc7ad501cb43bbc8c0bb8b423b5c00c691d2a65d2cf
```

vdo:
https://github.com/vocdoni/vocdoni-node/runs/2112106737

mvdan:
@vdo is there a way to also see the logs for the server the client was talking to? It returned an empty proof, and there isn't an error message either, so I have nothing to start an investigation with.

I also looked at the underlying implementation of genProof, and as far as I can tell it's not designed to return an empty proof unless something goes wrong. Plus, gravitontree never marshals a proof into zero bytes either.

So I think there's an error of some sort on the server side, and I've run the tests locally a bunch of times but I can't get it to fail like it did on CI. If this error is rare, I'd personally add printing of server logs if vochaintest fails, and next time we get a failure we can investigate better.

vdo:
Sorry for the late response.

For some reason, the post run logs output was removed from the test script, maybe to avoid too much logging in gitlab, here:
https://github.com/vocdoni/vocdoni-node/blob/master/dockerfiles/testsuite/start_test.sh#L46

I suggests enabling them back, if the bug happens again.

mvdan:
[...]

PR #608, opened 2022-09-21
fix vochaintest races

### LINKED ISSUE #575: bug: compose-test fails with different errors due to "not have enough balance"
occurrence:
https://github.com/vocdoni/vocdoni-node/runs/7181582646?check_suite_focus=true#step:3:1181
```
2022-07-04T13:40:47Z	INFO	vochaintest/vochaintest.go:1260	account 0xFE68F836c8Ff56A5b4A792bDC211F779451a375f succesfully created: infoURI:"ipfs://"
2022-07-04T13:40:48Z	FATAL	vochaintest/vochaintest.go:1135	cannot set account info: submitRawTx failed: addTx setAccountInfoTxCheck: unauthorized: not enough balance to transfer
main.testTokenTransactions
	go.vocdoni.io/dvote/cmd/vochaintest/vochaintest.go:1135
main.main
	go.vocdoni.io/dvote/cmd/vochaintest/vochaintest.go:176
runtime.main
	runtime/proc.go:255
```

altergui:
vochaintest is currently plagued by timing issues, where the code is doing time.Sleep() with the hope that a block is mined with a particular transaction included, which frequently fails to happen.

this fixes #575 where i collected a couple of frequent errors, here's my stab at solving all of them

i did refactor major parts of the code, since there were lots of copy&paste and things i have the feeling are not pretty idiomatic like functions embedded inside testVocli for reasons unclear. 

I'd love to refactor more to make the file more readable/maintainable but for now i'll address the urgent issue of the frequent failures

### DIFF
--- a/cmd/vochaintest/vochaintest.go
+++ b/cmd/vochaintest/vochaintest.go
@@ -1,3 +1,4 @@
+//nolint:lll
 package main
 
 import (
@@ -40,8 +41,22 @@ var ops = map[string]bool{
 	"initaccounts":      true,
 }
 
-var addressRegexp = regexp.MustCompile("Public address of the key:.*")
-var pathRegexp = regexp.MustCompile("Path of the secret key file:.*")
+var (
+	addressRegexp = regexp.MustCompile("Public address of the key:.*")
+	pathRegexp    = regexp.MustCompile("Path of the secret key file:.*")
+)
+
+// how many times to retry flaky transactions
+// * amount of blocks to wait for a transaction to be mined before giving up
+// * how many times to retry opening a connection to an endpoint before giving up
+const retries = 10
+
+// testClient is a *client.Client but with additional methods
+// that are useful only in tests,
+// like waitUntilAccountExists or ensureProcessCreated
+type testClient struct {
+	*client.Client
+}
 
 func opsAvailable() (opsav []string) {
 	for k := range ops {
@@ -192,6 +207,20 @@ func privKeyToSigner(key string) (*ethereum.SignKeys, error) {
 	return skey, nil
 }
 
+// newTestClient will open a connection to addr.
+// If it fails, it will retry 10 times (with 1s interval)
+// before giving up
+func newTestClient(addr string) (*testClient, error) {
+	for tries := retries; tries > 0; tries-- {
+		c, err := client.New(addr)
+		if err == nil {
+			return &testClient{c}, nil
+		}
+		time.Sleep(1 * time.Second)
+	}
+	return nil, fmt.Errorf("couldn't connect to gateway, tried %d times", retries)
+}
+
 func censusGenerate(host string, signer *ethereum.SignKeys, size int, filepath string, withWeight uint64) {
 	cl, err := client.New(host)
 	if err != nil {
@@ -222,17 +251,8 @@ func censusGenerate(host string, signer *ethereum.SignKeys, size int, filepath s
 }
 
 func censusImport(host string, signer *ethereum.SignKeys) {
-	var cl *client.Client
-	var err error
-
 	// Connect
-	for tries := 10; tries > 0; tries-- {
-		cl, err = client.New(host)
-		if err == nil {
-			break
-		}
-		time.Sleep(1 * time.Second)
-	}
+	cl, err := newTestClient(host)
 	if err != nil {
 		log.Fatal(err)
 	}
@@ -265,16 +285,43 @@ func censusImport(host string, signer *ethereum.SignKeys) {
 	log.Infof("Census created and published\nRoot: %x\nURI: %s", root, uri)
 }
 
-func waitUntilNextBlock(mainClient *client.Client) error {
-	h, err := mainClient.GetCurrentBlock()
-	if err != nil {
-		return fmt.Errorf("cannot get current height: %w", err)
+func (c *testClient) ensureTxCostEquals(signer *ethereum.SignKeys, txType models.TxType, cost uint64) error {
+	for i := 0; i < retries; i++ {
+		// if cost equals expected, we're done
+		newTxCost, err := c.GetTransactionCost(models.TxType_SET_ACCOUNT_INFO)
+		if err != nil {
+			log.Warn(err)
+		}
+		if newTxCost == cost {
+			return nil
+		}
+
+		// else, try to set
+		treasurerAccount, err := c.GetTreasurer()
+		if err != nil {
+			return fmt.Errorf("cannot get treasurer: %w", err)
+		}
+
+		log.Infof("will set %s txcost=%d (treasurer nonce: %d)", txType, cost, treasurerAccount.Nonce)
+		txhash, err := c.SetTransactionCost(
+			signer,
+			txType,
+			cost,
+			treasurerAccount.Nonce)
+		if err != nil {
+			log.Warn(err)
+			time.Sleep(time.Second)
+			continue
+		}
+		err = c.WaitUntilTxMined(txhash)
+		if err != nil {
+			log.Warn(err)
+		}
 	}
-	mainClient.WaitUntilBlock(h + 1)
-	return nil
+	return fmt.Errorf("tried to set %s txcost=%d but failed %d times", txType, cost, retries)
 }
 
-func ensureProcessCreated(mainClient *client.Client,
+func (c *testClient) ensureProcessCreated(
 	signer *ethereum.SignKeys,
 	entityID common.Address,
 	censusRoot []byte,
@@ -285,9 +332,9 @@ func ensureProcessCreated(mainClient *client.Client,
 	startBlockIncrement int,
 	duration int,
 	maxCensusSize uint64,
-	retries int) (uint32, []byte, error) {
+) (uint32, []byte, error) {
 	for i := 0; i < retries; i++ {
-		start, pid, err := mainClient.CreateProcess(
+		start, pid, err := c.CreateProcess(
 			signer,
 			entityID.Bytes(),
 			censusRoot,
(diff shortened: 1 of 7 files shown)


cmd/vochaintest/vochaintest.go
```diff
@@ -192,6 +198,27 @@ func privKeyToSigner(key string) (*ethereum.SignKeys, error) {
...
+// like waitUntilAccountExists or ensureProcessCreated
+type testClient struct {
+	*client.Client
+}
+
+// newTestClient will open a connection to addr.
+// If it fails, it will retry 10 times (with 1s interval)
+// before giving up
+func newTestClient(addr string) (*testClient, error) {
+	for tries := 10; tries > 0; tries-- {
```
p4u: make this a constant or use the already defined "retries"

cmd/vochaintest/vochaintest.go
```diff
@@ -265,16 +283,34 @@ func censusImport(host string, signer *ethereum.SignKeys) {
...
+
+		// else, try to set
+		treasurerAccount, err := mainClient.GetTreasurer()
+		if err != nil {
+			return fmt.Errorf("cannot get treasurer: %w", err)
+		}
+
+		// SetTransactionCost can return OK and then the tx be rejected anyway later.
+		// So, ignore errors, we only care about GetTransactionCost in the next loop
+		_ = mainClient.SetTransactionCost(
```
p4u: The error should be checked. 

You are correct, the transaction might fail afterwards (when DeliverTx() is executed) but there might be a first error detected by CheckTx() in the moment the TX reaches the mempool.

cmd/vochaintest/vochaintest.go
```diff
@@ -265,16 +283,34 @@ func censusImport(host string, signer *ethereum.SignKeys) {
...
 }
 
-func waitUntilNextBlock(mainClient *client.Client) error {
-	h, err := mainClient.GetCurrentBlock()
-	if err != nil {
-		return fmt.Errorf("cannot get current height: %w", err)
+func (mainClient testClient) ensureTxCostEquals(signer *ethereum.SignKeys, txType models.TxType, cost uint64) error {
+	for i := 0; i < retries; i++ {
+		// if cost equals expected, we're done
+		newTxCost, _ := mainClient.GetTransactionCost(models.TxType_SET_ACCOUNT_INFO)
```
p4u: Do not ignore errors please. Even if they look dummy. Cause afterwards debugging a code that hide some errors is pretty inconvenient.

cmd/vochaintest/vochaintest.go
```diff
@@ -265,16 +283,34 @@ func censusImport(host string, signer *ethereum.SignKeys) {
 	log.Infof("Census created and published\nRoot: %x\nURI: %s", root, uri)
 }
 
-func waitUntilNextBlock(mainClient *client.Client) error {
-	h, err := mainClient.GetCurrentBlock()
-	if err != nil {
-		return fmt.Errorf("cannot get current height: %w", err)
+func (mainClient testClient) ensureTxCostEquals(signer *ethereum.SignKeys, txType models.TxType, cost uint64) error {
```
p4u: shouldn't `testClient`be a pointer? @mvdan

cmd/vochaintest/vochaintest.go
```diff
@@ -345,71 +383,83 @@ func initAccounts(treasurer, oracle string, accountKeys []*ethereum.SignKeys, ho
...
 		log.Infof("minted 100000 tokens to account %s", k.Address())
 	}
 	return nil
 }
 
-func ensureAccountExists(mainClient *client.Client, account *ethereum.SignKeys, retries int) error {
+func (mainClient testClient) ensureAccountInfoEquals(account *ethereum.SignKeys, infoURI string) error {
+	for i := 0; i < retries; i++ {
+		// if account InfoURI is as expected, we're done
+		acct, _ := mainClient.GetAccount(account.Address())
```
p4u: error?

cmd/vochaintest/vochaintest.go
```diff
@@ -345,71 +383,83 @@ func initAccounts(treasurer, oracle string, accountKeys []*ethereum.SignKeys, ho
...
+		if acct != nil && acct.InfoURI == infoURI {
+			return nil
+		}
+
+		// else, set it
+		// CreateOrSetAccount can return OK and then the tx be rejected anyway later.
+		// So, ignore errors, we only care about GetAccount in the next loop
+		_ = mainClient.CreateOrSetAccount(account, common.Address{}, infoURI, acct.Nonce, nil)
```
p4u: error

p4u (approved):

p4u:
Sure, this package is full of shit and requires a big refactor. So feel free to advance on this direction when there are not other more urgent issues to fix. Thanks!

altergui:
one more to dig into, part of this same issue:
https://github.com/vocdoni/vocdoni-node/actions/runs/3112036601/jobs/5045013019#step:3:10833
```
2022-09-23T10:24:23Z	INFO	vochaintest/vochaintest.go:1551	vocli account set alice
2022-09-23T10:24:25Z	INFO	client/util.go:21	waiting for block 2...
2022-09-23T10:24:37Z	INFO	vochaintest/vochaintest.go:1564	account 0xfe10DAB06D636647f4E40dFd56599da9eF66Db1c fetched: infoURI:"ipfs://aliceinwonderland"

2022-09-23T10:24:37Z	INFO	vochaintest/vochaintest.go:1566	vocli account mint alice (this lets one reuse node states across test runs, because subsequent SetAccountInfoTxs are not free
2022-09-23T10:24:38Z	INFO	vochaintest/vochaintest.go:1573	vocli keys changepassword (a new key)
2022-09-23T10:24:45Z	INFO	vochaintest/vochaintest.go:1593	vocli mint alice -> newAccount
2022-09-23T10:24:48Z	INFO	client/util.go:21	waiting for block 5...
2022-09-23T10:24:54Z	INFO	vochaintest/vochaintest.go:1516	account 0xfCc7842BA2b4b473a10784f25c5B0Abc0a784ead info: infoURI:"ipfs://"

2022-09-23T10:24:56Z	INFO	client/util.go:21	waiting for block 6...
2022-09-23T10:25:03Z	INFO	client/util.go:21	waiting for block 7...
2022-09-23T10:25:09Z	INFO	client/util.go:21	waiting for block 8...
2022-09-23T10:25:15Z	INFO	client/util.go:21	waiting for block 9...
2022-09-23T10:25:21Z	INFO	client/util.go:21	waiting for block 10...
2022-09-23T10:25:33Z	INFO	client/util.go:21	waiting for block 11...
2022-09-23T10:25:39Z	INFO	client/util.go:21	waiting for block 12...
2022-09-23T10:25:45Z	INFO	client/util.go:21	waiting for block 13...
2022-09-23T10:25:51Z	INFO	client/util.go:21	waiting for block 14...
2022-09-23T10:25:57Z	INFO	client/util.go:21	waiting for block 15...
2022-09-23T10:26:09Z	FATAL	vochaintest/vochaintest.go:1604	mint failed: 0xfCc7842BA2b4b473a10784f25c5B0Abc0a784ead balance is still 0 (gave up waiting after 10 blocks)
main.testVocli.func4
	go.vocdoni.io/dvote/cmd/vochaintest/vochaintest.go:1604
main.testVocli
	go.vocdoni.io/dvote/cmd/vochaintest/vochaintest.go:1606
main.main
	go.vocdoni.io/dvote/cmd/vochaintest/vochaintest.go:184
runtime.main
	runtime/proc.go:255
```

jordipainan:
All of this can be resolved knowing if a tx given it;s txhash is mined. This is not easy with the current tendermint version, do you know @altergui if with the new version you're applying will be possible or more easy to get this information?

p4u:
> All of this can be resolved knowing if a tx given it;s txhash is mined. This is not easy with the current tendermint version, do you know @altergui if with the new version you're applying will be possible or more easy to get this information?

I think that should be possible already. The sumitRawTx JSON-RPC method returns the txHash (see https://github.com/vocdoni/vocdoni-node/blob/master/rpcapi/votehandlers.go#L33) and the transaction hash can be queried by `getTxByHash` method (https://github.com/vocdoni/vocdoni-node/blob/master/rpcapi/indexerhandlers.go#L185)

p4u:
Please @altergui prioritize this PR since right now is quite impossible to use the CI testing due these race issues.

altergui:
one more, from master (not this branch, maybe already solved)
https://github.com/vocdoni/vocdoni-node/actions/runs/3126574866/jobs/5072227806#step:3:10946
```
2022-09-26T09:28:27Z	INFO	vochaintest/vochaintest.go:1260	account 0xEF57Ce71BF107A1f485DAA7774436E850Aa967F5 succesfully created: infoURI:"ipfs://"
2022-09-26T09:28:29Z	FATAL	vochaintest/vochaintest.go:1135	cannot set account info: submitRawTx failed: addTx setAccountInfoTxCheck: unauthorized: not enough balance to transfer
main.testTokenTransactions
	go.vocdoni.io/dvote/cmd/vochaintest/vochaintest.go:1135
main.main
	go.vocdoni.io/dvote/cmd/vochaintest/vochaintest.go:176
runtime.main
	runtime/proc.go:250
Error: Process completed with exit code 1.
```

altergui:
from master (probably solved)
https://github.com/vocdoni/vocdoni-node/actions/runs/3126574866/jobs/5072374705#step:3:11395
```
2022-09-26T09:38:20Z	FATAL	vochaintest/vochaintest.go:1735	submitRawTx failed: addTx collectFaucetTxCheck: faucet does not have enough balance 0 < 810
main.testVocli.func8
	go.vocdoni.io/dvote/cmd/vochaintest/vochaintest.go:1735
main.testVocli
	go.vocdoni.io/dvote/cmd/vochaintest/vochaintest.go:1737
main.main
	go.vocdoni.io/dvote/cmd/vochaintest/vochaintest.go:178
runtime.main
	runtime/proc.go:250
	
```

altergui:
i'm noting here races i see "in the wild", which are probably solved in this branch, but just in case i leave them here to investigate

altergui:
> I think that should be possible already. The sumitRawTx JSON-RPC method returns the txHash (see https://github.com/vocdoni/vocdoni-node/blob/master/rpcapi/votehandlers.go#L33) and the transaction hash can be queried by `getTxByHash` method (https://github.com/vocdoni/vocdoni-node/blob/master/rpcapi/indexerhandlers.go#L185)

nice! i'll explore this way.

i'm not so confident this would simplify the code, because the logic `hash := sendTx() ; waitUntilTxIsMined(hash) ; sendTx2()" is not enough: the tx might never be mined (fail because of a conflicting tx in mempool) and then we need to retry the original tx anyway.

that's why for me it's not such a bad idea to say "retry this mint tx until balance is not zero" because in the end, that's all we care about. that the balance is not zero. because that's what enables us to use that balance and continue with the test.

but i'm often wrong, so i'll definitely explore this possibility of using getTxByHash

altergui:
well, believe it or not, it seems the CachePurge issue (fix proposed in #615) was the origin of all the compose-test failures.

i just ran 10 times in a row the compose-test, after applying the fix, and all of them succeded. Before the fix, i'd say almost half of the runs failed.
image

altergui:
i almost can't believe it, but it seems i managed to make the compose-test solid again.
out of 20 runs (all queued up, so maximum load on CI servers) only 1 of them failed and due to a very rare race https://github.com/vocdoni/vocdoni-node/issues/129#issuecomment-1260585615

image

i'm not happy with the current PR because the git history is a huge mess impossible to review. I'll clean it up and ask for review

p4u:
> > I think that should be possible already. The sumitRawTx JSON-RPC method returns the txHash (see https://github.com/vocdoni/vocdoni-node/blob/master/rpcapi/votehandlers.go#L33) and the transaction hash can be queried by `getTxByHash` method (https://github.com/vocdoni/vocdoni-node/blob/master/rpcapi/indexerhandlers.go#L185)
> 
> nice! i'll explore this way.
> 
> i'm not so confident this would simplify the code, because the logic `hash := sendTx() ; waitUntilTxIsMined(hash) ; sendTx2()" is not enough: the tx might never be mined (fail because of a conflicting tx in mempool) and then we need to retry the original tx anyway.
> 
> that's why for me it's not such a bad idea to say "retry this mint tx until balance is not zero" because in the end, that's all we care about. that the balance is not zero. because that's what enables us to use that balance and continue with the test.
> 
> but i'm often wrong, so i'll definitely explore this possibility of using getTxByHash

Well, the Tx should not be available until it is actually past of a block.

cmd/vochaintest/vochaintest.go
altergui: ok, new constant

cmd/vochaintest/vochaintest.go
altergui: true, good catch, fixed

cmd/vochaintest/vochaintest.go
altergui: right, error checked

cmd/vochaintest/vochaintest.go
altergui: done. i added this snippet in some cases where an error is not frequent
```
		if err != nil {
			log.Warn(err)
		}
```
and log.Debug if the error is more common. in any case the loop block should continue trying another iteration.

PR #609, opened 2022-09-22
[Breaking] f/delegates tx support

jordipainan:
Adding support for txs with delegates was something pending for the ethless approach.
This PR adds support for NewProcess, SetProcess and SetAccountInfoUri to work with delegates.

Given that the code was modified for adding tests for delegates:
- Add unit tests for NewProcess
- Add more tests for SetProcess operations with delegates 
It also introduces some changes to process_test.go:
- Use quicktest lib
- Wrap some repetitive code

With the deleted code with just 21 more lines of code we have support for delegates and its unit testing :)

### DIFF
--- a/vochain/account_test.go
+++ b/vochain/account_test.go
@@ -39,6 +39,7 @@ func TestSetAccountInfoTx(t *testing.T) {
 	// set tx costs
 	app.State.SetTxCost(models.TxType_SET_ACCOUNT_INFO, 100)
 	app.State.SetTxCost(models.TxType_COLLECT_FAUCET, 100)
+	app.State.SetTxCost(models.TxType_ADD_DELEGATE_FOR_ACCOUNT, 100)
 
 	// should create an account
 	if err := testSetAccountInfoTx(t, &signer, common.Address{}, nil, app, infoURI); err != nil {
@@ -49,17 +50,19 @@ func TestSetAccountInfoTx(t *testing.T) {
 	if err := testSetAccountInfoTx(t, &signer, common.Address{}, nil, app, ""); err == nil {
 		t.Fatal(err)
 	}
+	if err := app.State.MintBalance(signer.Address(), 1000); err != nil {
+		t.Fatal(err)
+	}
 
 	// should always create an account for the tx sender without taking into account the Account field
 	if err := testSetAccountInfoTx(t, &signer2, signer.Address(), nil, app, infoURI); err != nil {
 		t.Fatal(err)
 	}
-
-	// should create an account with tokens if faucet pkg is provided
-	err := app.State.MintBalance(signer.Address(), 1000)
-	if err != nil {
+	if err := app.State.MintBalance(signer2.Address(), 1000); err != nil {
 		t.Fatal(err)
 	}
+
+	// should create an account with tokens if faucet pkg is provided
 	faucetPkg, err := GenerateFaucetPackage(&signer, signer3.Address(), 100, rand.Uint64())
 	if err != nil {
 		t.Fatal(err)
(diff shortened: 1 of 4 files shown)


vochain/process.go
```diff
@@ -453,18 +453,25 @@ func (app *BaseApplication) NewProcessTxCheck(vtx *models.Tx, txBytes,
...
-	// check if process entityID matches tx sender
+	// check if process entityID matches tx sender, if not
+	// check if recovered address is an Oracle and if not
+	// check if recovered address is a delegate of the entityID provided
 	if !bytes.Equal(tx.Process.EntityId, addr.Bytes()) && !isOracle {
-		// Check if the transaction comes from an oracle
-		// Oracles can create processes with any entityID
-		return nil, common.Address{}, fmt.Errorf(
-			"unauthorized to create a new process, recovered addr is %s", addr.Hex())
+		entityIDAddress := common.BytesToAddress(tx.Process.EntityId)
```
p4u: this var name should be "entityAddress" not "entityIDAddress"

vochain/process.go
```diff
@@ -453,18 +453,25 @@ func (app *BaseApplication) NewProcessTxCheck(vtx *models.Tx, txBytes,
...
+		entityIDAccount, err := state.GetAccount(entityIDAddress, true)
```
p4u: Are we sure the account should be taken from the last comited state? Please review this (I'm not sure either, but in any case a comment should be addressed to let the decision clear)

vochain/transaction.go
```diff
@@ -130,7 +130,8 @@ func (app *BaseApplication) AddTx(vtx *VochainTx, commit bool) (*AddTxResponse,
 			if err := app.State.AddProcess(p); err != nil {
 				return nil, fmt.Errorf("newProcess: addProcess: %w", err)
 			}
-			if err := app.State.IncrementAccountProcessIndex(txSender); err != nil {
+			entityIDAddr := common.BytesToAddress(p.EntityId)
```
p4u: same

vochain/process.go
```diff
@@ -453,18 +453,25 @@ func (app *BaseApplication) NewProcessTxCheck(vtx *models.Tx, txBytes,
...
+		if err != nil {
+			return nil, common.Address{}, fmt.Errorf(
+				"cannot get entityID account for checking if the sender is a delegate: %w", err,
```
p4u: "cannot get entity account"

p4u (changes requested):

p4u:
@jordipainan please set yourself to "resolved" if the comment its already addressed

p4u:
I think this commit is not breaking the consensus. But could you check it?

vochain/process.go
jordipainan: In this case IMO it make sense. We want to know if an account is a delegate and just in the case that the transaction for adding a delegate and the transaction that allow someone to act as a delegate, if in the list, happen in the same block and are valid will make sense, because they are going to use the volatile state.
Said so it always feels more easy to handle/secure to wait at least to the next block to act as a delegate when added and I cannot imagine a situation in which can be a problem.

vochain/transaction.go
```diff
@@ -130,7 +130,8 @@ func (app *BaseApplication) AddTx(vtx *VochainTx, commit bool) (*AddTxResponse,
 			if err := app.State.AddProcess(p); err != nil {
 				return nil, fmt.Errorf("newProcess: addProcess: %w", err)
 			}
-			if err := app.State.IncrementAccountProcessIndex(txSender); err != nil {
+			entityAddr := common.BytesToAddress(p.EntityId)
+			if err := app.State.IncrementAccountProcessIndex(entityAddr); err != nil {
```
jordipainan, line 134: @p4u It is breaking by this line, now that I think more.. Here the AccountProcessIndex will be incremented in different accounts (oracles vs entityID)

vochain/account_test.go
```diff
@@ -91,6 +94,38 @@ func TestSetAccountInfoTx(t *testing.T) {
...
+	acc2, err = app.State.GetAccount(signer2.Address(), false)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if acc2 == nil {
+		t.Fatal(ErrAccountNotExist)
+	}
+	if acc2.Balance != 900 {
+		t.Fatalf("expected balance to be %d got %d", 900, acc2.Balance)
+	}
```
mvdan, line 125: FYI: you can avoid repeating the values with:

```
if got, want := acc2.Balance, 900; got != want {
    t.Fatalf("want balance to be %v, got %v", want, got)
}
```

or with quicktest, `qt.Assert(t, acc2.Balance, qt.Equal, 900)` :)

mvdan (commented):
Without understanding the tx change super well, this looks fine to me

jordipainan:
All comments addressed on DVOT-499, creating a new PR #652 and closing this.

PR #612, opened 2022-09-23
fix getEnvelopeList empty voterID

jordipainan:


jordipainan:
closes #611

mvdan (approved):
I guess this is because of our sql migration, where voter_id might be left empty?

jordipainan:
> I guess this is because of our sql migration, where voter_id might be left empty?

On all the voting processes that does not have an ethereum address (not yet), but I supposed that defining it as not null the voter_id have at least one byte, to know that a key is not supported calling the Address() method.

p4u (approved):

PR #614, opened 2022-09-26
bump go 1.17.11 -> go 1.18.6

altergui:
simple bump


mvdan (approved):
Double check that you ran `go mod tidy` :)

altergui:
ah! i didn't! good one! i did just now and luckily it doesn't change anything. :relieved:

PR #615, opened 2022-09-27
vochain/cache: fix condition to purge txs from cache

altergui:
the idea is that a tx will be purged from cache when the blockchain height is greater than the vote.Height plus a voteCachePurgeThreshold. the comparison operator was reversed and the transactions were removed from cache on every CachePurge()

### DIFF
--- a/vochain/cache.go
+++ b/vochain/cache.go
@@ -107,7 +107,7 @@ func (v *State) CachePurge(height uint32) {
 			log.Warn("vote cache index is not [32]byte")
 			continue
 		}
-		if vote.Height+voteCachePurgeThreshold >= height {
+		if height >= vote.Height+voteCachePurgeThreshold {
 			v.voteCache.Remove(cacheGetNullifierKey(vote.Nullifier))
 			v.voteCache.Remove(id)
 			removeFromMempool = append(removeFromMempool, vid)


altergui:
here's a run that reproduces the issue consistently (i changed the code to make CachePurge to run on every block instead of every 18 blocks)
https://github.com/vocdoni/vocdoni-node/actions/runs/3135845632/jobs/5092017998

anonvoting times out (exit code 124) because it waits for 8 votes to be counted, which never happens because some votes are lost (due to buggy cache purges)
```
2022-09-27T13:07:49Z	INFO	client/api.go:1021	waiting for votes to be validated...
Error: Process completed with exit code 124.
```

altergui:
here's an improved version (thanks to e1b1bef4f7898ac95ed13333251b3d74e72605cc) where it's evident that EnvelopeHeight only reaches 7 (so, 1 vote purged)
https://github.com/vocdoni/vocdoni-node/actions/runs/3135955002/jobs/5092259995#step:3:5003
```
2022-09-27T13:23:46Z	INFO	client/api.go:1021	waiting for votes to be validated...
2022-09-27T13:23:46Z	INFO	client/api.go:1033	process 5734e11003c5776d858d17c8018f07899db535866ebf805a32e0010200000000 envelope height: 2 (want 8)
2022-09-27T13:23:49Z	INFO	client/api.go:1033	process 5734e11003c5776d858d17c8018f07899db535866ebf805a32e0010200000000 envelope height: 7 (want 8)
Error: Process completed with exit code 124.
```

altergui:
and here are 5 runs, all of them successful, due to the hotfix applied

https://github.com/vocdoni/vocdoni-node/actions/runs/3135982868
https://github.com/vocdoni/vocdoni-node/actions/runs/3135999391
https://github.com/vocdoni/vocdoni-node/actions/runs/3136000179
https://github.com/vocdoni/vocdoni-node/actions/runs/3136000801
https://github.com/vocdoni/vocdoni-node/actions/runs/3136001393

(i ran this 5 times, using `git commit --amend && git push -f` given that it's a race condition)

altergui:
i also understood why this "cache purge" is not simply a cache purge,
 	`v.mempoolRemoveTxKeys(removeFromMempool, true)` actually removes the transaction from tendermint mempool
https://github.com/vocdoni/vocdoni-node/blob/bbb9a8cfa60aecf02ef8cfbad47d57f7ac264c26/vochain/cache.go#L116-L120

if the transaction is lucky, it will be purged at slightly different times on different nodes, and as long as it survives on 1 node, it will be rebroadcasted again to the rest of the nodes, and processed on the next block (since the buggy purge only happens every 18 blocks)

but if all the nodes are in sync (which i believe happens more often in docker-compose than in production) that's bad news for the transaction: it will be purged away from existence.

altergui:
just for completion, this bug was introduced on 2021/05/12
https://github.com/vocdoni/vocdoni-node/commit/b1857e1158ff3e88a1b53c1e8a2a25039d6ddb6e#diff-5def5afd87dc4943c14f87645fc78291e93d23c40500c387420c427d61f4005dR57

p4u:
Great catch @altergui, this bug has been here for some time and some weird behaviors observed in the past where probably because of it.

PR #625, opened 2022-09-29
[Breaking] Allow create accounts without InfoURI and force FaucetPayload

jordipainan:


cmd/vochaintest/vochaintest.go
```diff
@@ -1211,19 +1218,28 @@ func testSetTxCost(mainClient *client.Client, treasurerSigner *ethereum.SignKeys
...
+	txCost, err = mainClient.GetTransactionCost(models.TxType_COLLECT_FAUCET)
+	if err != nil {
+		return err
+	}
+	log.Infof("tx cost of %s fetched successfully (%d)", models.TxType_COLLECT_FAUCET, txCost)
 
-	// create account without faucet package
-	if err := mainClient.CreateOrSetAccount(signer,
+	// create account with faucet package
+	faucetPkg, err := mainClient.GenerateFaucetPackage(signer, signer2.Address(), 5000, rand.Uint64())
```
mvdan, line 1234: what is the 5000? perhaps consider a constant like `const startingBalance = 5000` or something like that, to make it clear

cmd/vochaintest/vochaintest.go
```diff
@@ -1614,6 +1575,20 @@ func testVocli(url, treasurerPrivKey string) {
 	}
 	log.Infof("alice key path: %s", aliceKeyPath)
 
+	if err := ioutil.WriteFile(path.Join(dir, "oraclePrivateKey"), []byte(oraclePrivateKey), 0700); err != nil {
```
mvdan, line 1578: any reason the private key should be executable? I think we want 0600

cmd/vochaintest/vochaintest.go
```diff
@@ -1614,6 +1575,20 @@ func testVocli(url, treasurerPrivKey string) {
...
+		log.Fatal(err)
+	}
+	_, stdout, _, err = executeCommand(vocli.RootCmd, append([]string{"keys", "import", path.Join(dir, "/oraclePrivateKey")}, stdArgs...), "", true)
```
mvdan, line 1581: if you're using path.Join, you don't need the slash. but also, I think you want filepath.Join.

cmd/vochaintest/vochaintest.go
mvdan: this should also be filepath.Join.

vochain/transaction.go
```diff
@@ -192,43 +192,43 @@ func (app *BaseApplication) AddTx(vtx *VochainTx, commit bool) (*AddTxResponse,
...
+					return nil, fmt.Errorf("setAccountInfoURITx: burnTxCost %w", err)
+				}
+				// consume faucet payload
+				if err := app.State.ConsumeFaucetPayload(
+					txValues.FaucetPayloadSigner,
+					&models.FaucetPayload{
+						Identifier: tx.GetFaucetPackage().GetPayload().GetIdentifier(),
+						To:         tx.GetFaucetPackage().GetPayload().GetTo(),
+						Amount:     tx.GetFaucetPackage().GetPayload().GetAmount(),
+					},
```
mvdan, line 211: can you not reuse `tx.GetFaucetPackage().GetPayload()` entirely here?

vochain/transaction.go
```diff
@@ -277,49 +276,54 @@ func (app *BaseApplication) AddTx(vtx *VochainTx, commit bool) (*AddTxResponse,
 		if commit {
 			switch vtx.Tx.GetSetAccountDelegateTx().Txtype {
 			case models.TxType_ADD_DELEGATE_FOR_ACCOUNT:
+				err = app.State.BurnTxCostIncrementNonce(txValues.From, models.TxType_ADD_DELEGATE_FOR_ACCOUNT)
```
mvdan, line 279: please be careful that you are ignoring these errors. that's why I'm a fan of `if err := f(); err != nil {...}`, because it makes it much more obvious.

vochain/transaction.go
```diff
@@ -277,49 +276,54 @@ func (app *BaseApplication) AddTx(vtx *VochainTx, commit bool) (*AddTxResponse,
...
 				if err := app.State.SetAccountDelegate(
 					txValues.From,
 					txValues.Delegate,
 					models.TxType_ADD_DELEGATE_FOR_ACCOUNT,
 				); err != nil {
 					return nil, fmt.Errorf("setAccountDelegate: %w", err)
 				}
-				err = app.State.SubtractCostIncrementNonce(txValues.From, models.TxType_ADD_DELEGATE_FOR_ACCOUNT)
 			case models.TxType_DEL_DELEGATE_FOR_ACCOUNT:
+				err = app.State.BurnTxCostIncrementNonce(txValues.From, models.TxType_DEL_DELEGATE_FOR_ACCOUNT)
```
mvdan, line 288: ditto

altergui:
i'm sorry you'll have to rewrite the vochaintest part, but it's for a good cause :)

altergui:
> i'm sorry you'll have to rewrite the vochaintest part, but it's for a good cause :)

in fact, the PR i just merged (causing you conflicts) fixes precisely the testsuite failure that happened in your PR run :angel:  
```
2022-09-29T21:16:40Z	FATAL	vochaintest/vochaintest.go:1647	submitRawTx failed: addTx setAccountInfoTxCheck: unauthorized: not enough balance to transfer
```

jordipainan:
Resuming this... Thanks for the comments @mvdan and @altergui  im going to close this PR in favor of another one, all the comments made here are addressed.

PR #640, opened 2022-10-26
vochain/scrutinizer: stop warning on inconsistent DBs outside tests

mvdan:
(see commit message)


mvdan:
"database is locked" appears to be a real bug I should fix, but I think it's unrelated here.

mvdan:
Split into https://github.com/vocdoni/vocdoni-node/issues/642 and re-running CI.

p4u (approved):

PR #644, opened 2022-10-26
introduce a new package censusdb

p4u:
The database operations previously implemented in the API census handler, are now moved to a standalone package.

The reasons are:
1) cleaner code since we do not mix API and database 
2) the censusdb can now be used by other components,
   such as the vochain/censusdownloader in order to
   import new census automatically when an election
   is created.

censusdb and api/census will deprecate censusManager


api/api.go
```diff
@@ -34,9 +34,8 @@ type API struct {
 	//lint:ignore U1000 unused
 	metricsagent *metrics.Agent
 	vocinfo      *vochaininfo.VochainInfo
-	censusMap    sync.Map
-
-	db db.Database
+	censusdb     *censusdb.CensusDB
+	db           db.Database // used for internal db operations
```
p4u, line 38: This database is currently used by the wallet handler. But ideally the wallet handler would have its own db so this "general" database can be removed from the API package.

api/censusdb/censusdb.go
```diff
@@ -0,0 +1,199 @@
...
+// SetTree sets the censustree.Tree object to the census reference.
+func (cr *CensusRef) SetTree(tree *censustree.Tree) {
+	cr.tree = tree
+}
+
+// CensusDB is a database of census trees.
+type CensusDB struct {
+	db           db.Database
+	censusMap    sync.Map
+	censusDBlock sync.RWMutex
```
mvdan: what is this mutex for? note that pebble is safe for concurrent use, as long as each goroutine uses a separate batch (transaction), which is the case.

api/censusdb/censusdb.go
```diff
@@ -0,0 +1,199 @@
...
+	ref, err := c.addCensusRefToDB(censusID, authToken, censusType, indexed, public)
+	if err != nil {
+		return nil, err
+	}
+	ref.tree = tree
+	if public {
+		ref.tree.Publish()
+	}
+	// add the tree in memory so we can quickly access to it afterwards
+	c.censusMap.Store(string(censusID), *ref)
```
mvdan: I'm confused by censusMap. If we're OK with in-memory storage, we don't need pebble. If we want to persist the data across restarts, then I think censusMap is incomplete, because it doesn't load data from an existing database when dvotenode is started.

mvdan: Also, assuming we want persistent storage, I think we could use pebble alone. As a key-value database which supports concurrent operations, one "get key" operation will be pretty fast. so I don't think a sync.Map is necessary.

api/censusdb/censusdb.go
```diff
@@ -0,0 +1,199 @@
...
+	// the removal of the tree from the disk is done in a separate goroutine.
+	// This is because the tree is locked and we don't want to block the operations,
+	// and depending on the size of the tree, it can take a while to delete it.
+	go func() {
+		c.censusMap.Delete(string(censusID))
+		_, err := censustree.DeleteCensusTreeFromDatabase(c.db, censusName(censusID))
+		if err != nil {
+			log.Warnf("error deleting census tree %x: %s", censusID, err)
+		}
+	}()
```
mvdan: If we keep pebble alone, without mutexes, then I would delete this goroutine. you would not be preventing other operations by keeping the transaction open for a little longer, to delete the entire census tree. you'd also save work - you'd use one goroutine and one transaction, whereas now you use two goroutines and two transactions.

api/censusdb/censusdb.go
```diff
@@ -0,0 +1,199 @@
...
+			[]byte(censusDBreferencePrefix),
+			censusID...,
+		))
+	if err != nil {
+		if errors.Is(err, db.ErrKeyNotFound) {
+			return nil, fmt.Errorf("census id not found in local storage")
+		}
+		return nil, err
+	}
+	dec := gob.NewDecoder(bytes.NewBuffer(b))
```
mvdan: nit: bytes.NewReader is enough

api/censusdb/censusdb.go
```diff
@@ -0,0 +1,199 @@
...
+
+// Tree returns the censustree.Tree object of the census reference.
+func (cr *CensusRef) Tree() *censustree.Tree {
+	return cr.tree
+}
+
+// SetTree sets the censustree.Tree object to the census reference.
+func (cr *CensusRef) SetTree(tree *censustree.Tree) {
+	cr.tree = tree
+}
```
mvdan, line 52: just make the tree field public :) to avoid encoding it in the gob, you could do something like:

```
type CensusRef struct {
    Tree ...
    Encoded EncodedCensusRef
}

type EncodedCensusRef struct {
    AuthToken ...
    ...
}
```

Then you only encode and decode the `Encoded` field. No real difference in terms of memory, but you do avoid the get/set methods.

api/censusdb/censusdb.go
p4u: Ummm, you are right, this lock might probably be removed. I'll check it.

api/censusdb/censusdb.go
p4u: Using pebble alone would work. But why not use a in-memory data store? It adds a bit more of complexity to the code, but might have a positive impact on the performance. These queries are handled by the API, for a large census with thousands of voters, there might be many Read access. So having a copy in memory for those references that are actually used, looks like a good approach.

api/censusdb/censusdb.go
p4u: The Delete operation might take many seconds (depending on the size of the census) since the only way to achieve it is by iterating over all tree leafs. This goroutine ensures the HTTP request is replied before any timeout is triggered. And once the reference is removed (that's executed before launching the routine) at practical effects the census is not available anymore for the API.

api/censusdb/censusdb.go
p4u: Nice trick. I was looking for a struct tag able to exclude a value from gob but it does not exist. However I like more the get/set approach since we are only avoiding 1 field from serialization (fore more, I'd see your approach better).

api/censusdb/censusdb.go
p4u: Do you think Pebble cache will offer the same performance as having this sync.Map @mvdan ?

mvdan: I'm willing to bet that brief reads on pebble are not going to be noticeably slow. I can see on their benchmarks that write operations execute at tens of thousands per second, and read operations at hundreds of thousands per second, for example.

Ultimately there aren't benchmarks or performance measurements here, so we're both just guessing. Which is why I'd personally go with less code until we can prove that there's a bottleneck that needs fixing. But I would be very surprised if the read throughput of pebble is not enough for our load.

PR #648, opened 2022-10-27
F/data downloader

p4u:
Introduce new package data downloader
    
    Most of the code is copyied from census/importqueue.go
    
    The idea is to make an async generic file import package for the
    data interface (IPFS currently). This way we can use it for
    importing census but also other files such as entity or election
    metadata.


p4u:
The CI found a data race related to the random number generator used in the test. 
Can you help here @mvdan? any glue why is that happening? 

https://github.com/vocdoni/vocdoni-node/issues/649

PR #650, opened 2022-10-27
mod: bump a few more dependencies

mvdan:
(see commit message)


mvdan:
big diff because we somehow had two split `require` sections with indirect lines, so I joined them again.

PR #651, opened 2022-10-28
[BREAKING] Dvot 500/f/set account

jordipainan:
[BREAKING CHANGES]

This PR changes de behavior of setAccountInfo, refactors the mentioned tx to setAccount and adapts the required code for having all working properly.

Now with set account you can:
- Create an account only if a valid faucet package is provided
- Create an account without an infoURI
- Add/Del delegates in batch just changing the inner SetAccountTx TxType

The inner txtypes are maintained for backwards compatibility.

As a requirement for making the CI tests pass it also deletes the vocli
package and its tests, see commit message.


jordipainan:
There are 3 breaking PRs, merge this first into master.

p4u:
@jordipainan please correct me if I'm wrong (I have not yet went deep into the code), but I'd suggest to use only SET_ACCOUNT to configure any information related with the account (maybe even delegates?). I think it makes more sense and makes things simpler. 

SET_ACCOUNT could be executed only if:
1. A faucet package is provided
2. There is enough funds into the wallet for paying the fee
3. The fee cost set is zero

Does this makes sense to you?

apiclient/account.go
```diff
@@ -46,7 +46,7 @@ func (c *HTTPclient) Transfer(to common.Address, amount uint64) (types.HexBytes,
 	stx.Tx, err = proto.Marshal(&models.Tx{
 		Payload: &models.Tx_SendTokens{
 			SendTokens: &models.SendTokensTx{
-				Txtype: models.TxType_SET_ACCOUNT_INFO,
+				Txtype: models.TxType_SET_ACCOUNT_INFO_URI,
```
p4u, line 49: Why do we need a set_account_info_uri transaction if we can already set the account using SET_ACCOUNT?

jordipainan:
> @jordipainan please correct me if I'm wrong (I have not yet went deep into the code), but I'd suggest to use only SET_ACCOUNT to configure any information related with the account (maybe even delegates?). I think it makes more sense and makes things simpler.
> 
> SET_ACCOUNT could be executed only if:
> 
>     1. A faucet package is provided
> 
>     2. There is enough funds into the wallet for paying the fee
> 
>     3. The fee cost set is zero
> 
> 
> Does this makes sense to you?

Hey @p4u, i do not fully understand 3 :)
But maybe if I give a bit more context we can go for it.

What we had:
- `SetAccountInfoTx`: 
     - Can create accounts, with or without a faucet package, a valid info uri must be provided
     - Can change the info uri of the account.
     - Cannot add delegates in this same tx when creating the account
- `AddDelegateForAccountTx`: adds a delegate 
- `DelDelegateForAccountTx`: remove a delegate

What we have with this PR
- `SetAccountTx` with 3 inner Tx.Type `[SetAccountInfoURI, AddDelegateForAccount, DelDelegateForAccount]`

An account can only be created if a valid faucet package is provided and the account that generated the faucet package
have funds for paying it.
An account can be created without any info URI
An account can be created with delegates and you do not need 2 txs as it is in the past.
Delegates can be added/removed in batch

I did not change the inner Tx.Types for not adding more changes, also on the SDK side, but if you want another naming im open to do so, despite that I think we can tackle this later.

p4u:
The approach LGTM

altergui (approved):
LGTM

api/wallet.go
```diff
@@ -207,9 +207,9 @@ func (a *API) walletCreateHandler(msg *bearerstdapi.BearerStandardAPIdata, ctx *
...
 	stx := models.SignedTx{}
 	stx.Tx, err = proto.Marshal(&models.Tx{
-		Payload: &models.Tx_SetAccountInfo{
-			SetAccountInfo: &models.SetAccountInfoTx{
-				Txtype:        models.TxType_SET_ACCOUNT_INFO,
+		Payload: &models.Tx_SetAccount{
+			SetAccount: &models.SetAccountTx{
+				Txtype:        models.TxType_SET_ACCOUNT_INFO_URI,
 				Nonce:         0,
 				InfoURI:       "none",
```
p4u: I understand this could be empty now.

altergui:
really bizarre. no difference at all, blockheight is increasing just fine on server (left) just as in laptop (right). but server gives "account 0x... does not exist"
image

if this is blocking other PRs (i understand it does), i'd say it's ok to just ignore this, remove the line so the test passes like you did, and continue with the rest. with some calm we'll sort this out in due time

api/wallet.go
p4u: is this solved @jordipainan?

api/wallet.go
```diff
@@ -206,12 +206,13 @@ func (a *API) walletCreateHandler(msg *bearerstdapi.BearerStandardAPIdata, ctx *
 	}
 
 	stx := models.SignedTx{}
+	infoURI := string("none")
```
p4u: I'd let this just an empty string. There is already a zero-value for a string type so I don't see the need for setting an own one.

vochain/account.go
```diff
@@ -67,9 +67,6 @@ func (a *Account) IsDelegate(addr common.Address) bool {
 
 // AddDelegate adds an address to the list of delegates for an account
 func (a *Account) AddDelegate(addr common.Address) error {
-	if a.IsDelegate(addr) {
```
p4u, line 70: Why this check is removed?

vochain/account.go
```diff
@@ -208,251 +223,248 @@ func (v *State) IncrementAccountProcessIndex(accountAddress common.Address) erro
...
-	if accountAddress == types.EthereumZeroAddress {
-		return fmt.Errorf("invalid address")
-	}
-	// check not created
-	acc, err := v.GetAccount(accountAddress, false)
-	if err != nil {
-		return fmt.Errorf("cannot create account %s: %w", accountAddress.String(), err)
+// ConsumeFaucetPayload consumes a given faucet payload storing
+// its key to the FaucetNonce tree so it can
+// only be used once
```
p4u: This line breaking does might not be necessary on the comment.

vochain/account.go
```diff
@@ -208,251 +223,248 @@ func (v *State) IncrementAccountProcessIndex(accountAddress common.Address) erro
...
+func (v *State) ConsumeFaucetPayload(from common.Address, faucetPayload *models.FaucetPayload) error {
+	// check faucet payload
+	if faucetPayload == nil {
+		return fmt.Errorf("invalid faucet payload")
 	}
-	if acc != nil {
-		return ErrAccountAlreadyExists
+	// check from account
+	if from == types.EthereumZeroAddress {
```
p4u: What's the reason for this check? The from comes from a signature, so in theory the zero Address  will never be a from.

vochain/account.go
```diff
@@ -208,251 +223,248 @@ func (v *State) IncrementAccountProcessIndex(accountAddress common.Address) erro
...
-			if !bytes.Equal(v.Bytes(), types.EthereumZeroAddress[:]) {
-				acc.DelegateAddrs = append(acc.DelegateAddrs, v.Bytes())
-			}
-		}
+	// store faucet identifier
+	b := make([]byte, 8)
+	binary.LittleEndian.PutUint64(b, faucetPayload.Identifier)
+	key := from.Bytes()
+	key = append(key, b...)
+	keyHash := crypto.Sha256(key)
```
p4u: Nit: you can do `keyHash := crypto.Sha256(append(from.Bytes(), b...))`

Is there a reason for using crypto.Sha256() here? Everywhere in the code we use ethereum.HashRaw() as default hashing function.

vochain/account.go
```diff
@@ -208,251 +223,248 @@ func (v *State) IncrementAccountProcessIndex(accountAddress common.Address) erro
...
+	return nil
 }
 
-// ConsumeFaucetPayload consumes a given faucet payload and sends the given amount of tokens to
-// the address pointed by the payload.
-func (v *State) ConsumeFaucetPayload(from common.Address, faucetPayload *models.FaucetPayload, txType models.TxType) error {
-	// check faucet payload
-	if faucetPayload == nil {
-		return fmt.Errorf("faucet payload is nil")
+// CreateAccountTxCheck is an abstraction of the ABCI CheckTx method for creating an account
```
p4u: This looks a wrong IA comment for this method.

vochain/account.go
```diff
@@ -208,251 +223,248 @@ func (v *State) IncrementAccountProcessIndex(accountAddress common.Address) erro
...
-	burnAcc, err := v.GetAccount(BurnAddress, false)
+	infoURI := tx.GetInfoURI()
+	if len(infoURI) > types.MaxURLLength {
+		return ErrInvalidURILength
+	}
+	if len(tx.Delegates) != 0 {
+		delegatesMap := make(map[common.Address]bool)
+		for _, v := range tx.Delegates {
+			delegateAddress := common.BytesToAddress(v)
+			if delegateAddress == types.EthereumZeroAddress {
```
p4u: I'd not consider EthereumZeroAddress as an special address. Its just too difficult to find it.

vochain/account.go
```diff
@@ -208,251 +223,248 @@ func (v *State) IncrementAccountProcessIndex(accountAddress common.Address) erro
...
+			faucetPayload.Amount+txCost,
+		)
+	}
+	return nil
+}
 
-	if err := v.SetAccount(accToAddr, accTo); err != nil {
-		return err
+// CreateAccount creates an account
+func (v *State) CreateAccount(accountAddress common.Address,
```
p4u: We do not use this line breaking for method definition.

vochain/account.go
```diff
@@ -208,251 +223,248 @@ func (v *State) IncrementAccountProcessIndex(accountAddress common.Address) erro
...
+	infoURI string,
+	delegates [][]byte) error {
+	return v.createAccount(accountAddress, infoURI, delegates, 0)
+}
+
+func (v *State) createAccount(accountAddress common.Address,
```
p4u: Same

vochain/account.go
```diff
@@ -208,251 +223,248 @@ func (v *State) IncrementAccountProcessIndex(accountAddress common.Address) erro
...
+func SetAccountInfoTxCheck(vtx *models.Tx, txBytes, signature []byte, state *State) error {
+	if vtx == nil || signature == nil || txBytes == nil || state == nil {
+		return fmt.Errorf("invalid transaction parameters provided")
+	}
+	tx := vtx.GetSetAccount()
+	if tx == nil {
+		return fmt.Errorf("invalid transaction")
+	}
+	if tx.Nonce == nil || tx.InfoURI == nil {
+		return fmt.Errorf("missing signature and/or transaction")
```
p4u: This error message seems wrong

vochain/account.go
```diff
@@ -208,251 +223,248 @@ func (v *State) IncrementAccountProcessIndex(accountAddress common.Address) erro
...
-	returnValues.TxSender, err = ethereum.AddrFromPublicKey(pubKey)
+	txSenderAddress, err := ethereum.AddrFromPublicKey(pubKey)
 	if err != nil {
-		return nil, fmt.Errorf("cannot extract address from public key: %w", err)
+		return fmt.Errorf("cannot extract address from public key: %w", err)
 	}
-	// get txSender account
-	txSender, err := state.GetAccount(returnValues.TxSender, false)
+	var txAccountAddress common.Address
+	if len(tx.Account) == 0 || bytes.Equal(tx.Account, types.EthereumZeroAddress.Bytes()) {
```
p4u: Same with zero address... But in the other side I don't understand the reason behind this check. Why not just take the txSenderAddress and ignore Account? Or the other option would be to enforce tx.Account to be equal to txSenderAddress.

vochain/account.go
```diff
@@ -208,251 +223,248 @@ func (v *State) IncrementAccountProcessIndex(accountAddress common.Address) erro
...
+	// get tx account Account
+	txAccountAccount, err := state.GetAccount(txAccountAddress, false)
 	if err != nil {
-		return nil, err
+		return fmt.Errorf("cannot get tx account: %w", err)
 	}
-	// check txSender balance
-	if txSender.Balance < cost {
-		return nil, fmt.Errorf("unauthorized: %s", ErrNotEnoughBalance)
+	if txAccountAccount == nil {
```
p4u, line 429: Isn't this check already performed?

vochain/account.go
```diff
@@ -208,251 +223,248 @@ func (v *State) IncrementAccountProcessIndex(accountAddress common.Address) erro
...
+		return ErrAccountNotExist
+	}
+	if infoURI == txAccountAccount.InfoURI {
```
p4u, line 432: This check also seems to be repeated

vochain/account.go
```diff
@@ -208,251 +223,248 @@ func (v *State) IncrementAccountProcessIndex(accountAddress common.Address) erro
...
+	// check info URI
+	infoURI := tx.GetInfoURI()
+	if len(infoURI) == 0 || len(infoURI) > types.MaxURLLength {
+		return fmt.Errorf("invalid URI, cannot be empty")
 	}
-	// check if delegate
-	if returnValues.TxSender != returnValues.Account {
-		if !acc.IsDelegate(returnValues.TxSender) {
-			return nil, fmt.Errorf("tx sender is not a delegate")
+	if txSenderAddress == txAccountAddress {
```
p4u, line 417: How can txSenderAddress and txAccountAddress be different? One is obtained from the other.

vochain/account.go
```diff
@@ -461,250 +473,253 @@ func (v *State) SetAccount(accountAddress common.Address, account *Account) erro
...
+		}`,
 		accountAddress.String(),
 		account.Nonce,
 		account.InfoURI,
 		account.Balance,
 		printPrettierDelegates(account.DelegateAddrs),
+		account.ProcessIndex,
 	)
+	for _, l := range v.eventListeners {
+		l.OnSetAccount(accountAddress.Bytes(), account)
```
p4u: I think passing a the account itself makes more sense than just the bytes of the address. But take care on not passing it as a pointer but as a value.

vochain/account.go
```diff
@@ -461,250 +473,253 @@ func (v *State) SetAccount(accountAddress common.Address, account *Account) erro
...
-	}
-	// get treasurer
-	treasurer, err := state.Treasurer(false)
-	if err != nil {
-		return common.Address{}, 0, err
-	}
-	// check nonce
-	if tx.Nonce != treasurer.Nonce {
-		return common.Address{}, 0, fmt.Errorf("invalid nonce %d, expected: %d", tx.Nonce, treasurer.Nonce)
+	if len(tx.To) == 0 || bytes.Equal(tx.To, types.EthereumZeroAddress.Bytes()) {
```
p4u: What the zero address did to you?? :)

vochain/account.go
```diff
@@ -461,250 +473,253 @@ func (v *State) SetAccount(accountAddress common.Address, account *Account) erro
...
-	sigAddress, acc, err := state.AccountFromSignature(txBytes, signature)
+	txSenderAddress, txSenderAccount, err := state.AccountFromSignature(txBytes, signature)
 	if err != nil {
-		return nil, err
+		return err
 	}
-	// check delegate to add is not itself
-	if delAcc == *sigAddress {
-		return nil, fmt.Errorf("cannot add self to delegates list")
+	delegatesToSetMap := make(map[common.Address]bool)
```
p4u: I saw this code snipped already. Is it necessary to duplicate it? In theory the TxCheck functions must be called before modifying the sate, so one can relay on the other.

vochain/transaction.go
```diff
@@ -185,54 +185,155 @@ func (app *BaseApplication) AddTx(vtx *VochainTx, commit bool) (*AddTxResponse,
...
-						0,
-					); err != nil {
-						return nil, fmt.Errorf("setAccountInfoTxCheck: createAccount %w", err)
+			switch tx.Txtype {
+			case models.TxType_CREATE_ACCOUNT:
+				txSenderAddress, err := ethereum.AddrFromSignature(vtx.SignedBody, vtx.Signature)
+				if err != nil {
+					return nil, fmt.Errorf("createAccountTx: txSenderAddress %w", err)
+				}
+				delegates := make([][]byte, 0)
```
p4u: Third time I see this code

p4u (changes requested):

api/wallet.go
jordipainan: Yes it CAN be empty, but this method needs to be refactored anyways for adding delegates too on the account bootstrap. I was planning to do it when tacking the client and the API with the tests.

I made it empty btw

vochain/account.go
jordipainan: Because is already checked on the on check and deliver tx, but I added it again as well on RemoveDelegate.

vochain/account.go
jordipainan: Yup good one, removed.

vochain/account.go
jordipainan: Applied suggestion and I've also changed the `crypto.Sha256` for our own `ethereum.HashRaw` it does not make sense to use another hash function, here the autocompletion played bad :S

vochain/account.go
jordipainan: hahah, I made it by myself. Changed for looking less IA and more human.

vochain/account.go
jordipainan: because it can be a delegate and the signature points to the delegate address

vochain/account.go
jordipainan: if delegates can be different, the txSenderAddress is the address sending the tx, the txAccountAddress is the address from the Account field if the tx if not empty, otherwise is the same as txSenderAddress

vochain/account.go
jordipainan: nope, first time

vochain/account.go
jordipainan: already passing the account, but changed to pass a copy

vochain/account.go
jordipainan: nothing haha, is a basic check all the eth code does to prevent operating with the zero account, and coming from that world i took this habit.
Anyways, deleted all zeroaddr checks.

vochain/account.go
jordipainan: they have different purposes, one checks duplicates against tx.delegates and the other against the account that will be eventually modified.

vochain/transaction.go
jordipainan: moved this check out of the commit phase, and added a generic checkDuplicateDelegates util function

vochain/account.go
jordipainan: update: still checking for delegates operation as the tx.Account is a field that can be different than txSender for those cases.

vochain/offchaindatahandler/offchaindatahandler.go
```diff
@@ -149,7 +149,12 @@ func (c *OffChainDataHandler) OnRevealKeys(pid []byte, priv string, txindex int3
 func (c *OffChainDataHandler) OnProcessStatusChange(pid []byte,
 	status models.ProcessStatus, txindex int32) {
 }
-
+func (c *OffChainDataHandler) OnSetAccount(addr []byte, account *vochain.Account) error {
```
p4u, line 152: AFAIK the address is found inside the account

vochain/transaction.go
```diff
@@ -185,54 +185,139 @@ func (app *BaseApplication) AddTx(vtx *VochainTx, commit bool) (*AddTxResponse,
 			return response, app.State.AddToRollingCensus(tx.ProcessId, tx.NewKey, weight)
 		}
 
-	case *models.Tx_SetAccountInfo:
-		txValues, err := SetAccountInfoTxCheck(vtx.Tx, vtx.SignedBody, vtx.Signature, app.State)
-		if err != nil {
-			return nil, fmt.Errorf("setAccountInfoTxCheck: %w", err)
+	case *models.Tx_SetAccount:
+		tx := vtx.Tx.GetSetAccount()
+		var err error
```
p4u: I think we don't need to declare this error var here

vochain/transaction.go
```diff
@@ -185,54 +185,139 @@ func (app *BaseApplication) AddTx(vtx *VochainTx, commit bool) (*AddTxResponse,
...
+					return nil, fmt.Errorf("setAccountDelegate: burnTxCostIncrementNonce %w", err)
+				}
+				if err := app.State.SetAccountDelegate(
+					txSenderAddress,
+					tx.Delegates,
+					models.TxType_ADD_DELEGATE_FOR_ACCOUNT,
+				); err != nil {
+					return nil, fmt.Errorf("setAccountDelegate: %w", err)
+				}
+			case models.TxType_DEL_DELEGATE_FOR_ACCOUNT:
```
p4u, line 302: Nit: let's add either a blank line between "case" statements or not (I think we should add them here).

p4u (approved):

vochain/offchaindatahandler/offchaindatahandler.go
jordipainan: it is not, the address have never been part of the account struct

vochain/transaction.go
jordipainan: I guess you are saying this because of the next switch case, I've added a new line on those because is less clear when reading because you have more code, but in the former I do not think is necessary cause someone can clearly distinguish between each switch case.
Said so, I've added it as you are recommending.

vochain/transaction.go
jordipainan: its a minor i fixed on #622 but lets fix it here too :)

jordipainan:
added a few more test cases, even if seems a bit redundant it always help :)

vochain/offchaindatahandler/offchaindatahandler.go
jordipainan: @p4u Let me know if you are thinking about to change this because it has lots of implications on the code and I wont go ahead with this PR.

p4u: You'r right, the address is the key for the merkle tree storing the Account values.

p4u:
I thinks this PR should also update the DEV genesis since it includes breaking changes @jordipainan

jordipainan:
> 

yes correct, I can upload the chainID too, actually changes the 3 environments genesis.

PR #652, opened 2022-10-28
[BREAKING] Dvot 499/f/add delegates logic

jordipainan:
Adds delegates logic on process related txs.

Refactors the process_test.go file for better readability.
Adds more tests to process_test.go.

PR #653, opened 2022-10-28
[BREAKING] vochain: use the process entityID for entity process count

jordipainan:
The entity process counter increase uses the tx sender for increasing the entity process counter.
This does not work if an oracle or a delegate is used as the increament on the counter is made on the oracle or on the process.entityID delegate account.

In the past was done this way because for token based voting the entityID is the contract address. This is going to be solved with the use of the field sourceNetworkContractAddress but for now just ignore the case where the oracle
created the tx (no increase done) and use correcly the increase on the entity process counter when the
process is created by the entity itself or by a delegate.


jordipainan:
There are 3 breaking PRs, merge this the second (not to master)

jordipainan:
Ready to merge after #652 @p4u

p4u (approved):

PR #666, opened 2022-11-05
bump Go, action, and staticcheck versions

mvdan:
(see commit message)


p4u:
the link is wrong @mvdan 

here the good one: https://github.com/dominikh/go-tools/releases/download/v0.3.3/staticcheck_linux_amd64.tar.gz

mvdan:
Yeah, upstream changed how they release assets. I asked them about it two days ago, because it feels unintentional, but they haven't responded. I'll just fix the link for now.

PR #667, opened 2022-11-07
Enhanced version of main.yml to improve performance

chilcano:
- removed runs-on:container section: it is not required since github actions already runs ci/cd tasks in isolated environment or workspaces
- all github actions are been updated to latest stable version: i.e. actions/checkout from v2 to v3
- used actions/setup-go@v3 to setup golang environment (even cache)
- configured docker/build-push-action@v3 to work with cache
- workflow prepared to be tested, once tested the older one should be disabled

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,41 +1,63 @@
 ---
-name: Main
+name: MainCI
 
 on: [push, pull_request]
 
 jobs:
-  go-tests:
-    runs-on: self-hosted
-    container:
-      image: golang:1.18.6
-      options: --user 1000
+  job_go_tests:
+    runs-on: ubuntu-latest      # if using self-hosted, make sure the runner has not conflicts PATH, CACHE, etc. with actions/setup-xyz
     if: github.repository == 'vocdoni/vocdoni-node'
     defaults:
       run:
         shell: bash
     steps:
-      - name: Checkout code
-        uses: actions/checkout@v2
-      - name: gofmt
+      - 
+        name: Checkout code
+        uses: actions/checkout@v3
+      - 
+        name: Set up Go environment
+        uses: actions/setup-go@v3
+        with:
+          go-version: '1.18.6'
+          cache: true
+      - 
+        name: Check Go and tidy cache
+        run: |
+          echo "* Golang version: $(go version)"
+          echo "* HOME value in this node: $HOME"
+          echo "* Running go mod tidy"
+          go mod tidy
+      - 
+        name: Run gofmt
         # Run gofmt first, as it's quick and issues are common.
         run: diff -u <(echo -n) <(gofmt -s -d $(git ls-files '*.go'))
-      - name: go test
+      - 
+        name: Run go test cover
         run: |
           # check that log lines contains no invalid chars (evidence of format mismatch)
           export LOG_PANIC_ON_INVALIDCHARS=true
           # we run vet in another step
           go test -vet=off -timeout=1m -coverprofile=covprofile ./...
+      - 
+        name: go test race
+        run: |
+          # check that log lines contains no invalid chars (evidence of format mismatch)
+          export LOG_PANIC_ON_INVALIDCHARS=true
           # -race can easily make the crypto stuff 10x slower
           go test -vet=off -timeout=15m -race ./...
-      - name: go vet
+      - 
+        name: Run go vet
         run: go vet ./...
-      - name: Download staticcheck
+      - 
+        name: Download staticcheck
         run: |
           curl -L https://github.com/dominikh/go-tools/releases/download/2022.1.2/staticcheck_linux_amd64.tar.gz | tar -xzf -
-      - name: staticcheck
+      - 
+        name: Run staticcheck
         run: |
           ./staticcheck/staticcheck ./... 2> staticcheck/stderr
-      - name:  check staticcheck stderr
+      - 
+        name: Check staticcheck stderr (this step isn't needed because we are using actions/setup-go@v3 on GitHub hosted runner)
         run: |
           if cat staticcheck/stderr | grep "matched no packages" ; then
             echo "staticcheck step did nothing, due to https://github.com/vocdoni/vocdoni-node/issues/444"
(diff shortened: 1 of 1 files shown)


p4u:
@mvdan @altergui since you were involved with the CI file definition, it would be great if you can take a look on the new proposal.

mvdan:
This has conflicts as I recently touched this file, please fix them.

Is there a reason to change the whitespace and rename the jobs? I don't see an advantage either way, so it's probably best to keep it as is to keep the changes small :)

chilcano:
Hi @mvdan
Just fixed the conflicts and reverted to `- name: xyz...` syntax to reduce the number of changes, however doing that is complicated since there are many things that need be updated in this workflow and others ( CC/ @emmdim ).
Anyway, these changes should reduce the execution time less of 10 mins.
If you need further explanation, I'm happy to setup a call to explain you the changes I've introduced.

.github/workflows/main.yml
```diff
@@ -1,41 +1,54 @@
...
       - name: Download staticcheck
+        # staticcheck provides a github action, use it: https://staticcheck.io/docs/running-staticcheck/ci/github-actions/
         run: |
           curl -L https://github.com/dominikh/go-tools/releases/download/v0.3.3/staticcheck_linux_amd64.tar.gz | tar -xzf -
-      - name: staticcheck
+      - name: Run staticcheck
         run: |
           ./staticcheck/staticcheck ./... 2> staticcheck/stderr
-      - name:  check staticcheck stderr
+      - name: Check staticcheck stderr (this step isn't needed because we are using actions/setup-go@v3 on GitHub hosted runner)
```
mvdan, line 51: This is all done manually due to a bug in Go, where bad permissions can lead to staticcheck silently doing nothing. See https://github.com/golang/go/issues/50869.

We can switch to just running staticcheck normally or via its github action, but then we risk silently not running staticcheck again in the future. IMO we should leave this workaround in place, even if it's clunky, until we use a Go version with that bugfix included.

.github/workflows/main.yml
```diff
@@ -1,41 +1,54 @@
...
         uses: actions/checkout@v3
-      - name: gofmt
+      - name: Set up Go environment
+        uses: actions/setup-go@v3
+        with:
+          go-version: '1.18.8'
+          cache: true
+      - name: Check Go and tidy cache
+        run: |
+          echo "* Golang version: $(go version)"
```
mvdan, line 23: doesn't setup-go already print the Go environment?

.github/workflows/main.yml
```diff
@@ -1,41 +1,54 @@
...
+          echo "* HOME value in this node: $HOME"
+          echo "* Running go mod tidy"
```
mvdan, line 25: don't the logs already say what commands are being run?

mvdan (approved):
SGTM, just a few more minor thoughts

p4u:
@chilcano the reason behind using our own runner is to bypass the maximum number of minutes CI runners Github have. Currently we are in the first paid tier level which includes 3000 minutes per month. But that limit applies to all repositories within the vocdoni organization. Thus is not hard to reach it (last month we reached more than 2000).

To this end I'd re-enable the self-hosted runners, at least for one of both tests. If there is something wrong happening in our CI server, we might fix it (the current CI at hetzner might be reinstalled from scratch if there are issues with the current setup).

.github/workflows/main.yml
chilcano: Since it is a limitation of Go, using another Linter or StaticCheck is not going to fix the problem.
Fair enough!

.github/workflows/main.yml
chilcano: Unfortunately, `setup-go` doesn't print go envs.

.github/workflows/main.yml
chilcano: yes, they already are shown in the console, these lines can be removed.

chilcano:
> @chilcano the reason behind using our own runner is to bypass the maximum number of minutes CI runners Github have. Currently we are in the first paid tier level which includes 3000 minutes per month. But that limit applies to all repositories within the vocdoni organization. Thus is not hard to reach it (last month we reached more than 2000).
> 
> To this end I'd re-enable the self-hosted runners, at least for one of both tests. If there is something wrong happening in our CI server, we might fix it (the current CI at hetzner might be reinstalled from scratch if there are issues with the current setup).

We should create/install fresh self-hosted runners and should be used for specialized GitHub jobs. 
My recommendations are:
1. Group in a GitHub job smaller and specialized GitHub steps and assign it a self-hosted runner. For example, move all specialized steps (own tests -stress, load, security, ...-, build containers, etc.) which are steps that take too much time to run (see compose-test, docker-release, go test/coverage, ...), into a GitHub job with an exclusive self-hosted runner.
2. For other steps like checkout, setup-go, cache, etc., even staticcheck, group them in a different job and run it in a GitHub public runner. These kind of steps are common and don't take much time to run if they are configured properly.

In brief, split the workflow in smaller steps, group steps based on specialization to run on public or self-hosted runners. At the same time, prepare fresh self-hosted runners (based on hetzner or google cloud) for these group of steps.

p4u:
> > @chilcano the reason behind using our own runner is to bypass the maximum number of minutes CI runners Github have. Currently we are in the first paid tier level which includes 3000 minutes per month. But that limit applies to all repositories within the vocdoni organization. Thus is not hard to reach it (last month we reached more than 2000).
> > To this end I'd re-enable the self-hosted runners, at least for one of both tests. If there is something wrong happening in our CI server, we might fix it (the current CI at hetzner might be reinstalled from scratch if there are issues with the current setup).
> 
> We should create/install fresh self-hosted runners and should be used for specialized GitHub jobs. My recommendations are:
> 
> 1. Group in a GitHub job smaller and specialized GitHub steps and assign it a self-hosted runner. For example, move all specialized steps (own tests -stress, load, security, ...-, build containers, etc.) which are steps that take too much time to run (see compose-test, docker-release, go test/coverage, ...), into a GitHub job with an exclusive self-hosted runner.
> 2. For other steps like checkout, setup-go, cache, etc., even staticcheck, group them in a different job and run it in a GitHub public runner. These kind of steps are common and don't take much time to run if they are configured properly.
> 
> In brief, split the workflow in smaller steps, group steps based on specialization to run on public or self-hosted runners. At the same time, prepare fresh self-hosted runners (based on hetzner or google cloud) for these group of steps.

Thanks for your comments. Your general idea makes sense to me.

I have some comments regarding some specific tasks you mention:
- docker-release should take less time than tests, since both compile the whole repository but the second performs also testing.
- docker-compose is already a different job
- go coverage might be removed since we do not use it, but if we want to keep it, I agree on moving it to a separate job.

That said, I think `go test -race` is one of the heavy actions that might be split into another job, however I'm not sure its worth it taking into account the time of preparing the environment (create the container, checkout code and compile all packages).

IMHO the existing split is good. I'd just take "go coverage" out.

I'd do a new CI installation and use if as self-runner at least for 2/3 of the jobs. But in any case I'd merge this PR since it already adds some important improvements and let the self-runner CI as an upcoming task.

p4u (approved):

.github/workflows/main.yml
mvdan: It is - expand "go env" at the end of the setup-go step.

PR #672, opened 2022-11-14
data/ipfs: upgrade IPFS CID to v1

altergui:
data/ipfs: upgrade IPFS CID to v1 

 * reimplement Publish using i.CoreApi
 * move unixfsFilesNode out of AddAndPin
 * reimplement AddAndPin using i.CoreAPI
 * lint Pin and Unpin
 * simpler PublishIPNSpath, refactoring addAndPin to return corepath.Resolved
 * use IPFSCIDv1, drop IPFScontentIdentifier that returned v0
 * new func MockStorage needed by api/census_test

### DIFF
--- a/data/ipfs.go
+++ b/data/ipfs.go
@@ -8,7 +8,7 @@ import (
 	"strings"
 	"time"
 
-	"github.com/ipfs/go-cid"
+	ipfscid "github.com/ipfs/go-cid"
 	files "github.com/ipfs/go-ipfs-files"
 	keystore "github.com/ipfs/go-ipfs-keystore"
 	ipfslog "github.com/ipfs/go-log"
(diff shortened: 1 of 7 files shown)


altergui:
(base is wrong, i need to rebase on master, solving conflicts)

data/ipfs.go
```diff
@@ -303,13 +295,12 @@ func (i *IPFSHandle) Retrieve(ctx context.Context, path string, maxSize int64) (
...
 // so the caller must handle properly the logic by using goroutines, channels or other
 // mechanisms in order to not block the whole program execution.
-func (i *IPFSHandle) PublishIPNSpath(ctx context.Context, path string,
-	keyalias string) (coreiface.IpnsEntry, error) {
-	root, err := i.AddAndPin(ctx, path)
-	if err != nil {
-		return nil, err
-	}
-	c, err := cid.Parse(root)
+func (i *IPFSHandle) PublishIPNSpath(
```
p4u: I know some linters want to split the parameters this way, but we decided not to follow the approach. Instead we split the lines in the minimum number while trying not to surpass 111 chars and make them as symmetric as possible.

p4u (changes requested):
@altergui there were some commits on the original branch (the one I started) that I'd like to maintain. Such as the ones modifying ipfsSync to allow "onlyConnect" mode.

p4u:
> 

Ah sorry, I miss read the branch name. So my changes are here https://github.com/vocdoni/vocdoni-node/tree/f/ipfs_sync_fixes
Could you cherry-pick those that do not collide with your work?

data/ipfs.go
altergui: good to know, i dropped that gofumptd commit

altergui:
> > 
> 
> Ah sorry, I miss read the branch name. So my changes are here https://github.com/vocdoni/vocdoni-node/tree/f/ipfs_sync_fixes Could you cherry-pick those that do not collide with your work?

ok but they seem out of scope for this PR so i'll prepare a separate one, review this as-is

p4u (approved):
The changes LGTM, the whole implementation can probably be improved but that's nothing we can do now. @mvdan can you also take a look?

mvdan (approved):
SGTM

altergui:
squashed everything and added a simple unit test FTW

PR #685, opened 2022-11-15
DOPS-396 add job go tests extra to PR

chilcano:
Removed `if: github.ref == 'refs/heads/master'` condition of job_go_tests_extra to force its execution


altergui:
could i request, while you're at it, to reorder the job_go_build_for_mac down? this test is skipped on PRs and the interesting result of job_compose_test is pushed down, needing to scroll to view result. really trivial issue, i was preparing a PR https://github.com/vocdoni/vocdoni-node/commit/61a88ce613f64176fddb987900324e78085d185d but i thought didn't really merit a PR
image

chilcano:
Hi @altergui 
I'm going to make your recommended changes in this PR.
Regards.

altergui:
i just realized that job_go_tests doesn't really run `go test` anymore :confused: 
maybe rename to `job_go_checks`then?
and `job_go_tests_extra` wouldn't be "extra" but actually `job_go_test`

p4u:
> i just realized that job_go_tests doesn't really run `go test` anymore confused maybe rename to `job_go_checks`then? and `job_go_tests_extra` wouldn't be "extra" but actually `job_go_test`

Umm, "go test" should be run on every push (no matter if PR or not). If the current "go_tests" job does not make tests, we should add it.

In the other side, go test extra can perform more time-consuming testing, such as the "-race" one.

Please, can you correct this behavior @chilcano ?

altergui:
AFAIU, we do want to `go test -race` on every push.
the "non -race" command was split by @mvdan in e10f01cdc7d18e75f25b275758f79e03e20aeb90 to just generate coverprofile (which we're not using right now)
i think it's pointless to run `go test` and then `go test -race` if we're not using coverprofile, and thus i believe it's reasonable to have job_go_test just run `go test -race` and job_go_checks run code checks `go mod tidy; gofmt; staticcheck; etc`

if we ever want coverprofile that could be actually a separate job 
```
job_go_coverprofile:
       - name: Run go test cover
         run: |
           # check that log lines contains no invalid chars (evidence of format mismatch)
           export LOG_PANIC_ON_INVALIDCHARS=true
           # we run vet in another step
           go test -vet=off -timeout=1m -coverprofile=covprofile ./...
```

chilcano:
Hi @p4u and @altergui 

> i just realized that job_go_tests doesn't really run `go test` anymore   

1. This PR includes `go test -vet=off -timeout=15m -race ./...` in `job_go_tests_extra` and it is already executed on every `push` or `pr` event. 
2. Should `job_go_tests_extra` be renamed? and Should `job_go_tests_extra` include a new step? which one?
3. Should `job_go_tests` be renamed? and Should `job_go_tests_extra` include a new step? which one?

I only need 5 minutes, could any of you both join me tomorrow to confirm these changes?

altergui:
my vote is:
* rename job_go_tests -> job_go_checks
* rename job_go_tests_extra -> job_go_test (no plural)

no new steps in any of them

we might be bikeshedding at this point, but.... @p4u WDYT?

p4u:
Ouch, I did explain myself very bad, sorry.

So, this is how I'd split the jobs:

+ job_go_test: runs "go test" WITHOUT race => executed on every push, but NOT if job_go_test_race is executed

+ job_go_checks: runs linters, staticcheck, vet, etc. => executed on every push, always.

+ job_go_test_race: runs "go test -race"  => executed only on commits to master, release-* and the rest of "special" branches.

+ job_docker_compose => executed only on PR, commits to master and the rest of "special" branches

Regarding coverprofile I think we might skip it.

p4u:
Sorry for the mess @chilcano and thanks for your patience

chilcano:
Hi @p4u 
Just updated the main workflow with your recommendations.

> * job_go_test: runs "go test" WITHOUT race => executed on every push, but NOT if job_go_test_race is executed.
> * job_go_test_race: runs "go test -race" => executed only on commits to master, release-* and the rest of "special" branches.

1. The `job_go_test` now contains 2 main steps (`Run Go test` and `Run Go test race`) which are executed on `push` event only. The `Run Go test race` step is executed if there is a `push` event on a special branch, and if it is not a special branch then `Run Go test` step is executed.
2. The `Run Go test` step execute `go test ./...`. Is that right?
3. I'm considering `commit` as `push` event.

> * job_go_checks: runs linters, staticcheck, vet, etc. => executed on every push, always.

Done.

> * job_docker_compose => executed only on PR, commits to master and the rest of "special" branches

Done.

Other thing to consider is that `job_docker_release` depends on `needs: [job_go_test, job_compose_test]`, in other words that job will be executed after `job_go_test` and `job_compose_test`.

Hope this helps.
Anything, just let me know.

p4u (approved):
Please squash the commits when merging it to master (github allows to do it)

PR #695, opened 2022-11-22
download account metadata and small vochain refactor

p4u:
The offchain data downloader is now subscribed to setAccount events in order to download the metadata (if exist) of an account.

Small refactor on vochian package.
- Prefix with "state_" the files that contain state related methods.
- Split in more files
- Move some functions to transactions.go (those that perform tx checks)


cmd/cli/main.go
```diff
@@ -433,17 +433,30 @@ func hostHandler(cli *vocdoniCLI) error {
 }
 
 func accountSetMetadata(cli *vocdoniCLI) error {
+	currentAccount, err := cli.api.Account("")
+	if err != nil {
+		return err
+	}
+	if currentAccount == nil {
+		return fmt.Errorf("account does not exist yet")
```
jordipainan, line 441: Can we use the `ErrAccountNotExists` in order to have unified error messages ?

cmd/cli/main.go
```diff
@@ -433,17 +433,30 @@ func hostHandler(cli *vocdoniCLI) error {
...
+	}
+
 	accMeta := api.AccountMetadata{
 		Version:     "1.0",
 		Name:        map[string]string{"default": "vocdoni account " + cli.getCurrentAccount().Address.Hex()},
 		Description: map[string]string{"default": "my account description"},
-		Media: api.AccountMedia{
+		Media: &api.AccountMedia{
 			Logo:   "https://upload.wikimedia.org/wikipedia/commons/f/f6/HAL9000.svg",
```
jordipainan, line 449: Do we want to have this default values ? If so, use a const here please.

cmd/cli/main.go
```diff
@@ -433,17 +433,30 @@ func hostHandler(cli *vocdoniCLI) error {
...
 			Avatar: "https://upload.wikimedia.org/wikipedia/commons/f/f6/HAL9000.svg",
 			Header: "https://images.unsplash.com/photo-1543000968-1fe3fd3b714e",
 		},
 	}
 
+	if currentAccount.Metadata != nil {
+		log.Infof("account has metadata (%s) let's update it", currentAccount.InfoURL)
```
jordipainan, line 456: Use a better message `let's update it` is too informal, also I think this should be a `Debugf`

vochain/keykeeper/keykeeper.go
```diff
@@ -507,3 +481,27 @@ func (k *KeyKeeper) signAndSendTx(tx *models.AdminTx) error {
...
 }
+
+// OnProcessKeys does nothing
+func (k *KeyKeeper) OnProcessKeys(pid []byte, pub string, txindex int32) {
+	// do nothing
+}
+
+// OnRevealKeys does nothing
+func (k *KeyKeeper) OnRevealKeys(pid []byte, priv string, txindex int32) {
+	// do nothing
```
jordipainan: Delete this message and unify with other methods.

`func (k *KeyKeeper) OnRevealKeys(pid []byte, priv string, txindex int32) {}`

vochain/offchaindatahandler/offchaindatahandler.go
```diff
@@ -140,6 +140,22 @@ func (c *OffChainDataHandler) OnProcessesStart(pids [][]byte) {
 	}
 }
 
+// OnSetAccount is triggered when a new account is created or modifyied. If metadata info is present, it is enqueued.
+func (c *OffChainDataHandler) OnSetAccount(addr []byte, account *vochain.Account) {
+	c.queueLock.Lock()
+	defer c.queueLock.Unlock()
+	if !c.importOnlyNew || !c.isFastSync {
```
jordipainan, line 147: What if the tx is `AddDelegateForAccount` ? Will this be executed with the same metadata and will enqueue the same InfoURI ?

vochain/offchaindatahandler/offchaindatahandler.go
```diff
@@ -149,13 +165,7 @@ func (c *OffChainDataHandler) OnRevealKeys(pid []byte, priv string, txindex int3
 func (c *OffChainDataHandler) OnProcessStatusChange(pid []byte,
 	status models.ProcessStatus, txindex int32) {
 }
-func (c *OffChainDataHandler) OnSetAccount(addr []byte, account *vochain.Account) error {
-	return nil
-}
-func (c *OffChainDataHandler) OnTransferTokens(from, to []byte, amount uint64) error {
-	return nil
-}
+func (c *OffChainDataHandler) OnTransferTokens(from, to []byte, amount uint64) {}
```
jordipainan, line 167: I remember to discuss this months ago and the idea is to introduce error in all the callbacks to handle possible failures. That is why I introduced those `error` in the functions and the idea was to modify the others at some point for having all of them returning error.
Let me know if you think we must never return `error` so I wont add them in the future.

vochain/state_trees.go
```diff
@@ -161,3 +198,116 @@ func StateParentChildTreeCfg(parent, child string, key []byte) (statedb.TreeConf
 	childTree := StateChildTreeCfg(child)
 	return parentTree, childTree.WithKey(key)
 }
+
+// rootLeafGetRoot is the GetRootFn function for a leaf that is the root
+// itself.
+func rootLeafGetRoot(value []byte) ([]byte, error) {
```
jordipainan, line 204: In state_votes you made this functions methods of the state  (i.e `func (v *State) voteCountInc()`) but here are not State methods. Do this follow any specific logic?

Same for the functions below.
Also I would like to have `func (s *State)` rather than `func (v *State)`, looks nicer :)

jordipainan (changes requested):
Added some comments :)

cmd/cli/main.go
p4u: Umm, this is a main.go program, its not importable by other packages. So the errors are just informative for the end user (and printed in the screen).

vochain/offchaindatahandler/offchaindatahandler.go
p4u: Yes, I think these event callbacks should not return errors (maybe only Commit). But I'm open to discuss it.

That's because the State eventListener is not gonna stop processing things because an arbitrary handler returns an error, as most it would print it, but then this can be probably better handled by the handler itself. 

So I think its more explicit for the handler developer to not include errors. So the developer knows that returning an error won't have any real effect.

vochain/state_trees.go
p4u: In general methods are used when the logic requires access to the struct object state. Functions are used when the logic can be run standalone and no state is required. I don't think abusing from methods is a good idiomatic strategy.

What I tried to do on this minor refactor is to group the methods and functions by functionality. So in this case, everything related to "state merkle trees" is moved to "state_trees.go".

vochain/offchaindatahandler/offchaindatahandler.go
p4u: If the file is already downloaded it will do nothing. However you are right, it would be more convenient to check the kind of transaction before adding the URI to the queue. But the event handler does not provide this information, so it should be added.

vochain/state_trees.go
p4u: Regarding the renaming of s/v I agree but this is not related to this commit (you can field a new issue if you want).

In any case a major refactor on naming and structure must be perform in some moment (such as move all the State methods to a separated package).

vochain/offchaindatahandler/offchaindatahandler.go
p4u: A new issue should be created to fix this, first on the Event handler layer.

jordipainan: Task created

PR #697, opened 2022-11-22
Removed pull_request_target event from GH workflow

chilcano:


p4u:
@mvdan can you take a look? This PR solves the double check execution.

@chilcano there is still an issue to fix, which is this "job_go_tests" which is never executed. Any ideas what's going on?

chilcano:
> @chilcano there is still an issue to fix, which is this "job_go_tests" which is never executed. Any ideas what's going on?

Do you mean why we have `job_go_tests` as an _`Expected — Waiting for status to be reported `_ in the PR review page.
This is because this repo has a legacy configuration. The Github is expecting to run a job (`job_go_tests`) which doesn't exist in the current workflow. In the workflow we have `job_go_test`(without char `s`).
To be clear, the current job `job_go_test` implemented in the workflow is always executed in `push` and `pull_request` events, that means that removing or renaming the Branch Protection Rules in the repo will remove the message int he PR review page.
I'll update this.
image

chilcano:
Done!
This message  **`job_go_tests` as an _`Expected — Waiting for status to be reported `_** in the PR review page will never appear again.

mvdan (approved):

PR #700, opened 2022-11-23
Dvot 594 token transfer indexer

jordipainan:
Do you approve this approach? What do you think about the endpoint? Any suggestion is very welcome.

This is a draft PR so do not review the code extensively, just the approach .. (pagination, endpoint, db schema, api call)


api/accounts.go
```diff
@@ -80,6 +81,14 @@ func (a *API) enableAccountHandlers() error {
 	); err != nil {
 		return err
 	}
+	if err := a.endpoint.RegisterMethod(
+		"/accounts/{organizationID}/token/transfers/page/{page}",
```
p4u: LGTM but I'd avoid the `/token/` word.

api/accounts.go
```diff
@@ -304,3 +313,30 @@ func (a *API) electionCountHandler(msg *bearerstdapi.BearerStandardAPIdata, ctx
...
 }
+
+// /accounts/<organizationID>/token/transfers/page/<page>
+// Returns the token transfers for an organization
+func (a *API) tokenTransfersHandler(msg *bearerstdapi.BearerStandardAPIdata, ctx *httprouter.HTTPContext) error {
+	organizationID, err := hex.DecodeString(util.TrimHex(ctx.URLParam("organizationID")))
+	if err != nil || organizationID == nil {
+		return fmt.Errorf("organizationID (%q) cannot be decoded", ctx.URLParam("organizationID"))
+	}
+	page, err := strconv.Atoi(ctx.URLParam("page"))
```
p4u: I'd check if the account actually exists (by querying the state) before making the scrutinizer "expensive" call.

types/types.go
```diff
@@ -17,3 +17,7 @@ type CensusRoot = []byte
 type EncodedProtoBuf = []byte
 
 type Nullifier = []byte
+
+type FromAccount = []byte
```
p4u: Why this type redefinition? Is this a requirement for sqlite?

vochain/scrutinizer/indexertypes/types.go
```diff
@@ -325,6 +325,14 @@ func (b *BlockMetadata) String() string {
 	return builder.String()
 }
 
+type TokenTransfer struct {
+	ID        int64          `json:"id"`
+	From      types.HexBytes `json:"from,omitempty"`
+	To        types.HexBytes `json:"to"`
+	Amount    uint64         `json:"amount"`
+	Timestamp time.Time      `json:"timestamp"`
```
p4u: I think we might add "height", its cheap and can be useful for finding the transaction.

p4u (changes requested):
Yes, the approach looks good! Just some minor comments.

I'd like @mvdan to review too since this is modifying SQLite stuff.

api/accounts.go
jordipainan: oki i've added it because transfers can lead someone to think "transfer of what?" but since the only thing we can transfer are balances probably is fine to get rid of `token`

types/types.go
jordipainan: Not a strong requirement but following @mvdan approach

api/accounts.go
jordipainan: oki, will do! thank you

vochain/scrutinizer/indexertypes/types.go
jordipainan: yup, good one, so we can have the block number

jordipainan:
thanks for the insights @p4u

types/types.go
p4u: I'd avoid this if its not required.

mvdan: this is currently required because sqlc will otherwise generate `string` as a type for the sql `BLOB`, which will cause problems - strings in SQL are null-terminated, for example, so then any query containing a null byte in a blob will be truncated.

mvdan: at the same time, assuming FromAccount and ToAccount are both the same type, perhaps use `type AccountID = []byte`, then we save on the repetition.

mvdan (approved):
sql bits SGTM. I assume you're not adding code that uses badgerhold on purpose :)

p4u:
@jordipainan what is the status of this PR? I think we should prioritizie it.

jordipainan:
> @jordipainan what is the status of this PR? I think we should prioritizie it.

on it, added further adjustments based on the latest code

jordipainan:
Summary:

- `api`
    - new api endpoint -> `GET`: `/accounts/{organizationID}/transfers/page/{page}`
    - implemented api endpoint handler `tokenTransfersByFromAccountHandler`
    - new api type `TokenTransfer`
- `apiClient`
    - new method `TokenTransfersByFromAccount`
- `types`
    - create `AccountID` type as per @mvdan suggestion
- `vochain/indexer`
   - sql -> create token_transfers table, migrations and autogenerated code
   - types -> create `TokenTransfer` type
   - core -> add tokenTransfers pool and implement `OnTransferTokens` event

- `vochain/core`

I've made some changes on the event for token transfers, these changes implied to change also some code for the account `TransferBalance`. The rationale for the changes is the following:
Currently we have an event `OnTransferTokens(from, to, amount)` which are information that we can access from the transaction itself, the problem is that if we want to store other useful information we need or to pass the entire transaction or too many params that go agains code readability. Examples are the `txhash` or the `height`.
Notice this is done in a similar way on the event (`onNewTx`).
Also `OnTransferTokens` was executed inside the `account.TransferBalance` method and there was no way to access information of the transaction or the current height of the blokchain.
So I've decided to change what information we send to the `TransferBalance` method from (from, to, amount) to (from, to, amount, txHash, height) wrapped in a struct called `TransferTokensMeta`. That way we can pass useful information for all methods that imply token transfers: `CreateAccount` with a faucet payload, `mint` tokens, `claim` a faucet payload or just a regular `sendTokens`.
We can now index any transfer of tokens to the indexer with all of its information.

- Ran `gofmt` to several files.

jordipainan:
@p4u ready for review too :)

api/accounts.go
```diff
@@ -89,6 +90,14 @@ func (a *API) enableAccountHandlers() error {
 	); err != nil {
 		return err
 	}
+	if err := a.endpoint.RegisterMethod(
```
p4u, line 93: Please add a second endpoint to `/accounts/{accountID}/transfers` that will show the first page of transfers. You can use the same handler function.

api/accounts.go
```diff
@@ -333,3 +342,37 @@ func (a *API) electionCountHandler(msg *apirest.APIdata, ctx *httprouter.HTTPCon
 	}
 	return ctx.Send(data, apirest.HTTPstatusCodeOK)
 }
+
+// /accounts/<organizationID>/transfers/page/<page>
+// Returns the token transfers for an organization
+func (a *API) tokenTransfersHandler(msg *apirest.APIdata, ctx *httprouter.HTTPContext) error {
+	organizationID, err := hex.DecodeString(util.TrimHex(ctx.URLParam("organizationID")))
```
p4u: lets call it "accountID", its more accurated.

api/accounts.go
```diff
@@ -333,3 +342,37 @@ func (a *API) electionCountHandler(msg *apirest.APIdata, ctx *httprouter.HTTPCon
...
+	if err != nil || organizationID == nil {
+		return fmt.Errorf("organizationID (%q) cannot be decoded", ctx.URLParam("organizationID"))
+	}
+	acc, err := a.vocapp.State.GetAccount(common.BytesToAddress(organizationID), true)
+	if acc == nil {
+		return fmt.Errorf("organization not found")
```
p4u: account not found

api/api_types.go
```diff
@@ -242,3 +242,12 @@ func CensusTypeToOrigin(ctype CensusTypeDescription) (models.CensusOrigin, []byt
 	}
 	return origin, root, nil
 }
+
+type TokenTransfer struct {
```
p4u: This type is declared twice. I think there is no need to declare it here (in the API package).

vochain/indexer/indexertypes/types.go
```diff
@@ -336,6 +336,15 @@ func (b *BlockMetadata) String() string {
 	return builder.String()
 }
 
+type TokenTransfer struct {
```
p4u: Document the exported type properly please.

vochain/state/eventlistener.go
```diff
@@ -28,7 +28,7 @@ type EventListener interface {
 	OnProcessResults(pid []byte, results *models.ProcessResult, txIndex int32)
 	OnProcessesStart(pids [][]byte)
 	OnSetAccount(addr []byte, account *Account)
-	OnTransferTokens(from, to []byte, amount uint64)
+	OnTransferTokens(tx *vochaintx.TransferTokensMeta)
```
p4u: Why `meta`? Looks like the main type is defined in the indexer and this is a subset. That does not look very correct according to the actual hierarchy.

Here an idea: Consider calling the vochaintx type just "TokenTransfer". Then the indexer type could be named something like "TokenTranferMeta" and would heritage from this type (and add the extra fields required by the indexer).

```
type TokenTransferMeta struct {
   vochaintx.TokenTransfer
   extraFields...
}
 ```

File: vochain/transaction/transaction.go [hunk 11]
```diff
@@ -249,12 +249,17 @@ func (t *TransactionHandler) CheckTx(vtx *vochaintx.VochainTx, forCommit bool) (
 					); err != nil {
 						return nil, fmt.Errorf("setAccountTx: consumeFaucet %w", err)
 					}
+					if err := t.state.TransferBalance(&vochaintx.TransferTokensMeta{
+						FromAddress: faucetIssuerAddress,
+						ToAddress:   txSenderAddress,
+						Amount:      faucetPayload.Amount,
+						TxHash:      vtx.TxID[:],
+						Height:      uint64(t.state.CurrentHeight()),
```
p4u: Why do we need the Height here and the transaction hash?

vochain/transaction/vochaintx/vochaintx.go
```diff
@@ -43,3 +44,12 @@ func (tx *VochainTx) Unmarshal(content []byte, chainID string) error {
 func TxKey(tx tmtypes.Tx) [32]byte {
 	return sha256.Sum256(tx)
 }
+
+// TransferTokensMeta wraps useful information about a token transfer
```
p4u: Please, make the documentation of the exported types more extense (Copilot can suggest much more) and add dots to the end of the statements.

p4u (changes requested):

jordipainan:
ack, im going to tackle the comments asap

vochain/state/eventlistener.go
jordipainan: I've renamed the types, but I did not use inheritance given that the inner types are different for convenience (i.e Address as common.Address vs Address as types.HexBytes)

vochain/transaction/transaction.go
jordipainan: For having this information available on the state that is later consumed by the indexer

vochain/transaction/transaction.go
```diff
@@ -360,7 +375,16 @@ func (t *TransactionHandler) CheckTx(vtx *vochaintx.VochainTx, forCommit bool) (
 			if err != nil {
 				return nil, fmt.Errorf("sendTokensTx: burnTxCostIncrementNonce %w", err)
 			}
-			return response, t.state.TransferBalance(from, to, tx.Value)
+			if err := t.state.TransferBalance(&vochaintx.TokenTransfer{
+				FromAddress: from,
+				ToAddress:   to,
+				Amount:      tx.Value,
+				TxHash:      vtx.TxID[:],
```
mvdan, line 380: assuming TxID is also a []byte, these `[:]` expressions are redundant

vochain/state/balances.go
```diff
@@ -221,43 +222,82 @@ func (v *State) ConsumeFaucetPayload(from common.Address, faucetPayload *models.
...
 	}
-	acc, err := v.GetAccount(address, false)
+	acc, err := v.GetAccount(tx.ToAddress, false)
+	if err != nil {
+		return fmt.Errorf("mintBalance: %w", err)
+	}
+	if acc == nil {
+		return ErrAccountNotExist
+	}
+	if acc.Balance+tx.Amount <= acc.Balance {
```
mvdan: nit: you want `<` to actually catch an overflow, otherwise this will also trigger if `tx.Amount == 0`.

mvdan: (I think the existing code also has this potential flaw)

vochain/indexer/migrations/0005_create_table_token_transfers.sql
```diff
@@ -0,0 +1,17 @@
...
+  transfer_time DATETIME NOT NULL
+);
+
+CREATE INDEX index_from_account_token_transfers
+ON token_transfers(from_account);
+
+-- +goose Down
+DROP TABLE token_transfers
+
+DROP INDEX index_from_account_token_transfers
```
mvdan: nit: include trailing newlines. perhaps configure your editor for that?

mvdan (approved):
sqlite bits SGTM

vochain/transaction/transaction.go
jordipainan: is a [32]byte

mvdan: ack :)

vochain/state/balances.go
jordipainan: At this point, we've already checked that `tx.Amount` is not 0, so I cannot see when this case can happen.

vochain/indexer/migrations/0005_create_table_token_transfers.sql
jordipainan: I've just missed it :S

vochain/state/balances.go
mvdan: just a matter of defensive programming :) even if we can be sure that today Amount==0 is impossible, it might be possible in the future after refactors or we might have got it wrong. up to you of course, I just prefer to write defensive code in general.

jordipainan: fair enough, added the suggested change :) Thank you dani!

vochain/indexer/indexer.go
```diff
@@ -490,6 +493,12 @@ func (idx *Indexer) Commit(height uint32) error {
 			}
 		}
 	}
+	// index token transfers
+	for _, tt := range idx.tokenTransferPool {
+		if err := idx.newTokenTransfer(tt); err != nil {
+			log.Errorf("commit: cannot create new token transfer: %v", err)
```
p4u: Use the new `log.Errorw()` function for printing errors.

vochain/indexer/indexer.go
```diff
@@ -718,7 +727,63 @@ func (idx *Indexer) OnProcessesStart(pids [][]byte) {
...
+		TxHash:       tt.TxHash,
+		Height:       int64(tt.Height),
+		FromAccount:  tt.From,
+		ToAccount:    tt.To,
+		Amount:       int64(tt.Amount),
+		TransferTime: tt.Timestamp,
+	}); err != nil {
+		return err
+	}
+	log.Debugf("adding new token transfer from %x to %x: %d into the indexer database", tt.From, tt.To, tt.Amount)
```
p4u: `log.Debugw` will look nicer :) This is a clear example on where this function is useful.

`log.Debugw("new token transfer", map[string]interface{}{ "from": fmt.Sprintf("%x, tt.From"), "amount": tt.Amount, etc...)`

vochain/indexer/indexertypes/types.go
```diff
@@ -336,6 +336,17 @@ func (b *BlockMetadata) String() string {
 	return builder.String()
 }
 
+// TokenTransferMeta contains the information of a token transfer and some extra useful information.
+// The types are compatible with the SQL defined schema.
+type TokenTransferMeta struct {
+	Amount    uint64          `json:"amount"`
+	From      types.AccountID `json:"from"`
```
p4u, line 346: If this needs to be JSON encoded I think we should add a Method on types.AccountID to Marshal the []byte address correctly (in hex format). Else declare it as `type.HexBytes`

vochain/state/balances.go
```diff
@@ -221,62 +222,105 @@ func (v *State) ConsumeFaucetPayload(from common.Address, faucetPayload *models.
...
 	if accTo == nil {
 		return ErrAccountNotExist
 	}
-	if err := accFrom.Transfer(accTo, amount); err != nil {
+	if err := accFrom.Transfer(accTo, tx.Amount); err != nil {
 		return err
 	}
-	log.Debugf("transferring %d tokens from %s to %s", amount, from.String(), to.String())
-	if err := v.SetAccount(from, accFrom); err != nil {
+	log.Debugf("transferring %d tokens from %s to %s", tx.Amount, tx.FromAddress.String(), tx.ToAddress.String())
```
p4u: `log.Debugw` :)

vochain/state/balances.go
```diff
@@ -221,62 +222,105 @@ func (v *State) ConsumeFaucetPayload(from common.Address, faucetPayload *models.
...
+		return fmt.Errorf("mintBalance: %w", err)
+	}
+	if acc == nil {
+		return ErrAccountNotExist
+	}
+	if acc.Balance+tx.Amount < acc.Balance {
+		return ErrBalanceOverflow
+	}
+	acc.Balance += tx.Amount
+	log.Debugf("minting %d tokens to account %s", tx.Amount, tx.ToAddress.String())
```
p4u, line 287: `log.Debugw`

vochain/transaction/transaction.go
p4u: But the indexer already have access to the blockchain Height, I don't see the need for adding it here. Is there another reason behind this?

p4u (approved):
Just minor things, approving.

jordipainan:
@p4u  addressed all comments.  Changed AccountID from []byte to HexBytes and get rid of storing height on the state as the indexer have access to it. Thank you for the insights.

jordipainan:
Adapted to #761

PR #702, opened 2022-11-24
vochain: rename scrutinizer to indexer

jordipainan:
Scrutinizer was the naming used when the function
of this package was to just compute results, this was a long ago and the package now acts as an indexer for state related data in order to access relevant information without needing to make expesive queries to the state. For this reason we decided to rename the scrutinizer package to indexer.

This commit changes all references to the scrutinizer in favor of indexer.


mvdan (approved):

altergui (commented):
the name change SGTM. didn't read the changes in detail nor tested the code

PR #703, opened 2022-11-24
remove old unused packages

jordipainan:
With the last commits, some packages became unnecessary, delete references to them and update go.mod/go.sum


mvdan:
This overlaps with https://github.com/vocdoni/vocdoni-node/pull/702. Is that intentional? Perhaps because you want to change this separately?

mvdan (approved):

jordipainan:
> This overlaps with #702. Is that intentional? Perhaps because you want to change this separately?

yup, gonna merge this first and then rebase 702

PR #706, opened 2022-11-25
apiv2: add some more endpoints

p4u:
- /chain/validators
- /chain/blocks/<height>
- /chain/blocks/hash/<hash>
- /accounts/treasurer

### DIFF
--- a/api/accounts.go
+++ b/api/accounts.go
@@ -14,6 +14,7 @@ import (
 	"go.vocdoni.io/dvote/httprouter"
 	"go.vocdoni.io/dvote/httprouter/bearerstdapi"
 	"go.vocdoni.io/dvote/log"
+	"go.vocdoni.io/dvote/types"
 	"go.vocdoni.io/dvote/util"
 	"go.vocdoni.io/proto/build/go/models"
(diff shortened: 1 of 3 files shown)


api/accounts.go
```diff
@@ -214,6 +223,23 @@ func (a *API) accountSetHandler(msg *bearerstdapi.BearerStandardAPIdata, ctx *ht
 	return ctx.Send(data, bearerstdapi.HTTPstatusCodeOK)
 }
 
+// GET /accounts/treasurer
+// get the treasurer address
+func (a *API) treasurerHandler(msg *bearerstdapi.BearerStandardAPIdata, ctx *httprouter.HTTPContext) error {
+	acc, err := a.vocapp.State.Treasurer(true)
+	if err != nil || acc == nil {
```
jordipainan: If `err != nil` does not mean that the treasurer does not exist but that there was some problem on the backend fetching it. Can we return a different error ? or is this done on purpose ?

p4u: right

jordipainan (approved):

Issue #687, opened 2022-11-17
DVOT-605: Vochain allows to create a process in a UNKNOWN state


jpaulet:
A process can only be created in READY or PAUSED status.

altergui:
i proposed PR #719

altergui:
fixed in #719

PR #719, opened 2022-11-29
vochain: AddProcess now returns error if process status is UNKNOWN

altergui:


altergui:
fixes #687

altergui:
actually i missed the comment about "A process can only be created in READY or PAUSED status."
i'll redo the PR, but first i'd love to have feedback if this is the right place to implement such checks

p4u (approved):
I think this change is state breaking, since some processes might be already created at UNKNOWN state. In this case you should upgrade the Development Genesis file to make the testnet blockchain auto-restart.

p4u:
The check here looks good. But this will only be triggered when Commit()  It would be better to discard the transaction already on mempool's CheckTx(), that's in the transactions.go file NewProcessTxCheck() if I recall correctly.

p4u (changes requested):
Let's apply this changes before approving.

altergui:
perfect feedback @p4u :heart: 

rewritten from scratch, now check happens where it should (CheckTx) and fails on any state except READY or PAUSED
also dev genesis version is bumped

p4u (approved):

jordipainan (approved):

PR #720, opened 2022-11-29
vochain/indexer: implement more sqlite queries

mvdan:
(see commit message)


mvdan:
I think you're all busy this evening, so I took advantage and pushed a couple more commits. Six sqlite TODOs removed in total :)

p4u:
There is a minor issue with VoterID{} empty value spotted by the tests @mvdan

p4u (approved):

mvdan:
You're right, I was testing with badgerhold disabled only :) Fixed now.

jordipainan (approved):

PR #721, opened 2022-11-30
census: add zkweighted census type

p4u:
see commit message


jordipainan:
Hi! Sorry if not my business but I was reading the code just for curiosity.
I assume this is commit adds partial support for weighted census right ? Saying this because in transaction.go#L544 the weight is always 1, I did not review if this happens on other places :)

lucasmenendez (approved):
According to my current knowledge of the project its perfect, all the steps as we have discussed.

PR #725, opened 2022-12-06
mod: update a couple of deps

mvdan:
(see commit message)


mvdan:
Hm, is compose-test broken in master?

p4u:
Yes, probably it is.

PR #727, opened 2022-12-07
apiv2: convert tendermint block headers to camelCase

p4u:
So we unify to camelCase all APIv2 replies.


p4u:
Is this too dirty @mvdan? That's the best way I found to fix the snake/camel case issue with the tendermint blocks definition.

mvdan:
Can you show me what the original JSON looks like, and what you're converting it to?

Your implementation looks like it would work, although it's very inefficient if the values are nested. For example, three levels of objects means that you decode and encode the innermost object three times, because the top call does the entire value, and recuses into each sub-value which then does the same.

One easy improvement would be to decode the entire thing into `any`, then have a recursive function which type switches over it to see if it's `map[string]any` (object, so rewrite the keys and recurse into the values), `[]any` (array, recurse), or anything else (do nothing). You at least avoid the duplication of work mentioned above.

It would be even faster and easier to do this at the token level, but encoding/json only has a token decoder, not a token encoder.

mvdan:
Also, I see that you're using json.RawMessage, so I imagine your intention was to do lazy decoding at each level. It doesn't actually work very well, because the json decoder still has to tokenize the whole value to "decode" into a RawMessage. So you're still tokenizing nested values multiple times in a quadratic way.

p4u:
Wow, this is a great input for improving the function. Would you make a proposal by adding a new commit? It looks like it's clear for you, I'd need more time to get to implement properly your comments. You can just write the code and I will test it.

The original JSON looks like this:

```json
{
  "header": {
    "version": {
      "block": 11,
      "app": 0
    },
    "chain_id": "vocdoni-stage-24",
    "height": 123,
    "time": "2022-12-14T16:40:47.963888306Z",
    "last_block_id": {
      "hash": "74930754291C5739B01F38B80B5C6E9618C0885816A18115E98221DF9EF6A27F",
      "parts": {
        "total": 1,
        "hash": "1B43BBD217D9A8AE723093265C3198A894A727E7C5ED25662B20DC36E0B2514A"
      }
    },
    "last_commit_hash": "D674CD12566D765975DC9657CE374450C2ED87F63E7912E9DB1CB1DAAD0EAACA",
    "data_hash": "E3B0C44298FC1C149AFBF4C8996FB92427AE41E4649B934CA495991B7852B855",
    "validators_hash": "37CCC8CBF23951172EB530ACAC20B2756C8D7E3AA817A3E7ABC8B9424D100B40",
    "next_validators_hash": "37CCC8CBF23951172EB530ACAC20B2756C8D7E3AA817A3E7ABC8B9424D100B40",
    "consensus_hash": "4B289DA08458EE8CEA88609F8CFB704E69D78AF70D6127B4AD94D09FFCE897A2",
    "app_hash": "5308EAD6BF0DF7481B38141D1046BE0E8E1C871D8B2CA5271443EFD030CB64AE",
    "last_results_hash": "E3B0C44298FC1C149AFBF4C8996FB92427AE41E4649B934CA495991B7852B855",
    "evidence_hash": "E3B0C44298FC1C149AFBF4C8996FB92427AE41E4649B934CA495991B7852B855",
    "proposer_address": "92C9A63172DFB4E9637309DFBFE20B1D11EDC4E7"
  },
  "data": {
    "txs": []
  },
  "evidence": {
    "evidence": []
  },
  "last_commit": {
    "height": 122,
    "round": 0,
    "block_id": {
      "hash": "74930754291C5739B01F38B80B5C6E9618C0885816A18115E98221DF9EF6A27F",
      "parts": {
        "total": 1,
        "hash": "1B43BBD217D9A8AE723093265C3198A894A727E7C5ED25662B20DC36E0B2514A"
      }
    },
    "signatures": [
      {
        "block_id_flag": 2,
        "validator_address": "2DECD25EBDD6E3FAB2F06AC0EE391C16C292DBAD",
        "timestamp": "2022-12-14T16:40:47.963888306Z",
        "signature": "ayj6CnGcD1zImfCiSIHNXaAujx4uxZdQ/NgWU/yxQJz/GJAMGIyD3C704cGtx3e2zJaXYDtZvwj5C+/Q/v2FBw=="
      },
      {
        "block_id_flag": 2,
        "validator_address": "3C6FF3D424901733818B954AA3AB3BC2E3695332",
        "timestamp": "2022-12-14T16:40:47.972896402Z",
        "signature": "QVZjYpjyVNjbLWaP552MyU3pRZ4Lw8FF98tsGKO7DAUN/QanEf6QJwK7maCesvgfeISG34tYWDKL/p+fUjYFAQ=="
      },
      {
        "block_id_flag": 2,
        "validator_address": "92C9A63172DFB4E9637309DFBFE20B1D11EDC4E7",
        "timestamp": "2022-12-14T16:40:47.967578082Z",
        "signature": "BKCU3QXdEAJP9oc/PXe8suPQupHXJtbQxT+axlxT9jWITLsGTnDFe/rQIOiiQhKh+l5nDq/ZwMuGs4Gn/3AyBw=="
      },
      {
        "block_id_flag": 2,
        "validator_address": "B04F5541E932BB754B566969A3CD1F8E4193EFE8",
        "timestamp": "2022-12-14T16:40:47.961444995Z",
        "signature": "FhlEUaeMACn8pg1WzkzIgaOQAsu9XIXwxxl7mRjphX8uoY2xD+i44WgVeYfL9DDCVb9lIFCBhJ2+DHxzKqygAA=="
      }
    ]
  }
}
```

mvdan:
Here: https://github.com/vocdoni/vocdoni-node/pull/753

PR #729, opened 2022-12-13
RFC: allow Nonce=0

altergui:
after thinking about the "mempool ordering" thing for the Nth time, i still don't see how setting a Priority field would help.

in fact, i wonder what's the current issue or problem. is it only race conditions during tests?

because in that case, AFAIU the main problem is getting the nonce right. for that matter, i thought why not let the node choose the nonce? so here's a dirty POC for RFC and with that feedback prepare a "real" PR or discard the idea altogether and go back to brainstorming

### DIFF
--- a/cmd/vochaintest/vochaintest.go
+++ b/cmd/vochaintest/vochaintest.go
@@ -283,12 +283,12 @@ func (c *testClient) ensureTxCostEquals(signer *ethereum.SignKeys, txType models
 			return fmt.Errorf("cannot get treasurer: %w", err)
 		}
 
-		log.Infof("will set %s txcost=%d (treasurer nonce: %d)", txType, cost, treasurerAccount.Nonce)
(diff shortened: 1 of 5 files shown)


p4u:
The Nonce is necessary for:

1. letting the sender choose the order of transactions, since this could be relevant 
2. avoid replay attacks (a transaction is executed twice)

So, the Nonce must be part of the Transaction payload and signed by the sender in order to achieve these properties. 

Tendermint 0.35 in theory allows transaction prioritization. So the transactions, once in the mempool, can be ordered by nonce. The one with lowest nonce will be executed before.

altergui:
perfect feedback @p4u, the replay attack prevention thing was not clear to me and now i see this approach is a no-go, so i'm discarding it. i'm happy i posted the WIP early and didn't develop it further, agile FTW, failing ASAP :)

PR #730, opened 2022-12-13
refactor configuration and services structure

p4u:
- remove old flags and config options
- remove all related RPC/API options
- add two new simple falgs: --enableAPI --enableRPC
- small nit fixes and new comments
- rename all env var files
- add a function to build a genesis file and all the node keys `--vochainCreateGenesis=outputDir:8` will create a new genesis with 8 validators in directory `outputDir`


service/indexer.go
```diff
@@ -0,0 +1,24 @@
...
+		filepath.Join(vs.Config.DataDir, "indexer"),
+		vs.App,
+		!vs.Config.Indexer.IgnoreLiveResults,
+	); err != nil {
+		return err
+	}
+	// launch the indexer after sync routine (executed when the blockchain is ready)
+	go vs.Indexer.AfterSyncBootstrap()
+	return nil
+}
```
altergui: given the multiline, i feel it would be more readable (idiomatic even?) to just do 
```
vs.Indexer, err = ...(multiline)...
if err != nil {
  return err
}
```

altergui (commented):
reviewing (partial)

vochain/genesis.go
```diff
@@ -15,6 +24,13 @@ type VochainGenesis struct {
 	Genesis           string
 }
 
+type AppState struct {
+	Oracles []string `json:"oracles"`
+	//  Accounts  [] `json:"accounts"`
+	// HO TENIM A ALGUN LLOC DEFINIT?
```
altergui: :rofl:

vocone/vocone.go
```diff
@@ -126,10 +126,11 @@ func NewVocone(dataDir string, keymanager *ethereum.SignKeys, disableIpfs bool)
...
-	// Create the IPFS storage layer
-	if vc.storage, err = service.IPFS(&config.IPFSCfg{
-		ConfigPath: filepath.Join(dataDir, "ipfs"), NoInit: disableIpfs,
-	}, nil, nil); err != nil {
+	// Create the IPFS storage layer (we use the Vocdoni general service)
+	srv := service.VocdoniService{}
+	if vc.storage, err = srv.IPFS(&config.IPFSCfg{
+		ConfigPath: filepath.Join(dataDir, "ipfs"),
+	}); err != nil {
 		return nil, err
```
altergui, line 134: same here, separate the `if err != nil {` and keep the `vc.storage` assignment uncluttered

altergui (commented):
all in all, LGTM

i do believe the env vars prefix change will bring some surprises in current deployments (it happened to me already in the past, forgot to update the env file and didn't understand what was happening), something to keep in mind during upgrades.

altergui:
mmh do note this panic.

```
### Running test tokentransactions ###
Creating 2257917647_testsuite_test_run ... 
Creating 2257917647_testsuite_test_run ... done
2022-12-13T10:45:50Z	INFO	vochaintest/vochaintest.go:107	logger construction succeeded at level info with output stdout
2022-12-13T10:45:50Z	INFO	vochaintest/vochaintest.go:1146	connecting to main gateway http://gateway0:9090/dvote
2022-12-13T10:45:50Z	INFO	vochaintest/vochaintest.go:1185	treasurer fetched 0xfe10DAB06D636647f4E40dFd56599da9eF66Db1c with nonce 0
2022-12-13T10:45:50Z	INFO	vochaintest/vochaintest.go:1192	tx cost of SET_ACCOUNT_INFO_URI fetched successfully (10)
2022-12-13T10:45:50Z	INFO	vochaintest/vochaintest.go:286	will set SET_ACCOUNT_INFO_URI txcost=1000 (treasurer nonce: 0)
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x40 pc=0x62b7c2]

goroutine 1 [running]:
go.vocdoni.io/dvote/crypto/ethereum.(*SignKeys).SignVocdoniTx(0x40?, {0xc000138460?, 0xc000138458?, 0x1?}, {0xc000148090?, 0x0?})
	go.vocdoni.io/dvote/crypto/ethereum/ethereum.go:147 +0x22
go.vocdoni.io/dvote/rpcclient.(*Client).SubmitRawTx(0x1507a20?, 0xc000a435c0?, 0xc000525da0)
	go.vocdoni.io/dvote/rpcclient/api.go:1486 +0xc6
go.vocdoni.io/dvote/rpcclient.(*Client).SetTransactionCost(0x111c51f?, 0x2b?, 0x11, 0x3e8, 0x0)
	go.vocdoni.io/dvote/rpcclient/api.go:1396 +0x13d
main.(*testClient).ensureTxCostEquals(0xc000520218, 0x27?, 0x8e7bc8?, 0x3e8)
	go.vocdoni.io/dvote/cmd/vochaintest/vochaintest.go:287 +0x1a6
main.(*testClient).testSetTxCost(0xc000520218, 0x1a?)
	go.vocdoni.io/dvote/cmd/vochaintest/vochaintest.go:1195 +0x205
main.testTokenTransactions({0x7ffcc9053e9f, 0x1a}, {0x7ffcc9053ef7?, 0x6?}, 0x1116436?)
	go.vocdoni.io/dvote/cmd/vochaintest/vochaintest.go:1154 +0x225
main.main()
	go.vocdoni.io/dvote/cmd/vochaintest/vochaintest.go:174 +0xc05
2
```

p4u:
> mmh do note this panic.
> 
> ```
> ### Running test tokentransactions ###
> Creating 2257917647_testsuite_test_run ... 
> Creating 2257917647_testsuite_test_run ... done
> 2022-12-13T10:45:50Z	INFO	vochaintest/vochaintest.go:107	logger construction succeeded at level info with output stdout
> 2022-12-13T10:45:50Z	INFO	vochaintest/vochaintest.go:1146	connecting to main gateway http://gateway0:9090/dvote
> 2022-12-13T10:45:50Z	INFO	vochaintest/vochaintest.go:1185	treasurer fetched 0xfe10DAB06D636647f4E40dFd56599da9eF66Db1c with nonce 0
> 2022-12-13T10:45:50Z	INFO	vochaintest/vochaintest.go:1192	tx cost of SET_ACCOUNT_INFO_URI fetched successfully (10)
> 2022-12-13T10:45:50Z	INFO	vochaintest/vochaintest.go:286	will set SET_ACCOUNT_INFO_URI txcost=1000 (treasurer nonce: 0)
> panic: runtime error: invalid memory address or nil pointer dereference
> [signal SIGSEGV: segmentation violation code=0x1 addr=0x40 pc=0x62b7c2]
> 
> goroutine 1 [running]:
> go.vocdoni.io/dvote/crypto/ethereum.(*SignKeys).SignVocdoniTx(0x40?, {0xc000138460?, 0xc000138458?, 0x1?}, {0xc000148090?, 0x0?})
> 	go.vocdoni.io/dvote/crypto/ethereum/ethereum.go:147 +0x22
> go.vocdoni.io/dvote/rpcclient.(*Client).SubmitRawTx(0x1507a20?, 0xc000a435c0?, 0xc000525da0)
> 	go.vocdoni.io/dvote/rpcclient/api.go:1486 +0xc6
> go.vocdoni.io/dvote/rpcclient.(*Client).SetTransactionCost(0x111c51f?, 0x2b?, 0x11, 0x3e8, 0x0)
> 	go.vocdoni.io/dvote/rpcclient/api.go:1396 +0x13d
> main.(*testClient).ensureTxCostEquals(0xc000520218, 0x27?, 0x8e7bc8?, 0x3e8)
> 	go.vocdoni.io/dvote/cmd/vochaintest/vochaintest.go:287 +0x1a6
> main.(*testClient).testSetTxCost(0xc000520218, 0x1a?)
> 	go.vocdoni.io/dvote/cmd/vochaintest/vochaintest.go:1195 +0x205
> main.testTokenTransactions({0x7ffcc9053e9f, 0x1a}, {0x7ffcc9053ef7?, 0x6?}, 0x1116436?)
> 	go.vocdoni.io/dvote/cmd/vochaintest/vochaintest.go:1154 +0x225
> main.main()
> 	go.vocdoni.io/dvote/cmd/vochaintest/vochaintest.go:174 +0xc05
> 2
> ```

I don't understand this panic. But looks like a race (the node is not yet ready).  However the tests are passing...

service/indexer.go
p4u: Thanks for the observation, but I'm not certain why it would be more idiomatic.

We usually try to use inline `if` when possible. And we also try to declare the vars in the most reduced scope.

These are the two rules that we are usually following. They are properly applied here.

altergui:
my understanding is that this change is causing the panic. Why, i don't know. i can't understand it. but it's the only change in your commit that is touching this area of code.
image

https://github.com/vocdoni/vocdoni-node/pull/730/commits/e8178150bb19d8c6c41d6f9c917a0c3fbb6638ae#diff-f186fc1c51faf6f6b63e73942f19d97223044661fe1a39e73bfababd04c1023eL42-L43

altergui:
ok, now i know. my guess was wrong.
k is nil
and of course vochaintest is buggy:
```
func privKeyToSigner(key string) (*ethereum.SignKeys, error) {
	var skey *ethereum.SignKeys
	if len(key) > 0 {
		skey = ethereum.NewSignKeys()
		if err := skey.AddHexKey(key); err != nil {
			return nil, fmt.Errorf("cannot create key %s with err %s", key, err)
		}
	}
	return skey, nil
}

```

if `len(key) == 0` a nil k is returned and ends up panicking

altergui:
or, well, we could fix `crypto/ethereum/ethereum.go`, in all methods check
`if k != nil && k.Private.D == nil {`

altergui:
anyway, here we go. double fix

> or, well, we could fix `crypto/ethereum/ethereum.go`, in all methods check `if k != nil && k.Private.D == nil {`

this sounds sensible in any case, so i included this fix as well.

also fixed the reason why k was nil in the first place, which was start_test.sh was reading the old env var names

p4u:
> my understanding is that this change is causing the panic. Why, i don't know. i can't understand it. but it's the only change in your commit that is touching this area of code. image
> 
> [e817815#diff-f186fc1c51faf6f6b63e73942f19d97223044661fe1a39e73bfababd04c1023eL42-L43](https://github.com/vocdoni/vocdoni-node/commit/e8178150bb19d8c6c41d6f9c917a0c3fbb6638ae#diff-f186fc1c51faf6f6b63e73942f19d97223044661fe1a39e73bfababd04c1023eL42-L43)

I made this change precisely to try to fix the panic. I cannot understand how "*k" can be nil and still execute methods available on k :-1:  That's why I did not add the check "if k == nil"

crypto/ethereum/ethereum.go
```diff
@@ -132,7 +132,7 @@ func (k *SignKeys) AddressString() string {
 
 // SignEthereum signs a message. Message is a normal string (no HexString nor a Hash)
 func (k *SignKeys) SignEthereum(message []byte) ([]byte, error) {
-	if k.Private.D == nil {
+	if k != nil && k.Private.D == nil {
```
p4u: @mvdan do this check over `k` make sense to you?

mvdan: depends on how the methods are meant to be used. _in general_, it's up to the caller to check that the receiver isn't nil when calling methods on a pointer, so the method itself shouldn't have a nil check. I'd say that applies here too. It's only in cases where the nil value is unavoidable where you want a nil check inside the method, e.g. a `String() string` method which may return `<nil>`.

p4u: @altergui I think I'd prefer to remove this check. And instead expect the caller is doing the proper check (as all other struct types defined in our code)

PR #731, opened 2022-12-13
oracle: enqueue trnsactions and check nonce

p4u:
Currently, due design limitations, an account can only send one transaction per block. If more than one transaction is send, the nonces will probably collide and the second is discarted.

To bypass this issue, now the oracle transactions are handled by a queue, which will wait until the nonce is correct before sending the transaction.


p4u:
@mvdan can you review the logic behind the channel?

PR #737, opened 2022-12-16
Feature/zk prover custom struct

lucasmenendez:
Includes a new `Proof` struct that wraps all the required data to perform a verification, but also some helper functions to encode and decode that struct.

### DIFF
--- a/crypto/zk/prover/prover.go
+++ b/crypto/zk/prover/prover.go
@@ -28,23 +28,23 @@ var (
 )
 
 // ProofData struct contains the calculated parameters of a Proof. It allows to
-// encode and decode the go-rapidsnark inputs and outputs easily.
+// encode and decode go-rapidsnark inputs and outputs easily.
 type ProofData struct {
 	A []string   `json:"pi_a"`
 	B [][]string `json:"pi_b"`
 	C []string   `json:"pi_c"`
 }
 
-// Prove struct envolves the ProofData struct and its associated public signals.
-// Contains all required information to perform a proof verification.
+// Proof struct wraps the ProofData struct and its associated public signals.
+// Contains all the required information to perform a proof verification.
 type Proof struct {
 	Data       ProofData `json:"data"`
 	PubSignals []string  `json:"pubSignals"`
 }
 
-// ParseProof function encoding the provided proof data and public signals into
-// a Proof struct performing an unmarshal operation over them. Returns an error
-// if something was wrong.
+// ParseProof function encodes the provided proof data and public signals into a
+// Proof struct, performing an unmarshal operation over them. Returns an error
+// if something is wrong.
 func ParseProof(proofData, pubSignals []byte) (*Proof, error) {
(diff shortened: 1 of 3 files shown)


crypto/zk/prover/prover.go
```diff
@@ -21,10 +21,59 @@ var (
...
+	signals := []string{}
+	if err := json.Unmarshal(pubSignals, &signals); err != nil {
+		return nil, ErrEncodingProof
+	}
+	return &Proof{Data: data, PubSignals: signals}, nil
+}
+
+// Bytes function returns the current Proof struct parameters Data and
+// PubSignals as []byte. It returns an error if something fails.
+func (p *Proof) Bytes() (proofData []byte, pubSignals []byte, err error) {
```
p4u: We usually prefer not to use named return parameters. It can be sometimes confusing for someone reading or modifying the code. Declare variables in the method body is much more direct and might avoid some mistakes.

crypto/zk/prover/prover.go
```diff
@@ -21,10 +21,59 @@ var (
...
+	ErrParsePubSignals = fmt.Errorf("error during zksnark public signals proof generation, it must be a json with array of strings")
+	ErrEncodingProof   = fmt.Errorf("error encoding prove result into a proof struct")
+	ErrDecodingProof   = fmt.Errorf("error decoding prove as []byte")
 	ErrVerifyProof     = fmt.Errorf("error during zksnark verification")
 )
 
+// ProofData struct contains the calculated parameters of a Proof. It allows to
+// encode and decode go-rapidsnark inputs and outputs easily.
+type ProofData struct {
+	A []string   `json:"pi_a"`
```
p4u, line 33: As far as I understand, these are hexadecimal strings. A better option might be to use our `types.HexBytes{}` which is a []byte wrapper that marshals the bytes as hexadecimal (instead of base64) and has a `.String()` method to transform the byte slice to hex string.

p4u (changes requested):

crypto/zk/prover/prover.go
lucasmenendez: Are you suggesting that i should convert each item of each ProofData parameter into a `types.HexBytes{}` or just the inputs of `prover.ParseProof()` and results of `prover.Bytes()` functions? 
I ask this because if each parameter must be a `type.HexByte{}`, this can complicate the code quite a bit, since they should be simple or nested lists.

p4u: Ideally, the type would be `types.HexBytes{}`, but you're right it requires an extra logic which might not be worth it. So, it's up to you to decide.

p4u (approved):

crypto/zk/prover/prover.go
lucasmenendez: I will check the possibility of change it to `types.HexBytes{}` and i will post a comment with the result.

PR #740, opened 2022-12-20
api: add results endpoint

jordipainan:
@p4u I'm getting the info from the indexer, it is also an option to get it directly from the state.


p4u:
You need to rebase with master. There is a change that affects your PR, there has been a package renaming from "bearerstandardapi" to "apirest"

p4u:
Although your code looks good, I think it makes more sense to use the `/results` endpoint for returning the results contained within the blockchain State. Specially, if the scope of this new endpoint is to be used by Chainlink, that makes more sense. 

I think in the near future election results might be computed by the same blockchain logic so there will be a consensus involved. Currently the Oracles are the actors sending the results transaction, the blockchain does not validate them, just stores (one results for each oracle).

Since in the near future we would have a consensus over a single result, we might just return a JSON object with a single result.

Thus for now the results endpoint might do the following:
1. Get all oracle results stored in state
2. Compare them
3. If not match, return an error explaining the results do not match
4. If match, return a single results object

We might also add some other state related information such as the number of votes.

```json
{
 "electionId": 0x1234,
 "results": [  "1", "2", "1" ],
 "voteCount: 4
 // weight? do we have it stored in the state?
 }
 ```
Let's also think if there exist more information that could be useful for chainlink, such as the "censusOrigin" and the "censusRoot" (probably not?).

We might also consider renaming the results endpoint to something more self-explanatory, such as "consensusResults"

**Comment by p4u** [comment]
@jordipainan let me know when this is ready for review

**Comment by jordipainan** [comment]
@p4u ready :)

File: api/elections.go [hunk 1]
```diff
@@ -250,6 +261,71 @@ func (a *API) electionVotesHandler(msg *apirest.APIdata, ctx *httprouter.HTTPCon
...
+		return fmt.Errorf("cannot get election results: %w", err)
+	}
+	if process == nil {
+		return fmt.Errorf("cannot fetch process %x", electionID)
+	}
+	// do not make distinction between live results elections and encrypted results elections
+	// since we fetch the results from the blockchain state, elections must be terminated and
+	// results must be available
+	if process.Status != models.ProcessStatus_RESULTS {
+		return fmt.Errorf("the election is not yet finalized")
```
p4u: the error message should be something like "no results yet results for the election"

api/elections.go
```diff
@@ -250,6 +261,71 @@ func (a *API) electionVotesHandler(msg *apirest.APIdata, ctx *httprouter.HTTPCon
...
+	}
+	// check process results are the same between each other
+	var results [][]*types.BigInt
+	if process.Results == nil || len(process.Results) == 0 {
+		return fmt.Errorf("this election has no results, may be that no votes were cast")
```
p4u: if not votes cast, results should be calculated anyway. The error message looks not correct

api/elections.go
```diff
@@ -250,6 +261,71 @@ func (a *API) electionVotesHandler(msg *apirest.APIdata, ctx *httprouter.HTTPCon
...
+	}
+	prevProcessResult := process.Results[0]
+	signers := make([]types.HexBytes, 0)
```
p4u: no need for using make if the length is 0, just declare the slice

api/elections.go
```diff
@@ -250,6 +261,71 @@ func (a *API) electionVotesHandler(msg *apirest.APIdata, ctx *httprouter.HTTPCon
...
+			}
+			continue
+		}
+		// if consensus results do not match, return error
+		for k, questionResult := range processResult.Votes {
+			for kk, questionOption := range questionResult.Question {
+				if !bytes.Equal(questionOption, prevProcessResult.Votes[k].Question[kk]) {
+					log.Debugf("election results missmatch %s != %s, current signer %s prev signer %s",
+						questionResult.String(), prevProcessResult.Votes[k].String(),
+						common.BytesToAddress(processResult.OracleAddress), common.BytesToAddress(prevProcessResult.OracleAddress))
```
p4u: this line seems to long

api/elections.go
```diff
@@ -250,6 +261,71 @@ func (a *API) electionVotesHandler(msg *apirest.APIdata, ctx *httprouter.HTTPCon
...
+		return fmt.Errorf("the election is not yet finalized")
+	}
+	// check process results are the same between each other
+	var results [][]*types.BigInt
+	if process.Results == nil || len(process.Results) == 0 {
+		return fmt.Errorf("this election has no results, may be that no votes were cast")
+	}
+	prevProcessResult := process.Results[0]
+	signers := make([]types.HexBytes, 0)
+	for i, processResult := range process.Results {
```
p4u: I don't fully understand the logic behind this nested loops. Isn't an approach like the following more simple?

1. Take the first available results (no need to iterate, just get them)
2. For the next available results, iterate and compare to the one we already have (no need to compare with the previous, just with the first one).
3. Either all of them matches or not. If a result do not match, return error.

api/elections.go
```diff
@@ -250,6 +261,71 @@ func (a *API) electionVotesHandler(msg *apirest.APIdata, ctx *httprouter.HTTPCon
...
+					return fmt.Errorf("reported election results missmatch")
+				}
+			}
+		}
+		prevProcessResult = processResult
+	}
+	electionResults := &ConsensusElectionResults{
+		ElectionID: process.ProcessId,
+		Results:    results,
+		Signers:    signers,
```
p4u: I'd not include Signers to the reply since this might be deprecated in the future once the results are computed inside the blockchain logic.

p4u (changes requested):

api/elections.go
```diff
@@ -250,6 +261,64 @@ func (a *API) electionVotesHandler(msg *apirest.APIdata, ctx *httprouter.HTTPCon
...
+	// results must be available
+	if process.Status != models.ProcessStatus_RESULTS {
+		return fmt.Errorf("no results yet results for the election")
+	}
+	// check process results are the same
+	for _, processResult := range process.Results {
+		// if consensus results do not match, return error
+		for k, questionResult := range processResult.Votes {
+			for kk, questionOption := range questionResult.Question {
+				if !bytes.Equal(questionOption, process.Results[0].Votes[k].Question[kk]) {
```
p4u: I think we should check Reaults[0], Votes[k] and Question[kk] are actually not nil. Since that would end in a panic.

Nit: I'd create a new var firstResults := Results[0] for better code readibility before entering he loop and set the index to +1 (since it is not needed to compare the firstResults with themselves). But if len(0) you can just skip the check.

p4u (changes requested):

api/api_types.go
```diff
@@ -32,6 +33,17 @@ type ElectionSummary struct {
 	Results      [][]*types.BigInt `json:"result,omitempty"`
 }
 
+type ElectionResults struct {
```
p4u, line 37: Please add comments and descriptions to all exported types, methods and fields.

api/api_types.go
```diff
@@ -32,6 +33,17 @@ type ElectionSummary struct {
...
+	// ABIEncoded is the abi encoded election results
+	ABIEncoded string `json:"abiEncoded"`
+
+	CensusRoot types.HexBytes `json:"censusRoot"`
+	// SourceContractAddress is the address of the smart contract containing the census
+	OrganizationID        types.HexBytes `json:"organizationId"`
+	Results               [][]*big.Int   `json:"results"`
+	SourceContractAddress types.HexBytes `json:"sourceContractAddress"`
```
p4u: this field should have the JSON decorator `omitempty` since it won't be always available.

api/elections.go
```diff
@@ -250,6 +261,87 @@ func (a *API) electionVotesHandler(msg *apirest.APIdata, ctx *httprouter.HTTPCon
...
+		return fmt.Errorf("cannot get election results: %w", err)
+	}
+	if process == nil {
+		return fmt.Errorf("cannot fetch process %x", electionID)
+	}
+	// do not make distinction between live results elections and encrypted results elections
+	// since we fetch the results from the blockchain state, elections must be terminated and
+	// results must be available
+	if process.Status != models.ProcessStatus_RESULTS {
+		return fmt.Errorf("no results yet results for the election")
```
p4u: rephrase

api/elections.go
```diff
@@ -250,6 +261,87 @@ func (a *API) electionVotesHandler(msg *apirest.APIdata, ctx *httprouter.HTTPCon
...
+	}
+	// check process results are the same
+	var results [][]*big.Int
+	if len(process.Results) > 0 {
+		firstResult := process.Results[0]
+		for _, processResult := range process.Results[1:] {
```
p4u, line 288: process.Results[1] might not exists (panic), in this case we just exit the if.

api/elections.go
```diff
@@ -250,6 +261,87 @@ func (a *API) electionVotesHandler(msg *apirest.APIdata, ctx *httprouter.HTTPCon
...
+			if processResult == nil {
+				return fmt.Errorf("invalid process result")
```
p4u: that should be a "continue" with an Error message, since we can just ignore the wrong results

api/elections.go
```diff
@@ -250,6 +261,87 @@ func (a *API) electionVotesHandler(msg *apirest.APIdata, ctx *httprouter.HTTPCon
...
+			}
+			// if consensus results do not match, return error
+			for k, questionResult := range processResult.Votes {
+				if questionResult == nil {
+					return fmt.Errorf("invalid question result")
```
p4u: same as before, we don't want to exit the function here, just continue (also apply to the rest of returns inside the loops)

api/elections.go
```diff
@@ -250,6 +261,87 @@ func (a *API) electionVotesHandler(msg *apirest.APIdata, ctx *httprouter.HTTPCon
...
+					if len(questionOption) == 0 {
+						return fmt.Errorf("invalid question option")
+					}
+					if !bytes.Equal(questionOption, firstResult.Votes[k].Question[kk]) {
+						log.Debugf("election results for signer %s missmatch %s != %s",
+							questionResult.String(),
+							firstResult.Votes[k].String(),
+							common.BytesToAddress(processResult.OracleAddress),
+						)
+						return fmt.Errorf("reported election results missmatch")
```
p4u, line 310: This is the only return that should persist IMHO

p4u (changes requested):

api/elections.go
jordipainan: if not exists the slice will be empty and the for loop wont be executed

p4u (approved):

p4u:
@jordipainan there are some conflicts that you should address.

Whenever it is fixed, you can merge the PR. Congrats!

api/elections.go
p4u: You'r right. For some reason I don't understand this creates a Panic:

```golang
	a := []string{"1"}
	for _, x := range a[2:] {
		fmt.Print(x)
	}
```
But `for _, x := range a[1:]` does not.

jordipainan:
added electionID as return param

PR #743, opened 2023-01-03
api: add faucet endpoint for testing

p4u:
New endpoint on /faucet/:network/:address on the Gateway API and the Vocone API.

This endpoint is enabled if the flag --enableFaucetWithAmount is set to a number higher than 0, in that case the number indicates the amount of tokens the faucet endpoint sends.

Note that there is no protection for using this faucet, its aim is to be used only for testing purposes.


jordipainan (approved):
lgtm

PR #746, opened 2023-01-04
switch to zerolog

p4u:
Keeping the current compatibility, use Zerolog to improve the console log messaging. The output is now prettier, the old `info` messages are now better structured (using key/values) and the tendermint log is also much easy to read thanks to the colored key-value output.

Log can still improve, this is a first step.

New functions are available to have a better log output, such as:

- `log.Errorw(err, "some message")`
- `log.Infow("connecting...", map[string]interface{}{"ip": ipAddr, "protocol":"quic"}`
- `log.Monitor("ipfs info", map[string]interface{}{"peers": len(peers), "networkSize": len(network.KnownPeers)})`

The old ones still exist and have the same behavior, but the key-value based logging provides more understandable messages. It is recommended to use them if it makes sense for new code. 


vochain/start.go
```diff
@@ -57,45 +57,56 @@ func NewVochain(vochaincfg *config.VochainCfg, genesis []byte) *BaseApplication
...
+	case "debug":
+		l.logLevel = 0
+	case "info":
+		l.logLevel = 1
+	case "error":
+		l.logLevel = 2
+	case "disabled", "none":
+		l.logLevel = 3
+	}
+}
```
altergui, line 76: some constants here would help readability and replace the logLevel comment

altergui: but not requesting changes, LGTM as-is

altergui (approved):
sounds great, i didn't test the code but i remember back in the day i was trying to improve the log package, i remember coming across zerolog and it looked so good i considered using it, just never got down to it. very nice work!

jordipainan:
ltgm, but can we delete the new `require` section on the go.mod and put all imports on the same `require`? @mvdan pointed this out on the past.

jordipainan (approved):

p4u:
Added JSON format for file output, now you can do the following (looks pretty neat):

`go run ./cmd/node --chain=stage --logLevel=debug --logOutput=/tmp/out.log`

`tail -f /tmp/out.lo  | jq .`

PR #753, opened 2023-01-12
api: test and rewrite convertKeysToCamel

mvdan:


p4u (approved):
Much more neat than my version, thanks!

PR #755, opened 2023-01-12
F/support overwrite vote

p4u:


vochain/app.go
```diff
@@ -90,6 +90,7 @@ func NewBaseApplication(dbType, dbpath string) (*BaseApplication, error) {
 		TransactionHandler: transactionHandler,
 		blockCache:         lru.NewAtomic(32),
 		dataDir:            dbpath,
+		chainID:            "test",
```
jordipainan, line 93: Are you sure that we want to hardcode the chainID value as “test” here? Even if it can be set later, seems strange.

vochain/state/process.go
```diff
@@ -68,6 +69,21 @@ func (v *State) AddProcess(p *models.Process) error {
...
+		"endBlock":      p.BlockCount + p.StartBlock,
+		"mode":          p.Mode,
+		"envelopeType":  p.EnvelopeType,
+		"voteOptions":   p.VoteOptions,
+		"censusRoot":    fmt.Sprintf("%x", p.CensusRoot),
+		"censusOrigin":  models.CensusOrigin_name[int32(p.CensusOrigin)],
+		"maxCensusSize": p.MaxCensusSize,
+		"status":        p.Status,
+		"height":        v.CurrentHeight(),
+		"censusURI":     censusURI,
```
jordipainan, line 85: Some info like SourceNetworkAddress is ignored, maybe it is useful to add it here for certain processes.

jordipainan (approved):

PR #756, opened 2023-01-17
several changes on vochain and testing

p4u:
The main change (the point to start this PR) was to replace the current Protobuf message "Vote" by an internal GoLang type.

There is no need to declare a protobuf model on this type and it is messy with other protobuf types such as Envelope.

The only reason to have this type as Protobuf model, was to be able to serialize it in order to compute the Hash, which is stored into the blockchain state. However this has been replaced by a new method Hash() that serializes the most rellevant fields of the Vote type (which are already byte arrays).

On this point, the state.AddVote() method has been changed, instead of just storing the vote hash, we also store now the Weight and VotePackage. Thus it will make possible to compute an election results using only state information (without the Indexer).

Then I got into the rabit hole... fixing and improving testing and other stuff.


api/accounts.go
```diff
@@ -211,7 +211,7 @@ func (a *API) accountSetHandler(msg *apirest.APIdata, ctx *httprouter.HTTPContex
 		} else {
 			resp.MetadataURL = a.storage.URIprefix() + cid
 		}
-		if cid != metadataCID {
+		if strings.TrimPrefix(cid, "ipfs://") != strings.TrimPrefix(metadataCID, "ipfss://") {
```
jordipainan: Is `ipfss://` correct here? Seems a typo

test/api_test.go
```diff
@@ -165,23 +165,24 @@ func TestAPIcensusAndVote(t *testing.T) {
...
-	   qt.Assert(t, code, qt.Equals, 200)
-	   v2 := &api.Vote{}
-	   err = json.Unmarshal(resp, v2)
-	   qt.Assert(t, err, qt.IsNil)
-	   qt.Assert(t, v2.VoteID.String(), qt.Equals, v.VoteID.String())
-	   qt.Assert(t, v2.BlockHeight, qt.Equals, uint32(3))
-	   qt.Assert(t, v2.TransactionIndex, qt.Equals, uint32(0))
-	   qt.Assert(t, v2.VoterID.String(), qt.Equals, voterKey.AddressString())
-	*/
+	//  This test is disabled because the indexer is not properly set up in the test environment.
```
jordipainan: Can you please add the `TODO` keyword so we can later filter all TODO's ?

test/api_test.go
```diff
@@ -202,6 +203,7 @@ func TestAPIaccount(t *testing.T) {
 	waitUntilHeight(t, c, 1)
 
 	// create a new account
+
```
jordipainan: Unnecessary empty line

vochain/app.go
```diff
@@ -554,8 +579,8 @@ func (app *BaseApplication) Query(req abcitypes.RequestQuery) abcitypes.Response
 	return abcitypes.ResponseQuery{}
 }
 
-// EndBlock updates the app height and timestamp at the end of the current block
-func (app *BaseApplication) EndBlock(req abcitypes.RequestEndBlock) abcitypes.ResponseEndBlock {
+// fnEndBlockDefault updates the app height and timestamp at the end of the current block
+func (app *BaseApplication) fnEndBlockDefault(req abcitypes.RequestEndBlock) abcitypes.ResponseEndBlock {
```
jordipainan, line 583: Why the `default` naming ?

p4u: Do you have a better suggestion?

jordipainan: I would just get rid of `default` and use `fnEndBlock` which is the default, and then you have the option of setting it for changing the behaviour

p4u: This collides with the definition on BaseApplication

jordipainan: ok, so ignore my comment. I do not have a better suggestion.

test/api_test.go
p4u: Actually, I re-enabled this test, the comment was just a garbage.

jordipainan (approved):

PR #757, opened 2023-01-17
vochain/indexer: implement WalkEnvelopes with sqlite, add tests for GetEnvelopes

mvdan:


mvdan:
With this done, there are only two main queries left to do in sqlite: GetEnvelopes (hence the added tests), and AfterSyncBootstrap. I assume we want AfterSyncBootstrap to work with sqlite too - if for some reason it isn't needed or can be simplified, let me know.

jordipainan (approved):

PR #758, opened 2023-01-18
use Informal System's Tendermint Core fork

jordipainan:
It appears that the official repo is being abandoned and now Informal is taking over the development with their [fork](https://github.com/tendermint/tendermint/issues/9972).

For now, let's maintain the same version and only change the source repository.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -363,3 +363,13 @@ require (
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 	lukechampine.com/blake3 v1.1.7 // indirect
 )
+
+// Use Informal System's fork of Tendermint Core
+//
+// It seems that the official repo is being abandoned and now Informal is
+// taking over the development with their fork:
+// https://github.com/tendermint/tendermint/issues/9972
+//
+// Twitter thread:
+// https://twitter.com/informalinc/status/1613580954383040512
+replace github.com/tendermint/tendermint => github.com/informalsystems/tendermint v0.35.9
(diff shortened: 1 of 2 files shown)


mvdan:
If we're keeping the same version, do we win anything by switching? It seems to me that there's no benefit if we keep the same version, because we can rely on go.sum to ensure that the old repo isn't tampered with. And even if they delete it, proxy.golang.org will have it for a long time.

There is one disadvantage to switching now; they havent renamed the module in the fork, so we need to use a replace directive to use the fork. This is hacky and will need to be copied into any vocdoni-node downstreams, so we should avoid it if we can.

Once the fork releases a new version under the module name `github.com/informalsystems/tendermint`, then we can swap the dependency properly by renaming all the imports, and then we don't need a `replace` at all. But until then, I'm not sure that we want to do this :)

jordipainan:
@mvdan Gotcha, I did not think about the disadvantage you are pointing out, and I did not know that a replacement was a hacky thing. Let's make the update whenever there are changes that are useful for us.
As a side note, 0.35.X is now an abandoned version and the only version that informal is going to maintain is the 0.34.24. For us, the improvements made in 0.35 were worth the update. I think that downgrading the version is not an option because of these useful changes we are making use of, and better if we wait for the ABCI++ (Tendermint2) release that should happen in the following months.
Closing this for now.

mvdan:
Ah, we're in an awkward spot then. Hopefully they release tendermint2 soon, as otherwise we might start running into compatibility or security issues over time.

p4u:
Lol, this is sad. Tendermint 0.35 had so many improvements in relation to 0.34.X. The code was being refactored in a much idiomatic way and they started to adopt standard frameworks instead of custom non-maintained ones such as AMINO (replaced by Protobuf) or the p2p layer (replaced by libp2p).

That are bad news for us, since we use Tendermint in a very raw way (Cosmos relies on the ABCI connection protocol and all this shitty abstraction layer) and 0.35 was precisely improving this.

I'd love to speak with some other projects interested in keeping the development and ideas of 0.35

PR #759, opened 2023-01-18
api: add overWriteCount

p4u:
The votes endpoint now returns `overwriteCount` with the number of vote overwrites already consumed by the voter.

This PR extends the Indexer with this new functionallity.

TODO: modify the live results computation so the overwrites of votes are taken into account. Currently the live votes just adds the new (overwritten) votes instead of replacing them. However the final results will be correct.


vochain/indexer/queries/vote_reference.sql
```diff
@@ -1,10 +1,10 @@
 -- name: CreateVoteReference :execresult
-INSERT INTO vote_references (
+REPLACE INTO vote_references (
```
mvdan, line 2: why replace? why would we ever want to replace an existing vote?

vochain/indexer/migrations/0003_alter_table_processes_votes.sql
```diff
@@ -1,5 +1,6 @@
 -- +goose Up
 ALTER TABLE vote_references ADD voter_id BLOB NOT NULL DEFAULT X'';
+ALTER TABLE vote_references ADD overwrite_count INTEGER NOT NULL;
```
mvdan: you should add a new migration with a higher number in the filename, otherwise sqlite databases which are up to date will not get the new column.

you also need to add a default, given the NOT NULL, e.g. `DEFAULT 0`.

you should also add the "down" variant.

vochain/indexer/process.go
```diff
@@ -24,7 +24,7 @@ import (
 var zeroBytes = []byte("")
 
 func nonNullBytes(p []byte) []byte {
-	if p == nil {
+	if len(p) == 0 {
```
mvdan: why? if the incoming byte slice is non-nil and length zero, then we don't need to replace it, as far as I can tell.

vochain/indexer/migrations/0003_alter_table_processes_votes.sql
mvdan: cc @jordipainan since we recently investigated why a dvote node was failing to upgrade the db, given we had modified an existing migration once it hit master.

jordipainan: yup, and this should be added after #700

vochain/indexer/queries/vote_reference.sql
p4u: Yes, that's the objective of this PR. We want to allow voters to send multiple time its vote (if the election is configured to allow so).

vochain/indexer/migrations/0003_alter_table_processes_votes.sql
p4u: > you should add a new migration with a higher number in the filename, otherwise sqlite databases which are up to date will not get the new column.

Since we don't need backwards compatibility at this point, I'd take profit of it and remove all migrations. This is the reason why I did not create a new file and just used an existing one. However, that should be done in a separated commit.

> you also need to add a default, given the NOT NULL, e.g. DEFAULT 0.

In theory this is already addressed by Golang since an uint32 has the default value 0. But probably its more correct to add it also here, so I'll do. Thanks.

> you should also add the "down" variant.

What's that?

vochain/indexer/process.go
p4u: Right, it was not necessary.

vochain/indexer/migrations/0003_alter_table_processes_votes.sql
mvdan: You do need a default, because these are sql values, not go :)

see the "Down" bits at the bottom of the file.

PR #760, opened 2023-01-23
vochain/indexer: fix presumed bugs in GetEnvelopes

mvdan:
(see commit message)


jordipainan:
Agree on the points, thank you @mvdan

jordipainan (approved):

mvdan:
The bug is actually worse than what I wrote in the commit message. For empty or short pids, the returned structs have the empty or short pid filled in, rather than each pid retrieved from the database :see_no_evil:

PR #771, opened 2023-01-26
AfterSyncBootstrap sqlite

mariajdab:
`GetProcessIDsByFinalResults` was added on `process.sql`  to obtain the process ids
`if enableBadgerhold` then will obtain the process ids using the  previous database system

`UpdateProcessResultByID` was added on `process.sql` to update by process ID
`if enableBadgerhold` then will update the process using the  previous database system

### DIFF
--- a/vochain/indexer/db/process.sql.go
+++ b/vochain/indexer/db/process.sql.go
@@ -197,6 +197,34 @@ func (q *Queries) GetProcessEnvelopeHeight(ctx context.Context, id types.Process
 	return results_envelope_height, err
 }
 
+const getProcessIDsByFinalResults = `-- name: GetProcessIDsByFinalResults :many
+SELECT id FROM processes
+WHERE final_results = ?
+`
+
+func (q *Queries) GetProcessIDsByFinalResults(ctx context.Context, finalResults bool) ([]types.ProcessID, error) {
+	rows, err := q.db.QueryContext(ctx, getProcessIDsByFinalResults, finalResults)
+	if err != nil {
+		return nil, err
+	}
+	defer rows.Close()
+	var items []types.ProcessID
+	for rows.Next() {
+		var id types.ProcessID
+		if err := rows.Scan(&id); err != nil {
+			return nil, err
+		}
+		items = append(items, id)
+	}
+	if err := rows.Close(); err != nil {
+		return nil, err
+	}
+	if err := rows.Err(); err != nil {
+		return nil, err
+	}
+	return items, nil
+}
+
 const getProcessStatus = `-- name: GetProcessStatus :one
 SELECT status FROM processes
 WHERE id = ?
@@ -219,34 +247,6 @@ func (q *Queries) GetTotalProcessEnvelopeHeight(ctx context.Context) (interface{
 	var sum interface{}
 	err := row.Scan(&sum)
 	return sum, err
-}
-
-const getProcessByFinalResults = `-- name: GetProcessByFinalResults :many
-SELECT id FROM processes
-WHERE final_results = ?
-`
-
-func (q *Queries) GetProcessByFinalResults(ctx context.Context, finalResults bool) ([]types.ProcessID, error) {
-	rows, err := q.db.QueryContext(ctx, getProcessByFinalResults, finalResults)
-	if err != nil {
-		return nil, err
-	}
-	defer rows.Close()
-	var prcs []types.ProcessID
-	for rows.Next() {
-		var id types.ProcessID
-		if err := rows.Scan(&id); err != nil {
-			return nil, err
-		}
-		prcs = append(prcs, id)
-	}
-	if err := rows.Close(); err != nil {
-		return nil, err
-	}
-	if err := rows.Err(); err != nil {
-		return nil, err
-	}
-	return prcs, nil
 }
 
 const searchEntities = `-- name: SearchEntities :many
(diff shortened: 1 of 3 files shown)


vochain/indexer/db/process.sql.go
```diff
@@ -221,6 +221,34 @@ func (q *Queries) GetTotalProcessEnvelopeHeight(ctx context.Context) (interface{
 	return sum, err
 }
 
+const getProcessByFinalResults = `-- name: GetProcessByFinalResults :many
+SELECT id FROM processes
+WHERE final_results = ?
+`
+
+func (q *Queries) GetProcessByFinalResults(ctx context.Context, finalResults bool) ([]types.ProcessID, error) {
```
mvdan: did you write this by hand? these files are generated by sqlc, and they have a large "DO NOT EDIT" at the top :) the source for the SQL queries is in `vochain/indexer/queries`, and you can use `go generate` to run sqlc to generate the code.

For example, see a past change: d33b0d14a6372b23fd52190c3ead9fe3280cc369

mvdan (commented):
Also note that there's a second query that needs to be ported; see `idx.db.Upsert` inside the loop, around line 388.

vochain/indexer/db/process.sql.go
mariajdab: Thanks!

mariajdab:
> Also note that there's a second query that needs to be ported; see `idx.db.Upsert` inside the loop, around line 388.

Get it. I will check that. Thank you!

vochain/indexer/indexer.go
```diff
@@ -344,25 +341,38 @@ func (idx *Indexer) AfterSyncBootstrap() {
...
+	queries, ctx, cancel := idx.timeoutQueries()
+	defer cancel()
+
+	prcIDs, err := queries.GetProcessIDsByFinalResults(ctx, false)
 	if err != nil {
 		log.Error(err)
 	}
-	log.Infof("recovered %d live results processes", len(prcs))
+
+	if !enableBadgerhold {
```
p4u: I think this condition should be reversed

vochain/indexer/indexer.go
```diff
@@ -344,25 +341,38 @@ func (idx *Indexer) AfterSyncBootstrap() {
...
+		// Find those processes which do not have yet final results,
+		// they are considered live so we need to compute the temporary
+		// results (or only its weight in case of Encrypted)
+		//prcs, err := idx.ProcessList(nil, 0, 0, "", 0, "", "READY", true)
```
p4u: This commented code was here to help implement the SQL migration, since querying ProcessList with the state=READY would have the same output as your custom method `GetProcessIDsByFinalResults`. However, your solution is better, so you can remove this comment.

vochain/indexer/indexer.go
mariajdab: Yep, you are right

p4u:
Please, @mariajdab close yourself the opened conversations once addressed.

mvdan:
> Get it. I will check that. Thank you!

let me know when you're done with the upsert query and I can give a second review.

mvdan:
> Doubts: Not sure if the following code needs to be implemented with SQLite too.

if you mean `queryWithRetries`, that was only to work around a bug in badgerhold, the old database, so that's not needed for sqlite.

vochain/indexer/indexer.go
```diff
@@ -374,18 +380,62 @@ func (idx *Indexer) AfterSyncBootstrap() {
...
+			CensusOrigin:      int64(process.GetCensusOrigin()),
+			Status:            int64(process.GetStatus()),
+			Namespace:         int64(process.GetNamespace()),
+			ModePb:            encodedPb(process.GetMode()),
+			PrivateKeys:       strings.Join(process.EncryptionPrivateKeys, ","),
+			PublicKeys:        strings.Join(process.EncryptionPublicKeys, ","),
+			CreationTime:      currentBlockTime,
+			SourceBlockHeight: int64(process.GetSourceBlockHeight()),
+			SourceNetworkID:   process.SourceNetworkId.String(),
+			Metadata:          process.GetMetadata(),
```
mvdan: I _think_ we only need to update the results columns, similar to what the old database did. the new database uses fewer tables, but you can still use SQL to only update some fields.

For that same reason, I think we want the sqlite query to be an `UPDATE ... WHERE id = ?` rather than a REPLACE; that way we don't have to provide _all_ the column values from scratch.

mvdan:
Looks good from a quick skim. Have you run the tests with and without badgerhold enabled, to check that your sqlite queries work properly?

mariajdav:
> Looks good from a quick skim. Have you run the tests with and without badgerhold enabled, to check that your sqlite queries work properly?

Yeah, both are passing the tests

mariajdab:
> Looks good from a quick skim. Have you run the tests with and without badgerhold enabled, to check that your sqlite queries work properly?

Yeah, both are passing the tests

mvdan (approved):
SGTM! I think we could probably use more consistent names for the SQL queries, but there's already a lack of consistency with the existing ones, so we can fix that up later.

mvdan:
Also, do you want to squash your commits? Otherwise @p4u can do a squash-merge. I don't think we want to end up with 9 commits in master for this change. When you squash, you can also leave your final commit with a good commit message.

p4u:
Great job Maria!

It is good to squash commits once they are going to master branch. Please @mariajdab do squash them.

Thanks.

mariajdab:
> Also, do you want to squash your commits? Otherwise @p4u can do a squash-merge. I don't think we want to end up with 9 commits in master for this change. When you squash, you can also leave your final commit with a good commit message.

Sure! I'll do it

mariajdab:
> Great job Maria!
> 
> It is good to squash commits once they are going to master branch. Please @mariajdab do squash them.
> 
> Thanks.

Thanks @p4u

PR #774, opened 2023-01-26
all: start using some generics

mvdan:
(see commit message)


p4u (approved):
Good, so from now on lets use `atomic.X` types instead of declaring standard variables and using `atomic.StoreXXX` @jordipainan @altergui @lucasmenendez 

Btw, I love this statement: "it's better to be safe than sorry". That should be written in stone.

mvdan:
pretty well known saying in english :)

PR #735, opened 2022-12-15
end2endtest: refactor and extend

altergui:
I'll be pushing here my WIP on end2endtest package

- apiclient: named returns
- apiclient: add context to all the Wait* methods
- end2endtest: use api.Wait* methods, dropping duplicated code in helpers.go

PR #778, opened 2023-01-27
(WIP) e2etest: extend to the point of replacing vochaintest

altergui:
followup to #735

will be posting here any progress


altergui:
closed due to protected branch. recreated in #791

PR #785, opened 2023-01-27
db/badgerdb: remove the code and its use in keykeeper

mvdan:
(see commit message)


mvdan:
This removes our only direct use of badgerdb; only badgerhold remains.

p4u (approved):

PR #787, opened 2023-01-30
all: remove unnecessary type conversions

mvdan:
(see commit message)


p4u:
Nice. 
I just created a new wiki page so we can list this kind of tooling: https://github.com/vocdoni/vocdoni-node/wiki/Repository-tools Feel free to add more @mvdan

PR #788, opened 2023-01-31
api/scrutiny: return non empty results

jordipainan:
A bit hacky until state computed results


api/elections.go
```diff
@@ -281,58 +281,84 @@ func (a *API) electionScrutinyHandler(msg *apirest.APIdata, ctx *httprouter.HTTP
...
+
+	// Results are being compared because we want to make sure that
+	// all oracles are sending the same results.
+	equalResults := func(baseResult, otherResult *models.ProcessResult) bool {
+		// if the number of votes is different, the results are different
+		if len(baseResult.Votes) != len(otherResult.Votes) {
+			log.Warnf("different number of votes")
+			return false
+		}
+		for k, questionResult := range otherResult.Votes {
```
mvdan: to be consistent, you could name the parameters `result1, result2`, and then the votes `vote1, vote2` at the top of this loop. You're indexing `Votes[k]` in multiple places otherwise.

api/elections.go
```diff
@@ -281,58 +281,84 @@ func (a *API) electionScrutinyHandler(msg *apirest.APIdata, ctx *httprouter.HTTP
...
-				if questionResult == nil {
-					log.Errorf("invalid question result")
+			for kk, questionOption := range questionResult.Question {
+				if questionOption == nil {
+					log.Warnf("invalid question option")
 					continue
 				}
-				if len(questionResult.Question) != len(firstResult.Votes[k].Question) {
-					log.Errorf("question length do not match")
+				if len(baseResult.Votes[k].Question[kk]) != len(questionOption) {
```
mvdan: FYI: bytes.Equal already checks lengths, so this is in theory not necessary

mvdan (approved):
Perhaps explain in the commit message why it's wanted? but in any case the code seems fine to me.

api/elections.go
jordipainan: yup!

jordipainan:
@mvdan comments addressed and added an extensive commit message. Thank you :)

api/elections.go
```diff
@@ -281,58 +281,82 @@ func (a *API) electionScrutinyHandler(msg *apirest.APIdata, ctx *httprouter.HTTP
...
+		return fmt.Errorf("no results available")
+	}
+	firstResult := intermediateResults[0]
+
+	// Results are being compared because we want to make sure that
+	// all oracles are sending the same results.
+	equalResults := func(baseResult, otherResult *models.ProcessResult) bool {
+		// if the number of votes is different, the results are different
+		if len(baseResult.Votes) != len(otherResult.Votes) {
+			log.Warnf("different number of votes")
```
p4u: I think its good to print some more debug info and to use the new Warnw() function for this

api/elections.go
```diff
@@ -281,58 +281,82 @@ func (a *API) electionScrutinyHandler(msg *apirest.APIdata, ctx *httprouter.HTTP
...
+		if len(processResult.OracleAddress) == 0 { // check oracle sent the results otherwise ignore
+			continue
+		}
+		intermediateResults = append(intermediateResults, processResult)
+	} // now we have an array of non void results set by a set of oracles via SetProcessResults
+
+	if len(intermediateResults) == 0 {
+		return fmt.Errorf("no results available")
+	}
+	firstResult := intermediateResults[0]
```
p4u: This assignment should go next to the place where it is used, so line 342

api/elections.go
```diff
@@ -281,58 +281,82 @@ func (a *API) electionScrutinyHandler(msg *apirest.APIdata, ctx *httprouter.HTTP
...
+	if len(intermediateResults) > 1 {
+		for _, processResult := range intermediateResults[1:] {
+			if !equalResults(firstResult, processResult) {
+				return fmt.Errorf("reported election results missmatch")
 			}
 		}
+	} // at this point results are equal
+
+	// cast results to big.Int
+	results := make([][]*big.Int, len(firstResult.Votes))
```
p4u: don't we have a function somewhere that already does that?

p4u (approved):

api/elections.go
jordipainan: Yup, I've changed to Warnw() and I've added more useful information

api/elections.go
jordipainan: Right! Changed

api/elections.go
jordipainan: No AFAIK, but I've created a new one just below the `GetFriendlyResults` so now we can reuse it when is needed :)

vochain/state/utils.go
```diff
@@ -31,6 +31,18 @@ func GetFriendlyResults(results []*models.QuestionResult) [][]string {
 	return r
 }
 
+// ResultsToBigIntMatrix converts the results of a process to a big.Int matrix.
+func ResultsToBigIntMatrix(results []*models.QuestionResult) [][]*big.Int {
```
p4u: We are repeating code. What about modifying GetFriendlyResults to return a `[][]*types.bigInt` instead of [][]string ?

When types.bigInt is stringified or json-marshaled, it correctly converts the contented to a String.

jordipainan: Idk, then we are going to repeat code in other parts of the code. 2 times 2 nested loops.

jordipainan: Made the proper changes for enabling this changes in the last commit :)

api/censuses.go
```diff
@@ -201,7 +201,7 @@ func (a *API) censusAddHandler(msg *apirest.APIdata, ctx *httprouter.HTTPContext
 		if p.Weight == nil {
 			p.Weight = new(types.BigInt).SetUint64(1)
 		}
-		if ref.Indexed && p.Weight.ToInt().Uint64() != 1 {
+		if ref.Indexed && p.Weight.ToStdBigInt().Uint64() != 1 {
```
p4u: What's the point for replacing "ToInt()"  to "ToStdBigInt" ?

rpcclient/api.go
```diff
@@ -246,11 +246,11 @@ func (c *Client) TestResults(pid []byte, totalVotes int, withWeight uint64) (res
 		}
 		log.Infof("no results yet")
 	}
-	total := fmt.Sprintf("%d", uint64(totalVotes)*withWeight)
-	if results[0][1] != total ||
-		results[1][2] != total ||
-		results[2][3] != total ||
-		results[3][4] != total {
+	total := uint64(totalVotes) * withWeight
```
p4u: You can make total a big.Int here so you don't need to declare new variables on each comparison.

types/big.go
```diff
@@ -23,11 +23,11 @@ func (i *BigInt) UnmarshalText(data []byte) error {
 }
 
 func (i *BigInt) GobEncode() ([]byte, error) {
-	return i.ToInt().GobEncode()
+	return i.ToStdBigInt().GobEncode()
```
p4u: This naming change is out of the scope for this PR :(

api/elections.go
```diff
@@ -281,66 +281,104 @@ func (a *API) electionScrutinyHandler(msg *apirest.APIdata, ctx *httprouter.HTTP
...
-				if questionResult == nil {
-					log.Errorf("invalid question result")
-					continue
-				}
-				if len(questionResult.Question) != len(firstResult.Votes[k].Question) {
-					log.Errorf("question length do not match")
+			for kk, question1 := range vote1.Question {
+				question2 := vote2.Question[kk]
+				if question1 == nil {
+					log.Warnw("invalid question option (nil)",
```
p4u: Please rebase with master. The syntax for using log key/value functions has changed, now it expects a list of key/values instead of a map (this way is more standard and comfortable). The correct way to use this functions would be something like: `log.Warnw("invalid question", "question", nil, "index", kk)`

api/elections.go
```diff
@@ -281,66 +281,104 @@ func (a *API) electionScrutinyHandler(msg *apirest.APIdata, ctx *httprouter.HTTP
...
-							questionResult.String(),
-							firstResult.Votes[k].String(),
-							common.BytesToAddress(processResult.OracleAddress),
-						)
-						return fmt.Errorf("reported election results missmatch")
-					}
+				if !bytes.Equal(question1, question2) {
+					log.Warnw("question option bytes do not match",
+						map[string]interface{}{
+							"baseResult question option":  fmt.Sprintf("%x", question2),
```
p4u: So, there is no need to make so explicit the name of the value. On the first parameter you are adding the context message and the next just print useful variable information in key/value format.:

`"question option do not match", "option1", fmt.Sprintf("%x", question2), "option2", fmt.Sprintf("%x", question1))`

api/helpers.go
```diff
@@ -115,7 +116,15 @@ func encodeEVMResultsArgs(electionId common.Hash,
...
-	abiEncodedResultsBytes, err := args.Pack(electionId, organizationId, censusRoot, sourceContractAddress, results)
+	// change results from *types.BigInt to *bigInt as args.Pack requires math/big.Int type
+	resultsStd := make([][]*big.Int, len(results))
+	for i, r := range results {
+		resultsStd[i] = make([]*big.Int, len(r))
+		for j, v := range r {
+			resultsStd[i][j] = v.ToStdBigInt()
+		}
+	}
+	abiEncodedResultsBytes, err := args.Pack(electionId, organizationId, censusRoot, sourceContractAddress, resultsStd)
```
p4u, line 127: @mvdan is this loop required for transforming types.BigInt to big.Int? Isn't a way to make a direct casting? (just wondering)

oracle/oracle.go
```diff
@@ -153,11 +153,20 @@ func (o *Oracle) OnComputeResults(results *indexertypes.Results, proc *indexerty
...
+	resultsBigInt := state.GetFriendlyResults(setprocessTxArgs.Results.GetVotes())
+	// convert results to string matrix
+	resultsString := make([][]string, len(resultsBigInt))
+	for i, v := range resultsBigInt {
+		resultsString[i] = make([]string, len(v))
+		for j, w := range v {
+			resultsString[i][j] = w.String()
+		}
+	}
 	signedResultsPayload := OracleResults{
```
p4u: OracleResults.Results could be of type `[][]*types.BigInt` since JSON marhals it as String.

api/helpers.go
mvdan: unfortunately the loop is required right now. even though the types are the same in memory, they are different named types in the Go type system, so you can't go from one slice to another directly. we could probably avoid this awkwardness with either generics or avoiding the need for `types.BigInt` to exist at all, but probably for another PR. perhaps raise an issue and I can look at it when we're done with sqlite :)

api/censuses.go
jordipainan: When I was coding this PR, specifically the part touching the abi.Encode I needed a method to cast results from types.BigInt to math/big.
When I looked at the code at first glance I did not find anything and later on, I found that ToInt()  what it actually does is to cast to math/big. This is, at least, anti-intuitive and I cannot find a single reason on why someone should expect that a method called ToInt() instead of casting a bigInt to Int cast from types.BigInt to a math/big BigInt.

So, I've decided to change the name to a more suited one.

rpcclient/api.go
jordipainan: yup! changed

types/big.go
jordipainan: Well, the scope of this PR was to quickly fix the API retuning empty results, and we've spent 1 week iterating over some things that are not in the scope of this PR. 
I agree that this naming change could be included in another PR, but given how this PR evolved, the change does not hurt anyone and IMHO can be included here.
At this point, I prefer not to roll back the changes.

jordipainan:
Commits can be squashed

api/censuses.go
p4u: I see. So `ToInt()` method name did not provide enough info about what it actually does. However if we are changing this, I'd not use`ToStdBigInt`, taking ideas from other big.Int wrappers, a better naming IMO would be `MathBigInt()` which clearly indicates what is the method returning.

types/big.go
p4u: Ok. But a separated commit with this change would have been appreciated.

api/censuses.go
jordipainan: I agree that MathBigInt() is a better naming. 
When I'm searching for methods, my brain would expect ToMathBigInt() but is just opinionated I guess, the next thing I would do is to search without the To keyword so I guess MathBigInt() is fine too :)
Thank you for the suggestion on the naming.

jordipainan: Changed :)

types/big.go
jordipainan: Gotcha, I will do it better next time.

p4u:
Squash it yourself with a proper message please.

PR #791, opened 2023-02-01
e2etest: implement testTokenTransactions from vochaintest

altergui:
this also includes a small refactor to allow more tests to be added in the future

### DIFF
--- a/api/elections.go
+++ b/api/elections.go
@@ -32,7 +32,15 @@ func (a *API) enableElectionHandlers() error {
 		"/elections",
 		"GET",
 		apirest.MethodAccessTypePublic,
-		a.electionHandler,
+		a.electionFullListHandler,
+	); err != nil {
+		return err
+	}
+	if err := a.endpoint.RegisterMethod(
+		"/elections/page/{page}",
+		"GET",
+		apirest.MethodAccessTypePublic,
+		a.electionFullListHandler,
 	); err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


altergui:
i'm soooo confident this works, i even added to testsuite so it's run in CI :innocent:

apiclient/blockchain.go
```diff
@@ -8,13 +8,89 @@ import (
...
+	case models.TxType_NEW_PROCESS:
+		return txscost.Costs.NewProcess, nil
+	case models.TxType_ADD_DELEGATE_FOR_ACCOUNT:
+		return txscost.Costs.AddDelegateForAccount, nil
+	case models.TxType_DEL_DELEGATE_FOR_ACCOUNT:
+		return txscost.Costs.DelDelegateForAccount, nil
+	case models.TxType_COLLECT_FAUCET:
+		return txscost.Costs.CollectFaucet, nil
+	default:
+		return 0, fmt.Errorf("transaction type unknown")
```
altergui: this is sooo horrible. any ideas?

p4u: see vochain/state/balances.go

cmd/end2endtest/account.go
```diff
@@ -1,32 +1,277 @@
...
+	}
+}
+
+func testGetTxCost(api *apiclient.HTTPclient) error {
+	// get treasurer nonce
+	treasurer, err := api.Treasurer()
+	if err != nil {
+		return err
+	}
+	log.Infof("treasurer is %s", common.BytesToAddress(treasurer.Address))
```
p4u: from now on, we are trying to use the key/value log functions when printing information with values, such as `log.Infow("treasurer", map[string]interface{}{"address": common.BytesToAddress(treasurer.Address)}`

Also use `log.Errorw(err, "message")` for printing errors.

api/api_types.go
```diff
@@ -158,6 +158,23 @@ type TransactionReference struct {
 	Height uint32 `json:"blockHeight"`
 	Index  uint32 `json:"transactionIndex"`
 }
+type TransactionsCost struct {
+	Costs transactionCosts `json:"costs,omitempty"`
+}
+type transactionCosts struct {
```
p4u: Is this definition actually needed? Cannot we manage it with a `map[string]uint64` api/chain.go does?

apiclient/blockchain.go
```diff
@@ -8,13 +8,89 @@ import (
...
+	}
+}
+
+// SubmitTx POSTs the given tx to the API, and returns txHash and response
+func (c *HTTPclient) SubmitTx(stx *models.SignedTx) (types.HexBytes, []byte, error) {
```
p4u: Is this method needed?

Usually in the apiclient we try to abstract the user from dealing with transactions. So if there is a need for sending a signed transaction, a specific method should be implemented. 

For dealing with multiple keys to perform different signatures you can use: `c.Clone("newPrivKey").Whatever()`

cmd/end2endtest/account.go
```diff
@@ -1,32 +1,277 @@
...
+	if err := bob.Generate(); err != nil {
+		log.Fatal(err)
+	}
+
+	// Connect to the API host
+	hostURL, err := url.Parse(c.host)
+	if err != nil {
+		log.Fatal(err)
+	}
+	log.Debugf("connecting to %s", hostURL.String())
```
p4u: log.Debugw() will look nice here :)

p4u (approved):

altergui:
closed due to branch rename, followup in #793

apiclient/blockchain.go
altergui: definitely not needed, actually unused. dropped it

api/api_types.go
altergui: loved this solution. made everything so simpler, thanks a million

cmd/end2endtest/account.go
altergui: right, forgot about that. will do :+1:

PR #793, opened 2023-02-02
e2etest: implement testTokenTransactions from vochaintest

altergui:
this also includes a small refactor to allow more tests to be added in the future

i'm soooo confident this works, i even added to testsuite so it's run in CI :innocent:
and removed the redundant tests from vochaintest

### DIFF
--- a/cmd/end2endtest/account.go
+++ b/cmd/end2endtest/account.go
@@ -1,28 +1,275 @@
 package main
 
 import (
+	"context"
+	"encoding/hex"
+	"fmt"
+	"net/url"
+	"strings"
+	"time"
+
+	"github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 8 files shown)


cmd/end2endtest/account.go
```diff
@@ -1,32 +1,277 @@
...
+	if err := bob.Generate(); err != nil {
+		log.Fatal(err)
+	}
+
+	// Connect to the API host
+	hostURL, err := url.Parse(c.host)
+	if err != nil {
+		log.Fatal(err)
+	}
+	log.Debugw("connecting to API", "host", hostURL.String())
```
altergui, line 42: :sunglasses:

altergui (commented):
now using log.Debugw

p4u (approved):

PR #794, opened 2023-02-04
all: remove badgerhold and badger

mvdan:
I implemented computePendingProcesses as well, which I had missed.


p4u:
This is a great moment, we should celebrate :partying_face:

p4u (approved):

PR #795, opened 2023-02-06
vochain/indexer: fix WalkEnvelopes race with async==true

mvdan:
(see commit message)


mvdan:
This should fix the failure in CI on master.

p4u (approved):

PR #796, opened 2023-02-06
ci: analyze code coverage in 'go test', and send to coverall.io

altergui:
adapted from @p4u 4799f58a add coverage analysis using coverall.io API and gh actions

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -68,7 +68,7 @@ jobs:
         run: go mod tidy
       - name: Run Go test
         if: github.ref != 'refs/heads/master' && github.ref != 'refs/heads/stage' && !startsWith(github.ref, 'refs/heads/release') && github.ref != 'refs/heads/dev' && github.ref != 'refs/heads/apiv2'
-        run: go test ./...
+        run: go test -coverprofile=covprofile ./...
       - name: Run Go test race
         if: github.ref == 'refs/heads/master' || github.ref == 'refs/heads/stage' || startsWith(github.ref, 'refs/heads/release') || github.ref == 'refs/heads/dev' || github.ref == 'refs/heads/apiv2'
         run: |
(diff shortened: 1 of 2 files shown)


altergui:
i'll use this PR to do a couple of tests

specifically i'm interested in comparing the coverage collected by `go test` vs the one collected by the new feature of go v1.20 that can make binaries collect code coverage (so we can see code coverage of vochaintest, end2endtest)

altergui:
i see mvdan removed `-covermode atomic` since it's expensive, so i'll leave the default mode `set` as well

from https://go.dev/blog/cover#heat-maps

The go test command accepts a -covermode flag to set the coverage mode to one of three settings:

    set: did each statement run?
    count: how many times did each statement run?
    atomic: like count, but counts precisely in parallel programs

The default is ‘set’, which we’ve already seen. The atomic setting is needed only when accurate counts are required when running parallel algorithms. It uses atomic operations from the [sync/atomic](https://go.dev/pkg/sync/atomic/) package, which can be quite expensive. For most purposes, though, the count mode works fine and, like the default set mode, is very cheap.

mvdan:
please don't upgrade to Go 1.20 at the same time; that is a large change that will require updating dependencies.

mvdan:
Also, use `go install github.com/mattn/goveralls@latest`, and remove `GO111MODULE: off`.

altergui:
> please don't upgrade to Go 1.20 at the same time; that is a large change that will require updating dependencies.

absolutely, i'm just playing with collecting coverage from testsuite in CI and i need go 1.20 for that, but i would definitely create a separate PR for bumping go

so no worries, no, you won't see a commit from me saying "whitespace fix" that actually upgrades go :rofl:

mvdan:
Ah I see. FYI, @p4u is already looking at updating to Go 1.20, but we can't yet as go-libp2p need to do a release first.

p4u:
> Ah I see. FYI, @p4u is already looking at updating to Go 1.20, but we can't yet as go-libp2p need to do a release first.

I just tried, and it didn't work (cause of go-libp2p) so we'll need to wait. But I reverted the changes, so there is nothing to commit.

altergui:
left #799 and will keep this current PR just covering 'go test'. trying was worth it, sometimes the version bump is trivial

altergui:
greatly superseded by #853

Issue #792, opened 2023-02-02
api: elections/count and total transactions


p4u:
Perform the following modifications to the API
- [] add new endpoint /elections/count
- [] extend chain/info so it includes the total count of transactions

Open the issues on interoperability so SDK and Docs update.

jordipainan:
> add new endpoint /elections/count 

What this method returns? The total process count of a given organization or the total process count of all organizations?
If it is the later why we want to have 2 endpoints that return the same info? The consumer can just use chain/info.

Thank you! @p4u

p4u:
The total since the election count for an org is already available.

p4u:
Sorry @jordipainan my initial description was wrong. There is no need to duplicate the transaction count in two different endpoints.

p4u:
So, I'm being very confusing, sorry. 

I'd just extend `/chain/info` with the total number of elections and transactions. That's it.

jordipainan:
just saw this message, oki!

PR #798, opened 2023-02-06
f/api/792

jordipainan:
(see commit messages)

Partially address #792 (missing SDK and Docs, will add in interoperability)

### DIFF
--- a/api/elections.go
+++ b/api/elections.go
@@ -93,14 +93,6 @@ func (a *API) enableElectionHandlers() error {
 		return err
 	}
 	if err := a.endpoint.RegisterMethod(
-		"/elections/count",
-		"GET",
-		apirest.MethodAccessTypePublic,
-		a.electionsCountHandler,
-	); err != nil {
-		return err
-	}
-	if err := a.endpoint.RegisterMethod(
 		"/elections",
 		"POST",
 		apirest.MethodAccessTypePublic,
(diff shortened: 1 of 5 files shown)


api/elections.go
```diff
@@ -92,6 +92,14 @@ func (a *API) enableElectionHandlers() error {
 	); err != nil {
 		return err
 	}
+	if err := a.endpoint.RegisterMethod(
```
p4u: If `chain/info` already returns this info, we don't need this new endpoint. I think we should try to avoid duplicate info. IMO, we can remove this new endpoint.

apiclient/election.go
```diff
@@ -363,3 +363,21 @@ func (c *HTTPclient) ElectionResults(electionID types.HexBytes) (*api.ElectionRe
 	}
 	return electionResults, nil
 }
+
+// ElectionsCount returns the total number of elections
+func (c *HTTPclient) ElectionsCount() (uint64, error) {
```
p4u: Already on Info too.

api/elections.go
jordipainan: but chain/info returns the number of transactions not the number of voting processes

jordipainan:
done @p4u

jordipainan:
self remainder squash

p4u (approved):

PR #800, opened 2023-02-07
f/interop/11

jordipainan:
See https://github.com/vocdoni/interoperability/issues/11

### DIFF
--- a/api/api_types.go
+++ b/api/api_types.go
@@ -163,6 +163,7 @@ type ChainInfo struct {
 	BlockTime        *[5]int32 `json:"blockTime,omitempty"`
 	ElectionCount    *uint64   `json:"electionCount,omitempty"`
 	Height           *uint32   `json:"height,omitempty"`
+	Syncing          *bool     `json:"syncing,omitempty"`
 	Timestamp        *int64    `json:"blockTimestamp,omitempty"`
 	TransactionCount *uint64   `json:"transactionCount,omitempty"`
 	ValidatorCount   *uint32   `json:"validatorCount,omitempty"`
--- a/api/api_types.go
+++ b/api/api_types.go
@@ -162,6 +162,7 @@ type ChainInfo struct {
 	ID               string    `json:"chainId,omitempty"`
 	BlockTime        *[5]int32 `json:"blockTime,omitempty"`
 	ElectionCount    *uint64   `json:"electionCount,omitempty"`
+	GenesisTime      time.Time `json:"genesisTime,omitempty"`
 	Height           *uint32   `json:"height,omitempty"`
 	Syncing          *bool     `json:"syncing,omitempty"`
 	Timestamp        *int64    `json:"blockTimestamp,omitempty"`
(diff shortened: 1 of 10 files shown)


api/api_types.go
```diff
@@ -162,9 +162,13 @@ type ChainInfo struct {
 	ID               string    `json:"chainId,omitempty"`
 	BlockTime        *[5]int32 `json:"blockTime,omitempty"`
 	ElectionCount    *uint64   `json:"electionCount,omitempty"`
+	GenesisTime      time.Time `json:"genesisTime,omitempty"`
 	Height           *uint32   `json:"height,omitempty"`
+	Syncing          *bool     `json:"syncing,omitempty"`
```
p4u: I'm not sure about declaring all these pointers here. I think ChainInfo is only used by the info endpoint, thus all fields must have a valid value. Is there a specific reason? If not, I'd just declare all of them as standard vars and remove omitempty (I know this is out of the scope here, but I think it is worth to fix it).

api/chain.go
```diff
@@ -185,22 +186,50 @@ func (a *API) organizationCountHandler(msg *apirest.APIdata, ctx *httprouter.HTT
...
 		Height:           &height,
-		Timestamp:        &timestamp,
+		Syncing:          &isSyncing,
 		TransactionCount: &transactionCount,
-	})
+		ValidatorCount:   numValidators,
+		Timestamp:        &timestamp,
+		VoteCount:        &voteCount,
+	}
+	tctx, cancel := context.WithTimeout(context.TODO(), 2*time.Second)
```
p4u: Why a timeout context? Is Genesis() performing an operation that might have a timeout?

api/chain.go
```diff
@@ -185,22 +186,50 @@ func (a *API) organizationCountHandler(msg *apirest.APIdata, ctx *httprouter.HTT
...
+	defer cancel()
+	if a.vocapp.Node != nil {
```
p4u: Clients are calling info() several times, I think it is worth it to store the genesis as a static value on the vocapp struct. Then we can access it directly through `a.vocapp.Genesis.GenesisTime`

p4u (changes requested):

api/api_types.go
jordipainan: The reason is that the other struct fields were declared as pointers and I followed the previous work done.
I assume there were declared as pointers because we want 0/false to be valid values and not ignored in the response, and only ignored if such a specific case is required (by assigning them a nil value).

But ack, going to change the values to get rid of pointers.

api/chain.go
jordipainan: Because the method requires a context to be passed.

api/chain.go
jordipainan: gotcha, going to add this info to the vocapp struct.

jordipainan:
Comments addressed, thank you!

p4u (approved):

p4u:
Can you fix the compose test and squash?

PR #801, opened 2023-02-07
log: millisecond timestamps

altergui:
made with love, very precise love

### DIFF
--- a/log/log.go
+++ b/log/log.go
@@ -2,13 +2,13 @@ package log
 
 import (
 	"fmt"
+	"io"
 	"os"
 	"strconv"
 	"strings"
 	"time"
 
 	"github.com/rs/zerolog"
-	zlog "github.com/rs/zerolog/log"
 	"google.golang.org/protobuf/encoding/protojson"
 	"google.golang.org/protobuf/reflect/protoreflect"
 )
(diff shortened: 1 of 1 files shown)


p4u:
Do we really need nanoseconds? Wouldn't be millisecond enough? Even seconds should work....

The point for reducing the precision is to reduce noise on the log output.

altergui:
100% agree. console actually logs milliseconds.
the setting is confusing tho. needs `w.TimeFormat = time.RFC3339Nano` (or else the precision is just seconds) and then `zerolog.TimeFieldFormat = zerolog.TimeFormatUnixMs` reduces that precision to milliseconds

as it is, the log shows milliseconds, which is what we want. just ignore the `RFC3339Nano` thing (it is needed, tho)

altergui:
just in case this is not clear from my last message:

this PR makes the log show milliseconds (not nanoseconds) exactly as you (and me) want

the `RFC3339Nano` const might confuse you, but it's actually the right way to do it (there's no RFC3339Ms const, only RFC3339 and RFC3339Nano). let's say that sets the "internal" precision of zerolog package, and then the TimeFieldFormat dictates the precision (milliseconds) of the printed log.

p4u (approved):

Issue #802, opened 2023-02-07
api: handle errors in an idiomatic way


altergui:
in commit https://github.com/vocdoni/vocdoni-node/commit/2464a8340617e468b77eae11018c52549b515565 it became evident we could handle errors better
@jordipainan suggested using constants.

at some point we need to match strings, since it's a reply coming from an API. but the idea would be as soon as possible to turn that into a native `error` and use golang error wrapping and matching facilities.

i'm happy to implement something along those lines.
@mvdan you most likely have some valuable pointers, opinion or ready-made solution?

mvdan:
Assuming that this is an error string being returned by our server on an API request, we could make the indexer (which uses sqlite) catch "not found" errors returned by sqlite, and return those as a special error code to our API clients - akin to an HTTP 404.

altergui:
now the first part is merged, the remaining task would be to somehow assign HTTP error codes (404 for example) to some of the errors. i'll explore this today and propose a PR

altergui:
second part done.

apiclient is still untouched, until now i've focused on the api part, and reporting this in interop so that the SDK can adapt to those changes.

i'll start with integrating this apiclient now, if there are no objections

altergui:
apiclient still pending, that second PR was totally half-baked and i fixed it in #832 , also added unit tests

PR #804, opened 2023-02-07
api: refactor errors: replace hardcoded strings with constants of APIError type

altergui:
here's an initial proposal for fixing #802 

i just touched accounts.go (many more files make sense to be refactored) to gather early feedback on the idea


api/accounts.go
```diff
@@ -182,22 +182,22 @@ func (a *API) accountSetHandler(msg *apirest.APIdata, ctx *httprouter.HTTPContex
...
+		return ErrNoMetadataURI
 	}
 
 	var metadataCID string
 	if req.Metadata != nil {
 		// if election metadata defined, check the format
 		metadata := AccountMetadata{}
 		if err := json.Unmarshal(req.Metadata, &metadata); err != nil {
-			return fmt.Errorf("wrong metadata format: %w", err)
+			return fmt.Errorf("%w: %w", ErrWrongMetadataFormat, err)
```
mvdan: I think we can ignore the warnings here; it's likely because our go vet is still go 1.19.

mvdan (commented):
I agree with this approach in general, but one thought: will the API clients be matching returned errors against these ErrX strings?

That's going to be a bit hacky in the long term. For example, if we ever spot a typo in one of these messages, or if we want to make the messages a bit clearer for end users, we'll have to update _all the clients_ at once to prevent breaking them.

One alternative would be to use something like:

```
package api

type ErrorCode int

const (
    ErrAddressMalformed = 1001 // address malformed
    ErrAccountNotFound = 1002 // account not found
    [...]
)
```

Then we can use `//go:generate go run golang.org/x/tools/cmd/stringer@v0.5.0 -linecomment -type ErrorCode` to generate `String` methods with those strings from the comments.

Then, the server can respond with JSON like:

```
"error": {"code": 1001, "message": "address malformed"}
```

and the API clients can compare the code to known constants rather than matching strings directly.

Finally, we could also assign these error code numbers by categories, like HTTP does for 2XX, 3XX, 4XX, etc. For example, we could say that error codes in the 4001-4999 range are the user's fault and the error codes 5001-5999 are the server's fault, mimicking HTTP.

mvdan:
I also realise that `stringer` generates `String` methods and not `Error` methods. If it's easier for us to use these error codes as error values directly, we can always do `func (e ErrorCode) Error() string { return e.String() }`.

api/accounts.go
altergui: seems we can't, go 1.19 fails to build, it's not just go vet acting out. so it seems we'll need go 1.20, until then i'll fallback to "%w %v"

altergui:
@mvdan that's amaaaaaaaaaaazing. i was totally unaware of go:generate and of tools/cmd/stringer

also would likely not have thought of `func (e ErrorCode) Error() string { return e.String() }`

now 
```
func blah() error {
	return ErrCantDecodeOrgID
}

func main() {
     fmt.Println(blah())
}
```

returns
```
"error": {"code": 5002, "message": "cannot decode organizationID"}
```

altergui:
my Error() method looks hacky to me, but for some reason i thought doing a JSON marshal would be overkill. if i'm wrong, just say so and i'm happy to code it

api/errors.go
```diff
@@ -0,0 +1,41 @@
...
+	ErrCantDecodeAccountID                              // cannot decode accountID
+	ErrCantDecodeElectionID                             // cannot decode electionID
+	ErrCantParsePageNumber                              // cannot parse page number
+	ErrCantFetchElectionList                            // cannot fetch election list
+	ErrCantFetchElectionID                              // cannot fetch electionID
+	ErrCantFetchTokenTransfers                          // cannot fetch token transfers
+)
+
+func (e ErrorCode) Error() string {
+	return fmt.Sprintf("\"error\": {\"code\": %d, \"message\": \"%s\"}", e, e.String())
```
mvdan: to be clear, I'm not saying that the Error method should return JSON. This kind of `{code: x, message: y}` encoding should be done by the REST HTTP layer, which presumably responds to all JSON requests with a JSON response. The error value can be that object with two fields rather than just one string.

mvdan: The Error method here should just return the message, which is what will be useful when logging errors in Go code, for example. We could always make it something like `code: message` if we want to include both.

altergui: makes absolute sense

altergui:
so, this also took waaaaaaaaay longer than expected, since i tried my best to deduplicate similar errors, and grouping them in some sensible order (could be better but i think it's good enough)

altergui:
one issue i encountered is that censusdb/censusdb.go is a different package (`censusdb`) and if i want to reference `api.ErrWhatever` I can't `import api` from censusdb because that generates a circular import.

i recall seeing in other projects a solution for this, if i'm not mistaken, which is declare the errors on `censusdb` and then "clone" them (`ErrWhatever = censusdb.ErrWhatever`) in `api` package. So that other packages (like `apiclient`) can reference them as api.ErrWhatever anyway, without worrying if it's "really" from censusdb.

i might do this, but wanted first to push this "finished" PR (if we ignore censusdb for now) for feedback already

p4u:
Is this ready @altergui ?

altergui:
yes, ready for review/merge

(forgot to remove the "WIP")

p4u:
Ouch, needs rebase @altergui

altergui:
np, done

p4u (approved):

PR #805, opened 2023-02-09
vochain/indexer: make processes.source_network_id an integer

mvdan:
(see commit message)


mvdan:
By the way, this search/query parameter was never used by any of the API endpoints, as far as I can tell. It was always left empty, which was `""` and is now `0`.

mvdan:
Also, a reminder that this is part of the backwards incompatible cleanups we want to do to our schemas before we commit to stable migrations and SQL data. I have another couple of changes like this over the next week that will break the schema, I think.

p4u:
This value is legacy. In some moment we need to perform a clean refactor on the proto models.

p4u (approved):

PR #806, opened 2023-02-09
vochain/indexer: remove empty fields and clean up code some more

mvdan:


api/elections.go
```diff
@@ -183,12 +183,11 @@ func (a *API) electionHandler(msg *apirest.APIdata, ctx *httprouter.HTTPContext)
 			FinalResults: proc.FinalResults,
 			VoteCount:    count,
 		},
-		MetadataURL:   proc.Metadata,
-		ElectionCount: proc.EntityIndex,
```
p4u, line 187: Not 100% about removing ElectionCount from the API response. It is documented on the developer portal, so probably expected by the SDK and clients: https://developer.vocdoni.io/vocdoni-api/election-by-id

From the RPC it can be removed, for sure.

mvdan: Hmm, OK, but what actually _is_ the election count for one process? That page doesn't mention what its value is meant to be, either :)

FWIW in master this is always 0 right now.

p4u: I think it is the election number for the organization. So the process number 23 out of 50 for example.

mvdan: Right, so that's roughly what I had described on Discord. With badgerhold, when a process was created, EntityIndex was a snapshot of the number of processes created by that organization.

Is that still what we want? I can implement it in sqlite if that's the case, but I had gotten the impression that we didn't need EntityIndex nor ElectionCount (which are the same value).

p4u: Yes, I think that was introduced because the product people wanted to show this information on voice.aragon.org but since we are refactoring the whole API and the frontends, I'd remove them, and just re-add if they are actually needed.

However we should add a new issue on interoperability since this is gonna change the API and could break the SDK (which expects a specific JSON).

mvdan: ACK. I can leave the fields around always set to 0 with a TODO, if it helps making the API transition less bumpy.

p4u: I'd remove them, let's clean old stuff... It should be quick to change it on the SDK side.

I can open the interoperability issue once PR merged.

mvdan: ack, then this is ready when you review, I think :)

p4u (approved):

PR #807, opened 2023-02-09
Anonymous Weighted Voting Support

lucasmenendez:
This PullRequest includes the support for anonymous voting for the `vocdoni-node` project. To achieve this, it uses a new circom circuit ([hosted here](https://github.com/vocdoni/zk-franchise-proof-circuit/tree/feature/merging_repos_and_new_tests)) and includes the implementation of [`go-rapidsnark` project](https://github.com/iden3/go-rapidsnark) that supports the generation and verification of zk-proofs using that circuit. 

Main challenges achieved:
- Centralizing the zk-snark stuff into a single package, including not only the circuit configuration and the download of the its artifacts, but also the logic to use it with the new prover ([`go-rapidsnark`](https://github.com/iden3/go-rapidsnark)).
- Transform the Census of the anonymous voting from type *indexed* to *weighted* (*zk-weighted*). Which involves to modify what the merkle tree leaves key and value contains. Until now the key of the leaf was the index of the participant into the census (*indexed* census). Now, the key of the leaf contains the `ZkAddress`, and leaf the value the weight assigned to the participant (*zk-weighted* census).
- *End-to-end* tests of a anonymous election process.

## Major changes
* **`crypto/zk`**:
    * New struct `ZkAddress` (`crypto/zk/address.go`) which simplifies the generation and managment of BabyJubJub key pair and Merkle Tree key.
    * New package `crypto/zk/prover` that wraps `go-rapidsnark` API to make it consistent and easy to use.
    * New package `crypto/zk/circuit` that includes the circuit struct (that contains the specific parameters of the ZkSnark circuit according to the `crypto/zk/prover` API) and the circuit configuration struct definitions. Also supports the circuit artifacts download, and contains the configuration of the diferent circuits, that allows to download these artifacts. Until now, the circuit configurations was included into the Vochain genesis (`vochain/genesis/genesis.go`), now the genesis only includes the tag that identifies the circuit configuration inside the `crypto/zk/circuit/config.go` file, allowing to consult that information from the apiclient without read the Vochain code.
    * Some changes on `utils.go` file to update it to the new prover and circuit implementation.
* **`apiclient/census.go`**:
    * New struct `CensusProofZk`: This struct is similar to the old one `CensusProof`, but includes some required parameters to proof that a participant is registered into a census in anonymous way (using ZkSnark's).
    * New function `CensusGenProofZk`: Allows to generate a ZkSnark based CensusProof for any previusly registered participant. It wraps `CensusGenProof` and complete the proof generation computing the ZkSnark circuit inputs with the parameters provided (censusID, electionID and the private key of the participant).

## Minor changes
* **`Dockerfile`**: Including some required dependencies to the container to support the use of `go-rapidsnark`.
* **`vochain/app`**: 
    * The correct circuit tag (according to the execution environment) is included into the `BaseApplication` struct, and it is accesible using its own getter (`app.CircuitConfigurationTag()`) and setter (`app.SetCircuitConfigurationTag(circuitTag)`). 
    * The circuit will be loaded every time that the tag changes. If a wrong tag is provided, the tag will be setted with `"dev"` circuit tag. 
* **`vochain/keykeeper`**: Updating the implementation to support anonymous voting and introducing some checks to prevent panic errors.
* **`api/chain`**: The current circuit tag could be consulted via api at chainInfo endpoint (`/chain/info/circuit`).
* **`api/chain/circuit`**: The current circuit configuration could be consulted via api at chainCircuitInfo endpoint (`/chain/info`).
* **`api/censuses`**: Check if the census Merkle Tree is based on Poseidon Hash function (used into the anonymous voting) to prevent double hashing of the participant key on the census, because the key used in anonymous voting (BabyJubJub) is already hashed during its generation.
* **`apiclient/client`**: Now, the `HTTPClient` struct has its own `ZkAddress` included, generated during the struct initialization based on its account keys (`ethereum.SignKeys`).
* **`apiclient/vote`**: Now check the `election.VoteMode` or the `election.Census.CensusOrigin` to parse the votes according to its type, instead check for nil vote parameters. Anonymous voting flow is also implemented. 
* **`cmd/end2endtest`** (and other test files): Updated unit and end-to-end tests and split the end-to-end tests into both files `weighted.go` and `zkweighted.go` (currently disabled). The last one with some modifications to prevent memory leaks.


lucasmenendez:
First of all, i want to say sorry about the size of this Pull Request. I have to learn about make tiny value deliveries instead of this kind of "epics". Also this is my very first "important" contribution to a project so any kind of tip will be very welcome.
Ask or let me know anything to improve the approach, the solution design, the process, the code and so.

PS: Finally I had decide to not squash some commits because my way implementing this changes has been more in circles than linear, and had not many sense

api/censusdb/censusdb.go
```diff
@@ -74,16 +80,16 @@ func NewCensusDB(db db.Database) *CensusDB {
...
 // New creates a new census and adds it to the database.
 func (c *CensusDB) New(censusID []byte, censusType models.Census_Type,
-	indexed bool, uri string, authToken *uuid.UUID) (*CensusRef, error) {
+	indexed bool, uri string, authToken *uuid.UUID, maxLevels int) (*CensusRef, error) {
 	if c.Exists(censusID) {
 		return nil, ErrCensusAlreadyExists
 	}
 	tree, err := censustree.New(censustree.Options{Name: censusName(censusID), ParentDB: c.db,
-		MaxLevels: 256, CensusType: censusType, IndexAsKeysCensus: indexed})
+		MaxLevels: maxLevels, CensusType: censusType, IndexAsKeysCensus: indexed})
```
p4u, line 88: Let's open an issue to remove all references to "indexed" since this is now deprectaed.

api/censuses.go
```diff
@@ -239,6 +254,39 @@ func (a *API) censusAddHandler(msg *apirest.APIdata, ctx *httprouter.HTTPContext
...
+
+	// Get the current census type from the disk
+	ref, err := a.censusdb.Load(censusID, nil)
+	if err != nil {
+		return err
+	}
+
+	// Set unknown as default type and check other available types
+	censusType := CensusTypeUnknown
+	switch ref.CensusType {
```
p4u, line 272: I think we already have this swtich in a helper function. If not I think it would be good to have it, so other endpoints can reuse it.

api/censuses.go
```diff
@@ -529,18 +578,31 @@ func (a *API) censusProofHandler(msg *apirest.APIdata, ctx *httprouter.HTTPConte
...
+
+	leafV, siblings, err := ref.Tree().GenProof(leafKey)
 	if err != nil {
 		return err
 	}
-	leafV, siblings, err := ref.Tree().GenProof(keyHash)
+
+	// Get the leaf siblings from arbo based on the key received and include
+	// them into the response
+	circomSiblings, err := ref.Tree().GetCircomSiblings(leafKey)
```
p4u: Shouldn't we only return the circom siblings if the census type is zkWeighted?

api/censuses.go
```diff
@@ -529,18 +578,31 @@ func (a *API) censusProofHandler(msg *apirest.APIdata, ctx *httprouter.HTTPConte
 	if err != nil {
 		return err
 	}
-	keyHash, err := ref.Tree().Hash(key)
+
+	leafKey := key
+	if ref.CensusType != int32(models.Census_ARBO_POSEIDON) {
```
p4u, line 601: Please document why we are not performing a hash if the census type is ARBO_POSEIDON

api/censuses.go
```diff
@@ -576,11 +638,16 @@ func (a *API) censusVerifyHandler(msg *apirest.APIdata, ctx *httprouter.HTTPCont
 	if err != nil {
 		return err
 	}
-	keyHash, err := ref.Tree().Hash(cdata.Key)
-	if err != nil {
-		return err
+
+	leafKey := cdata.Key
+	if ref.CensusType != int32(models.Census_ARBO_POSEIDON) {
```
p4u, line 668: Same here

api/censuses.go
```diff
@@ -576,11 +638,16 @@ func (a *API) censusVerifyHandler(msg *apirest.APIdata, ctx *httprouter.HTTPCont
...
+		leafKey, err = ref.Tree().Hash(cdata.Key)
+		if err != nil {
+			return err
+		}
 	}
-	valid, err := ref.Tree().VerifyProof(keyHash, cdata.Value, cdata.Proof, cdata.Root)
+
+	valid, err := ref.Tree().VerifyProof(leafKey, cdata.Value, cdata.Proof, cdata.Root)
 	if err != nil {
 		return err
```
p4u: Please wrap the errors, not all of them but at least those that require explicit information such as `generated proof verification failed: %w`

apiclient/census.go
```diff
@@ -17,6 +24,20 @@ type CensusProof struct {
...
+// CensusProofZk represents the proof for a voter in a zk census. It contains
+// the encoded proof and the circuit public tokens to verify it. It also keeps
+// the weight of the vote and the key type of the merkle tree, and also includes
+// the voter's nullifier. This last parameter is necessary because now, to
+// compute the nullifier, the private key of the ZkAddress is required, so it
+// cannot be computed by anyone but the voter.
+type CensusProofZk struct {
+	Proof      types.HexBytes
+	PubSignals types.HexBytes
+	Weight     uint64
```
p4u: I think weight should be a `types.BigInt`

apiclient/census.go
```diff
@@ -112,3 +133,113 @@ func (c *HTTPclient) CensusGenProof(censusID, voterKey types.HexBytes) (*CensusP
...
+	if code != 200 {
+		return nil, fmt.Errorf("%s: %d (%s)", errCodeNot200, code, resp)
+	}
+	censusData = &api.Census{}
+	err = json.Unmarshal(resp, censusData)
+	if err != nil {
+		return nil, fmt.Errorf("could not unmarshal response: %w", err)
+	}
+
+	log.Debugw("zk census data received, starting to generate the proof inputs...", map[string]interface{}{
```
p4u: Log format changed a week a go, not you must use `"key1", value1, "key2", value2,...)`
```
log.Debugw("zk census data received, starting to generate the proof inputs...",
  "censusRoot", censusRoot.String(),
  "electionId", electionId.String(),
  "zkAddress",  c.zkAddr.String(),
  )
  ```

File: apiclient/census.go [hunk 9]
```diff
@@ -112,3 +133,113 @@ func (c *HTTPclient) CensusGenProof(censusID, voterKey types.HexBytes) (*CensusP
...
+
+	// Calculate and encode vote hash -> sha256(voteWeight)
+	encElectionId := zk.BytesToArboStr(electionId)
+	voteHash := zk.BytesToArboStr(weight.Bytes())
+
+	// Create the inputs and encode them into a JSON
+	rawInputs := map[string]interface{}{
+		"censusRoot":     strCensusRoot,
+		"censusSiblings": censusData.Siblings,
+		"weight":         weight.String(),
```
p4u: So if you use weight as `types.BigInt` it will automatically marshal JSON to string.

apiclient/census.go
```diff
@@ -112,3 +133,113 @@ func (c *HTTPclient) CensusGenProof(censusID, voterKey types.HexBytes) (*CensusP
...
+	if censusData.Weight != nil {
+		weight = censusData.Weight.MathBigInt()
+	}
+
+	// Calculate and encode vote hash -> sha256(voteWeight)
+	encElectionId := zk.BytesToArboStr(electionId)
+	voteHash := zk.BytesToArboStr(weight.Bytes())
+
+	// Create the inputs and encode them into a JSON
+	rawInputs := map[string]interface{}{
```
p4u: I think this struct should be declared somewhere else. Probably in the crypto/zk package.

apiclient/census.go
```diff
@@ -112,3 +133,113 @@ func (c *HTTPclient) CensusGenProof(censusID, voterKey types.HexBytes) (*CensusP
...
+		return nil, fmt.Errorf("error encoding inputs: %w", err)
+	}
+	log.Debugw("zk circuit proof inputs generated", rawInputs)
+
+	// Get artifacts of the current circuit
+	currentCircuit, err := circuit.LoadZkCircuit(context.Background(), c.circuit)
+	if err != nil {
+		return nil, fmt.Errorf("error loading circuit: %w", err)
+	}
+	log.Debugw("zk circuit loaded", map[string]interface{}{
```
p4u: format change, sorry :/

apiclient/client.go
```diff
@@ -63,6 +67,9 @@ func NewHTTPclient(addr *url.URL, bearerToken *uuid.UUID) (*HTTPclient, error) {
 		return nil, fmt.Errorf("cannot get chain ID from API server")
 	}
 	c.chainID = info.ID
+
+	// Get the default circuit config
+	c.circuit = circuit.CircuitsConfigurations[info.CircuitConfigurationTag]
```
p4u: Since this is a map, you should receive the second parameter `ok` and check if its true (else raise an error)

benchmark/zk_census_benchmark_test.go
```diff
@@ -0,0 +1,179 @@
...
+			newSibling = arbo.BytesToBigInt(unpackedSiblings[i]).String()
+		}
+		strSiblings[i] = newSibling
+	}
+	// Get artifacts of the current circuit
+	currentCircuit, err := circuit.LoadZkCircuit(context.Background(), zkCircuitTest)
+	qt.Assert(b, err, qt.IsNil)
+
+	// Create the inputs and encode them into a JSON
+	rawInputs := map[string]interface{}{
```
p4u: uops, repeated struct definition.

censustree/censustree.go
```diff
@@ -51,7 +51,7 @@ type Options struct {
 }
 
 // TMP to be defined the production circuit nLevels
-const nLevels = 256
+const DefaultMaxLevels = 256
```
p4u, line 59: Document properly

censustree/censustree.go
```diff
@@ -76,18 +76,23 @@ func DeleteCensusTreeFromDatabase(kv db.Database, name string) (int, error) {
 // New returns a new Tree, if there already is a Tree in the
 // database, it will load it.
 func New(opts Options) (*Tree, error) {
+	var maxLevels = DefaultMaxLevels
```
p4u, line 79: Reviewed until here from now, please do not remove this comment.

p4u (changes requested):

apiclient/vote.go
```diff
@@ -28,11 +31,12 @@ type VoteData struct {
...
 	ProofMkTree *CensusProof
 	ProofCSP    types.HexBytes
+	ProofZkTree *CensusProofZk
 }
 
 // Vote sends a vote to the Vochain. The vote is a VoteData struct,
-// which contains the electionID, the choices and the proof.  The
-// return value is the voteID (nullifier).
+// which contains the electionID, the choices and the proof. The
+// return value is the fcvoteID (nullifier).
```
altergui: this looks like a typo? given the `return voteAPI.VoteID, nil`

apiclient/vote.go
```diff
@@ -47,9 +51,46 @@ func (c *HTTPclient) Vote(v *VoteData) (types.HexBytes, error) {
...
 	}
 
+	// Get de election metadata
+	election, err := c.Election(v.ElectionID)
+	if err != nil {
+		return nil, err
+	}
+
 	// Build the proof
+	log.Debugw("generating a new vote", map[string]interface{}{"electionId": v.ElectionID})
```
altergui: format change here, blame @p4u :grin:

crypto/zk/address.go
```diff
@@ -0,0 +1,122 @@
...
+
+// AddressFromBytes functions returns a new ZkAddress based on the seed
+// provided. Using it, a BabyJubJub key pair is calculated and based on it. The
+// ZkAddress components are:
+//   - ZkAddress.PrivKey: The BabyJubJub private key big.Int value, following
+//     the EdDSA standard, and using blake-512 hash.
+//   - ZkAddress.PublicKey: The poseidon hash of the BabyJubJub public key
+//     components (X and Y coordinates as big.Int).
+//   - ZkAddress.addr: The truncated to 20 bytes (defaultZkAddrLen) little
+//     endian version of the ZkAddress.PublicKey.
```
altergui: wording is weird. did you mean:
The ZkAddress.PublicKey, little endian, truncated to 20 bytes (defaultZkAddrLen)


or not really sure where does little endian apply

altergui (commented):
can't provide a deep review since most of the code is out of my scope, but i'm reading it like it was a novel and providing some comments

crypto/zk/address.go
```diff
@@ -0,0 +1,122 @@
...
+func AddressFromBytes(seed []byte) (*ZkAddress, error) {
+	if len(seed) < 16 {
+		return nil, fmt.Errorf("the seed provided does not have 32 bytes at least")
```
altergui: 16 or 32?

altergui: also, it might be sensible to mention this lower limit in the func documentation

"based on the seed provided (16 bytes or longer)"

crypto/zk/address.go
```diff
@@ -0,0 +1,122 @@
...
+	}, nil
+}
+
+// AddressFromString wraps AddressFromBytes function transforming the seed from
+// string to []byte.
+func AddressFromString(seed string) (*ZkAddress, error) {
+	return AddressFromBytes([]byte(seed))
+}
+
+// AddressFromSignKeys function gets the private key from the ethereum.SignKeys
```
altergui: in all func documentations, you can omit the "function" word, just `AddressFromSignKeys gets the private key...`

for example `// NewRandAddress returns a ZkAddress...` is perfect

crypto/zk/address.go
```diff
@@ -0,0 +1,122 @@
...
+// types.HexBytes using the arbo.BigIntToBytes() function.
+func (zk *ZkAddress) Bytes() types.HexBytes {
+	return types.HexBytes(arbo.BigIntToBytes(defaultZkAddrLen, zk.addr))
+}
+
+// String function returns the current ZkAddress.Bytes() result transformed to
+// string using the types.HexBytes.String() function.
+func (zk *ZkAddress) String() string {
+	addr := zk.Bytes()
+	return addr.String()
```
altergui: why not make `Bytes()` return `[]byte`?
* would be more consistent with other "Bytes()" method i see in the repo.
* in all calls `Key:    zkAddr.Bytes()`, it will be converted to types.HexBytes anyway
* the only place that would need change is the String() method, but couldn't it be simply `return hex.EncodeToString(zk.Bytes())`?

don't take my suggestion like a good idea necessarily, maybe there's a reason for what you did, i'm commenting this just in case this is a leftover from you going in circles and then leaving this behind. if not, just ignore my comment, no need to reply :grin:

altergui (commented):
more comments

crypto/zk/circuit/circuit.go
```diff
@@ -0,0 +1,278 @@
...
+	// remote location
+	if err := circuit.LoadRemote(ctx); err != nil {
+		return nil, err
+	}
+
+	// checks hashes of current files
+	if correct, err := circuit.VerifiedCircuitArtifacts(); err != nil {
+		return nil, err
+	} else if !correct {
+		return nil, fmt.Errorf("download artifacts does not match with the expected ones")
```
altergui: "hashes from downloaded artifacts don't match the expected ones"

crypto/zk/circuit/circuit.go
```diff
@@ -0,0 +1,278 @@
...
+	return circuit, nil
+}
+
+// LoadLocal tries to read the content of current circuit artifacts from its
+// local path (provingKey, verificationKey and wasm). If any of the read
+// operatios fails, returns an error.
+func (circuit *ZkCircuit) LoadLocal() error {
+	var err error
+
+	log.Debugw("loading circuit locally...", map[string]interface{}{
```
altergui: log.Debugw format change

crypto/zk/circuit/circuit.go
```diff
@@ -0,0 +1,278 @@
...
+	} else if !correct {
+		return nil, fmt.Errorf("download artifacts does not match with the expected ones")
+	}
+
+	return circuit, nil
+}
+
+// LoadLocal tries to read the content of current circuit artifacts from its
+// local path (provingKey, verificationKey and wasm). If any of the read
+// operatios fails, returns an error.
```
altergui: operations fail, returns an error

crypto/zk/circuit/circuit.go
```diff
@@ -0,0 +1,278 @@
...
+
+	// Compose provingKey remote and local locations
+	provingKeyUri := fmt.Sprintf("%s/%s", remotePath, circuit.Config.ProvingKeyFilename)
+	provingKeyLocalPath := filepath.Join(localPath, circuit.Config.ProvingKeyFilename)
+	// Compose verificationKey remote and local locations
+	verificationKeyUri := fmt.Sprintf("%s/%s", remotePath, circuit.Config.VerificationKeyFilename)
+	verificationKeyLocalPath := filepath.Join(localPath, circuit.Config.VerificationKeyFilename)
+	// Compose wasm remote and local locations
+	wasmUri := fmt.Sprintf("%s/%s", remotePath, circuit.Config.WasmFilename)
+	wasmLocalPath := filepath.Join(localPath, circuit.Config.WasmFilename)
```
altergui: this vars init could certainly be refactored to be reused in LoadLocal

and even built into some struct that lets you then do some kind of `for _, key := range keys`
```
	// Download and store locally key
	key.dest, err = downloadFile(ctx, key.uri)
	if err != nil {
		return fmt.Errorf("error downloading %s: %w", key.name, err)
	}
	if err := storeFile(key.dest, key.localPath); err != nil {
		return fmt.Errorf("error storing %s: %w", key.name, err)
	}
	```

Comment (altergui) [reply crypto/zk/circuit/circuit.go]: instead of the triplicated code below

File: crypto/zk/circuit/circuit.go [hunk 26]
```diff
@@ -0,0 +1,278 @@
...
+		return false, fmt.Errorf("no content provided to check")
+	} else if expected == nil {
+		return false, fmt.Errorf("no hash provided to compare")
+	}
+
+	hash := sha256.New()
+	if n, err := hash.Write(content); err != nil {
+		return false, fmt.Errorf("error computing hash function of %s: %w", content, err)
+	} else if n != len(content) {
+		return false, fmt.Errorf("the number of writted bytes does not match with the content provied")
```
altergui: written

provided

crypto/zk/circuit/circuit.go
```diff
@@ -0,0 +1,278 @@
...
+	return content, nil
+}
+
+// storeFile helper function allows to write the file content provided into a
+// new file created at the path provided.
+func storeFile(content []byte, dstPath string) error {
+	if content == nil {
+		return fmt.Errorf("no content provided")
+	} else if _, err := os.Stat(filepath.Dir(dstPath)); err != nil {
+		return fmt.Errorf("destination path parent folder does not exist")
```
altergui: no need for `else`, can be two separate ifs

crypto/zk/circuit/circuit.go
```diff
@@ -0,0 +1,278 @@
...
+	}
+
+	fd, err := os.Create(dstPath)
+	if err != nil {
+		return fmt.Errorf("something was wrong creating the artifact file: %w", err)
```
altergui: why the verbosity?
"error creating artifact file"

crypto/zk/circuit/circuit.go
```diff
@@ -0,0 +1,278 @@
...
+	}
+
+	if nBytes, err := fd.Write(content); err != nil {
+		return fmt.Errorf("something was wrong writting the artifact file: %w", err)
+	} else if len(content) != nBytes {
+		return fmt.Errorf("something was wrong writting the artifact file: the length of the provided content does not match with the bytes writted")
```
altergui: writing

written

also, according to fd.Write() docs, the whole check `len(content) != nBytes` is unnecessary:
`// Write returns a non-nil error when n != len(b).`

crypto/zk/utils.go
```diff
@@ -2,61 +2,119 @@
 package zk
```
altergui, line 2: my brain dumped at this point, will finish reviewing next week, starting from here

p4u:
@lucasmenendez as soon as you apply the suggestions, please mark them as "resolved".

api/censusdb/censusdb.go
lucasmenendez: Done! https://github.com/vocdoni/vocdoni-node/issues/812

api/censuses.go
lucasmenendez: I think we have the inverse function, that transforms a census type from string to models.Census_Type, but the naming of the function (https://github.com/vocdoni/vocdoni-node/blob/feature/new_prover_integration/api/census_helpers.go#L11) could be problematic so I proupose to rename it and create the new one. I'm going to open another issue.

api/censuses.go
lucasmenendez: Yes, I am agree

api/censuses.go
lucasmenendez: Included!

api/censuses.go
lucasmenendez: Included!

apiclient/client.go
lucasmenendez: Solved!

api/censuses.go
lucasmenendez: I included some messages but probably I need to review the code in deep to include more.

apiclient/census.go
lucasmenendez: The old proof parameter was uint64, should I change both types?

apiclient/census.go
lucasmenendez: Solved!

apiclient/census.go
lucasmenendez: Solved

apiclient/census.go
lucasmenendez: Reviewed! Do you think that I should do the same with ZkAddress bigInts (keys and nullifier)?

benchmark/zk_census_benchmark_test.go
lucasmenendez: Sorry... Where? O.O

censustree/censustree.go
lucasmenendez: Solved!

apiclient/vote.go
lucasmenendez: I think all the references to the old syntax have been fixed

apiclient/vote.go
lucasmenendez: Solved! (I think I fell asleep on the keyboard).

crypto/zk/address.go
lucasmenendez: Something like this would be better?: "The ZkAddress.PublicKey truncated to the default length (20 bytes). The ZkAddress.PublicKey is a number formatted with little  endian, so it will be the last 20 bytes."

crypto/zk/circuit/circuit.go
lucasmenendez: included!

crypto/zk/circuit/circuit.go
lucasmenendez: fixed!

crypto/zk/circuit/circuit.go
lucasmenendez: fixed! I dont know that... :D

crypto/zk/circuit/circuit.go
lucasmenendez: messages simplified

crypto/zk/circuit/circuit.go
lucasmenendez: Done!

crypto/zk/circuit/circuit.go
lucasmenendez: updated

crypto/zk/address.go
lucasmenendez: I have not any reason to keep it as types.HexBytes, I only supposed that use it was a team agreement and it is the correct type to be returned by default (when the original one is []bytes).

lucasmenendez: But I am agree with you, so i've updated it.

crypto/zk/address.go
lucasmenendez: Ok, I've fixed every comment that follows this format, so if you find another place where it follows, let me know and I'll fix it.

crypto/zk/address.go
altergui: oooooh now that's perfectly clear to me (i definitely didn't understand it before). so, SGTM how you put it now.

i'm even thinking now, why not start with the conclusion? 
"ZkAddress.addr: The last 20 bytes of the ZkAddress.PublicKey. This is calculated by truncating PublicKey (a little endian number) to the defaultZkAddrLen"

crypto/zk/circuit/circuit.go
lucasmenendez: Agree! I tried to implement changes to improve the code in this way. Please, check it @altergui and let me know your opinion.

crypto/zk/address.go
elboletaire: why use `fmt.Errorf` if you're not formatting the error message? Wouldn't be better to just use `errors.New` here?

lucasmenendez: I try to answer to you @elboletaire, but I am not an expert. I discovered `fmt.Errorf` when I started to work on Vocdoni so I don't know all the advantages or disadvantages between both ways, buy from my experience, creating errors with `fmt.Errorf` allows to wrap an error (like `errors.Wrap`) but also to include a formatted string (using templates). For example, if you want to wrap a received error, and also add more context with some variables:

- Using `errors.Wrap`:
```go
input := "some parameter"
err := someFuncThatFails(input)
if err != nil {
    errMsg := fmt.Sprintf("something fails with input '%s'", input)
    return errors.Wrap(err, errMsg)
}
```

- Using `errors.New`:
```go
input := "some parameter"
err := someFuncThatFails(input)
if err != nil {
    errMsg := fmt.Sprintf("something fails with input '%s': %w", input, err)
    return errors.New(errMsg)
}
```

- Using `fmt.Errorf`:
```go
input := "some parameter"
err := someFuncThatFails(input)
if err != nil {
    return fmt.Errorf("something fails with input '%s': %w", input, err)
}
```

elboletaire: Yeah but you're not using a string template, so what benefits does it have since you're not using it?

crypto/zk/address.go
lucasmenendez: Updated!

crypto/zk/address.go
lucasmenendez: @altergui I updated the restriction to the correct number of bytes (32)

apiclient/census.go
p4u: here @lucasmenendez

benchmark/zk_census_benchmark_test.go
p4u: apiclient/census.go L199

censustree/censustree.go
```diff
@@ -50,8 +50,13 @@ type Options struct {
...
 
-// TMP to be defined the production circuit nLevels
-const nLevels = 256
+// By default, the maximum number of levels will be 256, which allows to add
+// up to 2^256 leaves to the tree, with keys with up to 32 bytes. However the
+// number of levels could be setted up during the tree initialization which
+// allows to support uses cases with specific restrictions such as the
+// zkweighted census, which needs to optimize some artifacts size that depends
+// of the number of levels of the tree.
+const DefaultMaxLevels = 256
```
p4u, line 59: I think we might use a default max levels up to 20 bytes.
On non zk censuses we currently support pubkeys (32 bytes) and addresses (20 bytes). We might skip the pubkey use case. We should create an issue to change this, so we dont forget.

censustree/censustree.go
```diff
@@ -76,18 +81,23 @@ func DeleteCensusTreeFromDatabase(kv db.Database, name string) (int, error) {
...
 // database, it will load it.
 func New(opts Options) (*Tree, error) {
+	var maxLevels = DefaultMaxLevels
 	var hashFunc arbo.HashFunction
 	switch opts.CensusType {
 	case models.Census_ARBO_BLAKE2B:
 		hashFunc = arbo.HashFunctionBlake2b
 	case models.Census_ARBO_POSEIDON:
+		if opts.MaxLevels < maxLevels {
+			maxLevels = opts.MaxLevels + 1
```
p4u, line 96: This is a tricky behavior. Let's add a comment.

crypto/zk/address.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+// NewRandAddress returns a ZkAddress based on a random BabyJubJub private key.
+func NewRandAddress() (*ZkAddress, error) {
+	return AddressFromBytes([]byte(util.RandomHex(32)))
+}
+
+// Bytes returns the current ZkAddress.addr transformed to types.HexBytes using
+// the arbo.BigIntToBytes() function.
+func (zk *ZkAddress) Bytes() []byte {
```
p4u: For better readability, move the methods next to its struct definition.

crypto/zk/address.go
```diff
@@ -0,0 +1,121 @@
...
+	return arbo.BigIntToBytes(defaultZkAddrLen, zk.addr)
+}
+
+// String function returns the current ZkAddress.Bytes() result transformed to
+// string using the types.HexBytes.String() function.
+func (zk *ZkAddress) String() string {
+	addr := zk.Bytes()
```
p4u: nit: you can do this in a single step

crypto/zk/address.go
```diff
@@ -0,0 +1,121 @@
...
+// AddressFromString wraps AddressFromBytes function transforming the seed from
+// string to []byte.
+func AddressFromString(seed string) (*ZkAddress, error) {
+	return AddressFromBytes([]byte(seed))
+}
+
+// AddressFromSignKeys gets the private key from the ethereum.SignKeys provided
+// and pass its string representation as []byte to AddressFromBytes function.
+func AddressFromSignKeys(acc *ethereum.SignKeys) (*ZkAddress, error) {
+	privKey := acc.PrivateKey()
```
p4u: nit: you can do this in a single step

crypto/zk/address.go
```diff
@@ -0,0 +1,121 @@
...
+// components are:
+//   - ZkAddress.PrivKey: The BabyJubJub private key big.Int value, following
+//     the EdDSA standard, and using blake-512 hash.
+//   - ZkAddress.PublicKey: The poseidon hash of the BabyJubJub public key
+//     components (X and Y coordinates as big.Int).
+//   - ZkAddress.addr: The last 20 bytes of the ZkAddress.PublicKey. This is
+//     calculated by truncating PublicKey (a little endian number) to the
+//     defaultZkAddrLen.
+func AddressFromBytes(seed []byte) (*ZkAddress, error) {
+	if len(seed) < 32 {
```
p4u, line 79: Either receive a [32]byte or make it clear on the function description

crypto/zk/address.go
```diff
@@ -0,0 +1,121 @@
...
+	// Calculate nullifier hash: poseidon(babyjubjub(privKey) + sha256(ElectionId))
+	nullifier, err := poseidon.Hash([]*big.Int{
+		zk.PrivKey,
+		intElectionId[0],
+		intElectionId[1],
+	})
+	if err != nil {
+		return nil, fmt.Errorf("error generating nullifier: %w", err)
+	}
+	return nullifier, nil
```
p4u: nit: you can return directly the result of `poseidon.Hash()`

crypto/zk/circuit/circuit.go
```diff
@@ -0,0 +1,263 @@
...
+	"path/filepath"
+	"time"
+
+	"go.vocdoni.io/dvote/log"
+)
+
+// By default set the circuit base dir empty which means that the artifacts will
+// be downloaded or loaded from the LocalDir path associated to the circuit config
+var circuitsBaseDir = ""
+var BaseDir = &circuitsBaseDir
```
p4u: Is this pointer necessary? Why not just declaring BaseDir as string?

crypto/zk/circuit/circuit.go
```diff
@@ -0,0 +1,263 @@
...
+// LoadZkCircuitByTag gets the circuit configuration associated to the provided
+// tag or gets the default one and load its artifacts to prepare the circuit to
+// be used.
+func LoadZkCircuitByTag(configTag string) (*ZkCircuit, error) {
+	circuitConf := GetCircuitConfiguration(configTag)
+
+	ctx, cancel := context.WithTimeout(context.Background(), downloadCircuitsTimeout)
+	defer cancel()
+
+	zkCircuit, err := LoadZkCircuit(ctx, circuitConf)
```
p4u: nit: you can return directly `LoadZkCircuit()`

crypto/zk/circuit/circuit.go
```diff
@@ -0,0 +1,263 @@
...
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, fileUrl, nil)
+	if err != nil {
+		return nil, fmt.Errorf("error creating the file request: %w", err)
+	}
+
+	res, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return nil, err
+	} else if res.StatusCode != http.StatusOK {
```
p4u: This else is not necessary. Also the `defer` should go before this check.

crypto/zk/circuit/circuit.go
```diff
@@ -0,0 +1,263 @@
...
+
+	return content, nil
+}
+
+// storeFile helper function allows to write the file content provided into a
+// new file created at the path provided.
+func storeFile(content []byte, dstPath string) error {
+	if content == nil {
+		return fmt.Errorf("no content provided")
+	} else if _, err := os.Stat(filepath.Dir(dstPath)); err != nil {
```
p4u: Please, simplify if/else cases. If there is a return, no need for an else.

crypto/zk/circuit/circuit.go
```diff
@@ -0,0 +1,263 @@
...
+
+	return nil
+}
+
+// VerifiedCircuitArtifacts checks that the computed hash of every circuit
+// artifact matches with the expected hash, from the circuit config.
+func (circuit *ZkCircuit) VerifiedCircuitArtifacts() (bool, error) {
+	if circuit.ProvingKey == nil || circuit.Config.ProvingKeyHash == nil {
+		return false, fmt.Errorf("provingKey or its hash are nil")
+	} else if circuit.VerificationKey == nil || circuit.Config.VerificationKeyHash == nil {
```
p4u: These else cases can be removed.

crypto/zk/circuit/circuit.go
```diff
@@ -0,0 +1,263 @@
...
+	// if the circuit is not available locally, tries to download from its
+	// remote location
+	if err := circuit.LoadRemote(ctx); err != nil {
+		return nil, err
+	}
+
+	// checks hashes of current files
+	if correct, err := circuit.VerifiedCircuitArtifacts(); err != nil {
+		return nil, err
+	} else if !correct {
```
p4u: IMHO if/else cases should be reduced to improve code clarity. I'd prefer:

```
correct, err := circuit.VerifiedCircuitArtifacts()
if err != nil { return nil, err}
if !correct { ... }
```

crypto/zk/circuit/circuit_test.go
```diff
@@ -0,0 +1,313 @@
...
+	"path"
+	"path/filepath"
+	"testing"
+	"time"
+
+	qt "github.com/frankban/quicktest"
+)
+
+const (
+	filenameProvingKey      = "proving_key.zkey"
```
p4u: nit: add the `test` prefix to testing constants and global variables.

crypto/zk/circuit/config.go
```diff
@@ -0,0 +1,115 @@
...
+func (config ZkCircuitConfig) KeySize() int {
+	return config.Levels / 8
+}
+
+// CircuitsConfiguration stores the relation between the different vochain nets
+// and the associated circuit configuration. Any circuit configuration must have
+// the remote and local location of the circuits artifacts and their metadata
+// such as artifacts hash or the number of parameters.
+var CircuitsConfigurations = map[string]ZkCircuitConfig{
+	"bizono": {
```
p4u: I'd not use these namings for the circuit configurations. We might have 2 dev networks but use the same circuit. We might also have 2 production networks and use the same circuit config.

For now I'd just use `dev`. Once we do the ceremony and so on, we might have something like `prod2023`

crypto/zk/utils.go
p4u: Also here.

p4u (changes requested):

crypto/zk/address.go
p4u: @elboletaire long ago we decided to use only `fmt.Errorf()` for convention. Since it allows formatting, strings is much more convenient. For those cases where the string does not need to be formatted, it just works.

p4u:
> First of all, i want to say sorry about the size of this Pull Request. I have to learn about make tiny value deliveries instead of this kind of "epics". Also this is my very first "important" contribution to a project so any kind of tip will be very welcome. Ask or let me know anything to improve the approach, the solution design, the process, the code and so.
> 
> PS: Finally I had decide to not squash some commits because my way implementing this changes has been more in circles than linear, and had not many sense

No problem Lucas! This is a giant step towards having a full-featured decentralized voting system. It's complex and requires big PR occasionally.

Excellent work btw! Congratulations.

crypto/zk/address.go
lucasmenendez: If i pretend to use the (types.HexBytes).String() function I can't do it in a single step, however i can replace it with `return AddressFromBytes([]byte(hex.EncodeToString(acc.PrivateKey())))` to do it in a single step.

crypto/zk/address.go
lucasmenendez: I'll try to clarify this in the description, because the seed has to be at least  32 bytes, so [32]bytes won't do.

censustree/censustree.go
lucasmenendez: Done! https://github.com/vocdoni/vocdoni-node/issues/819

benchmark/zk_census_benchmark_test.go
lucasmenendez: Done! I created a new struct and function to simplify the circuit input generation: https://github.com/vocdoni/vocdoni-node/blob/feature/new_prover_integration/crypto/zk/circuit/inputs.go

crypto/zk/circuit/circuit.go
lucasmenendez: I thought so, but after some testing I think not. What I was trying to do was to keep the value of the variable between packages, to support that one package could update the value of the variable and ensure that other packages could read the same value. And I thought that the best way to do this would be to use pointers. Example: https://replit.com/@lucasmenendez/testGlobalVars

crypto/zk/utils.go
```diff
@@ -2,61 +2,118 @@
...
 )
 
-func LoadVkFromFile(path string) (*types.Vk, error) {
-	vkJSON, err := os.ReadFile(path)
-	if err != nil {
-		return nil, err
+// ProtobufZKProofToProverProof parses the provided protobuf ready proof struct
+// into a prover ready proof struct.
+func ProtobufZKProofToProverProof(p *models.ProofZkSNARK) (*prover.Proof, error) {
+	if len(p.A) != 3 || len(p.B) != 6 || len(p.C) != 3 {
```
p4u: These magic numbers should be constants.

crypto/zk/utils.go
```diff
@@ -2,61 +2,118 @@
...
-		return nil, nil, fmt.Errorf("error on zkProof format")
+// ProverProofToProtobufZKProof encodes the proof provided into a protobuf ready
+// struct using including the index of the circuit used. If the provided proof
+// does not contains a defined public signals, the rest of the arguments are
+// required to calculate that parameter. If the provided proof does not contains
+// a defined public signals and any of the rest of the parameters is nil, the
+// resulting struct will not contains any defined PublicInputs value.
+func ProverProofToProtobufZKProof(p *prover.Proof,
+	electionId, censusRoot, nullifier types.HexBytes, weight *big.Int) (*models.ProofZkSNARK, error) {
+	if len(p.Data.A) != 3 || len(p.Data.B) != 3 || len(p.Data.C) != 3 {
```
p4u: Same.

crypto/zk/utils.go
```diff
@@ -2,61 +2,118 @@
...
-	// parse zkProof & PublicInputs from tx.Proof
-	proof, err := parsers.ProofStringToProof(proofString)
-	if err != nil {
-		return nil, nil, err
-	}
-	publicInputs, err := parsers.PublicSignalsStringToPublicSignals(publicInputsString)
-	if err != nil {
-		return nil, nil, err
+	if p.PubSignals != nil && len(p.PubSignals) > 0 {
+		if len(p.PubSignals) != 7 {
```
p4u: 7 is also a magic number.

I think this if/else logic clause might be simplified with something like this.

```
proof.PublicInputs = p.PubSignals
if p.PubSignals != nil && len(p.PubSignals) > 0 && len(p.PubSignals) != 7 {
  return ....
 }

// if PubSignals is not provided, we generate the
if proof.PublicInputs == nil  {
 if electionId != nil || censusRoot != nil || nullifier != nil || weight != nil {
  return nil, fmt.Errorf("no enought arguments to generate the calc signals")
 }
 proof.PublicInputs = zkProofPublicInputs(electionId, censusRoot, nullifier, weight)
}
```

vochain/app.go
```diff
@@ -662,6 +670,14 @@ func (app *BaseApplication) SetChainID(chainId string) {
 	app.State.SetChainID(chainId)
 }
 
+func (app *BaseApplication) SetCircuitConfigTag(tag string) {
```
p4u: Comment this exported method please and specify where to find the string tag.

vochain/app.go
```diff
@@ -662,6 +670,14 @@ func (app *BaseApplication) SetChainID(chainId string) {
...
+	// Update the loaded circuit of the current app transactionHandler
+	if err := app.TransactionHandler.LoadZkCircuit(tag); err != nil {
```
p4u, line 685: If tag is not found, an error should be returned.

vochain/genesis/genesis.go
```diff
@@ -351,38 +301,7 @@ var Genesis = map[string]VochainGenesis{
...
-					"zk-circuits-artifacts/6afb7c22d856c8b727262b0a0ae8ab7ca534dd4e",
-				CircuitPath: "zkcensusproof/dev/1024",
-				Parameters:  []int64{1024},
-				LocalDir:    "./circuits",
-				ZKeyHash:    hexToBytes("0x1cd0c9225210700d4d6307493bbe5f98554e29339daba6d9bd08a4e0e78df443"),
-				WitnessHash: hexToBytes("0x61b40e11ece8de3fbfaf27dbd984e0e0b1fa05ee72d4faa0c2be06c1d7a9b845"),
-				VKHash:      hexToBytes("0xaed892ff98ab37b877cfcb678cb5f48f1be9d09dbbaf74b5877f46b54d10f9ad"),
-			},
-		},
+		CircuitsConfigTag: "stage",
```
p4u: is the correct tag stage?

vochain/transaction/election_tx.go
```diff
@@ -130,33 +128,23 @@ func (t *TransactionHandler) NewProcessTxCheck(vtx *vochaintx.VochainTx,
 	// check valid/implemented process types
 	// pre-regiser and anonymous must be either both enabled or disabled, as
 	// we only support a single scenario of pre-register + anonymous.
-	if tx.Process.Mode.PreRegister != tx.Process.EnvelopeType.Anonymous {
-		return nil, common.Address{}, fmt.Errorf("pre-register mode only supported " +
-			"with anonymous envelope type and viceversa")
-	}
+	// if tx.Process.Mode.PreRegister != tx.Process.EnvelopeType.Anonymous {
```
p4u: Remove comented code.

p4u (approved):

p4u:
Finished my review!

p4u:
Some conflicts need to be fixed @lucasmenendez

PR #820, opened 2023-02-16
api: return 204 (No Content) for some endpoints

p4u:
No content (204) is returned only if the queried asset is not found.
For any other error, the bad request 400 code is used.

List of endpoints returning 204:
- /chain/transactions/reference/:hash
- /chain/transactions/reference/index/:globalIndex
- /chain/transactions/:height/:index
- /votes/verify/:electionId/:voteId
- /votes/:voteId

### DIFF
--- a/api/chain.go
+++ b/api/chain.go
@@ -3,6 +3,7 @@ package api
 import (
 	"encoding/hex"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"strconv"
 	"time"
@@ -11,6 +12,7 @@ import (
 	"go.vocdoni.io/dvote/httprouter/apirest"
 	"go.vocdoni.io/dvote/util"
 	"go.vocdoni.io/dvote/vochain"
+	"go.vocdoni.io/dvote/vochain/indexer"
 )
 
 const (
@@ -18,8 +20,7 @@ const (
 )
 
 var (
-	ErrTransactionNotFound = fmt.Errorf("transaction hash not found")
-	ErrBlockNotFound       = fmt.Errorf("block not found")
+	ErrBlockNotFound = fmt.Errorf("block not found")
 )
 
 func (a *API) enableChainHandlers() error {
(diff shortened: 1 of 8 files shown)


marcvelmer (commented):
Why adding the 204 to this endpoints?

/votes/verify/:electionId/:voteId
/votes/:voteId

Here I think it's totally OK to return a 404 - Not found.

p4u:
> Why adding the 204 to this endpoints?
> 
> /votes/verify/:electionId/:voteId /votes/:voteId
> 
> Here I think it's totally OK to return a 404 - Not found.

I think it is good to unify behaviors. It is a potential 404 (not found), yes. But the same rules that apply for transaction/reference should apply on votes. Cause in some point the Client will need to check for the existence of a recently casted vote, querying by nullifier instead of transactionHash. So IMHO 204 is also useful here.

marcvelmer:
> > Why adding the 204 to this endpoints?
> > /votes/verify/:electionId/:voteId /votes/:voteId
> > Here I think it's totally OK to return a 404 - Not found.
> 
> I think it is good to unify behaviors. It is a potential 404 (not found), yes. But the same rules that apply for transaction/reference should apply on votes. Cause in some point the Client will need to check for the existence of a recently casted vote, querying by nullifier instead of transactionHash. So IMHO 204 is also useful here.

But if i'm asking for a voteId that doesn't exists (and will never exists) the return should be 404 IMHO. I would just use the 204 for the transactions as a special case as discussed yesterday.

p4u:
Updated commit, please @marcvelmer check it now

```
No content (204) is returned on the transaction handlers in
case the transaction is not (yet) found in the blockstore.

Not found (404) is returned on the /votes and /elections
endpoints to indicate the vote or election is not found.

For any other error, the bad request 400 code is used.

List of endpoints returning 204:
- /chain/transactions/reference/:hash
- /chain/transactions/reference/index/:globalIndex
- /chain/transactions/:height/:index

List of endpoints returning 404:
- /votes/verify/:electionId/:voteId
- /votes/:voteId
- /elections/:electionId
- /elections/:electionId/keys
- /elections/:electionId/votes
- /elections/:electionId/scrutiny
```

PR #823, opened 2023-02-22
api: return error codes along with error messages

altergui:
this is a further improvement to address #802

i'm starting here with the smallest possible PoC, that accomplishes the following

```
$ curl 'http://127.0.0.1:9090/v2/elections/c5d2460186f7d5d7fe684134f063b699719de9aaf2a308f7ad58020000000000'  -v
*   Trying 127.0.0.1:9090...
* Connected to 127.0.0.1 (127.0.0.1) port 9090 (#0)
> GET /v2/elections/c5d2460186f7d5d7fe684134f063b699719de9aaf2a308f7ad58020000000000 HTTP/1.1
> Host: 127.0.0.1:9090
> User-Agent: curl/7.74.0
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 404 Not Found
< Content-Length: 43
< Content-Type: application/json
< Vary: Origin
< Date: Wed, 22 Feb 2023 16:03:33 GMT
< 
{"error":"election not found","code":4045}
```


httprouter/apirest/errors.go
```diff
@@ -0,0 +1,129 @@
...
+)
+
+func (e APIError) Error() string {
+	return e.String()
+}
+
+func (e APIError) HTTPStatusCode() int {
+	switch e {
+	case ErrElectionNotFound:
+		return http.StatusNotFound
```
altergui: here we would add all the cases we want to treat specially (i.e. not return a simple 400, but a 404 or 204)

httprouter/apirest/apirest.go
```diff
@@ -142,6 +143,15 @@ func (a *API) RegisterMethod(pattern, HTTPmethod string,
...
+				err := msg.Context.Send(apierr.JSON(), apierr.HTTPStatusCode())
+				if err != nil {
+					log.Warn(err)
+				}
+				return
+			}
+
 			// catch some specific errors to return the HTTP status code
 			if errors.Is(err, httprouter.ErrNotFound) {
 				if err := msg.Context.Send(nil, HTTPstatusCodeNotFound); err != nil {
```
altergui: we could delete these two blocks `if errors.Is() {}`, since this PR is much more flexible, and returns (along with the desired http status code) a JSON body, instead of nil like these two blocks do

altergui:
i couldnt `import api` from package apirest (circular import), so i just copypasted errors.go into apirest, and referenced apirest.ErrElectionNotFound from api

i see a couple ways to properly solve this:
* move errors.go into `apirest` and change all references from api
* merge apirest package into api (so that apirest.API struct becomes api.API)
* move errors.go into it's own package `errors` and `import errors` from api and apirest

i would love opinions/discussions, in particular about merging apirest into api, since after a quick read of apirest, i had the feeling it would make a lot of sense to be part of api package, but for sure there's a reason they are two separate packages.
so i'd resort to the first option (move errors.go into apirest), if it makes sense to have those errors be part of apirest package.

only else, i'd consider moving everything into a yet another (independent) `errors` package

httprouter/apirest/apirest.go
altergui: to be clear, i'd reimplement in this PR the same behaviour done in https://github.com/vocdoni/vocdoni-node/pull/820 but with more extensibility, and obviously keep the 204 / 404 codes exactly as described in https://github.com/vocdoni/vocdoni-node/pull/820#issuecomment-1435219318

altergui:
this is ready for review

as always, it can be further improved, but it is reviewable/mergeable as-is, basically it keeps current behaviour but in a more extensible way

p4u:
@altergui I think the approach can be improved. Please, consider the following and let's discuss it if you want.

1. the package apirest (the handler of the httpRouter that manages the HTTP replies) might implement a type:
```golang
type APIerror struct {
  Message string
  ReturnCode int
}
```
2. the API package contains a file named errors.go with the custom errors definitions
```golang
var(
 ErrPageInvalid apirest.APIerror { Message: "page is invalid", ReturnCode: 400 }
 ...
 )
 ```
 3. Modify the apirest package so the endpoint func handlers return the type &APIerror instead of plain error.
 4. The apirest package would format the error message and return the status code if &APIerror not nil
 
 What's the issue with this approach? Is there something I'm missing?
 
Also I'd ONLY use this error wrapper for errors that are returned by the API endpoints, not when doing other internal stuff (such as calling to NewAPI or EnableHandlers).

Please, let me know your thoughts regarding this approach. I think this is what we already discussed by call.

**Comment by altergui** [comment]
SGTM, i'll drop the current commits and propose a new refactor

the only limitation i see in your approach is that we can't reference api.ErrPageInvalid inside package `apirest` (because api imports apirest already) for example to match errors.Is() in some logic inside apirest. But this is not a problem since we're not doing that.

on another note, making "endpoint func handlers return the type &APIerror" will change hundreds of lines that currently simply `return err`, that's why i resorted to typecasting. but no problem, i can make a simple wrapper method to be able to `return apirest.APIError(err)` and replace all returns

**Comment by p4u** [comment]
Let me add some clarity on the layers that we have reagarding the HTTP stack, and maybe it will help to improve the final design of the error handling for the API.

1. httpRouter => the basic HTTP layer handling requests and responses. It accepts handlers to define the data sets to exchange over HTTP and its rules.
2. apirest => a handler of the httpRouter that provides the functionallities for creating an API REST with bearer token authorization support
3. api => the final endpoints registered in apirest. Other packages rather than api can also register handlers to apirest.

The package handling the error messages is `apirest` (the one that abstracts an error type to an HTTP specific response). So the the APIerror type needs to be defined here.

However the error messages and return code, must be defined in a upper layer, so the `api` package or any other using the `apirest` handler.

**Comment by p4u** [comment]
As an example, another handler (also in the second layer) is the jsonRPC, that would not use REST but JSON messages and signatures as authentication mechanism.

Thus this architecture allows us to use two HTTP APIs with a single router:

```
                ___ jsonRPC ---> RPC handlers defined by multiple packages
httpRouter -|___ apiRest  ---> API endpoints defined by multiple packages
```

**Comment by altergui** [comment]
done, reimplemented using a struct. do note i added a field for a unique Code (4001, 4002, etc) in addition to HTTPstatus (mostly 400/500 with some 404, 204) since that's what i understood in that original conversation. i hope i got it right, feedback welcome.

i want to work further, for example i 100% agree with "Also I'd ONLY use this error wrapper for errors that are returned by the API endpoints, not when doing other internal stuff (such as calling to NewAPI or EnableHandlers)." but i need to review this in the whole package.

also i did not make endpoint func handlers return the type &APIerror, i'm still in doubt if it really worth it vs typecasting. i guess the answer will become evident during the process.

File: httprouter/apirest/apirest.go [hunk 3]
```diff
@@ -57,6 +57,33 @@ type APIhandler = func(*APIdata, *httprouter.HTTPContext) error
...
+}
+
+// Error returns the Message contained inside the APIerror
+func (e APIerror) Error() string {
+	return e.Message
+}
+
+// JSON returns a serialized ErrorMsg using APIerror.Error() and APIerror.Code
+func (e APIerror) JSON() []byte {
+	data, err := json.Marshal(&ErrorMsg{
```
p4u: No need to use a special struct `ErrorMsg` you can use APIerror if you declare it this way:

```go
type APIerror struct {
	Message    string `json:"error"`
	Code       int `json:"code"`
	HTTPstatus int `json:"-"`
}
```

p4u: No need to use a special struct `ErrorMsg` you can use APIerror if you declare it this way:

```go
type APIerror struct {
	Message    string `json:"error"`
	Code       int `json:"code"`
	HTTPstatus int `json:"-"`
}
```

httprouter/apirest/apirest.go
```diff
@@ -57,6 +57,33 @@ type APIhandler = func(*APIdata, *httprouter.HTTPContext) error
...
+	HTTPstatus int
+}
+
+// Error returns the Message contained inside the APIerror
+func (e APIerror) Error() string {
+	return e.Message
+}
+
+// JSON returns a serialized ErrorMsg using APIerror.Error() and APIerror.Code
+func (e APIerror) JSON() []byte {
```
p4u: This method should be named `MarshalJSON`

httprouter/apirest/apirest.go
```diff
@@ -142,6 +169,15 @@ func (a *API) RegisterMethod(pattern, HTTPmethod string,
 	routerHandler := func(msg httprouter.Message) {
 		bsaMsg := msg.Data.(*APIdata)
 		if err := handler(bsaMsg, msg.Context); err != nil {
+			// if err is an APIError, use those properties
+			if apierror, ok := err.(APIerror); ok {
+				err := msg.Context.Send(apierror.JSON(), apierror.HTTPstatus)
```
p4u: I'd reverse the functionallity.

The APIerror type might have a method named `Send` that takes the `msg.Context` as a parameter.
Then you can call `apierror.Send(msg.Context)` directly, which is nicer.

httprouter/apirest/apirest.go
p4u: Yes!

p4u (approved):

httprouter/apirest/apirest.go
altergui: lovely <3

altergui:
i did improve this further before merging.  on one hand:

> i want to work further, for example i 100% agree with "Also I'd ONLY use this error wrapper for errors that are returned by the API endpoints, not when doing other internal stuff (such as calling to NewAPI or EnableHandlers)." but i need to review this in the whole package.

done.

> 
> also i did not make endpoint func handlers return the type &APIerror, i'm still in doubt if it really worth it vs typecasting. i guess the answer will become evident during the process.

i left the typecasting. otherwise i'd mess with almost 200 lines of `return err` pointlessly, since they would be reported as generic errors anyway, pretty pointless. Now, if they happen, they generate "500 Internal Server Error" with the plaintext error as body.

another notable change is that i spotted an inconsistency: census handlers returned 400 "census not found in the local database" (unlike 404 from "election not found" and such). they now return ErrCensusNotFound (404) instead

httprouter/apirest/apirest.go
altergui: loooooove this idea, everything ended much prettier (and dropped the JSON / MarshalJSON method), thanks!!!

PR #824, opened 2023-02-23
Bug: wrap api lists into objects

lucasmenendez:
Attempt to fix the [`vocdoni/interoperability` issue #9](https://github.com/vocdoni/interoperability/issues/9), which explains how the API returns list results in different formats: a raw list of items or an object containing a single key with the list as a value.

The aim of this fix is to ensure that the API returns lists in a consistent way, wrapped in an object if the resulting list is not empty, or an empty object if it is.

Only two API handlers are involved in the fix:
* `api.chainTxListPaginated()` (api/chain.go#L331), that handles `/chain/transactions/page/<pageNumber>` endpoint.
* `api.electionFullListHandler()` (api/election.go#L119), that handles `/elections` endpoint.
* `api.electionVotesHandler()` (api/election.go#L301), that handles `elections/<electionID>/votes` and `elections/<electionID>/votes/page/<page>` endpoints.

Neither case has a struct for a list of results returned by the API handlers, so the implemented solution wraps the results in a inline defined struct with the key expected before marshalling. This solution was implemented in other API endpoints such as `/elections/<electionID>/votes/count`

Another possible solution would be to create two structs to wrap this particular type of item for both cases, such as a list of ElectionSummary's and a list of TxReference's.

### DIFF
--- a/api/chain.go
+++ b/api/chain.go
@@ -14,6 +14,7 @@ import (
 	"go.vocdoni.io/dvote/util"
 	"go.vocdoni.io/dvote/vochain"
 	"go.vocdoni.io/dvote/vochain/indexer"
+	"go.vocdoni.io/dvote/vochain/indexer/indexertypes"
 )
 
 const (
@@ -317,14 +318,11 @@ func (a *API) chainTxListPaginated(msg *apirest.APIdata, ctx *httprouter.HTTPCon
 	if err != nil {
 		return err
 	}
-	// wrap list in a map to consistently return list in a object, return empty
+	// wrap list in a struct to consistently return list in a object, return empty
 	// object if the list does not contains any result
-	res := map[string]interface{}{}
-	if len(refs) > 0 {
-		res["transactions"] = refs
-	}
-
-	data, err := json.Marshal(res)
+	data, err := json.Marshal(struct {
+		Txs []*indexertypes.TxReference `json:"transactions,omitempty"`
+	}{refs})
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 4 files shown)


lucasmenendez:
Mark as draft because other endpoints need code review

p4u (approved):

lucasmenendez:
After some team discussion, we decide that the expected behaviour is that the API endpoints that return a list of items, if the list is empty, will still to return an object with the expected key and an empty list, instead of an empty object.

To implement this behaviour consistently, I will review all the API endpoints that return a list of items. The following endpoints will need to be reviewed:

- [x] `/accounts/<accountID>/transfers/page/<page>`
- [x] `/chain/organizations/page/<page>`
- [x] `/chain/transactions/page/<page>`
- [x] `/chain/validators`
- [x] `/elections` & `/elections/page/<page>`
- [x] `/elections/<electionID>/votes` & `/elections/<electionID>/votes/page/<page>`

lucasmenendez:
@p4u @altergui Why does the API return a 404 error when no election has been created, instead an empty list? ([here](https://github.com/vocdoni/vocdoni-node/blob/bug/wrap_api_lists_into_objects/api/elections.go#L132))

altergui:
> @p4u @altergui Why does the API return a 404 error when no election has been created, instead an empty list? ([here](https://github.com/vocdoni/vocdoni-node/blob/bug/wrap_api_lists_into_objects/api/elections.go#L132))

git blame points to pau :man_shrugging: 

it is indeed inconsistent with endpoint /elections
maybe something related to paging? like the UI expects a 404 when requesting a page after the last page with actual content

p4u:
Agree that this behavior should be changed. Please feel free to take it.

p4u (approved):

PR #825, opened 2023-02-23
ApiClient WaitUntilElectionStatus helper updated to handle 404 errors

lucasmenendez:
The `HTTPclient.WaitUntilElectionStatus()` helper function locks the current function and waits until the provided election has reached the provided status. This is achieved, by making GET requests to `/elections/<electionID>` API endpoint in an infinite loop, comparing the status of this election with the data received.

If this helper function is used to wait for an election to reach "CREATED" status, the API will return a 404 error without any message until the election reaches it. This case was not supported by the current implementation of this helper, and may be updated.

To handle this error, the error string comparison may be updated.

### DIFF
--- a/apiclient/helpers.go
+++ b/apiclient/helpers.go
@@ -144,7 +144,9 @@ func (c *HTTPclient) WaitUntilElectionStatus(ctx context.Context,
 	for {
 		election, err := c.Election(electionID)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


altergui (approved):
this is a perfect example of why it's so important the error handling refactor we are doing :grin:
i wish we could already match for an error code, but as a hotfix this is perfect

PR #826, opened 2023-02-24
Indexed Census deprecation

lucasmenendez:
After implementing the zkWeighted census, the zkIndexed census must be deprecated and all legacy code removed.

After an initial code review, the following files contain logic about this type of census and will be updated:

- [x] `api/censuses.go`: where this type of census is created.
- [x] `api/census_helpers.go`: which contains some helper functions related with the censuses.
- [x] `censustree/censustree.go`: which wraps arbo trees and database transactions to retrieve and persist censuses.
- [x] `censustree/censustree_test.go`: which contains some unit tests.
- [x] `vochain/state/rollingcensus.go`: which contains some legacy code in a block comment.


censustree/censustree.go
```diff
@@ -18,10 +16,8 @@ import (
 )
 
 var (
-	censusWeightKey         = []byte("censusWeight")
-	censusIndexKey          = []byte("censusIndex")
-	isIndexAsKeysCensus     = []byte("isIndexAsKey")
-	censusKeysToIndexPrefix = []byte("keyIndex")
+	censusWeightKey = []byte("censusWeight")
+	censusIndexKey  = []byte("censusIndex")
```
p4u: Is this DB key necessary now?

vochain/offchaindatahandler/census.go
```diff
@@ -30,30 +26,3 @@ func (d *OffChainDataHandler) enqueueOffchainCensus(root, uri string) {
 	}
 	d.storage.AddToQueue(uri, d.importExternalCensus, true)
 }
-
-// importRollingCensus imports a rolling census (zkIndexed) from a remote URI into the censusDB storage.
-func (d *OffChainDataHandler) importRollingCensus(pid []byte) {
```
p4u, line 35: I'd not remove this method, we will need it for supporting rolling-census voting. Ideally this should be adapted to work without indexed.

vochain/offchaindatahandler/offchaindatahandler.go
```diff
@@ -79,9 +79,6 @@ func (d *OffChainDataHandler) Commit(height uint32) error {
 		case itemTypeElectionMetadata, itemTypeAccountMetadata:
 			log.Infof("importing metadata from %s", item.uri)
 			go d.enqueueMetadata(item.uri)
-		case itemTypeRollingCensus:
```
p4u: Same here, I'd keep this case woring if possible.

p4u (changes requested):

censustree/censustree.go
lucasmenendez: the key and also functions [censustree.updateCensusIndex()](https://github.com/vocdoni/vocdoni-node/blob/feature/indexed_census_deprecation/censustree/censustree.go#L223) and [censustree.GetCensusIndex()](https://github.com/vocdoni/vocdoni-node/blob/feature/indexed_census_deprecation/censustree/censustree.go#L357) no?

vochain/offchaindatahandler/census.go
lucasmenendez: I will restore the changes regarding to rolling census and open an issue to adapt it, ok?

p4u (approved):

PR #831, opened 2023-02-28
make all binaries now report their version (internal.Version)

altergui:
trying to debug an issue with voconed:latest docker image, i found out the only binary reporting its version is `node`.

here i replicate that very useful behaviour in all binaries

### DIFF
--- a/cmd/vochaininspector/inspector.go
+++ b/cmd/vochaininspector/inspector.go
@@ -15,6 +15,7 @@ import (
 	"go.vocdoni.io/dvote/config"
 	"go.vocdoni.io/dvote/db"
 	"go.vocdoni.io/dvote/db/metadb"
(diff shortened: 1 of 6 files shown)


altergui:
i changed all binaries (and not only voconed) since i know my future self will thank me for doing this :angel:

cmd/vochaininspector/inspector.go
```diff
@@ -51,6 +56,7 @@ func main() {
 
 	flag.Parse()
 	log.Init(logLevel, "stdout")
+	log.Infof("starting %s version %q", filepath.Base(os.Args[0]), internal.Version)
```
p4u: Infow would look nicer :) Also for previous `Infof()` calls

p4u (approved):

PR #834, opened 2023-03-01
F/validators as keykeepers

p4u:
This PR introduces three major changes:

1. Use secp256k1 EC for block validation (before ed25519). This will bring better Ethereum compatibility and allow us to use the same key for signing transactions.
2. Modify genesis format. The new genesis format is much simpler, it mainly relies on a custom APP state data. The validator set is configured on InitChain().
3. Use validators as keypeepers and add the `key_index` value into the genesis. Each validator has its own key index, no need to specify it by flag or env var as before.


cmd/node/main.go
```diff
@@ -488,6 +486,41 @@ func main() {
...
+	//
+	// Validator
+	//
+	if globalCfg.Mode == types.ModeMiner {
+		// create the key for the validator used to sign transactions
+		signer := ethereum.SignKeys{}
+		if err := signer.AddHexKey(globalCfg.Vochain.MinerKey); err != nil {
+			log.Fatal(err)
+		}
+		validator, err := srv.App.State.Validator(signer.Address(), true)
```
jordipainan, line 498: I have doubts about this part.
Imagine a validator is set via consensus at block 100. When the new validator starts the syncing process from scratch, L#503 will be executed and the Keykeeper will not be initialized.
The validator could keep syncing until, at block 100, acknowledges that is a validator, but the node must be restarted to enable the Keykeeper.

vochain/apputils.go
```diff
@@ -62,6 +63,7 @@ func NewNodeKey(tmPrivKey, nodeKeyFilePath string) (*tmtypes.NodeKey, error) {
 	if err != nil {
 		return nodeKey, fmt.Errorf("cannot decode private key: (%s)", err)
 	}
+	// We need to use ed25519 curve for node key since tendermint does not support secp256k1
```
jordipainan, line 66: Do we want to add a TODO in this part? It is somewhat for me strange that a validator specifies a `secp256k1` key but internally uses a `ed25519`. I feel that is a bit hacky and for consistency, the same crypto should be used.

jordipainan (commented):
Made some comments, not approving for the moment, as I need to revisit more in depth the logic. But if you can answer the ones already made, it will be helpful :)

vochain/types.go
```diff
@@ -186,3 +185,11 @@ type TendermintPubKey struct {
 	Type  string `json:"type"`
 	Value []byte `json:"value"`
 }
+
+type AppStateValidators struct {
+	Address  types.HexBytes `json:"signer_address"`
+	PubKey   types.HexBytes `json:"consensus_pub_key"`
+	Power    uint64         `json:"power"`
+	Name     string         `json:"name"`
+	KeyIndex uint8          `json:"key_index"`
```
jordipainan, line 194: Must `KeyIndex` always be incremental? Can an existing validator change its `KeyIndex` (by validator self tx or by other state change)?

vochain/apputils.go
p4u: This is a limitation on the Tendermint side. Nothing to do.

vochain/types.go
p4u: No, it cannot be changed for now.

cmd/node/main.go
p4u: That's a good observation. 
However, we don't have currently the transaction logic implemented for adding new validators. Thus, this is not a real problem right now.
When we implement the validator add/del logic, we will need to enable a mechanism to trigger actions on the validator to start processing blocks. But that's not for now.

jordipainan (approved):

PR #835, opened 2023-03-02
Feature/allow custom weight on voting

### LINKED ISSUE #827: Allow custom weight on voting
Currently, the zkSnarks circuit allows as input a weight. This weight is the leaf of the merkle-tree uses as Census. But if using very specific weights, voting anonymously can be an issue, since the weight can be correlated. 

A quick solution would be to add the following modification:

- The circuit takes two `weight` parameters: `factoryWeight` and `votingWeight`. The first is private, while the second is public.
- Internally, the circuit must check: `votingWithg <= factoryWeight`
- For checking the validity of the merkle-proof, the `factoryWeight` is used.

This way, the voter is free to select the weight used when casting the vote. While the census, tree weight is kept hidden as private input.

This modification partially solves the privacy issue while also adds an interesting new feature: now a user can decide to vote with less weight, thus not influencing the result too much in case of a whale.

The same functionality should also be implemented for non-anonymous elections.

A new flag parameter should be added to the protobuf models to allow enabling/disabling this feature `bool allowCustomWeight`.

lucasmenendez:
Original #827.

This feature allows a voter to use a different weight to vote, if that weight (votingWeight) is less than or equal to the weight used by the election creator to register that voter (factoryWeight). 
This feature helps to increase voter anonymity by avoiding possible correlations between the registered weight and the weight used, and also allows a voter to reduce the influence of his vote.

Initially we decided to introduce a new flag in the census definition to enable or disable this behaviour, something like CustomWeight boolean. If this flag is set to false, all these checks will be skipped. The introduction of this flag is already under discussion.


lucasmenendez:
I got a question for you @p4u...

To ensure that the `votingWeight` is less than or equal to the `factoryWeight` during the censusProof generation, this `votingWeight` must be provided by the voter.

Actually, the API endpoint for the censusProof is handled on the GET http method, so we have two ways to get the `votingWeight`:
- Change this API endpoint to POST http method and request it in the body.
- Request it as a URL query parameter.

What do you think is better?

I think it is more idiomatic to keep the GET method for this endpoint but i don't like using URL query parameters for this use case.

lucasmenendez:
Significant changes included:
* ApiClient:
  * Snark proof generation: moved to the `HTTPclient.Vote()` method, because it is comparable to the transaction sign.
  * `HTTPclient.Vote()`: transaction vote skipped on anonymous voting.
  * `CensusProofZk` struct & `HTTPclient.CensusGenProofZk()`: Deleted, they are no longer needed.
  * `CensusProof` struct: now includes the census siblings of the associated key, used on ZkCensusProof generation.
* Vochain:
  * `VerifyProofOffChainTree()` on `vochain/transaction/proof.go`: supports custom weight on signed weighted voting.

**This change requires changes to the protobuf models**, read more here: https://github.com/vocdoni/dvote-protobuf/pull/59.

apiclient/census.go
```diff
@@ -1,39 +1,22 @@
...
-// compute the nullifier, the private key of the ZkAddress is required, so it
-// cannot be computed by anyone but the voter.
-type CensusProofZk struct {
-	Proof      types.HexBytes
-	PubSignals types.HexBytes
-	Weight     types.BigInt
-	KeyType    models.ProofArbo_KeyType
-	Nullifier  types.HexBytes
+	Proof    types.HexBytes
+	Value    types.HexBytes
```
p4u: Value and Weight should be renamed to something more self explanatory

apiclient/vote.go
```diff
@@ -26,12 +29,12 @@ import (
...
+	Choices      []int
+	ElectionID   types.HexBytes
+	VotingWeight *big.Int
 
 	ProofMkTree *CensusProof
 	ProofCSP    types.HexBytes
-	ProofZkTree *CensusProofZk
 }
 
 // Vote sends a vote to the Vochain. The vote is a VoteData struct,
```
p4u, line 43: I think we should document on the description of this method the behavior of VotingWeight (how can it be used and what happens if it is nil).

p4u (approved):

p4u:
please also merge with master and fix conflicts

PR #837, opened 2023-03-02
B/api not found

p4u:


vochain/app.go
```diff
@@ -334,10 +339,10 @@ func (app *BaseApplication) GetTx(height uint32, txIndex int32) (*models.SignedT
 func (app *BaseApplication) getTxTendermint(height uint32, txIndex int32) (*models.SignedTx, error) {
 	block := app.GetBlockByHeight(int64(height))
 	if block == nil {
-		return nil, fmt.Errorf("unable to get block by height: %d", height)
```
altergui, line 337: really? why not expose these "internal" errors (AFAIU) to the consumer and let it decide whether to treat that as "tx not found"?

api/elections.go
```diff
@@ -586,3 +584,19 @@ func (a *API) computeCidHandler(msg *apirest.APIdata, ctx *httprouter.HTTPContex
 	}
 	return ctx.Send(data, apirest.HTTPstatusOK)
 }
+
+// getElection retrieves an election from the vochain state.
+// Checks if election exists and return 404 if not
+func getElection(electionID []byte, vs *state.State) (*models.Process, error) {
```
altergui, line 590: loved this code dedup, much needed in many places

altergui (approved):
overall looks good to me, my only question is why return "tx not found" when that's not exaaaaaactly true, but i trust your criteria :innocent:

vochain/app.go
p4u: The query is for a transaction identified by a height and an index (on this height). If the height does not exist, IMO is consistent to return "the transaction you were looking for does not exist".

PR #832, opened 2023-02-28
apirest: fix APIerror returns, since fmt.Errorf was not really working

altergui:
now APIerror struct contains an error instead of a string,
this allows error wrapping.

New APIerror methods: With, Withf, WithErr

All `return fmt.Errorf` in api are replaced with `APIerror.With`

(Also includes correcting a small nit for MarshalJSON failures, since they are really server-side and they returned 400 instead of 500)

PR #840, opened 2023-03-03
f/interop/8

jordipainan:
https://github.com/vocdoni/interoperability/issues/8

### DIFF
--- a/api/chain.go
+++ b/api/chain.go
@@ -10,6 +10,7 @@ import (
 	"go.vocdoni.io/dvote/crypto/zk/circuit"
 	"go.vocdoni.io/dvote/httprouter"
 	"go.vocdoni.io/dvote/httprouter/apirest"
+	"go.vocdoni.io/dvote/types"
 	"go.vocdoni.io/dvote/util"
 	"go.vocdoni.io/dvote/vochain"
 	"go.vocdoni.io/dvote/vochain/indexer"
@@ -146,6 +147,14 @@ func (a *API) enableChainHandlers() error {
 		"GET",
 		apirest.MethodAccessTypePublic,
 		a.chainBlockByHashHandler,
+	); err != nil {
+		return err
+	}
+	if err := a.endpoint.RegisterMethod(
+		"/chain/organizations/filter/page/{page}",
+		"POST",
+		apirest.MethodAccessTypePublic,
+		a.chainOrganizationsFilterPaginated,
 	); err != nil {
 		return err
 	}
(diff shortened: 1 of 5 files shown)


api/chain.go
```diff
@@ -467,3 +476,34 @@ func (a *API) chainBlockByHashHandler(msg *apirest.APIdata, ctx *httprouter.HTTP
...
 	return ctx.Send(convertKeysToCamel(data), apirest.HTTPstatusOK)
 }
+
+// POST /chain/organizations/filter/page/<page>
+// returns a list of organizations paginated by the given page
+func (a *API) chainOrganizationsFilterPaginated(msg *apirest.APIdata, ctx *httprouter.HTTPContext) error {
+	// get organizationId from the request body
+	var organizationId string
+	if err := json.Unmarshal(msg.Data, &organizationId); err != nil {
+		return fmt.Errorf("%w: %v", ErrCantParseOrgID, err)
```
altergui: fmt.Errorf doesn't really work, this was my fault. i fixed it yesterday and just merged PR #832
 
example
https://github.com/vocdoni/vocdoni-node/pull/832/commits/a5f01b52c6493f10cdd76207cdcf9b2b323663e0#diff-8a1c97a3c76047a21682fceabcf8e05e4fdb9b8909d5e2209ec1fa626e153466L125-L126

new syntax is much prettier, tho :sunglasses: 

`return ErrCantParseOrgID.WithErr(err)`

altergui: please replace all of the "return fmt.Errorf" inside all *Handler() func

altergui (changes requested):
my code needs fixing, sorry about that

api/elections.go
```diff
@@ -600,3 +610,63 @@ func getElection(electionID []byte, vs *state.State) (*models.Process, error) {
...
+func (a *API) electionFilterPaginatedHandler(msg *apirest.APIdata, ctx *httprouter.HTTPContext) error {
+	// get organizationId from the request body
+	body := struct {
+		OrganizationID types.HexBytes `json:"organizationId"`
+		ElectionID     types.HexBytes `json:"electionId"`
+		WithResults    bool           `json:"withResults"`
+		Status         string         `json:"status"`
+	}{}
+	if err := json.Unmarshal(msg.Data, &body); err != nil {
+		return fmt.Errorf("%w (%x): %v", ErrMarshalingServerJSONFailed, msg.Data, err)
```
altergui: here you would use Withf instead of WithErr

`return ErrMarshalingServerJSONFailed.Withf("(%x): %v", msg.Data, err)`

api/chain.go
```diff
@@ -466,3 +475,34 @@ func (a *API) chainBlockByHashHandler(msg *apirest.APIdata, ctx *httprouter.HTTP
...
+	page := 0
+	if ctx.URLParam("page") != "" {
+		page, err = strconv.Atoi(ctx.URLParam("page"))
+		if err != nil {
+			return ErrCantParsePageNumber.WithErr(err)
+		}
+	}
+	page = page * MaxPageSize
+	// get matching organization ids from the indexer
+	matchingOrganizationIds := a.indexer.EntityList(MaxPageSize, page, organizationId)
```
lucasmenendez, line 499: Other endpoints return 204 http status code (with an error associated to it) when no results are found. I don't know if this is the expected behaviour here.

lucasmenendez: I also have the same doubts about the following endpoints. We have already opened an issue with something similar https://github.com/vocdoni/interoperability/issues/16.

lucasmenendez (commented):
Looks good to me!

api/elections.go
```diff
@@ -623,37 +623,37 @@ func (a *API) electionFilterPaginatedHandler(msg *apirest.APIdata, ctx *httprout
 		Status         string         `json:"status"`
 	}{}
 	if err := json.Unmarshal(msg.Data, &body); err != nil {
-		return fmt.Errorf("cannot unmarshal request body: %w", err)
+		return ErrMarshalingServerJSONFailed.WithErr(err)
```
altergui: this one is most definitely conceptually wrong, should be ErrCantParseDataAsJSON

api/elections.go
```diff
@@ -623,37 +623,37 @@ func (a *API) electionFilterPaginatedHandler(msg *apirest.APIdata, ctx *httprout
...
 	}
 	if len(body.OrganizationID) == 0 {
-		return fmt.Errorf("organizationId is required")
+		return fmt.Errorf("%w", ErrOrgNotFound)
```
altergui: missed this one, should be `return ErrOrgNotFound` (yes, you can simply return the error, without `With` methods)

api/chain.go
```diff
@@ -482,15 +482,15 @@ func (a *API) chainOrganizationsFilterPaginated(msg *apirest.APIdata, ctx *httpr
 	// get organizationId from the request body
 	var organizationId string
 	if err := json.Unmarshal(msg.Data, &organizationId); err != nil {
-		return fmt.Errorf("cannot unmarshal organizationId: %w", err)
+		return ErrCantParseOrgID.WithErr(err)
```
altergui: i think this would be `ErrCantParseDataAsJSON.WithErr(err)`?

altergui (changes requested):
i promise i won't request more changes :see_no_evil:

api/chain.go
jordipainan: Changed the first one pointed out for now

p4u (approved):

PR #841, opened 2023-03-03
api: swagger annotations

altergui:
i nosedived into swagger and i'm in awe. pushing here my first dirty steps and will continue next week


altergui:
for context: since the first time i read `api` code, i found the `// /chain/organizations/count` comments absolutely useful and at the same time i was bothered about the inconsistencies (sometimes it mentions `POST` or `GET`, there are typos sometimes, etc) so for weeks i've been dreaming of preparing a PR to correct all those comment and make them uniform. so here i am, maybe this will also be useful somehow for API docs

p4u:
Very nice @altergui, I love you are taking care of these kind of things by your own. Thanks!

p4u (approved):

PR #849, opened 2023-03-08
Feature: pubkey as censustree key deprecation

### LINKED ISSUE #819: Deprecate PubKey as censustree leaf key to decrease the default number of levels of Arbo
We actually support two use cases. The old one, which uses a public key (32 bytes) as a leaf key on the census tree; and the new one, which uses the ZkAddress (20 bytes) as a leaf key. 
The number of levels of a Merkle tree (in this case the census tree), determines not only the number of leaves that the tree can store (2^nLevels), but also the length (in bytes) of the key of each leaf (nLevels/8).
To support keys of 32 bytes we need define the censustree with 256 levels, which is a very large number of levels, supporting up to 1,15*10^77 leaves. We need to adapt these keys to smaller length to reduce the number of levels.

lucasmenendez:
The goal of this PR is to reduce the number of levels in the census trees to 160 and to support census tree keys of up to 20 bytes, deprecating the use of publicKeys as census tree keys, thus attempting to fix issue #819.

To achieve this goal, I propose to create a new constant that defines the maximum length of any census key to 20 bytes, and to include some checks on the backend to ensure that longer keys are supported by truncating them.

### DIFF
--- a/api/censuses.go
+++ b/api/censuses.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/google/uuid"
 	"go.vocdoni.io/dvote/api/censusdb"
+	"go.vocdoni.io/dvote/censustree"
 	"go.vocdoni.io/dvote/data/compressor"
 	"go.vocdoni.io/dvote/httprouter"
 	"go.vocdoni.io/dvote/httprouter/apirest"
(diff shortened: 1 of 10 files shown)


p4u:
I don't see a commit on go.mod, so the protobuf models are not being upgraded.

p4u (approved):

PR #850, opened 2023-03-10
fix/api/filters

jordipainan:


api/elections.go
```diff
@@ -640,16 +640,17 @@ func getElection(electionID []byte, vs *state.State) (*models.Process, error) {
 func (a *API) electionFilterPaginatedHandler(msg *apirest.APIdata, ctx *httprouter.HTTPContext) error {
 	// get organizationId from the request body
 	body := struct {
-		OrganizationID types.HexBytes `json:"organizationId"`
-		ElectionID     types.HexBytes `json:"electionId"`
-		WithResults    bool           `json:"withResults"`
-		Status         string         `json:"status"`
+		OrganizationID types.HexBytes `json:"organizationId,omitempty"`
```
p4u: I think this might be an abuse of struct types defined inside a function scope. IMO that makes sense for a very simple case (such as returning a json body with a single `Count` field). But for defining bigger structs, we have the api_types.go file.

api/elections.go
```diff
@@ -640,16 +640,17 @@ func getElection(electionID []byte, vs *state.State) (*models.Process, error) {
...
+		WithResults    *bool          `json:"withResults,omitempty"`
+		Status         string         `json:"status,omitempty"`
 	}{}
 	if err := json.Unmarshal(msg.Data, &body); err != nil {
 		return ErrCantParseDataAsJSON.WithErr(err)
 	}
-	if len(body.OrganizationID) == 0 {
-		return ErrOrgNotFound
+	// check that at least one filter is set
+	if len(body.OrganizationID) == 0 && len(body.ElectionID) == 0 && body.WithResults == nil && body.Status == "" {
```
p4u: It should be fine to compare HexBytes to `nil`

api/elections.go
```diff
@@ -661,8 +662,12 @@ func (a *API) electionFilterPaginatedHandler(msg *apirest.APIdata, ctx *httprout
 		}
 	}
 	page = page * MaxPageSize
+
 	// get election list
-	elections, err := a.indexer.ProcessList(body.OrganizationID, page, MaxPageSize, body.ElectionID.String(), 0, 0, body.Status, body.WithResults)
+	if body.WithResults == nil {
+		body.WithResults = new(bool)
```
p4u: If you need a default value of `false` for WithResults, you might simply declare it as a non pointer variable, you will get default false for free :)

api/elections.go
```diff
@@ -661,8 +662,12 @@ func (a *API) electionFilterPaginatedHandler(msg *apirest.APIdata, ctx *httprout
...
+	}
+	elections, err := a.indexer.ProcessList(body.OrganizationID, page, MaxPageSize, body.ElectionID.String(), 0, 0, body.Status, *body.WithResults)
```
p4u: This is a 144 chars line, it should be broken. Please try not to surpass the 120.

p4u (changes requested):

p4u:
It would be nice (and necessary) to have tests for this code, you can add them here https://github.com/vocdoni/vocdoni-node/blob/master/test/api_test.go

Moreover, please join commits or squash all of them in a single one when merging.

p4u:
hey @jordipainan I don't see the tests. Did you forget to include the files when commit?

PR #853, opened 2023-03-13
Collect coverage of both unit and integration tests

altergui:
now that we have go 1.20 i want to finally see the coverage of our integration test (https://go.dev/testing/coverage/)
ideally, integrate both unit and integration coverage (as in https://dustinspecker.com/posts/go-combined-unit-integration-code-coverage/)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -80,16 +80,16 @@ jobs:
           go-version: '1.20'
       - name: Run Go test (and collect code coverage)
         if: github.ref != 'refs/heads/master' && github.ref != 'refs/heads/stage' && !startsWith(github.ref, 'refs/heads/release') && github.ref != 'refs/heads/dev' && github.ref != 'refs/heads/apiv2'
-        run: go test -coverprofile=unit.covprofile ./...
+        run: go test -coverprofile=unit.covdata.txt ./...
       - name: Run Go test race
         if: github.ref == 'refs/heads/master' || github.ref == 'refs/heads/stage' || startsWith(github.ref, 'refs/heads/release') || github.ref == 'refs/heads/dev' || github.ref == 'refs/heads/apiv2'
         env:
           LOG_PANIC_ON_INVALIDCHARS: true # check that log lines contains no invalid chars (evidence of format mismatch)
         run: go test -vet=off -timeout=15m -race ./... # note that -race can easily make the crypto stuff 10x slower
       - name: Send unit test coverage to coveralls.io
-        uses: coverallsapp/github-action@v2
+        uses: shogo82148/actions-goveralls@v1
         with:
-          file: unit.covprofile
+          path-to-profile: unit.covdata.txt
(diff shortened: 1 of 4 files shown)


altergui:
hey @chilcano the github-action of coveralls works ok on github runners
https://github.com/vocdoni/vocdoni-node/actions/runs/4409107548/jobs/7724980858#step:7:10

but fails on our self-hosted
https://github.com/vocdoni/vocdoni-node/actions/runs/4409107548/jobs/7724981170#step:5:11

```
  curl -sL https://github.com/coverallsapp/coverage-reporter/releases/latest/download/coveralls-linux.tar.gz | tar xz -C /usr/local/bin
  shell: /usr/bin/bash --noprofile --norc -e -o pipefail {0}
tar: coveralls: Cannot open: Permission denied
tar: Exiting with failure status due to previous errors
```

that `tar xz -C /usr/local/bin` is unfortunately hardcoded :cry: 
https://github.com/coverallsapp/github-action/blob/release/v2/action.yml#L64

any ideas? i don't have access to ci2
i guess the github-runner in our self-hosted is running as a non-root user, and in github hosted runners it runs as root?

alternatives i can think of:
* fork coverallsapp/github-action to make it write to /tmp instead of /usr/local/bin
* don't use coverallsapp/github-action at all and instead reproduce the same behaviour with a `run` step, that does the `curl | tar xz -C /tmp`

altergui:
on another note, the coverage calculation is absurdly precise but flaky (consecutive builds of the exact same code give slightly different results) so half of the time it gives a false negative as in:
"Coverage decreased (-0.008%) to 41.482"
which blocks merging since it's a required check :facepalm: 

is there any way to configure this, repo-wide? like modify what is considered a significant percentage of decrease to consider a "fail", or, if that's not possible, make the check NOT a blocker, just informative

altergui:
also, it looks like coveralls should add comments to PRs, but that's not happening i guess

https://github.com/marketplace/actions/coveralls-github-action#coveralls-comments-arent-added-to-my-pull-request

Ensure that:

  1.  Your workflow invokes the Coveralls action runs on pull requests, e.g.: on: ["push", "pull_request"]

  2.  You have invited the [@coveralls](https://github.com/coveralls) user to your repository as a collaborator
  3.  You have enabled the "LEAVE COMMENTS?" setting in the "Pull Request Alerts" area in your Repo Settings page inside the Coveralls app.
  
  
  point 1 is OK, i'm not sure about 2 and 3

altergui:
> don't use coverallsapp/github-action at all and instead reproduce the same behaviour with a `run` step, that does the `curl | tar xz -C /tmp`

i did this and it's working, i say this PR is ready for review/merge as-is

well, except for the other issue
> is there any way to configure this, repo-wide? like modify what is considered a significant percentage of decrease to consider a "fail", or, if that's not possible, make the check NOT a blocker, just informative"

p4u (approved):

PR #854, opened 2023-03-16
vochain: LoadZkCircuit now downloads zkCircuits into ~/.cache/vocdoni/

altergui:
until now, it would use ./zkCircuits, being relative this would create many copies, pollute the git repo while doing `go test`, and redownload the same files many times
    
`node` binary still puts them into -dataDir

also, i included a trivial refactor for idiomatic code, unrelated to the main objective of this PR, but not big enough to warrant a PR


altergui:
example behaviour until now, after running `go test ./...` and then `git status`

```
Untracked files:
  (use "git add <file>..." to include in what will be committed)
	api/zkCircuits/
	test/zkCircuits/
	vochain/indexer/zkCircuits/
	vochain/zkCircuits/
	vocone/zkCircuits/

```

altergui:
this also saves tens of seconds on each run, since that's the time it took to download the ~30Mb files on my internet connection

lucasmenendez (approved):
Thanks for the improvements and sorry for the mistakes!

PR #855, opened 2023-03-17
Use batch transaction on indexer

mariajdab:


mariajdab:
When you have some chance @p4u @mvdan @lucasmenendez @jordipainan @altergui Could you please check this approach? And if it's possible, provide some feedback. Thanks!

vochain/indexer/indexer.go
```diff
@@ -62,6 +62,7 @@ type Indexer struct {
 	voteIndexPool []*VoteWithIndex
 	// votePool is the list of votes that should be live counted, grouped by processId
 	votePool map[string][]*state.Vote
+	blockTx  *sql.Tx
 	// newProcessPool is the list of new process IDs on the current block
 	newProcessPool []*indexertypes.IndexerOnProcessData
```
mvdan: should be deleted now

vochain/indexer/indexer.go
```diff
@@ -329,26 +330,6 @@ func (idx *Indexer) AfterSyncBootstrap() {
...
-	}
-
-	// Update existing processes
-	for _, p := range idx.updateProcessPool {
-		if err := idx.updateProcess(p); err != nil {
-			log.Errorw(err, "commit: cannot update process")
-			continue
-		}
-		log.Debugw("updated process", "processID", hex.EncodeToString(p))
-	}
```
mvdan: the sql transaction at blockTx should be committed here, in the Commit method, unless I'm missing something about your approach. if you commit every time you call newEmptyProcess, then we're writing to the DB before the entire block is committed here.

vochain/indexer/process.go
```diff
@@ -394,13 +402,22 @@ func (s *Indexer) setResultsHeight(pid []byte, height uint32) error {
...
-		ResultsHeight: int64(height),
+
+	if err := s.ExecTx(ctx, func(q *indexerdb.Queries) error {
+		if _, err := q.SetProcessResultsHeight(ctx, indexerdb.SetProcessResultsHeightParams{
+			ID:            pid,
+			ResultsHeight: int64(height),
+		}); err != nil {
+			return err
+		}
+		return nil
```
mvdan: does this change help in any way?

there is no need to modify the other queries; I would focus on just newProcessPool. the only query to update is newEmptyProcess.

vochain/indexer/indexer.go
```diff
@@ -493,8 +493,13 @@ func (idx *Indexer) Rollback() {
 func (idx *Indexer) OnProcess(pid, eid []byte, censusRoot, censusURI string, txIndex int32) {
 	idx.lockPool.Lock()
 	defer idx.lockPool.Unlock()
-	data := &indexertypes.IndexerOnProcessData{EntityID: eid, ProcessID: pid}
-	idx.newProcessPool = append(idx.newProcessPool, data)
+	if err := idx.newEmptyProcess(pid); err != nil {
```
mvdan, line 493: this one _does_ need to use `blockTx`, otherwise we're writing the process to the database right away without waiting for the call to `Indexer.Commit`.

mvdan (commented):
remember that the whole point of blockTx is to only commit changes to the database when `Indexer.Commit` is called. otherwise we wouldn't need the pool slices nor the sql transactions at all.

p4u:
Hey.
I will wait to fix Dani's comments before jumping in.

By the way, here is the document we prepared for your challenge, Maria. There are some basic explanations about the Indexer that might be nice to refresh.

https://hackmd.io/@p4u/SJUZ_Tios

vochain/indexer/indexer.go
```diff
@@ -514,14 +519,20 @@ func (idx *Indexer) OnVote(v *state.Vote, txIndex int32) {
...
 	idx.lockPool.Lock()
 	defer idx.lockPool.Unlock()
-	idx.updateProcessPool = append(idx.updateProcessPool, pid)
+	if err := idx.updateProcess(pid); err != nil {
+		log.Errorw(err, "commit: cannot update process")
+	}
+	log.Debugw("updated process", "processID", hex.EncodeToString(pid))
 }
 
 // OnProcessKeys does nothing
```
altergui: this func doc is not true anymore, is it?

p4u:
@mariajdab either you close the issues opened by the comments or reply asking for a new review on this matter.

vochain/indexer/indexer.go
mariajdab: I updated the code, now the `blockTx` is committed in the Commit method. But the `idx.blockTx.Rollback()` I think could not be called inside the Indexer Rollback method, because the Rollback method is always called in the code flow and if the Tx committed was successful, calling the Rollback will return the error: `Transaction is already committed or rollback`.

vochain/indexer/indexer.go
mariajdab: Removed

PR #860, opened 2023-03-20
rename util.CreateEthRandomKeysBatch -> ethereum.NewSignKeysBatch

altergui:
ethereum package has NewSignKeys, and util.CreateEthRandomKeysBatch was just a batch of NewSignKeys, so it makes more sense to have it in the same package

deduplicate trimHex as well, use util.TrimHex


altergui:
(totally unsolicited PR, yes, i just came across this while working on something else)

p4u (approved):

PR #873, opened 2023-04-04
end2endtest: new voteOverwriteTest()

altergui:
added overwrite end2endtest that test two cases using the `sameBlock` and using the `nexBlock`
added `sendVote` helper to reuse code.


mariajdab:
Seems like working with `CensusTypeWeighted` with keyType of  `ProofArbo_ADDRESS` and trying to overwrite a vote more times that the `maxVoteOverwrite` allows actually not race the error expected. Checking the unit test for `maxVoteOverwrite`, it is using `ProofArbo_PUBKEY`

p4u:
At some point this PR should be adapted to the new structure @altergui is creating for the end2end test package.

altergui:
i removed the step "Check the election status is actually ENDED" since it's redundant.

mariajdab:
> i removed the step "Check the election status is actually ENDED" since it's redundant.

It's getting an occasional error: `never published results: context deadline exceeded`, so I think is needed to add more time in `time.Second*60` I will update it

cmd/end2endtest/overwritenotwait.go
```diff
@@ -18,7 +16,7 @@ import (
 
 func init() {
 	ops["overwritenotwaitnextblock"] = operation{
-		test:        &E2EAnonElection{},
+		test:        &E2EOverwriteNotWaitElection{},
```
altergui: i'd suggest naming these tests
"VoteOverwriteInSameBlock" and "VoteOverwriteInNextBlock" (E2E...Election)

BUT, wouldn't even be simpler to merge both tests (SameBlock and NextBlock) into overwrite.go?
i mean during the E2EOverwriteElection (or E2EVoteOverwriteElection), say, make t.voterAccounts[0] overwrite in the same block and t.voterAccounts[1] in next block.

maybe there's a reason to keep the tests separate, but if not, i believe merging them makes sense

cmd/end2endtest/overwritenotwait.go
```diff
@@ -65,45 +63,20 @@ func (t *E2EOverwriteNotWaitElection) Run() error {
...
-				delete(accountsMap, i)
 
-			}
-			if len(accountsMap) == 0 {
-				log.Infof("sent %d/%d votes... got %d HTTP errors", votesSent, len(accounts), contextDeadlines)
-				time.Sleep(time.Second * 2)
+		votesSent := 0
+		contextDeadlines := 0
+		for _, acc := range accounts {
+			ctxDeadline, err := t.sentVote(acc, []int{0}, nil)
```
altergui: i didn't catch this in the previous commits (it's my first review) but i think this method should be "sendVote"?

cmd/end2endtest/helpers.go
```diff
@@ -0,0 +1,402 @@
...
+						Title: map[string]string{"default": "No"},
+						Value: 1,
+					},
+				},
+			},
+		},
+	}
+}
+
+func (t e2eElection) createAccount(address string) (*vapi.Account, error) {
```
altergui: i think i'm nitpicking on sugar here, but:
 all methods are `(t e2eElection)` except setupElection which is `(t *e2eElection)`
  i'd change all to be  `(t *e2eElection)`, consistent with the other E2E structs

mariajdab: Yeah, only `setupElection` uses pointer because for now it was the only method needed that assigns struct `e2eElection` values, but to be more consistent I'll update all methods to be `(t *e2eElection)`

cmd/end2endtest/overwritenotwait.go
mariajdab: Right! Thanks!

cmd/end2endtest/helpers.go
altergui: peeerfect, as idiomatic as can be :partying_face: 

https://go.dev/doc/faq#methods_on_values_or_pointers
Next is consistency. If some of the methods of the type must have pointer receivers, the rest should too, so the method set is consistent regardless of how the type is used. See the section on [method sets](https://go.dev/doc/faq#different_method_sets) for details.

mariajdab:
> added overwrite end2endtest that test two cases using the `sameBlock` and using the `nexBlock` added `sendVote` helper to reuse code.

TODO in other PR: remove missing `log.fatal`

Issue #846, opened 2023-03-06
encrypt elections support for end2end test


p4u:
Encrypted elections should be supported by the end2end test suit.

Ideally, the apiclient `Vote` method would transparently encrypt the vote if the election is `secretUntilTheEnd`

PR #874, opened 2023-04-04
end2endtest refactoring

altergui:
* move mkTreeVoteTest out of main.go into its own file, rename to something less cryptic (plaintextElectionTest), operation `plaintextelection` instead of `vtest`

### DIFF
--- a/apiclient/helpers.go
+++ b/apiclient/helpers.go
@@ -18,7 +18,11 @@ import (
 	"google.golang.org/protobuf/proto"
 )
 
-const pollInterval = 4 * time.Second
+const (
+	TimeBetweenBlocks = 6 * time.Second
+	WaitTimeout       = 3 * TimeBetweenBlocks
+	PollInterval      = TimeBetweenBlocks / 6
+)
 
 func (c *HTTPclient) DateToHeight(date time.Time) (uint32, error) {
 	resp, code, err := c.Request(HTTPGET, nil, "chain", "dateToBlock", fmt.Sprintf("%d", date.Unix()))
@@ -67,7 +71,7 @@ func (c *HTTPclient) WaitUntilNBlocks(ctx context.Context, n uint32) {
 		info, err := c.ChainInfo()
 		if err != nil {
 			log.Error(err)
-			time.Sleep(pollInterval)
+			time.Sleep(PollInterval)
 			continue
 		}
 		c.WaitUntilHeight(ctx, info.Height+n)
@@ -92,7 +96,7 @@ func (c *HTTPclient) WaitUntilHeight(ctx context.Context, height uint32) error {
 			return nil
 		}
 		select {
-		case <-time.After(pollInterval):
+		case <-time.After(PollInterval):
 			continue
 		case <-ctx.Done():
 			return ctx.Err()
@@ -137,7 +141,7 @@ func (c *HTTPclient) WaitUntilElectionStatus(ctx context.Context,
 			return election, nil
 		}
 		select {
-		case <-time.After(pollInterval):
+		case <-time.After(PollInterval):
 			continue
 		case <-ctx.Done():
 			return nil, fmt.Errorf("election %v never reached status %s: %w", electionID, status, ctx.Err())
@@ -154,10 +158,29 @@ func (c *HTTPclient) WaitUntilTxIsMined(ctx context.Context,
 			return tr, nil
 		}
 		select {
-		case <-time.After(pollInterval):
+		case <-time.After(PollInterval):
 			continue
 		case <-ctx.Done():
 			return nil, ctx.Err()
+		}
+	}
+}
+
+// WaitUntilElectionKeys waits until the election has published its encryption keys,
+// and returns them.
+func (c *HTTPclient) WaitUntilElectionKeys(ctx context.Context, electionID types.HexBytes) (
+	*api.ElectionKeys, error) {
+	log.Infof("waiting for election %s to publish keys...", electionID)
+	for {
+		ek, err := c.ElectionKeys(electionID)
+		if err == nil {
+			return ek, nil
+		}
+		select {
+		case <-time.After(PollInterval):
+			continue
+		case <-ctx.Done():
+			return nil, fmt.Errorf("election %s keys not yet published: %w", electionID, ctx.Err())
 		}
 	}
 }
--- a/apiclient/vote.go
+++ b/apiclient/vote.go
@@ -7,6 +7,7 @@ import (
 	"math/big"
 
 	"go.vocdoni.io/dvote/api"
+	"go.vocdoni.io/dvote/crypto/nacl"
 	"go.vocdoni.io/dvote/crypto/zk"
 	"go.vocdoni.io/dvote/crypto/zk/circuit"
 	"go.vocdoni.io/dvote/crypto/zk/prover"
@@ -45,21 +46,12 @@ type VoteData struct {
 // which contains the electionID, the choices and the proof. The
 // return value is the voteID (nullifier).
 func (c *HTTPclient) Vote(v *VoteData) (types.HexBytes, error) {
-	votePackage := &vochain.VotePackage{
-		Votes: v.Choices,
-	}
-	votePackageBytes, err := json.Marshal(votePackage)
+	election, err := c.Election(v.ElectionID)
 	if err != nil {
 		return nil, err
 	}
-	vote := &models.VoteEnvelope{
-		Nonce:       util.RandomBytes(16),
-		ProcessId:   v.ElectionID,
-		VotePackage: votePackageBytes,
-	}
 
-	// Get de election metadata
-	election, err := c.Election(v.ElectionID)
+	vote, err := c.prepareVoteEnvelope(v.Choices, election)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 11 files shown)


altergui:
i want to make a couple of small improvements yet:
* [x] running some tests concurrently so integration test doesn't take 12m :dizzy_face:
* [x] cleaning git history

but i consider this ready for review already, feedback welcome

altergui:
notable things done until now in this PR:
* fix https://github.com/vocdoni/vocdoni-node/issues/846
* implement VochainTest interface that @p4u proposed

apiclient/helpers.go
```diff
@@ -162,6 +162,42 @@ func (c *HTTPclient) WaitUntilTxIsMined(ctx context.Context,
 	}
 }
 
+// WaitUntilElectionKeys waits until the election has published its encryption keys,
+// and returns them along with keyIndexes.
+func (c *HTTPclient) WaitUntilElectionKeys(pid []byte,
+) (keys []types.HexBytes, keyIndexes []uint32, err error) {
+	waitTimeout := 10 * time.Second
```
p4u: Timeout is too narrow.

apiclient/vote.go
```diff
@@ -187,6 +179,70 @@ func (c *HTTPclient) Verify(electionID, voteID types.HexBytes) (bool, error) {
...
+		EncryptionKeyIndexes: keyIndexes,
+	}, nil
+
+}
+
+// prepareVotePackageBytes returns a plaintext json.Marshal(vp) if keys is nil,
+// else assigns a random hex string to vp.Nonce
+// and encrypts the vp bytes for each given key as recipient
+func (c *HTTPclient) prepareVotePackageBytes(vp *vochain.VotePackage, keys []types.HexBytes) ([]byte, error) {
+	var vpb []byte
```
p4u: no need for declaring these two vars here

apiclient/vote.go
```diff
@@ -187,6 +179,70 @@ func (c *HTTPclient) Verify(electionID, voteID types.HexBytes) (bool, error) {
...
+	vpb, err = json.Marshal(vp)
+	if err != nil {
+		return nil, err
+	}
+
+	for i, k := range keys { // skipped if len(keys) == 0
+		if len(k) == 0 {
+			continue
+		}
+		log.Debugf("encrypting with key %s", k)
```
p4u: use `Debugw`

cmd/end2endtest/account.go
```diff
@@ -20,51 +19,76 @@ import (
...
 		log.Fatal(err)
 	}
 
 	// check send tokens
-	if err := testSendTokens(api, alice, bob); err != nil {
+	if err := testSendTokens(t.api, t.alice, t.bob); err != nil {
 		log.Fatal(err)
 	}
+
+	return time.Since(start), nil
```
p4u: if the test is named "token transactions" it would  make more sense to return the duration of the testSendokens operation only, since the rest is actually a setup phase.

cmd/end2endtest/account.go
```diff
@@ -185,35 +215,26 @@ func testSendTokens(api *apiclient.HTTPclient, aliceKeys, bobKeys *ethereum.Sign
...
+	log.Infow("current state", "account", acc.Address.String(), "nonce", acc.Nonce, "balance", acc.Balance)
+
+	if expBalance != acc.Balance {
+		log.Fatalf("expected %s to have balance %d got %d", acc.Address.String(), expBalance, acc.Balance)
 	}
-	if aliceAcc.Nonce+1 != aliceAccAfter.Nonce {
-		log.Fatalf("expected %s to have nonce %d got %d",
-			aliceKeys.Address(), aliceAcc.Nonce+1, aliceAccAfter.Nonce)
+	if expNonce != acc.Nonce {
+		log.Fatalf("expected %s to have nonce %d got %d", acc.Address.String(), expNonce, acc.Nonce)
```
p4u: I don't think none of these functions should fatal. They should return error and be correctly propagated until the initial caller of the test.

cmd/end2endtest/encrypted.go
```diff
@@ -0,0 +1,381 @@
...
+			log.Fatal("account balance is 0")
+		}
+	}
+
+	log.Infof("account %s balance is %d", api.MyAddress().Hex(), acc.Balance)
+
+	// Create a new census
+	censusID, err := api.NewCensus("weighted")
+	if err != nil {
+		log.Fatal(err)
```
p4u: let's return errors...

cmd/end2endtest/encrypted.go
```diff
@@ -0,0 +1,381 @@
...
+	}
+	log.Infof("new census created with id %s", censusID.String())
+
+	// Generate 10 participant accounts
+	voterAccounts := ethereum.NewSignKeysBatch(c.nvotes)
```
p4u: It looks like many of the following code might be reused by other tests. We might want to split it in small functions.

cmd/end2endtest/encrypted.go
```diff
@@ -0,0 +1,381 @@
...
+	ctx, cancel = context.WithTimeout(context.Background(), time.Second*300)
+	defer cancel()
+	election, err = api.WaitUntilElectionStatus(ctx, electionID, "RESULTS")
+	if err != nil {
+		log.Fatal(err)
+	}
+	log.Infof("election %s status is RESULTS", electionID.String())
+	log.Infof("election results: %v", election.Results)
+
+	return time.Since(start), nil
```
p4u: I think the returned time should be time.Since(startTime) instead.

cmd/end2endtest/main.go
```diff
@@ -31,39 +23,26 @@ import (
 // * how many times to retry opening a connection to an endpoint before giving up
 const retries = 10
 
+type VochainTest interface {
+	Setup(api *apiclient.HTTPclient, config *config) error
+	Run() (duration time.Duration, err error)
```
p4u: Document that duration is expected to be the actual time of the test, without considering any setup phase.

cmd/end2endtest/main.go
```diff
@@ -31,39 +23,26 @@ import (
 // * how many times to retry opening a connection to an endpoint before giving up
 const retries = 10
 
+type VochainTest interface {
```
p4u, line 31: This interface requires proper documentation.

cmd/end2endtest/plaintext.go
```diff
@@ -0,0 +1,381 @@
+package main
```
p4u, line 1: Many of the comments I made before can be applied on this file too. So I'm not reviewing further until all previous comments are addressed.

p4u (dismissed):
Great job @altergui, the code looks quite neat.

I'd suggest some changes on the general structure.

1. Add the test implementations into one ore several subdirectories and packages, they might be group by type, i.e "voting_tests" "account_tests"

2.  In the main package directory, you might have the interface definition and the cmd.

3. You might want to have a "shared" directory package, with the functions that are shared  among tests. Be aware of avoiding cyclic imports.

cmd/end2endtest/encrypted.go
```diff
@@ -0,0 +1,381 @@
...
+					},
+				},
+			},
+		},
+	})
+	if err != nil {
+		log.Fatal(err)
+	}
+	log.Infof("created new election with id %s - now wait until it starts", electionID.String())
+
```
mariajdab: I am not sure, but maybe is needed to also call `api.WaitUntilElectionCreated`

cmd/end2endtest/encrypted.go
```diff
@@ -0,0 +1,381 @@
...
+				}
+				// if the vote was sent successfully or already exists, we remove it from the accounts map
+				votesSent++
+				delete(accountsMap, i)
+
+			}
+			if len(accountsMap) == 0 {
+				break
+			}
+			log.Infof("sent %d/%d votes... got %d HTTP errors", votesSent, len(accounts), contextDeadlines)
```
mariajdab, line 103: I think this log maybe should before the break in order to show the vote information of each chunk of accounts passed to the function

apiclient/helpers.go
altergui: good catch, actually this whole func needed refactoring.

cmd/end2endtest/encrypted.go
altergui: conceptually it's ok but in practice is redundant, WaitUntilElectionStarts is what you want after all

cmd/end2endtest/encrypted.go
altergui: true, the contextDeadlines is not reported in the line below
("successfully sent %d votes")
but i guess it's not so important in case of success. so i'll leave it as is

cmd/end2endtest/encrypted.go
mariajdab: Get it! Thanks!

cmd/end2endtest/encrypted.go
```diff
@@ -0,0 +1,381 @@
...
+		}
+		if c.faucet != "" && acc.Balance == 0 {
+			log.Fatal("account balance is 0")
+		}
+	}
+
+	log.Infof("account %s balance is %d", api.MyAddress().Hex(), acc.Balance)
+
+	// Create a new census
+	censusID, err := api.NewCensus("weighted")
```
mariajdab: I think we can use `vapi.CensusTypeWeighted`

cmd/end2endtest/account.go
altergui: done in 2fcc2b5d35f8b5b9df78a87987246dfb561edad3

cmd/end2endtest/encrypted.go
altergui: good one, thanks!

cmd/end2endtest/encrypted.go
altergui: @mariajdab solved this

cmd/end2endtest/account.go
altergui: in general i'd agree, but in my experience with end2endtest it helps A LOT to call Fatal on the relevant line instead of returning the error and Fatal on the caller, because this obscures the source of the error.
being an integration test, i think we can make these kind of compromises for the sake of readability or debugging

on another hand, i just realized that since zerolog implementation, Fatal no longer spits a stack trace, which was tremendously useful precisely to handle this situation i just wrote.

altergui: i'll keep all the Fatals as-is for now, and consider this change for a second round of refactoring (after making zerolog Fatal spit stack traces) to not make this PR eternal

cmd/end2endtest/main.go
altergui: done. i actually refactored the code so that duration is calculated by the caller, makes everything cleaner. I also now report the duration of the Setup and Teardown steps, which might be relevant

altergui:
a most notable change that i want to highlight is that now job_compose_test runs the tests concurrently, which led to uncovering a couple of bugs like #890 that i either worked around or fixed. previously we simply said "ok let's not do this concurrently".

so the run time is halved to 4mins instead of the usual 8-12 minutes, which is simply GREAT. it takes roughly the same as job_go_test, which is also run concurrently. so now the whole CI takes 4 mins, instead of go_test taking 4 mins and compose_test taking ages

altergui:
for me this is ready to merge as-is: even though it could be improved (returns instead of Fatals, deduplicating code), it doesn't break anything and already brings a lot of value, so i believe it's better to merge it (also unblocking #873) and open a second iteration of refactoring in a new PR

mariajdab (approved):

mariajdab:
Just one conflict to resolve I think

PR #879, opened 2023-04-10
e2etest: replace hardcoded amounts, they don't work with api-dev

altergui:
quick fix for 

```
API error: 500 ({"error":"vochain replied with error code: (1) checkTx createAccountTx: issuer address does not have enough balance 49, required 51","code":5003}
```

when running end2endtest against https://api-dev.vocdoni.net/v2


mariajdab:
LGTM :eyes:

mariajdab (approved):

PR #881, opened 2023-04-11
oracle setprocess queue

altergui:


p4u:
@altergui I'm not sure fixing this bug is worth it. There is a pending development that will finally remove the oracle, the results will be computed automatically via an internal-state-transition as documented here https://hackmd.io/@vocdoni/ryHg_tslh

oracle/oracle.go
```diff
@@ -20,6 +22,9 @@ import (
 type Oracle struct {
 	VochainApp *vochain.BaseApplication
 	signer     *ethereum.SignKeys
+
+	resultQueue map[string]*models.ProcessResult
```
p4u, line 26: As it is a temporary workaround, it is fine. Just for the record, next time you can use a simple channel for achieving the same properties of the queue, such as in vocone: https://github.com/vocdoni/vocdoni-node/blob/master/vocone/vocone.go#L52

p4u (approved):

jordipainan (approved):

p4u:
@altergui feel free to merge it

oracle/oracle.go
altergui: nice! didn't consider this.
reading the vc.mempool code, i didn't understand how failed transactions are kept in the mempool  (to be retried in next block). it looks to me as they are not, i.e. they are dropped if they fail.
in my case i want them to stay in the queue until the process is RESULTS. i can imagine implementing this with a channel (re-adding them if they fail) but a map with a lock sounded easier.
as you said, it's just a temporary hotfix, so i'll merge as-is

PR #882, opened 2023-04-12
Bug/pubkey support legacy code

### LINKED ISSUE #875: bug: remove legacy tests with PubKey as census tree leaf key.
During end2end vote overwrite test implementation, @mariajdab found some legacy code that uses a Public Key as census tree leaf key:
- https://github.com/vocdoni/vocdoni-node/blob/7ca99b8171b8f298d0fd4eb73b0fb0000f4874a9/vochain/vote_test.go#L65
- https://github.com/vocdoni/vocdoni-node/blob/7ca99b8171b8f298d0fd4eb73b0fb0000f4874a9/vochain/indexer/indexer_test.go#L549
- https://github.com/vocdoni/vocdoni-node/blob/7ca99b8171b8f298d0fd4eb73b0fb0000f4874a9/test/testcommon/testvoteproof/voteproof.go#L97

There are likely to be other parts with legacy code that needs to be updated, so a prior review will be required.

lucasmenendez:
We found some parts of the vochain that contains legacy code regarding the support to PublicKey as census key. This feature is deprecated and this code needs to be removed or upgraded. Read more here: #875.

Changes made and related commits:
 - The references to old census key type (`models.ProofArbo_PUBKEY`) was removed and replaced to the new default type (`models.ProofArbo_ADDRESS`): 449d701475287c1ce9af84d8899cb2043e8242d7
 - The tests that implements public key as census type was replaced to use the voter address instead: 9b91fde9edccc3b9af1d0d2bd7e6f2bab134231a 40c92a993ab7825a08b73d0ba8cb60744dc48ae0
 - A proposal to create the VoterID using the voter address. To achieve this, a new VoterID type was introduced (`VoterIDTypeAddress `). Check out an example here: 4d05be309e9fd40ea76e415274d33842bf9c81da

### DIFF
--- a/api/censuses.go
+++ b/api/censuses.go
@@ -228,7 +228,7 @@ func (a *API) censusAddHandler(msg *apirest.APIdata, ctx *httprouter.HTTPContext
 
 		leafKey := p.Key
 		if len(leafKey) > censustree.DefaultMaxKeyLen {
-			return ErrParticipanKeyLengthInvalid.Withf("can not be longer than %d bytes", censustree.DefaultMaxKeyLen)
+			return ErrInvalidCensusKeyLength.Withf("the census key cannot be longer than %d bytes", censustree.DefaultMaxKeyLen)
 		}
 
 		if ref.CensusType != int32(models.Census_ARBO_POSEIDON) {
--- a/api/censuses.go
+++ b/api/censuses.go
@@ -6,7 +6,6 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"errors"
-	"fmt"
 	"time"
 
 	"github.com/google/uuid"
@@ -229,9 +228,7 @@ func (a *API) censusAddHandler(msg *apirest.APIdata, ctx *httprouter.HTTPContext
 
 		leafKey := p.Key
 		if len(leafKey) > censustree.DefaultMaxKeyLen {
-			deprecationMsg = []byte(fmt.Sprintf("census keys must not be longer "+
-				"than %d, the key provided has been truncated to this length",
-				censustree.DefaultMaxKeyLen))
+			return ErrParticipanKeyLengthInvalid.Withf("can not be longer than %d bytes", censustree.DefaultMaxKeyLen)
 		}
 
 		if ref.CensusType != int32(models.Census_ARBO_POSEIDON) {
@@ -240,12 +237,7 @@ func (a *API) censusAddHandler(msg *apirest.APIdata, ctx *httprouter.HTTPContext
 			if err != nil {
 				return ErrCantComputeKeyHash.WithErr(err)
 			}
-			// If the provided key is longer than the defined maximum length
-			// truncate it.
-			// TODO: return an error if the other key lengths are deprecated
-			if len(leafKey) > censustree.DefaultMaxKeyLen {
-				leafKey = leafKey[:censustree.DefaultMaxKeyLen]
-			}
+			leafKey = leafKey[:censustree.DefaultMaxKeyLen]
 		}
 
 		keys = append(keys, leafKey)
--- a/api/errors.go
+++ b/api/errors.go
@@ -67,6 +67,7 @@ var (
 	ErrNoElectionKeys                   = apirest.APIerror{Code: 4048, HTTPstatus: apirest.HTTPstatusNotFound, Err: fmt.Errorf("no election keys available")}
 	ErrKeyNotFoundInCensus              = apirest.APIerror{Code: 4049, HTTPstatus: apirest.HTTPstatusNotFound, Err: fmt.Errorf("key not found in census")}
 	ErrInvalidStatus                    = apirest.APIerror{Code: 4050, HTTPstatus: apirest.HTTPstatusBadRequest, Err: fmt.Errorf("invalid status")}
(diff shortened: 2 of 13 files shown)


p4u:
It looks weird and a bit inconsistent to have a type for ECDSA and a type for Address (which is in fact a hash of a ECDSA pubkey), when the VoterID type has already a method that returns the address "Address()".

I don't understand why this change is necessary. For creating a VoterID one needs to have the publicKey of the participant, but that's recovered from the ECDSA Signature. Usually we build the VoterID after verifying a Signature of a transaction. What's the use case to build a VoterID directly out of an Address instead of a Signature?

lucasmenendez:
> I don't understand why this change is necessary. For creating a VoterID one needs to have the publicKey of the participant, but that's recovered from the ECDSA Signature. Usually we build the VoterID after verifying a Signature of a transaction. What's the use case to build a VoterID directly out of an Address instead of a Signature?

@p4u This is not necessary at all, in fact all changes related to this are grouped together in a commit for easy undoing. I will revert it, make a second review trying to find more legacy code and request reviews to merge.

p4u:
> > I don't understand why this change is necessary. For creating a VoterID one needs to have the publicKey of the participant, but that's recovered from the ECDSA Signature. Usually we build the VoterID after verifying a Signature of a transaction. What's the use case to build a VoterID directly out of an Address instead of a Signature?
> 
> @p4u This is not necessary at all, in fact all changes related to this are grouped together in a commit for easy undoing. I will revert it, make a second review trying to find more legacy code and request reviews to merge.

I'm not saying revert the commit, I'm only trying to understand :)

I'd like to find a rationale behind the VoterID type. Until now, (to me) it makes sense so hold the publicKey as principal identifier, since this is the seed for any Address. We might consider changing this approach. But let's try to be clear on what's the purpose and scope of the VoterID.

I'm very open to discussion on this topic and happy to listen to your opinion.

lucasmenendez:
No no, I don't have a stronger argument for changing the current behaviour. I had not fully understood of where the VoterID is used, and with your explanation I think that these changes are not necessary.

My point was that when I read and updated the tests, I thought it was a bit tricky to use `voterAccount.Address()` to be registered in a census and be able to vote, but to verify that vote the `voterAccount.PublicKey()` is needed. Now, I understand that this is an internal process of the Vochain.

lucasmenendez:
I reviewed the code again and I can't find any more old stuff to update, but probably 4 or 6 eyes see more than 2 so, ready to review!

lucasmenendez:
After @marcvelmer comments regarding public key usage as census key I found a TODO that I miss. This TODO allows to create census with public keys as census keys if the census is not based on Poseidon hash algorithm. You can check it here: https://github.com/vocdoni/vocdoni-node/blob/449d701475287c1ce9af84d8899cb2043e8242d7/api/censuses.go#L245

Solving this TODO, the public keys will not longer be supported as census key, and requires to upgrade also the tests.

I will this PR back to draft to work on it and continue discussing this here: https://github.com/vocdoni/interoperability/issues/34.

api/censuses.go
```diff
@@ -229,9 +228,7 @@ func (a *API) censusAddHandler(msg *apirest.APIdata, ctx *httprouter.HTTPContext
 
 		leafKey := p.Key
 		if len(leafKey) > censustree.DefaultMaxKeyLen {
-			deprecationMsg = []byte(fmt.Sprintf("census keys must not be longer "+
-				"than %d, the key provided has been truncated to this length",
-				censustree.DefaultMaxKeyLen))
+			return ErrParticipanKeyLengthInvalid.Withf("can not be longer than %d bytes", censustree.DefaultMaxKeyLen)
```
altergui: typo: missing 't' in ErrParticipantKeyLengthInvalid

altergui: also 'cannot' is a single word

api/errors.go
```diff
@@ -67,6 +67,7 @@ var (
 	ErrNoElectionKeys                   = apirest.APIerror{Code: 4048, HTTPstatus: apirest.HTTPstatusNotFound, Err: fmt.Errorf("no election keys available")}
 	ErrKeyNotFoundInCensus              = apirest.APIerror{Code: 4049, HTTPstatus: apirest.HTTPstatusNotFound, Err: fmt.Errorf("key not found in census")}
 	ErrInvalidStatus                    = apirest.APIerror{Code: 4050, HTTPstatus: apirest.HTTPstatusBadRequest, Err: fmt.Errorf("invalid status")}
+	ErrParticipanKeyLengthInvalid       = apirest.APIerror{Code: 4051, HTTPstatus: apirest.HTTPstatusBadRequest, Err: fmt.Errorf("participant census key length is wrong")}
```
altergui: whenever possible, keep the var name exactly the same as the string.
suggestions:
ErrKeyLengthInvalid "key length invalid"
or (if Participant is an important word):
ErrParticipantKeyLengthInvalid "participant key length invalid"

altergui (commented):
i didn't test the code but leaving here some nitpicking

api/errors.go
lucasmenendez: I have renamed the error to `ErrInvalidCensusKeyLength` and I changed the message to `invalid census key length`, which I think is more descriptive.

altergui (approved):
LGTM

p4u (approved):

PR #886, opened 2023-04-12
rename branch master -> main

altergui:
subtask of https://github.com/vocdoni/interoperability/issues/27

this is a PR replacing a few ocurrences in code,
the main task would be changing the default branch in github project, in a coordinated way


p4u:
This changes looks a bit scary, but its true that now its probably a good moment for doing it... 
I've like to add some kind of backwards compatibility, we might for some time, push docker images on both tags `master` and `main`

altergui:
i created `main` branch (pointing to same commit as `master`), so that the actions for this PR work

i believe the following steps would be:
* merge this PR
* immediately after, change vocdoni-node github repo default branch to be `main`
* immediately after, delete `master` branch

i believe github rebases all open PRs when you change the default branch, but i'm going to test in a private repo to see how this works. if not, i'd manually edit each PR to target `main` branch, it's 5 minutes work.

altergui:
yep, i just tested and confirmed definitely github updates all open PRs to point to the new default automatically

lastly, after all the moves in github are done, each dev will need to do this on their local repos:
```sh
git branch -m master main
git fetch origin
git branch -u origin/main main
git remote set-head origin -a
```

p4u (approved):
let's abolish the slavery

PR #899, opened 2023-04-27
end2endtest: vote overwrite and max census size test

mariajdab:
end2endtest: add overwrite test using same block and next block
end2endtest: refactor add helper `sendvote` to reuse code
end2endtest: add max census size test

### DIFF
--- a/apiclient/helpers.go
+++ b/apiclient/helpers.go
@@ -149,6 +149,27 @@ func (c *HTTPclient) WaitUntilElectionStatus(ctx context.Context,
 	}
 }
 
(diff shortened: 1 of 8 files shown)


p4u (approved):

altergui:
please add the new tests to start_test.sh so that they are included in CI (and the code coverage is actually increased, not decreased :sweat_smile: )

p4u:
Also rebase with master and squash into a single commit before merging, please.

mariajdab:
Not sure why the `job_compose_test` fails, running all with docker seems everything ok

PR #900, opened 2023-04-27
F/ist

p4u:
Compute results as consensus would allow for the computation of results based on consensus among the nodes on the Vochain blockchain. This would ensure that the results are accurate and tamper-proof, and would also prevent any single node from having too much influence over the outcome. Currently, the results are not part of the state, but by making the Vochain nodes compute and verify the results, the results would be much more verifiable.

To enable this feature, we must introduce a new internal state transition mechanism (IST), which would make the Vochain transition to a new state without the need of external transactions. However all nodes execute the IST at exactly same height. Thus we need to create a new component, part of the state named Internal State Transition Controller (ISTC).


p4u:
Please, focus the reviews on the `ist` package

vochain/ist/ist.go
```diff
@@ -0,0 +1,303 @@
...
+	"go.vocdoni.io/proto/build/go/models"
+)
+
+const (
+	dbPrefix = "ist/"
+)
+
+// Controller is the internal state transition controller.
+type Controller struct {
+	st              *state.State
```
mvdan: use a longer name, otherwise `m.st` is not very useful :) perhaps `m.state`?

vochain/ist/ist.go
```diff
@@ -0,0 +1,303 @@
...
+
+// Actions is the model used to store the list of IST actions for
+// a specific height into state.
+type Actions struct {
+	Actions map[string]Action
+}
+
+// Action is the model used to store the IST actions into state.
+type Action struct {
+	Action     int
```
mvdan: assuming this gets marshalled, I would say pick either int32 or int64, otherwise the behavior may be different depending on the architecture.

vochain/ist/ist.go
```diff
@@ -0,0 +1,303 @@
...
+var ActionsToString = map[int]string{
+	ActionComputeResults: "compute-results",
+	ActionCommitResults:  "commit-results",
+	ActionEndProcess:     "end-process",
+}
+
+// Actions is the model used to store the list of IST actions for
+// a specific height into state.
+type Actions struct {
+	Actions map[string]Action
```
mvdan: any reason to make this a field of a struct, instead of `type Actions map[string]Action`?

or if you think you might need more fields in the future, rename the field, like `type Actions struct { ByName map...` to avoid the `Actions.Actions` stutter.

vochain/ist/ist.go
mvdan: this field name also stutters; perhaps something like `Action.ID` or `Action.Number` instead?

vochain/ist/ist.go
```diff
@@ -0,0 +1,303 @@
...
+type Action struct {
+	Action     int
+	ElectionID []byte
+	Attempts   uint32
+}
+
+// encode performs the encoding of the IST action using Gob.
+func (m *Actions) encode() []byte {
+	var buf bytes.Buffer
+	if err := gob.NewEncoder(&buf).Encode(m); err != nil {
```
mvdan, line 102: remember that encoding/gob is not a stable format; it may change with future Go versions. I think it's fine for in-memory stuff or temporary caches, but anything relatively permanent should use a stable format. For that, I'd either use go-bare if we don't need a schema, or protobuf if we want a schema.

vochain/ist/ist.go
```diff
@@ -0,0 +1,303 @@
...
+		return fmt.Errorf("cannot reschedule IST action: old height (%d) >= new height (%d)", oldHeight, newHeight)
+	}
+	actions, err := c.Actions(oldHeight)
+	if err != nil {
+		return err
+	}
+	// find the IST action
+	var actID string
+	for actID = range actions.Actions {
+		if string(id) == actID {
```
mvdan: why are you looping over a map to find a key? just index into the map?

vochain/ist/ist.go
```diff
@@ -0,0 +1,303 @@
...
+	// decode the IST actions
+	if err := actions.decode(actionsBytes); err != nil {
+		return nil, fmt.Errorf("could not decode actions: %w", err)
+	}
+	return &actions, nil
+}
+
+// Reschedule reschedules an IST action to the given new height.
+func (c *Controller) Reschedule(id []byte, oldHeight, newHeight uint32) error {
+	if id == nil {
```
mvdan: probably do `if len(id) == 0` instead, otherwise you'll happily accept `Reschedule(make([]byte, 0), ...` which is still an invalid ID

vochain/ist/ist.go
```diff
@@ -0,0 +1,303 @@
...
+func (c *Controller) endElection(electionID []byte) error {
+	process, err := c.st.Process(electionID, false)
+	if err != nil {
+		return fmt.Errorf("cannot get process: %w", err)
+	}
+	// if the process is canceled, ended or in results, we don't need to do
+	// anything else, smooth return.
+	if process.Status == models.ProcessStatus_CANCELED ||
+		process.Status == models.ProcessStatus_RESULTS ||
+		process.Status == models.ProcessStatus_ENDED {
```
mvdan: use a switch :)

vochain/ist/ist.go
```diff
@@ -0,0 +1,303 @@
...
+	}
+}
+
+const (
+	// ActionComputeResults computes and schedules the commit of results, for the given election and height.
+	ActionComputeResults = iota
+	// ActionCommitResults schedules the commit of the results to the state.
+	ActionCommitResults
+	// ActionEndProcess sets a process as ended. It schedules ActionComputeResults.
+	ActionEndProcess
```
mvdan, line 78: make these typed constants, to ensure that you don't mix them with other integer values

vochain/ist/store.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+	defer tx.Discard()
+	return tx.NoState().Get(key)
+}
+
+// dbIndex returns the IST action index in the state database.
+func dbIndex(height uint32) []byte {
+	hBytes := make([]byte, 4)
+	binary.LittleEndian.PutUint32(hBytes, height)
+	return append([]byte(dbPrefix), hBytes...)
```
mvdan: Or, since Go 1.19, you can now do: `return binary.LittleEndian.AppendUint32([]byte(dbPrefix), height)` :)

vochain/ist/ist.go
```diff
@@ -0,0 +1,303 @@
...
+					fmt.Sprintf("%x", id), "action", ActionsToString[action.Action], "attempt", action.Attempts)
+				if err := c.Reschedule([]byte(id), height, newHeight); err != nil {
+					return fmt.Errorf("cannot reschedule IST action: %w", err)
+				}
+				continue
+			}
+			if !isSynchronizing {
+				// compute results in a goroutine to avoid blocking the state transition
+				go func() {
+					if err := c.computeAndStoreResults(action.ElectionID); err != nil {
```
mvdan, line 212: you might have multiple of these calls happen concurrently, would that be safe?

mvdan (commented):
would be nice to also document which of these new `ist` methods are OK to call concurrently; APIs by default are not safe for concurrent use unless the docs say so.

vochain/ist/ist.go
p4u: Yes, it should be safe.

vochain/ist/ist.go
p4u: Since this information is not relevant for the Consensus State (its stored into a NoState() KV space), it is not a big issue if the Gob encoder in some point changes to a different format.

vochain/ist/results.go
```diff
@@ -0,0 +1,113 @@
...
+			r = new(results.Results)
+			if err := r.Decode(rEncoded); err != nil {
+				return fmt.Errorf("cannot decode results: %w", err)
+			}
+			break
+		}
+	}
+
+	// Delete the stored results
+	_ = c.deleteFromNoState(dbResultsIndex(electionID))
```
mariajdab, line 98: Just curious about why is not necessary check a possible error? Or maybe a log could be useful

vochain/ist/results.go
```diff
@@ -0,0 +1,113 @@
...
+	if err != nil {
+		return fmt.Errorf("cannot encode results: %w", err)
+	}
+	if err := c.storeToNoState(dbResultsIndex(electionID), rEncoded); err != nil {
+		return fmt.Errorf("cannot store results: %w", err)
+	}
+	return nil
+}
+
+// commitResults commits the results to the state. If the reults parameter is nil, it will
```
mariajdab: typo `reults`

vochain/ist/ist.go
```diff
@@ -0,0 +1,298 @@
...
+	ID         ActionID
+	ElectionID []byte
+	Attempts   uint32
+}
+
+// encode performs the encoding of the IST action using Gob.
+func (m *Actions) encode() []byte {
+	var buf bytes.Buffer
+	if err := gob.NewEncoder(&buf).Encode(m); err != nil {
+		// should never happend
```
mariajdab: should never happen :eyes:

vochain/ist/results.go
p4u: The reason is, there might be a situation where there are no results to delete (if the first condition of the loop is met). And in any case, if the deletion of results fails for some real reason, probably the database is corrupted and more errors will raise.

mariajdab: Got it! Thanks

cmd/end2endtest/main.go
```diff
@@ -103,6 +103,8 @@ func main() {
 			fmt.Fprintf(os.Stderr, op.example+"\n")
 			fmt.Fprintf(os.Stderr, "\n")
 		}
+		fmt.Fprintf(os.Stderr, "If the network is deployed locally using the docker test suit, ")
```
jordipainan:
```suggestion
		fmt.Fprintf(os.Stderr, "If the network is deployed locally using the docker test suite, ")
```

cmd/node/main.go
```diff
@@ -92,7 +91,7 @@ func newConfig() (*config.Config, config.Error) {
 	globalCfg.SaveConfig = *flag.Bool("saveConfig", false,
 		"overwrite an existing config file with the provided CLI flags")
 	globalCfg.Mode = *flag.StringP("mode", "m", types.ModeGateway,
-		"global operation mode. Available options: [gateway,oracle,ethApiOracle,miner,seed]")
+		"global operation mode. Available options: [gateway,miner,seed]")
```
jordipainan:
```suggestion
		"global operation mode. Available options: [gateway, miner, seed]")
```

config/config.go
```diff
@@ -5,7 +5,7 @@ import (
 	"go.vocdoni.io/dvote/types"
 )
 
-// DvoteCfg stores global configs for dvote
+// Config stores global configs for dvote
```
jordipainan:
```suggestion
// Config stores global configurations for vocdoni-node
```

db/batch.go
```diff
@@ -43,7 +43,7 @@ func NewBatchMaxSize(db Database, maxSize uint) *Batch {
 	}
 }
 
-// Get implements the WriteTx.Get interface method
+// Get implements the WriteTx.Get interface method.
```
jordipainan:
```suggestion
// Get implements the WriteTx.Get() interface method.
```

dockerfiles/testsuite/env.miner0
```diff
@@ -9,4 +9,4 @@ VOCDONI_VOCHAIN_GENESIS=/app/misc/genesis.json
 VOCDONI_VOCHAIN_MINERKEY=cda909c34901c137e12bb7d0afbcb9d1c8abc66f03862a42344b1f509d1ae4ce
 VOCDONI_METRICS_ENABLED=True
 VOCDONI_METRICS_REFRESHINTERVAL=5
-VOCDONI_VOCHAIN_MINERTARGETBLOCKTIMESECONDS=5
+VOCDONI_VOCHAIN_MINERTARGETBLOCKTIMESECONDS=6
```
jordipainan, line 12: Why this change ? Is it because you want to keep the refresh interval lower that the target block time ?

apiclient/helpers.go
```diff
@@ -191,17 +191,17 @@ func (c *HTTPclient) WaitUntilTxIsMined(ctx context.Context,
...
 		if err == nil {
 			return ek, nil
 		}
 		select {
-		case <-time.After(PollInterval):
-			continue
 		case <-ctx.Done():
 			return nil, fmt.Errorf("election %s keys not yet published: %w", electionID, ctx.Err())
+		default:
+			time.Sleep(PollInterval)
```
mvdan: why this change? it seems like the old code was much better. if the context is cancelled, the sleep should stop - that's what time.After did.

dockerfiles/testsuite/env.miner1
```diff
@@ -1,12 +1,12 @@
...
+VOCDONI_LOGLEVEL=debug
 VOCDONI_DEV=True
 VOCDONI_VOCHAIN_PUBLICADDR=miner1:26656
 VOCDONI_VOCHAIN_SEEDS=3c3765494e758ae7baccb1f5b0661755302ddc47@seed:26656
 VOCDONI_VOCHAIN_GENESIS=/app/misc/genesis.json
 VOCDONI_VOCHAIN_MINERKEY=d52a488fa1511a07778cc94ed9d8130fb255537783ea7c669f38292b4f53ac4f
 VOCDONI_METRICS_ENABLED=True
 VOCDONI_METRICS_REFRESHINTERVAL=5
-VOCDONI_VOCHAIN_MINERTARGETBLOCKTIMESECONDS=5
+VOCDONI_VOCHAIN_MINERTARGETBLOCKTIMESECONDS=6
```
jordipainan, line 11: Same

dockerfiles/testsuite/env.miner2
```diff
@@ -1,11 +1,11 @@
...
+VOCDONI_LOGLEVEL=debug
 VOCDONI_DEV=True
 VOCDONI_VOCHAIN_PUBLICADDR=miner2:26656
 VOCDONI_VOCHAIN_SEEDS=3c3765494e758ae7baccb1f5b0661755302ddc47@seed:26656
 VOCDONI_VOCHAIN_GENESIS=/app/misc/genesis.json
 VOCDONI_VOCHAIN_MINERKEY=e06976e5eaf3f147e12763eb140e3d5c2ed16a6fa747d787d8b92ca961fa7dc4
 VOCDONI_METRICS_ENABLED=True
 VOCDONI_METRICS_REFRESHINTERVAL=5
-VOCDONI_VOCHAIN_MINERTARGETBLOCKTIMESECONDS=5
+VOCDONI_VOCHAIN_MINERTARGETBLOCKTIMESECONDS=6
```
jordipainan, line 11: Same

dockerfiles/testsuite/env.miner3
```diff
@@ -1,11 +1,11 @@
...
+VOCDONI_LOGLEVEL=debug
 VOCDONI_DEV=True
 VOCDONI_VOCHAIN_PUBLICADDR=miner3:26656
 VOCDONI_VOCHAIN_SEEDS=3c3765494e758ae7baccb1f5b0661755302ddc47@seed:26656
 VOCDONI_VOCHAIN_GENESIS=/app/misc/genesis.json
 VOCDONI_VOCHAIN_MINERKEY=b8d258559adee836a43e964badf541ec106cc68a01f989d3c3c9a030ae3945a6
 VOCDONI_METRICS_ENABLED=True
 VOCDONI_METRICS_REFRESHINTERVAL=5
-VOCDONI_VOCHAIN_MINERTARGETBLOCKTIMESECONDS=5
+VOCDONI_VOCHAIN_MINERTARGETBLOCKTIMESECONDS=6
```
jordipainan, line 11: Same

dockerfiles/testsuite/env.miner2
```diff
@@ -1,11 +1,11 @@
 VOCDONI_DATADIR=/app/run
 VOCDONI_MODE=miner
-VOCDONI_LOGLEVEL=info
+VOCDONI_LOGLEVEL=debug
```
mvdan: wouldn't this make CI way too chatty?

vochain/indexer/migrations/0001_create_table_processes.sql
```diff
@@ -11,7 +11,6 @@ CREATE TABLE processes (
   results_votes           TEXT NOT NULL,
   results_weight          TEXT NOT NULL,
   results_envelope_height INTEGER NOT NULL,
-  results_signatures      TEXT NOT NULL,
```
mvdan, line 14: this is a hard breaking change on the indexer DB - all existing DBs will need to be deleted. are we still OK with that?

dockerfiles/testsuite/env.miner0
p4u: Because 6 is the reality. I have not been able to achieve 5s

dockerfiles/testsuite/genesis.json
```diff
@@ -4,7 +4,7 @@
   "initial_height": "0",
   "consensus_params": {
     "block": {
-      "max_bytes": "5242880",
+      "max_bytes": "2097152",
```
lucasmenendez, line 7: Just a doubt, what does this limit mean and why is it being reduced?

vochain/indexer/migrations/0001_create_table_processes.sql
p4u: Yes. This PR is breaking everything anyway. So we need a hard reset.

vochain/app.go
```diff
@@ -42,13 +42,14 @@ var (
...
+	Istc               *ist.Controller
 	Service            service.Service
 	Node               *tmcli.Local
 	TransactionHandler *transaction.TransactionHandler
-	IsSynchronizing    func() bool
+	isSynchronizingFn  func() bool
 	// tendermint WaitSync() function is racy, we need to use a mutex in order to avoid
 	// data races when querying about the sync status of the blockchain.
-	isSyncLock sync.Mutex
+	isSynchronizing uint32
```
lucasmenendez: Why not just use `atomic.Uint32`?

dockerfiles/testsuite/genesis.json
p4u: This is the size of the block. So the number of transactions that can be included in a period of time of 10-12 seconds.

I've performed several tests and benchmarking before creating this PR and 5 MiB seems too much. The block time easily increases to 30-40 seconds on high load (thousands of votes per minute continuously). However if using 2 MiB, the network performs stable, with a block time of 12-13 seconds on high load.

apiclient/helpers.go
p4u: tbh, I don't now. Probably this is some merge with master that went wrong.

p4u: ~~Ah, I think this is part of @mariajdab PR (that was merged recently)~~
No, this was me....

lucasmenendez (approved):
Great job! The improvements make the process shorter and easier to understand, at least for me.

vochain/app.go
p4u: No reason. But its kind of the same, no?

lucasmenendez: Yes, I think they are both the same, but probably using the atomic version would be clearer. That way, just by reading the variable declaration you know that it is atomic, currently you have to look into the `IsSynchronizing()` method to know that.

mvdan: you could also simply export an https://pkg.go.dev/sync/atomic#Bool, removing the need for uint32, the Fn field, your own method, etc.

dockerfiles/testsuite/start_test.sh
```diff
@@ -121,6 +117,8 @@ e2etest_cenususizelection () {
 
 log "### Starting test suite ###"
 $COMPOSE_CMD build
+$COMPOSE_CMD up -d seed # start the seed first so the nodes can properly bootstrap
+sleep 10
```
jordipainan, line 121: 10 sec seems too much, I think 5 sec is actually enough

statedb/treeupdate.go
```diff
@@ -129,6 +130,12 @@ func (u *TreeUpdate) Set(key, value []byte) error {
 	return u.tree.Set(u.tree.tx, key, value)
 }
 
+// Del removes a key-value from this tree.  `key` is the path of the leaf.
+func (u *TreeUpdate) Del(key []byte) error {
```
jordipainan, line 134: I assume every time you delete a key the tree "rebalances". Do you think that so many deletes can impact on tx processing performance?

statedb/treeview.go
```diff
@@ -30,6 +30,12 @@ func (v *databaseViewer) Get(key []byte) ([]byte, error) {
 	return tx.Get(key)
 }
 
+// IterateWithPrefix iterates over all the keys that start with the given prefix.
```
jordipainan, line 33:
```suggestion
// IterateWithPrefix iterates over all keys that start with the given prefix.
```

statedb/treeview.go
```diff
@@ -30,6 +30,12 @@ func (v *databaseViewer) Get(key []byte) ([]byte, error) {
...
+// The iration stops when the callback returns false.
```
jordipainan:
```suggestion
// The iteration stops when the callback returns false.
```

tree/tree.go
```diff
@@ -121,6 +121,24 @@ func (t *Tree) Add(wTx db.WriteTx, key, value []byte) error {
 	return nil
 }
 
+// Del deletes a key from the Tree.
```
jordipainan, line 124: What about: 

```
// Del deletes a key from the Tree.
func (t *Tree) Del(wTx db.WriteTx, key []byte) error {
	givenTx := wTx == nil
	if givenTx {
		wTx = t.DB().WriteTx()
		defer wTx.Discard()
	}
	if err := wTx.Delete(key); err != nil {
		return err
	}
	if givenTx {
		return wTx.Commit()
	}
	return nil
}
```
- If the negation operator is removed the code is more readable (IMO)
- In the last condition check we can just return the `wtx.Commit()` so there is no need to check the error

vochain/genesis/genesis.go
```diff
@@ -118,8 +104,8 @@ var devGenesis = GenesisDoc{
 }
 
 var stageGenesis = GenesisDoc{
-	GenesisTime: time.Date(2023, time.March, 10, 10, 0, 0, 0, time.UTC),
-	ChainID:     "vocdoni-stage-3",
+	GenesisTime: time.Date(2023, time.April, 10, 10, 0, 0, 0, time.UTC),
```
jordipainan, line 107: Please next time make it clear that DEV & STG envs will be affected as these are breaking consensus changes.

vochain/indexer/indexer_test.go
```diff
@@ -530,10 +531,7 @@ func TestResults(t *testing.T) {
 	})
 	qt.Assert(t, err, qt.IsNil)
 
-	vp, err := json.Marshal(vochain.VotePackage{
-		Nonce: fmt.Sprintf("%x", util.RandomBytes(32)),
-		Votes: []int{1, 1, 1, 1},
-	})
+	vp, err := state.NewVotePackage([]int{1, 1, 1, 1}).Encode()
```
jordipainan, line 534: Oh! that is a cool one! what a pleasure to create a vote pkg like this <3

vochain/indexer/indexer_test.go
```diff
@@ -1251,10 +1238,7 @@ func TestOverwriteVotes(t *testing.T) {
 	})
 
 	// Send the initial vote again (for third time)
-	vp, err = json.Marshal(vochain.VotePackage{
-		Nonce: fmt.Sprintf("%x", util.RandomHex(32)),
-		Votes: []int{0, 0, 0},
-	})
+	vp, err = state.NewVotePackage([]int{0, 0, 0}).Encode()
```
jordipainan, line 1241: What happens if I overflow the int value ?

vochain/indexer/process.go
```diff
@@ -375,6 +342,16 @@ func (s *Indexer) updateProcess(pid []byte) error {
 	}); err != nil {
 		return err
 	}
+
+	// If the process is in RESULTS status, and it was not in RESULTS status before, then finalize the results
```
jordipainan, line 346:
```suggestion
	// If the process is in RESULTS status, and it was not in RESULTS status before, finalize the results.
```

vochain/indexer/process.go
```diff
@@ -375,6 +342,16 @@ func (s *Indexer) updateProcess(pid []byte) error {
...
+	if p.Status == models.ProcessStatus_RESULTS &&
+		models.ProcessStatus(previousStatus) != models.ProcessStatus_RESULTS {
+		if err := s.finalizeResults(pid); err != nil {
+			return err
+		}
+	}
+
+	// If the process is in CANCELED status, and it was not in CANCELED status before, then remove the results
```
jordipainan, line 354:
```suggestion
	// If the process is in CANCELED status, and it was not in CANCELED status before, remove the results.
```

vochain/indexer/vote.go
```diff
@@ -214,50 +210,42 @@ func (s *Indexer) GetEnvelopeHeight(processID []byte) (uint64, error) {
 	return uint64(height), err
 }
 
-// ComputeResult process a finished voting, compute the results and saves it in the Storage.
+// finalizeResults process a finished voting, get the results from the state and saves it in the indexer Storage.
```
jordipainan, line 213: What is a finished voting? Maybe adds more clarity expressing it by STATUS

jordipainan, line 213:
```suggestion
// finalizeResults processes a finished voting, gets the results from the state and saves it in the indexer storage.
```

vochain/indexer/vote.go
```diff
@@ -214,50 +210,42 @@ func (s *Indexer) GetEnvelopeHeight(processID []byte) (uint64, error) {
...
 // Once this function is called, any future live vote event for the processId will be discarted.
```
jordipainan:
```suggestion
// Once this function is called, any future live vote event for the processId will be discarded.
```

vochain/indexer/vote.go
```diff
@@ -214,50 +210,42 @@ func (s *Indexer) GetEnvelopeHeight(processID []byte) (uint64, error) {
...
-func (s *Indexer) ComputeResult(processID []byte) error {
+func (s *Indexer) finalizeResults(processID []byte) error {
 	height := s.App.Height()
-	log.Debugf("computing results on height %d for %x", height, processID)
+	log.Debugw("finalize results", "processID", hex.EncodeToString(processID), "height", height)
```
jordipainan, line 217: I think the concept of finalize results should be better explained.

vochain/indexer/vote.go
```diff
@@ -214,50 +210,42 @@ func (s *Indexer) GetEnvelopeHeight(processID []byte) (uint64, error) {
...
 	if err != nil {
-		return fmt.Errorf("computeResult: cannot load processID %x from database: %w", processID, err)
+		return fmt.Errorf("finalizeResults: cannot load processID %x from state: %w", processID, err)
 	}
-	// Compute the results
-	var results *results.Results
-	if results, err = s.computeFinalResults(p); err != nil {
-		return err
+	if process.Status != models.ProcessStatus_RESULTS {
+		return fmt.Errorf("finalizeResults: processID %x is not in results status", processID)
```
jordipainan, line 225:
```suggestion
		return fmt.Errorf("finalizeResults: processID %x is not in RESULT status", processID)
```

vochain/indexer/vote.go
```diff
@@ -214,50 +210,42 @@ func (s *Indexer) GetEnvelopeHeight(processID []byte) (uint64, error) {
...
-		go l.OnComputeResults(results, p, height)
-	}
-	return nil
-}
 
-func joinHexBytes(list []types.HexBytes) string {
-	strs := make([]string, len(list))
-	for i, b := range list {
-		strs[i] = hex.EncodeToString(b)
+	// Remove the process from the live results
```
jordipainan, line 240:
```suggestion
	// Remove the process from live results
```

vochain/indexer/vote.go
```diff
@@ -285,12 +273,12 @@ func (s *Indexer) GetResultsWeight(processID []byte) (*big.Int, error) {
 }
 
 // unmarshalVote decodes the base64 payload to a VotePackage struct type.
-// If the vochain.VotePackage is encrypted the list of keys to decrypt it should be provided.
+// If the state.VotePackage is encrypted the list of keys to decrypt it should be provided.
```
jordipainan, line 276:
```suggestion
// If the state.VotePackage is encrypted the list of keys to decrypt should be provided.
```

vochain/state/process.go
```diff
@@ -237,8 +237,8 @@ func (v *State) SetProcessStatus(pid []byte, newstatus models.ProcessStatus, com
 			return fmt.Errorf("process %x already in ready state", pid)
 		}
 	case models.ProcessStatus_ENDED:
-		if currentStatus != models.ProcessStatus_READY {
-			return fmt.Errorf("process %x can only be ended from ready status", pid)
+		if currentStatus != models.ProcessStatus_READY && currentStatus != models.ProcessStatus_PAUSED {
```
jordipainan, line 240: Why this?

dockerfiles/testsuite/start_test.sh
p4u: Its not. The seed needs to be able to bootstrap the network, in a slow environment such as some CIs, it might require around 10 seconds.

statedb/treeupdate.go
p4u: The Del method is only used for the NoState() database currently.

vochain/state/process.go
```diff
@@ -295,66 +295,31 @@ func (v *State) SetProcessStatus(pid []byte, newstatus models.ProcessStatus, com
...
-// SetProcessResults sets the results for a given process
-func (v *State) SetProcessResults(pid []byte, result *models.ProcessResult, commit bool) error {
+// SetProcessResults sets the results for a given process and calls the event listeners.
+func (v *State) SetProcessResults(pid []byte, result *models.ProcessResult) error {
 	process, err := v.Process(pid, false)
 	if err != nil {
 		return err
 	}
 	// Check if the state transition is valid
 	// process must be ended, ready or results for setting the results
```
jordipainan, line 305:
```suggestion
	// process must be ended or ready for setting the results.
```

vochain/genesis/genesis.go
p4u: All commits will be squashed. The final message will contain the breaking notice.

vochain/indexer/indexer_test.go
p4u: In Go, the numeric overflows do not panic, instead it starts over from the zero value (so it's like applying modulus over the max possible value).

vochain/indexer/vote.go
p4u: What you mean? This is a debug message. Its scope is to provide technical information to the developer.

jordipainan: Sorry I was not referring to this debug message in particular, but added the message here. I was referring to the concept of finalizing the results. It is a new concept at least for me and was not easy to understand the change at first glance. That is why I suggested to add more context, maybe on top of the function.

vochain/state/process.go
p4u: I think it makes more sense. I can see a use case where the organizer pauses the process, waits some time to have some kind of confirmation, and then ENDs it directly. Actually, I don't see a reason to not support this transition. 
Do you have any strong point against it?

vochain/ist/ist.go
```diff
@@ -0,0 +1,298 @@
...
+		return nil
+	}
+	for id, action := range actions {
+		switch action.ID {
+		case ActionComputeResults:
+			log.Debugw("compute results", "height", height, "id", fmt.Sprintf("%x", id), "attempt", action.Attempts)
+			// check if we have all the revealed keys, else reschedule
+			if !c.checkRevealedKeys(action.ElectionID) {
+				// we are missing some keys, we cannot compute the results yet
+				newHeight := height + (1 + action.Attempts*2)
```
jordipainan, line 201: Maybe it is useful creating a function for this. Something like `RescheduleHeight()`

vochain/results/compute.go
```diff
@@ -0,0 +1,120 @@
...
+// If the vochain.VotePackage is encrypted the list of keys to decrypt it should be provided.
+// The order of the Keys must be as it was encrypted.
+// The function will reverse the order and use the decryption keys starting from the
+// last one provided.
+func unmarshalVote(VotePackage []byte, keys []string) (*state.VotePackage, error) {
+	var vote state.VotePackage
+	rawVote := make([]byte, len(VotePackage))
+	copy(rawVote, VotePackage)
+	// if encryption keys, decrypt the vote
+	if len(keys) > 0 {
```
jordipainan, line 93: I do think that this condition is not required, the loop wont be executed if len(keys) = 0

vochain/results/compute.go
```diff
@@ -0,0 +1,120 @@
...
+		Weight:       new(types.BigInt).SetUint64(0),
+		Final:        true,
+		VoteOpts:     p.VoteOptions,
+		EnvelopeType: p.EnvelopeType,
+		BlockHeight:  st.CurrentHeight(),
+	}
+
+	var nvotes atomic.Uint64
+	lock := sync.Mutex{}
+	starTime := time.Now()
```
jordipainan: typo `startTime`

vochain/results/compute.go
```diff
@@ -0,0 +1,120 @@
...
+		log.Infow("computed results",
+			"process", fmt.Sprintf("%x", electionID),
+			"votes", nvotes.Load(),
+			"results", results.String(),
+			"elapsed", time.Since(starTime).String(),
+		)
+	}
+	results.EnvelopeHeight = nvotes.Load()
+	return results, err
+}
```
jordipainan:
```suggestion
	startTime := time.Now()
	err = st.IterateVotes(electionID, true, func(vote *models.StateDBVote) bool {
		keys := []string{}
		if p.EnvelopeType.EncryptedVotes {
			for _, k := range vote.EncryptionKeyIndexes {
				if k < types.KeyKeeperMaxKeyIndex && k < len(p.EncryptionPrivateKeys) {
					keys = append(keys, p.EncryptionPrivateKeys[k])
				} else {
					log.Warn("wrong vote: key index overflow or too many fields")
					return false
				}
			}
		}
		vp, err := unmarshalVote(vote.VotePackage, keys)
		if err != nil {
			log.Debugf("vote invalid: %v", err)
			return false
		}
		if err = results.AddVote(vp.Votes, new(big.Int).SetBytes(vote.Weight), &lock); err != nil {
			log.Warnf("addVote failed: %v", err)
			return false
		}
		nvotes.Add(1)
		return false
	})
	if err == nil {
		log.Infow("computed results",
			"process", fmt.Sprintf("%x", electionID),
			"votes", nvotes.Load(),
			"results", results.String(),
			"elapsed", time.Since(startTime).String(),
		)
	}
	results.EnvelopeHeight = nvotes.Load()
	return results, err
}
```

- Solved typo `starTime`
- You can remove unnecessary else block and initialize keys as an empty slice before the condition check. Thus reducing the nesting level
- You can remove the separate conditions for "encryptionKeyIndexes has too many fields" and "key index overflow", and merge them into a single condition.

jordipainan (approved):
I'm so happy that we can finally remove the Oracle. Thank you so much for the hard effort you putted here.
It is a very big task.
I think the IST logic is correct and I did not find any edge case after the review.
Some functions in IST related code can be optimized but nothing is a big deal, I made some suggestions but feel free to ignore them.

Thank you <3

vochain/state/process.go
jordipainan: Oki, nope any strong opinion. Just wanted to understand the context. Thank you :)

PR #903, opened 2023-05-03
api: add new endpoint to get the transactions count

mariajdab:
vocdoni/interoperability#31

### DIFF
--- a/apiclient/blockchain.go
+++ b/apiclient/blockchain.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	"go.vocdoni.io/dvote/api"
+	"go.vocdoni.io/dvote/httprouter/apirest"
 	"go.vocdoni.io/dvote/types"
 	"go.vocdoni.io/dvote/vochain/genesis"
 	"go.vocdoni.io/proto/build/go/models"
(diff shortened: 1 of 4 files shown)


mariajdab:
Tried to replicate the error in `job_compose_test` but it was not possible

apiclient/blockchain.go
```diff
@@ -128,3 +129,23 @@ func (c *HTTPclient) OrganizationsBySearchTermPaginated(
...
+		return 0, err
+	}
+	if code != apirest.HTTPstatusOK {
+		return 0, fmt.Errorf("%s: %d (%s)", errCodeNot200, code, resp)
+	}
+	txsCount := new(struct {
+		Count uint64 `json:"count"`
+	})
+
+	err = json.Unmarshal(resp, txsCount)
```
p4u: We always try to do (when possible) `if err = ...`

p4u: Actually you can do:

`return txsCount.Count, json.Unmarshal(resp, txsCount)`

p4u (approved):
Great work.
Please, remember to post a message on the interoperability repository to announce the method is already implemented.

p4u:
squash your commits please

mariajdab:
For now, I will not include `censusize end2endtest`. And I will check with @altergui why the `job_compose_test` is failing with: `account not found`, but the docker local test and using the api seems ok

p4u:
I think you should also remove the changes to start_test.sh
The new tests have been introduced by the last PR merged, see https://github.com/vocdoni/vocdoni-node/blob/master/dockerfiles/testsuite/start_test.sh#L105

p4u:
@mariajdab I think we are fine for squash & merge

PR #905, opened 2023-05-04
error handling and redundant alias

mariajdab:


p4u:
Great, thanks for taking care of these little things :+1:

Issue #904, opened 2023-05-04
CI: account not found


p4u:
Occasionally, the CI fails as follows. 

```
2023-05-04 11:37:23+00:00 ### Logs from failed test (e2etest_cenususizelection) ###
2023-05-04T11:35:36.893Z INF end2endtest/main.go:113 > logger construction succeeded at level debug with output stdout
2023-05-04T11:35:36.893Z INF end2endtest/main.go:114 > starting end2endtest version=0e29576-dirty
2023-05-04T11:35:36.893Z INF end2endtest/main.go:118 > no keys passed, generated random private key: [675b5d6eed9dc11cf3a7e05951fb3f9738ffa198a14f27da0b11191d2f853916]
2023-05-04T11:35:36.893Z INF end2endtest/main.go:128 > privkey 675b5d6eed9dc11cf3a7e05951fb3f9738ffa198a14f27da0b11191d2f853916 = account 0x976c4215842f3b5Df91cb707f5eb92fF1A3Ee3e5
2023-05-04T11:35:36.893Z DBG end2endtest/main.go:173 > connecting to http://gateway0:9090/v2
2023-05-04T11:35:36.893Z DBG apiclient/client.go:166 > http request body=null path=/v2/chain/info type=GET
2023-05-04T11:35:36.895Z DBG apiclient/client.go:166 > http request body=null path=/v2/accounts/0x976c4215842f3b5Df91cb707f5eb92fF1A3Ee3e5 type=GET
2023-05-04T11:35:36.897Z INF end2endtest/helpers.go:65 > creating Vocdoni account 0x976c4215842f3b5Df91cb707f5eb92fF1A3Ee3e5
2023-05-04T11:35:36.897Z DBG data/ipfs.go:354 > computed cid: bagaaieraeryh2ucpq7amm42fpv6uaevlsri4hsb7io36e5hqwgreyoq3qw4a
[...]

PR #906, opened 2023-05-04
vochain/indexer: move transactions and token transfers to the sql blockTx

mvdan:

As a plus, the concept of "live goroutines" and waiting for them is now gone, because we now commit `blockTx` as part of `Indexer.Commit`, and that is synchronous.


mvdan:
Had a failure but it was https://github.com/vocdoni/vocdoni-node/issues/904.

p4u:
that's great! thanks

PR #908, opened 2023-05-07
vochain/indexer: deduplicate two App.State.Process calls

mvdan:
(see commit message)


mvdan:
I'm starting to realise just how inefficient `Indexer.Commit` is :)

p4u:
Definitely, the addLiveVote() change will make a difference… The rest of the optimizations are great, but I don't think they would affect the performance much. 

Good catch... that's what happens when a code grows without proper design.

p4u (approved):

PR #909, opened 2023-05-07
vochain: upgrade to cometBFT 0.37.1

p4u:
CometBFT is the new name for Tendermint.


p4u:
Need to figure how to sort out this:

```
vochain/start.go:202:2: tconfig.Mempool.TTLDuration is deprecated: Only used by priority mempool, which will be removed in the next major release.
vochain/start.go:203:2: tconfig.Mempool.TTLNumBlocks is deprecated: Only used by priority mempool, which will be removed in the next major release.
```

@altergui do you remember the mechanism we used on 0.34 before TTLDuration/TTLNumBlocks became available? 

Apart from that, the migration has been pretty smooth. The initial benchmarks and tests look good. The network feels more stable.

jordipainan (approved):

p4u:
So it looks like cometbft priority-mempool is gonna be deprecated in favor of the old implementation[ clist-mempool](https://github.com/cometbft/cometbft/tree/v0.37.1/mempool/v0).
This is why these TTL methods are going to disappear in the next release.

The best solution I found to implement the TTL in the application layer is:

1. Once a new transaction enters the mempool, it is indexed by its hash to a memory map, storing also the entry timestamp. This is performed by CheckTx() function, in case of success.
2. On DeliverTx() the transactions validated (they are gonna be included in a block) are marked to be deleted from the memory map
3. On Commit() all marked transactions are purged from the memory map

At this point, we keep the memory map clean from the transactions included in the blocks. However there are still transactions in the mempool that might be invalid (i.e because the voting process is already ended). So we need to perform a purge based on TTL (because its cheaper than running a full Check again and its fine if we keep those transactions for some minutes, even if they are useless).  To achieve this, we can do:

1. Enable transaction recheck (it was disabled in order to not re-check all existing mempool transactions on every new block which is very painful for performance)
2. Write a special case if recheck on CheckTx() method of the BaseApplication.
3. This recheck custom code, needs to perform the TTL purge. Walk over all transactions in the memory map and remove those who are too old.

This solution is a bit expensive but less than enabling full recheck. We need to constantly compute `hash(txData)` in order to build the key index of the memory map. However hashing is nowadays quite cheap.

p4u:
A potential modification of this approach that might be less expensive is to: instead of using recheck (so performing the TTL purge on every block) we might just have a goroutine which every N minuts triggers a TTL purge.

altergui:
regarding the approach without tx recheck:
what would be the purpose of keeping the txs in the mempool, if they are not going to be recheck? i mean, AFAIU, if the tx is kept into the mempool after a block is commited, is to be "reevaluated for inclusion" in a later block.
if that reevaluation doesn't happen, it seems to me all pending txs will simply be TTL'ed out.
or i am missing something between CheckTx, DeliverTx and Commit?

p4u:
> regarding the approach without tx recheck: what would be the purpose of keeping the txs in the mempool, if they are not going to be recheck? i mean, AFAIU, if the tx is kept into the mempool after a block is commited, is to be "reevaluated for inclusion" in a later block. if that reevaluation doesn't happen, it seems to me all pending txs will simply be TTL'ed out. or i am missing something between CheckTx, DeliverTx and Commit?

This "reevaluation" is the mempool.Recheck that we have disabled (its very expensive to keep). 

The TTL is implemented in the mempool v1 (priority-mempool) and not in the mempool v0 (clist-mempool). However the `v1` is gonna be deprecated thus we need to use `v0` (then no TTL so we need to manage it at the application layer).

p4u:
I already implemented the TTL mechanism here https://github.com/vocdoni/vocdoni-node/pull/909/commits/70b499643b8f0c870e2d4e76d08ca666f234e60a

PR #912, opened 2023-05-08
vochain/indexer: start benchmarking votes

mvdan:


mvdan:
this will let us see whether using blockTx for the votes really helps or not, and by how much.

vochain/indexer/bench_test.go
```diff
@@ -59,28 +38,31 @@ func benchmarkIndexTx(b *testing.B) {
...
+	b.ReportAllocs()
+	b.ResetTimer()
 	for i := 0; i < b.N; i++ {
 		for j := int32(0); j < 2000; j++ {
 			vote := &state.Vote{
 				Height:      uint32(util.RandomInt(10, 10000)),
 				ProcessID:   pid,
 				Nullifier:   util.RandomBytes(32),
-				VotePackage: []byte("{[\"1\",\"2\",\"3\"]}"),
+				VotePackage: []byte(`{"votes": [1,0,1]}`),
```
p4u, line 49: The correct way is now `vp, err := state.NewVotePackage([]int{1, 0}).Encode()`
However, there are more benchmarks that require this change, so I'll open an issue to make it at once.

Issue #914, opened 2023-05-08
votePackage json to struct


p4u:
Replace old VotePackage format `VotePackage: []byte("{[\"1\",\"2\",\"3\"]}")` by its new way of definition `vp, err := state.NewVotePackage([]int{1, 0}).Encode()`

There are some remaining files (specially benchmarks) that sill implement the old method.

- vochain/indexer/bench_test.go
- vochain/indexer/db_benchmark_test.go
- and maybe more

PR #915, opened 2023-05-09
test: update to use NewVotePackage and add some minor fixes

mariajdab:
- #914


mariajdab:
I updated `fnMempoolRemoveTxTendermint` but not sure if it will work for not test cases

p4u:
Looks good. Please rebase with master and then squash.
```
git fetch origin
git merge --ff origin/master
```

p4u (approved):

p4u:
> I updated `fnMempoolRemoveTxTendermint` but not sure if it will work for not test cases

I already fixed this issue. But your fix is also nice to have, so let's keep it for safety.

PR #920, opened 2023-05-11
F/upgrade ipfs && fix for files > 256 KiB

p4u:
see commit messages


data/downloader/downloader.go
```diff
@@ -131,20 +131,21 @@ func (d *Downloader) handleImport(item *DownloadItem) {
...
 	}
-	ctx, cancel = context.WithTimeout(context.Background(), ImportPinTimeout)
-	defer cancel()
-	data, err := d.RemoteStorage.Retrieve(ctx, item.URI, 0)
-	if err != nil {
-		log.Warnw("could not retrieve file", "uri", item.URI, "error", fmt.Sprintf("%v", err))
-	} else {
-		if item.Callback != nil {
-			go item.Callback(item.URI, data)
+	go func() {
```
mvdan, line 140: why a goroutine now? we should start documenting these; otherwise we're sprinkling goroutines everywhere and it's hard to keep track why they're needed.

data/downloader/downloader.go
```diff
@@ -131,20 +131,21 @@ func (d *Downloader) handleImport(item *DownloadItem) {
...
+		ctx, cancel := context.WithTimeout(context.Background(), ImportPinTimeout)
+		defer cancel()
+		if err := d.RemoteStorage.Pin(ctx, item.URI); err != nil {
+			log.Warnw("could not pin file", "uri", item.URI, "error", fmt.Sprintf("%v", err))
 		}
+	}()
+	if item.Callback != nil {
+		go item.Callback(item.URI, data)
```
mvdan, line 148: IMO callbacks should never be asynchronous. if the callback needs to do anything async, it can fire a goroutine on its own.

mvdan (commented):
overall SGTM, though I have little experience with the IPFS APIs. with how much ipfs evolves over time, we should also watch out for stuff getting deprecated, sometimes entire modules get replaced upstream.

data/downloader/downloader.go
p4u: The main objective of this callback is to retrieve the data, this is done Sync. 
Pining is a "nice-to-have" and not required for the own node (but useful for others so they can fetch also the data from other sources). This is the reasoning behind the pin.

I'll document it.

PR #923, opened 2023-05-13
log: add proper tests, add benchmark, small speed-up

mvdan:

I'd like us to move to Go's upcoming official structured logger (https://pkg.go.dev/golang.org/x/exp/slog), but first I wanted to make sure we properly covered the existing behavior with tests.

### DIFF
--- a/log/log_error_test_out.txt
+++ b/log/log_error_test_out.txt
@@ -1,0 +1,3 @@
+2006-01-02 15:04:05 +0000 UTC ERR log/log_test.go:29 > cannot commit to blockstore: some error
+2006-01-02 15:04:05 +0000 UTC WRN log/log_test.go:30 > various types duration=1000 list=[10,0,-10]
+2006-01-02 15:04:05 +0000 UTC ERR log/log_test.go:35 > some error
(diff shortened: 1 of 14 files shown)


mvdan:
I don't really understand why the gateway host isn't found in the compose tests. When running locally, the gateway seems to start OK. The logs don't show any gateway errors either.

p4u:
> I don't really understand why the gateway host isn't found in the compose tests. When running locally, the gateway seems to start OK. The logs don't show any gateway errors either.

Umm. You don't see the logs using `docker-compose logs -f gateway`  after executing `docker-compose up -d` ?

mvdan:
Locally I see the gateway starting normally with no errors in the logs. I mean that compose_test fails but it doesn't seem to show me the gateway logs, where presumably there's an error or a panic.

p4u:
Sometimes web log interpretation does not work properly (maybe too much data).
But Raw logs work (search for `panic`).

image

mvdan:
Thanks - I did not notice that in the raw logs. Should be fixed now.

log/log_error_test_out.txt
p4u, line 1: I don't understand the approach to these files. Could you clarify, please?

I suppose you are not raw comparing, since the file contains timestamps and other very static data.

mvdan: I am comparing contents directly. See log_test.go for the sample logs I do. The timestamps get replaced via a hook in log.go so that the output is stable, avoiding time.Now.

Issue #863, opened 2023-03-24
 Extend end-to-end tests to cover more voting scenarios


p4u:
We need to extend the end-to-end tests with a new voting test named something like "test voting protocol". This test should cover various voting scenarios and capabilities of the Vocdoni protocol to ensure the correct behavior and stability of the system.

The test should include the following scenarios and verifications:

- [x] Secret Until The End Elections: the apiclient Vote method would transparently encrypt the vote if the election is secretUntilTheEnd (#846)

- [x] Max Vote Overwrite: Verify that the same voter can send up to N votes, and each new vote should override the previous one. This will ensure the maxVoteOverwrite feature works as intended.

- [x] Max Census Size Restriction: Test the behavior when creating an election with a maxCensusSize smaller than the actual census size. Ensure that no additional votes can be cast once the limit is reached, thus validating the maxCensusSize restriction feature.

- [x] Ballot Protocol: Test different ballotProtocol configurations and validate that the final results are computed correctly based on the configuration parameters (uniqueValues, costFromWeight, maxCount, maxValue, maxTotalCost, costExponent). This will ensure the versatility and correctness of the ballot protocol (see https://hackmd.io/Sd08WpUPSVu6DEK9hSA-6Q)

- [x] Election Status: Test and verify the transitions of an election (READY, PAUSED, RESULTS, etc.) to ensure that only allowed transitions work correctly. This will validate the robustness of the election status management.

- [x] Un-interruptible Enforcement: Check if the interruptible flag works as expected by ensuring that elections with this flag set to false cannot be interrupted. This will validate the proper enforcement of the un-interruptible feature.

- [x] Dynamic Census: Test the dynamic census flag and feature by verifying that it works as expected in various scenarios. This will ensure the correct functionality of the dynamic census feature.

- [x] CSP Election: this was implemented in vochaintest as csptest. migrate it to end2endtest

altergui:
i'm doing "Secret Until The End Elections" and @mariajdab will start with "Max Vote Overwrite" today

p4u:
We should create a modular and flexible interface that all test implementations must follow. This will allow the main() function to call the interface methods of the selected test without having to modify the core code for each new test case.

### Goals
+ Design and implement a flexible and modular interface for all Vochain-related tests.
+ Refactor the current integration testing program to use the new interface.
+ Update documentation and provide examples for using the new interface.

### Interface

```go
type VochainTest interface {
[...]

PR #925, opened 2023-05-15
end2endtest: add ballot test

mariajdab:
- #863

### DIFF
--- a/cmd/end2endtest/ballot.go
+++ b/cmd/end2endtest/ballot.go
@@ -8,41 +8,52 @@ import (
 	"sync"
 	"time"
 
-	vapi "go.vocdoni.io/dvote/api"
 	"go.vocdoni.io/dvote/apiclient"
 	"go.vocdoni.io/dvote/crypto/ethereum"
 	"go.vocdoni.io/dvote/log"
-	"go.vocdoni.io/dvote/types"
+	"go.vocdoni.io/proto/build/go/models"
 )
 
 func init() {
-	ops["overwritelection"] = operation{
-		test: &E2EOverwriteElection{},
-		description: "Checks that the MaxVoteOverwrite feature is correctly implemented, even if a vote is consecutive " +
-			"overwrite without wait the next block, that means the error in checkTx: overwrite count reached, it's not raised",
-		example: os.Args[0] + " --operation=overwritelection --votes=1000",
+	ops["ballotelection"] = operation{
+		test:        &E2EBallotElection{},
+		description: "ballot election with unique values, maxCount, maxValue, maxTotalCost to test different ballotProtocol configurations",
+		example:     os.Args[0] + " --operation=ballotelection --votes=1000",
 	}
 }
 
-var _ VochainTest = (*E2EOverwriteElection)(nil)
+var _ VochainTest = (*E2EBallotElection)(nil)
 
-type E2EOverwriteElection struct {
+type E2EBallotElection struct {
 	e2eElection
 }
 
-func (t *E2EOverwriteElection) Setup(api *apiclient.HTTPclient, c *config) error {
+func (t *E2EBallotElection) Setup(api *apiclient.HTTPclient, c *config) error {
 	t.api = api
 	t.config = c
 
-	ed := newTestElectionDescription()
-	ed.ElectionType = vapi.ElectionType{
-		Autostart:     true,
-		Interruptible: true,
+	p := &models.Process{
+		StartBlock: 0,
+		BlockCount: 100,
+		Status:     models.ProcessStatus_READY,
+		EnvelopeType: &models.EnvelopeType{
+			EncryptedVotes: false,
+			UniqueValues:   true},
+		CensusOrigin: models.CensusOrigin_OFF_CHAIN_TREE_WEIGHTED,
+		VoteOptions: &models.ProcessVoteOptions{
+			MaxCount:     2,
+			MaxValue:     6,
+			MaxTotalCost: 10,
+			CostExponent: 1,
+		},
+		Mode: &models.ProcessMode{
+			AutoStart:     true,
+			Interruptible: true,
+		},
+		MaxCensusSize: uint64(t.config.nvotes),
 	}
-	ed.VoteType = vapi.VoteType{MaxVoteOverwrites: 2}
-	ed.Census = vapi.CensusTypeDescription{Type: vapi.CensusTypeWeighted}
 
-	if err := t.setupElection(ed); err != nil {
+	if err := t.setupElectionRaw(p); err != nil {
 		return err
 	}
 
@@ -50,18 +61,30 @@ func (t *E2EOverwriteElection) Setup(api *apiclient.HTTPclient, c *config) error
 	return nil
 }
 
-func (t *E2EOverwriteElection) Teardown() error {
+func (t *E2EBallotElection) Teardown() error {
 	// nothing to do here
 	return nil
 }
 
-func (t *E2EOverwriteElection) Run() error {
+func (t *E2EBallotElection) Run() error {
 	c := t.config
 	api := t.api
 
+	bdata := ballotData{
+		maxValue:     t.election.TallyMode.MaxValue,
+		maxCount:     t.election.TallyMode.MaxCount,
+		maxTotalCost: t.election.TallyMode.MaxTotalCost,
+		costExponent: t.election.TallyMode.CostExponent,
+	}
+
+	votes, expectedResults := ballotVotes(bdata, t.config.nvotes)
+
 	// Send the votes (parallelized)
 	startTime := time.Now()
+
 	wg := sync.WaitGroup{}
+	apiClientMtx := &sync.Mutex{}
+	indexv := 0
 	voteAccounts := func(accounts []*ethereum.SignKeys, wg *sync.WaitGroup) {
 		defer wg.Done()
 		log.Infof("sending %d votes", len(accounts))
@@ -69,16 +92,20 @@ func (t *E2EOverwriteElection) Run() error {
 		votesSent := 0
 		contextDeadlines := 0
 		for _, acc := range accounts {
-			ctxDeadline, err := t.sendVote(voteInfo{voterAccount: acc, choice: []int{0}}, nil)
+			apiClientMtx.Lock()
+			ctxDeadline, err := t.sendVote(voteInfo{voterAccount: acc, choice: votes[indexv]}, nil)
 			if err != nil {
+				apiClientMtx.Unlock()
 				log.Error(err)
 				break
 			}
+			indexv++
 			contextDeadlines += ctxDeadline
 			votesSent++
+
+			apiClientMtx.Unlock()
 		}
 		log.Infof("successfully sent %d votes... got %d HTTP errors", votesSent, contextDeadlines)
-		time.Sleep(time.Second * 4)
 	}
 
 	pcount := c.nvotes / c.parallelCount
@@ -90,26 +117,9 @@ func (t *E2EOverwriteElection) Run() error {
 		wg.Add(1)
 		go voteAccounts(t.voterAccounts[i:end], &wg)
 	}
-
 	wg.Wait()
 	log.Infof("%d votes submitted successfully, took %s (%d votes/second)",
 		c.nvotes, time.Since(startTime), int(float64(c.nvotes)/time.Since(startTime).Seconds()))
-
-	// overwrite the previous vote (choice 0) associated with account of index 0, using enough time to do it in the nextBlock
-	// try to make 3 overwrites (number of choices passed to the method). The last overwrite should fail due the maxVoteOverwrite constrain
-	ctxDeadlines, err := t.overwriteVote([]int{0, 1, 0}, 0, nextBlock)
-	if err != nil {
-		return err
-	}
-	log.Infof("the account %v send an overwrite vote, got %d HTTP errors", t.voterAccounts[0].Address(), ctxDeadlines)
-	time.Sleep(time.Second * 5)
-
-	// now the overwrite vote is done in the sameBlock using account of index 1
-	if ctxDeadlines, err = t.overwriteVote([]int{1, 1, 0}, 1, sameBlock); err != nil {
-		return err
-	}
-	log.Infof("the account %v send an overwrite vote, got %d HTTP errors", t.voterAccounts[1].Address(), ctxDeadlines)
-	time.Sleep(time.Second * 5)
 
 	// Wait for all the votes to be verified
 	log.Infof("waiting for all the votes to be registered...")
@@ -145,20 +155,16 @@ func (t *E2EOverwriteElection) Run() error {
 	// Wait for the election to be in RESULTS state
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second*100)
 	defer cancel()
+
 	elres, err := api.WaitUntilElectionResults(ctx, t.election.ElectionID)
 	if err != nil {
 		return err
 	}
 
-	// should count the first overwrite
-	expectedResults := [][]*types.BigInt{
-		votesToBigInt([]uint64{uint64(c.nvotes-2) * 10, 20, 0}),
+	if !matchResults(elres.Results, expectedResults) {
+		return fmt.Errorf("election result must match, expected Results: %v but got Results: %v", expectedResults, elres.Results)
 	}
 
-	// only the first overwrite should be valid in the results and must math with the expected results
-	if !matchResults(elres.Results, expectedResults) {
-		return fmt.Errorf("election result must match, expected Results: %s but got Results: %v", expectedResults, elres.Results)
-	}
 	log.Infof("election %s status is RESULTS", t.election.ElectionID.String())
 	log.Infof("election results: %v", elres.Results)
 
--- a/cmd/end2endtest/censusize.go
+++ b/cmd/end2endtest/censusize.go
@@ -12,6 +12,7 @@ import (
 	"go.vocdoni.io/dvote/apiclient"
 	"go.vocdoni.io/dvote/crypto/ethereum"
 	"go.vocdoni.io/dvote/log"
+	"go.vocdoni.io/dvote/types"
 )
 
 func init() {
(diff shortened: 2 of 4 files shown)


p4u:
please @altergui review

p4u:
We need to find out why the CI is failing @altergui @mariajdab

altergui:
> We need to find out why the CI is failing @altergui @mariajdab

it's timing out in the middle of plaintext election :thinking: taking more than 3-4 minutes to generate 100 votes.
this PR doesn't touch plaintext election test at all. so my only guess is that this new ballot test is overwhelming the network, is it resource-intensive in any way? if not, (it's as light as the other tests) it might be simply that the CI testsuite can't handle so many tests concurrently.

cmd/end2endtest/ballot.go
```diff
@@ -0,0 +1,195 @@
...
+}
+
+func (t *E2EBallotElection) Run() error {
+	c := t.config
+	api := t.api
+
+	voteValues, expectedR := t.randomBallotVotes()
+
+	// convert expected results int slice to string slice to be used on compare the results later
+	time.Sleep(time.Second * 6)
```
altergui: why wait here?

mariajdab: To have some time to return the `expectedR` before use it in the goroutine below, but seems is not necessary

altergui: `expectedR` is returned before the sleep even starts. no need to wait

p4u:
@altergui please review again and either accept it or request more changes.

@mariajdab you will need to merge with existing master and fix the conflicts.

mariajdab:
@altergui this CI shows the error that I mentioned you previously. I think the results expected are Ok. The code is using `WaitUntilElectionResults`

altergui:
i looked at the logs and i conclude there's one vote that you expect to be discarded, it's not being discarded
```
VoteMode:uniqueValues:true 
TallyMode:maxCount:2 
          maxValue:6
          maxTotalCost:10
```
```
vote values generated[[1 1] [2 6] [4 0] [1 6] [8 4] [6 5] [1 3] [0 6] [6 3] [3 3]]
results expected[[10 20 10 0 10 0 10] [10 0 0 20 0 0 30]]
```

you expect 4 votes to be discarded: 
 * `[1 1]` (fails uniqueValues)
 * `[8 4]` (8 + 4 = 12 is more than maxTotalCost=10)
 * `[6 5]` (6 + 5 = 11 is more than maxTotalCost=10)
 * `[3 3]` (fails uniqueValues)
 
 and yet,
 ```
 got Results: [[10 20 10 0 10 0 20] [10 0 0 20 0 10 30]]
 ```
 
 those results indicate that the `[6 5]` vote was not discarded, it was counted as valid.
 
 so there's "off by one" bug somewhere

altergui:
or, maxTotalCost is not being checked at all.
and the [8 4] vote failed because of maxValue

i'm still doing tests

mariajdab:
> i looked at the logs and i conclude there's one vote that you expect to be discarded, it's not being discarded
> 
> ```
> VoteMode:uniqueValues:true 
> TallyMode:maxCount:2 
>           maxValue:6
>           maxTotalCost:10
> ```
> 
> ```
> vote values generated[[1 1] [2 6] [4 0] [1 6] [8 4] [6 5] [1 3] [0 6] [6 3] [3 3]]
> results expected[[10 20 10 0 10 0 10] [10 0 0 20 0 0 30]]
> ```
> 
> you expect 4 votes to be discarded:
> 
>     * `[1 1]` (fails uniqueValues)
> 
>     * `[8 4]` (8 + 4 = 12 is more than maxTotalCost=10)
> 
>     * `[6 5]` (6 + 5 = 11 is more than maxTotalCost=10)
> 
>     * `[3 3]` (fails uniqueValues)
> 
> 
> and yet,
> 
> ```
> got Results: [[10 20 10 0 10 0 20] [10 0 0 20 0 10 30]]
> ```
> 
> those results indicate that the `[6 5]` vote was not discarded, it was counted as valid.
> 
> so there's "off by one" bug somewhere

Yeah, agree that's actually I realized trying to test repeatedly to make sure that it truly pass and there is not 'possible error'. And this is the error that some times appears that count a vote that should not be counted in the final results.

altergui:
indeed, the issue is that maxTotalCost is not being checked. but i believe this is a bug on the API, that is not respecting the parameter you're passing (MaxTotalCost=10) and instead defaulting to a hardcoded math that never fails. i'll post details in a bit

altergui:
indeed. the value you're passing is not respected at all, TWICE :rofl: 
first apiclient is ignoring what you passed, and pushing a hardcoded math to the api
https://github.com/vocdoni/vocdoni-node/blob/af85c0571e2834cf85a1906e1ece99c12f8939a2/apiclient/election.go#L174
and then the api ALSO ignores what apiclient pushed, and also recalculates the same math from scratch
https://github.com/vocdoni/vocdoni-node/blob/e5f82f9c8ad680ed952e4b17d19a9b14f01fe61b/api/wallet.go#L405

altergui:
mmh nevermind, now i see the NewElectionRaw uses a different codepath.

altergui:
ah, found the issue. 
you are missing `CostExponent: 1,` in your VoteOptions

the default CostExponent is 0 , which means the vote `[6 5]` is computed as `6**0 + 5**0 = 2`

mariajdab:
> ah, found the issue. you are missing `CostExponent: 1,` in your VoteOptions
> 
> the default CostExponent is 0 , which means the vote `[6 5]` is computed as `6**0 + 5**0 = 2`

Got it!

altergui:
i'm preparing a commit, i'll push it here and if you like it just squash it with yours

cmd/end2endtest/helpers.go
```diff
@@ -137,10 +138,21 @@ func (t *e2eElection) isCensusSizeValid(censusID types.HexBytes) bool {
...
 
-func (t *e2eElection) createElection(electionDescrip *vapi.ElectionDescription) (*vapi.Election, error) {
-	electionID, err := t.api.NewElection(electionDescrip)
-	if err != nil {
-		return nil, err
+func (t *e2eElection) createElection(ed *vapi.ElectionDescription, prc *models.Process) (*vapi.Election, error) {
+	var electionID types.HexBytes
+	var err error
+
+	// TODO: improve this
```
altergui: this really complicated debugging the MaxTotalCost issue, i didn't expect createElection take two different codepaths.

a more idiomatic alternative i can suggest is making a `func (t *e2eElection) createElectionRaw`

altergui (changes requested):

cmd/end2endtest/helpers.go
```diff
@@ -322,17 +356,17 @@ func (t *e2eElection) overwriteVote(choices []int, indexAcct int, waitType strin
...
 			if strings.Contains(err.Error(), "overwrite count reached") {
-				log.Infof("error expected: %s", err.Error())
+				log.Debugw("error expected: %s", err.Error())
 			} else {
 				return 0, errors.New("expected overwrite error")
 			}
 		}
 		contextDeadlines += ctxDeadLine
 		switch waitType {
-		case nextBlock:
```
altergui: this looks totally unrelated to "add ballot test", is this a fix for vote overwrite? in that case please make a separate PR

cmd/end2endtest/helpers.go
```diff
@@ -306,11 +339,12 @@ func (t *e2eElection) setupElection(ed *vapi.ElectionDescription) error {
 		return err
 	}
 
-	t.proofs = t.generateProofs(root, ed.ElectionType.Anonymous)
+	t.proofs = t.generateProofs(root, anonymousVoting)
 
 	return nil
 }
 
+// overwriteVote allow to try to overwrite a previous vote given the index of the account, it can use the sameBlock or the nextBlock
```
altergui: this func comment also looks unrelated. being just a comment, it would not merit a PR, but if you're doing a PR for the overwrite fixes (sameBlock vs nextBlock) please move these changes to that PR as well.

cmd/end2endtest/helpers.go
```diff
@@ -379,6 +414,59 @@ func (t *e2eElection) sendVote(v voteInfo, apiClientMtx *sync.Mutex) (int, error
...
 
+// randomBallotVotes allow to create random vote values that possibly exceed the max value, max total cost and is not unique
+func (t *e2eElection) randomBallotVotes() ([][]int, [][]int) {
+	maxCount := int(t.election.TallyMode.MaxCount)
+	maxValue := int(t.election.TallyMode.MaxValue)
+	maxTotalCost := int(t.election.TallyMode.MaxTotalCost)
+	costExponent := int(t.election.TallyMode.CostExponent)
+
+	expecResults := make([][]int, maxCount)
+	voteValues := make([][]int, 0)
```
altergui: for clarity, (and consistency with other parts of the code)
i'd suggest `expectedResults`

and `voteValues` could be simply `votes`

cmd/end2endtest/helpers.go
```diff
@@ -406,13 +494,25 @@ func getCensusParticipantKey(voterAccount *ethereum.SignKeys, censusType string)
 	return key, nil
 }
 
-// matchResult compare the expected vote results base in the overwrite applied, with the actual result returned by the API
-func matchResult(results [][]*types.BigInt, expectedResult [][]string) bool {
-	// only has 1 question
-	for q := range results[0] {
-		if !(expectedResult[0][q] == results[0][q].String()) {
-			return false
+func matchResults(results [][]*types.BigInt, expectedResult [][]string) bool {
```
altergui: missing s in `expectedResults`

cmd/end2endtest/helpers.go
```diff
@@ -406,13 +494,25 @@ func getCensusParticipantKey(voterAccount *ethereum.SignKeys, censusType string)
...
+	// iterate over each question to check if the results match with the expected results
+	for i := 0; i < len(results); i++ {
+		for q := range results[i] {
+			if !(expectedResult[i][q] == results[i][q].String()) {
```
altergui: i'd rather make expectedResults a `[][]*types.BigInt`, you're testing small ints so you don't mind but if results are BigInt let's respect that

cmd/end2endtest/helpers.go
```diff
@@ -406,13 +494,25 @@ func getCensusParticipantKey(voterAccount *ethereum.SignKeys, censusType string)
...
+				return false
+			}
 		}
 	}
 	return true
 }
+
+func sliceIntToStr(intSlice [][]int) [][]string {
```
altergui: you also can drop this whole func if you make expectedResults a `[][]*types.BigInt`

cmd/end2endtest/main.go
```diff
@@ -110,6 +110,10 @@ func main() {
 	flag.CommandLine.SortFlags = false
 	flag.Parse()
 
+	if c.parallelCount > c.nvotes {
+		c.parallelCount = c.nvotes
+	}
+
```
altergui: this is actually an unrelated fix but in this case it was me :sweat_smile: if you make a separate PR ("end2endtest small fixes") i'll move it there

altergui: along with voteOverwrite fixes and so on

dockerfiles/testsuite/start_test.sh
```diff
@@ -116,14 +117,18 @@ e2etest_tokentxs() {
 	e2etest tokentxs
 }
 
-e2etest_overwritelection () {
+e2etest_overwritelection() {
   e2etest overwritelection
 }
 
-e2etest_cenususizelection () {
+e2etest_cenususizelection() {
```
altergui: another candidate to be moved into "e2etest small fixes"
cenususizelection -> censusizelection

mariajdab: Yeah, sure I will create a small PR for those

cmd/end2endtest/helpers.go
mariajdab: yeah no problem

cmd/end2endtest/ballot.go
```diff
@@ -0,0 +1,180 @@
...
+	var wg sync.WaitGroup
+	wg.Add(1)
+
+	go func() {
+		defer wg.Done()
+		resultsChan <- sliceIntToStr(expectedR)
+	}()
+
+	results := <-resultsChan
+	wg.Wait()
```
altergui: also, this whole block looks like something you did while debugging the expectedResults bug. (i remember there was a time.Sleep here as well)
i must say it's a very peculiar way of doing `results := sliceIntToStr(expectedR)`
and as i said, you really should make `expectedResults` a slice of BigInts, so you don't even need that sliceIntToStr conversion,
and you can delete this whole thing entirely and just use `expectedR` (renamed to `expectedResults` for consistency)

mariajdab:
@altergui Seems that there is something wrong, is not a frequent error but if you try to run around 8 times it will appears, but now the error is in `expectedResults`. I'm checking the last changes added to `randomBallotVotes` [job-link](https://github.com/vocdoni/vocdoni-node/actions/runs/5050540323/jobs/9108586183)

cmd/end2endtest/helpers.go
mariajdab: PR created https://github.com/vocdoni/vocdoni-node/pull/943

altergui:
i see you're still working on this so i'm not reviewing it, when you finish just re-request review

a quick question, why `expectedResults` vs `resultExpected` vs `expecResults`? if they are all the same concept, please keep the var name consistent.
also, small abbreviations like `expecResults` are most likely not worth the loss of readability vs 3 letters less, better keep the words complete.

mariajdav:
> i see you're still working on this so i'm not reviewing it, when you finish just re-request review
> 
> a quick question, why `expectedResults` vs `resultExpected` vs `expecResults`? if they are all the same concept, please keep the var name consistent.
> also, small abbreviations like `expecResults` are most likely not worth the loss of readability vs 3 letters less, better keep the words complete.

Yes, I am still working on this.

mariajdab:
> i see you're still working on this so i'm not reviewing it, when you finish just re-request review
> 
> a quick question, why `expectedResults` vs `resultExpected` vs `expecResults`? if they are all the same concept, please keep the var name consistent. also, small abbreviations like `expecResults` are most likely not worth the loss of readability vs 3 letters less, better keep the words complete.

Yes, there are some small fixes that I need to push

mariajdab:
> > i see you're still working on this so i'm not reviewing it, when you finish just re-request review
> > a quick question, why `expectedResults` vs `resultExpected` vs `expecResults`? if they are all the same concept, please keep the var name consistent. also, small abbreviations like `expecResults` are most likely not worth the loss of readability vs 3 letters less, better keep the words complete.
> 
> Yes, there are some small fixes that I need to push

I just pushed

cmd/end2endtest/censusize.go
```diff
@@ -154,13 +155,19 @@ func (t *E2EMaxCensusSizeElection) Run() error {
...
+	choices := []*types.BigInt{
+		// should not count the last vote
+		new(types.BigInt).SetUint64(uint64(c.nvotes-1) * 10),
+		// should count the firsts overwrite
+		new(types.BigInt).SetUint64(0),
+		new(types.BigInt).SetUint64(0),
+	}
+
+	// according to the first overwrite
+	expectedResults := [][]*types.BigInt{choices}
```
altergui: you can merge both assignments 
```go
expectedResults := [][]*types.BigInt{
  new(types.BigInt)....
 }
```

since `choices` is not used anywhere else

cmd/end2endtest/helpers.go
```diff
@@ -393,6 +441,76 @@ func (t *e2eElection) sendVote(v voteInfo, apiClientMtx *sync.Mutex) (int, error
...
+		new(types.BigInt).SetUint64(10),
+	}
+	// default results on choice 2 for 10 votes
+	resultsCh2 := []*types.BigInt{
+		new(types.BigInt).SetUint64(10), new(types.BigInt).SetUint64(0),
+		new(types.BigInt).SetUint64(0), new(types.BigInt).SetUint64(10),
+		new(types.BigInt).SetUint64(10), new(types.BigInt).SetUint64(0),
+		new(types.BigInt).SetUint64(10),
+	}
+
```
altergui: there must be a way to make this less repetitive

cmd/end2endtest/helpers.go
```diff
@@ -393,6 +441,76 @@ func (t *e2eElection) sendVote(v voteInfo, apiClientMtx *sync.Mutex) (int, error
 	return contextDeadline, nil
 }
 
+// ballotVotes allow to create random vote values that possibly exceed the max value, max total cost and is not unique
+func (t *e2eElection) ballotVotes() ([][]int, [][]*types.BigInt) {
```
altergui: i have the feeling this is not generic enough to belong to e2eElection, isn't it something specific of E2EBallotElection?

mariajdab: Yeah, agree. It could be just a function

cmd/end2endtest/helpers.go
```diff
@@ -393,6 +448,76 @@ func (t *e2eElection) sendVote(v voteInfo, apiClientMtx *sync.Mutex) (int, error
...
+
+	switch {
+	// less than 10 votes
+	case nvotes < 10:
+		// default results on choice 1 for less than 10 votes
+		resultsCh1 = votesToBigInt([]uint64{0, 0, 0, 0, 0, 0, 0})
+		// default results on choice 2 for less than 10 votes
+		resultsCh2 = votesToBigInt([]uint64{0, 0, 0, 0, 0, 0, 0})
+
+	// greater o equal than 10 votes
```
altergui, line 470: i believe this two assignments could be de-indented (and instead of a `switch` a simple `if`)
or, even, that this whole block is redundant (since it assigns the zero value?), but i'm not sure.

cmd/end2endtest/helpers.go
```diff
@@ -420,13 +545,22 @@ func censusParticipantKey(voterAccount *ethereum.SignKeys, censusType string) ([
...
+		for q := range results[i] {
+			if !(expectedResults[i][q].String() == results[i][q].String()) {
+				return false
+			}
 		}
 	}
 	return true
 }
+
+func votesToBigInt(votes []uint64) []*types.BigInt {
```
altergui, line 560: it would be much prettier for the caller if this would be 
`votesToBigInt(votes ...int)`

so the calls would be

`resultsCh1 = votesToBigInt(0, 10, 20, 0, 0, 0, 10)`

cmd/end2endtest/helpers.go
```diff
@@ -393,6 +448,76 @@ func (t *e2eElection) sendVote(v voteInfo, apiClientMtx *sync.Mutex) (int, error
 	return contextDeadline, nil
 }
 
+// ballotVotes from a default list of 10 vote values that exceed the max value, max total cost and is not unique will
+func ballotVotes(b ballotData, nvotes int) ([][]int, [][]*types.BigInt) {
+	votes := make([][]int, 0, nvotes)
+	var resultsCh1, resultsCh2 []*types.BigInt
```
altergui, line 454: it's not immediately clear to me what "choice" ("Ch") means in this var name: are you talking about "Field" or "Value"? (in the terms described in https://developer.vocdoni.io/protocol/ballot)

altergui (approved):
i made some more comments/suggestions,
but this PR has been in the queue for so long, that i'll prefer to take it as-is and approve it. feel free to merge it without making any changes

cmd/end2endtest/helpers.go
altergui: i'd rather meant to move this out of helpers.go , into ballot.go

so keep in helpers.go the code that is used by more than one test (like setupElection), and whatever code is used by just a single test (like ballotVotes) is kept into that specific test (ballot.go)

cmd/end2endtest/helpers.go
mariajdab: Yes, it's needed, the zero values to convert to the type`[]*types.BigInt` and then sum the valid votes

mariajdab:
Last suggestions were added

PR #928, opened 2023-05-17
 Add endpoint to get transactions references of a specific block

jordipainan:
Returns:
```
{
    "blockNumber": 123456,
    "transactionCount": 3,
    "transactions": [
        {
            "type": "transfer",
            "index": 0,
            "height": 1234,
            "hash": "0x0123456789abcdef"
        },
        {
            "type": "approve",
            "index": 1,
            "height": 1235,
            "hash": "0x0123456789abcdef"
        },
        {
            "type": "transfer",
            "index": 2,
            "height": 1236,
            "hash": "0x0123456789abcdef"
        }
    ]
}
```

### DIFF
--- a/api/chain.go
+++ b/api/chain.go
@@ -16,6 +16,8 @@ import (
 	"go.vocdoni.io/dvote/vochain/genesis"
 	"go.vocdoni.io/dvote/vochain/indexer"
 	"go.vocdoni.io/dvote/vochain/indexer/indexertypes"
+	"go.vocdoni.io/proto/build/go/models"
+	"google.golang.org/protobuf/proto"
 )
 
 const (
@@ -84,6 +86,14 @@ func (a *API) enableChainHandlers() error {
 		"GET",
 		apirest.MethodAccessTypePublic,
 		a.chainTxByIndexHandler,
+	); err != nil {
+		return err
+	}
+	if err := a.endpoint.RegisterMethod(
+		"/chain/transactions/reference/height/{height}",
+		"GET",
+		apirest.MethodAccessTypePublic,
+		a.chainTxByHeightHandler,
 	); err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


api/chain.go
```diff
@@ -461,6 +471,59 @@ func (a *API) chainTxByIndexHandler(msg *apirest.APIdata, ctx *httprouter.HTTPCo
...
+				tx.ProtoReflect().Descriptor().Oneofs().Get(0)).Name())
+
+		txRef, err := a.indexer.GetTxHashReference(block.Txs[i].Hash())
+		if err != nil {
+			return ErrTransactionNotFound
+		}
+		blockTxs.Transactions = append(blockTxs.Transactions, TransactionMetadata{
+			Type:   txType,
+			Index:  int32(i),
+			Height: uint32(txRef.Index),
```
mvdan, line 516: index and height are both indices? confusing :)

api/chain.go
```diff
@@ -461,6 +471,59 @@ func (a *API) chainTxByIndexHandler(msg *apirest.APIdata, ctx *httprouter.HTTPCo
...
+		Transactions:      make([]TransactionMetadata, len(block.Txs)),
+	}
+	for i := range block.Txs {
+		signedTx := new(models.SignedTx)
+		tx := new(models.Tx)
+		var err error
+		if err = proto.Unmarshal(block.Txs[i], signedTx); err != nil {
+			return ErrUnmarshalingServerProto.WithErr(err)
+		}
+		if err = proto.Unmarshal(signedTx.Tx, tx); err != nil {
```
mvdan: nit: use `:=` for these two, then no need for `var err error`

mvdan (approved):

api/chain.go
jordipainan: yes it is a bit confusing, but there is the need for changing this names in multiple places. For now we agree that index is the tx index in the block, and height is the tx index in the overall chain.

Issue #797, opened 2023-02-06
indexer: index new blocks


p4u:
Blockchain blocks are currently not being handled by the indexer. Instead, we just retrieve them from the Tendermint blockstore once requested by the API endpoint `/chain/blocks/height/:height`.

There is a need on the Vocdoni explorer side (https://dev.explorer.vote) to retrieve a summary list of blocks; thus they can be listed. This summary needs to include little information, so we can aggregate several blocks on the same page. 

Ideally, the indexer SQL database would store the following information.

+ height (primary key)
+ size of the block
+ dataHash
+ number of transactions included in the block
+ number of validator signatures 
+ validator proposer address (indexed)
+ number of rounds

Currently, the Vochain application does not have an even handler (https://github.com/vocdoni/vocdoni-node/blob/master/vochain/state/eventlistener.go#L20) for allowing subscription to new blocks, but it might be added. Somehow, this even callback should be executed from `vochain.BeginNewBlock()`.

Finally, the `vochain/indexer` can use this event call to store new block indxes on its SQL database.

An example on how a block looks like.

```json
{
  "data": {
    "txs": [
      "CioiKAgCEAEaIMXSRgGG9xLCIDZlc5dzbPS+jVmyn3DANgm4AgAAAAAAIAISQWSK4t2xAl9k9PHUZUhUmqdDJKVWKq461fhTYxOnC3U4N0ct29IUhO9bT0KkeM0lFBQLiySB64z9JeR9oWKStlIA"
    ]
  },
  "evidence": {
    "evidence": []
  },
  "header": {
    "appHash": "994F3A12A64EF57F798B453B285177B245738ED562589F958711DB4D950B0779",
    "chainId": "vocdoni-development-76",
    "consensusHash": "9C51FE2637CF9460E108B5008DDEE1D25ECBC8C77926296836786D39C64686E4",
    "dataHash": "5458374E6E80F97E434A7E1AC9619A965E287F37A11CF7F80D656469C3D8F0C3",
    "evidenceHash": "E3B0C44298FC1C149AFBF4C8996FB92427AE41E4649B934CA495991B7852B855",
    "height": 244,
    "lastBlockId": {
      "hash": "0A75E30D4DB8A45E0FEE477448F7DEED1A50ABBB98C0BE4E96E50791F4A26776",
      "parts": {
        "hash": "D2803563C09030D4F89F65350AA71A8C5D661C797F789847B41D3E73262A84B3",
        "total": 1
      }
[...]

PR #929, opened 2023-05-17
api: return block hash on block info

jordipainan:
https://github.com/vocdoni/interoperability/issues/36

### DIFF
--- a/api/api_types.go
+++ b/api/api_types.go
@@ -3,6 +3,7 @@ package api
 import (
 	"time"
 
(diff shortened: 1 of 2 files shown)


mvdan (approved):

p4u:
Actually, after having some conversations with @mvdan, he might implement the block indexing described here https://github.com/vocdoni/vocdoni-node/issues/797

So, the hash will be retrieved from the SQL database. Then we can merge this PR as it is right now and change it once the indexing is implemented (so we don't need to recompute the hash).

PR #932, opened 2023-05-18
vochain/indexer: sort by time in descending order

mvdan:
(see commit message)


p4u:
> Note that, since we paginate via offset and limit,
and since new rows in the database will tend to be newer/higher,
this means that the first "page" of results will now tend to shift.
This problem already existed before, but now it will be more obvious.

Should be fine IMO. @selankon just be aware this might happen.

PR #933, opened 2023-05-19
vochain/indexer: return tx hash with votes, small optimizations

### LINKED ISSUE #764: Store the txHash into the indexer table voteRef
Store the txHash into the indexer voteRef so we don't need to access the tendermint blockstore anymore on GetEnvelope() and GetEnvelopes()

mvdan:

Fixes #764.

### DIFF
--- a/test/testcommon/testutil/util.go
+++ b/test/testcommon/testutil/util.go
@@ -57,6 +57,13 @@ func (r *Random) RandomBytes(n int) []byte {
 	return b
 }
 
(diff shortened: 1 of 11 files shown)


mvdan:
I started this by adding one more column to the vote_references table, but then I realised that this is SQL, so we should just use JOINs :) This also means that we don't need to worry about calling `OnNewTx` before `OnVote` - the info is stored in separate tables at separate times, and joined when we query.

p4u (approved):

PR #934, opened 2023-05-21
end2endtest: non-interruptible and interruptible election

mariajdab:
- #863

### DIFF
--- a/cmd/end2endtest/lifecycle.go
+++ b/cmd/end2endtest/lifecycle.go
@@ -1,0 +1,150 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"strings"
+	"sync"
+	"time"
+
+	vapi "go.vocdoni.io/dvote/api"
+	"go.vocdoni.io/dvote/apiclient"
+	"go.vocdoni.io/dvote/log"
+	"go.vocdoni.io/proto/build/go/models"
+)
+
+func init() {
+	ops["lifecyclelection"] = operation{
+		test:        &E2ELifecycleElection{},
+		description: "Publishes two elections one to validate that the election cannot be interrupted and the other validate that the election can be interrupted to valid status",
+		example:     os.Args[0] + " --operation=lifecyclelection --votes=1000",
+	}
+}
+
+var _ VochainTest = (*E2ELifecycleElection)(nil)
+
+type E2ELifecycleElection struct {
+	elections []e2eElection
+}
+
+func (t *E2ELifecycleElection) Setup(api *apiclient.HTTPclient, c *config) error {
+	t.elections = append(t.elections, e2eElection{api: api, config: c})
+	t.elections = append(t.elections, e2eElection{api: api, config: c})
+
+	electionDescriptions := []struct {
+		d             *vapi.ElectionDescription
+		interruptible bool
+	}{
+		{d: newTestElectionDescription(), interruptible: false},
+		{d: newTestElectionDescription(), interruptible: true},
+	}
+
+	for i, ed := range electionDescriptions {
+		ed.d.ElectionType.Autostart = true
+		ed.d.ElectionType.Interruptible = ed.interruptible
+		ed.d.Census = vapi.CensusTypeDescription{Type: vapi.CensusTypeWeighted}
+
+		if err := t.elections[i].setupElection(ed.d); err != nil {
+			log.Errorf("failed to setup election %d: %v", i, err)
+		}
+	}
+
+	log.Debugf("election details: %+v  %+v", *t.elections[0].election, *t.elections[1].election)
+	return nil
+}
+
+func (t *E2ELifecycleElection) Teardown() error {
+	// nothing to do here
+	return nil
+}
+
+func (t *E2ELifecycleElection) Run() error {
+	var wg sync.WaitGroup
+	errChan := make(chan error, 2)
+
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+
+		api := t.elections[0].api
+		electionID := t.elections[0].election.ElectionID
+
+		isExpectedError := func(err error, status string) bool {
+			switch status {
+			case "READY":
+				return strings.Contains(err.Error(), "cannot set process status from READY to ready")
+			case "PROCESS_UNKNOWN":
+				return strings.Contains(err.Error(), "process status PROCESS_UNKNOWN unknown")
+			case "RESULTS":
+				return strings.Contains(err.Error(), "not authorized")
+			default:
+				return strings.Contains(err.Error(), "is not interruptible")
+			}
+		}
+
+		for _, status := range models.ProcessStatus_name {
+			if _, err := api.SetElectionStatus(electionID, status); err != nil {
+				if !isExpectedError(err, status) {
+					errChan <- fmt.Errorf("unexpected error for status %s: %v", status, err)
+					return
+				}
+				log.Debugf("expected error message: %v", err.Error())
+			}
+		}
+	}()
+
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+
+		api := t.elections[1].api
+		electionID := t.elections[1].election.ElectionID
+
+		// Define the desired election status transition
+		transitions := []struct {
+			fromStatus string
+			toStatus   string
+			errMsg     string
+		}{
+			{"PAUSED", "PAUSED", "cannot set process status from PAUSED to paused"},
+			{"PAUSED", "READY", ""},
+			{"READY", "RESULTS", "not authorized to set process status to RESULTS"},
+			{"READY", "CANCELED", ""},
+			{"CANCELED", "PAUSED", "cannot set process status from CANCELED to paused"},
+			{"CANCELED", "ENDED", "only be ended from ready or paused status"},
+			{"CANCELED", "READY", "cannot set process status from CANCELED to ready"},
+			{"CANCELED", "CANCELED", "already in canceled state"},
+		}
+
+		// Test each status transition
+		for _, t := range transitions {
+			hash, err := api.SetElectionStatus(electionID, t.toStatus)
+			if err != nil {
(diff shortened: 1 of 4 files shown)


cmd/end2endtest/helpers.go
```diff
@@ -31,6 +31,14 @@ type voteInfo struct {
 	keys         []vapi.Key
 }
 
+var processStatusName = map[int32]string{
```
p4u: This is wrong, since 0 is `UNKNOWN`.

Furthermore, instead of declaring this map here, you might use the map from `models.ProcessStatus_name`.

Models is an import from "go.vocdoni.io/proto/build/go/models"

cmd/end2endtest/interruptible.go
```diff
@@ -0,0 +1,105 @@
...
+	wg.Wait()
+	log.Infof("%d votes submitted successfully, took %s (%d votes/second)",
+		c.nvotes, time.Since(startTime), int(float64(c.nvotes)/time.Since(startTime).Seconds()))
+
+	status := randomProcessStatus()
+	if _, err := api.SetElectionStatus(t.election.ElectionID, status); err != nil {
+		if strings.Contains(err.Error(), "is not interruptible") {
+			log.Debugf("error expected: %v", err.Error())
+		} else {
+			return errors.New("expected not interruptible error")
```
p4u: We use `fmt.Errorf()` to construct error messages instead of the `errors` library.

p4u (changes requested):
I think the `Election Status` and `Un-interruptible Enforcement` tests from https://github.com/vocdoni/vocdoni-node/issues/863 can be merged into a single one (election-lifecycle) that mainly tests all the possible transitions of an election (with interruptible flag and without). 

See the election lifecycle documentation https://developer.vocdoni.io/protocol/overview#election-lifecycle-states

cmd/end2endtest/interruptible.go
```diff
@@ -0,0 +1,105 @@
...
+		if end > len(t.voterAccounts) {
+			end = len(t.voterAccounts)
+		}
+		wg.Add(1)
+		go voteAccounts(t.voterAccounts[i:end], &wg)
+	}
+
+	wg.Wait()
+	log.Infof("%d votes submitted successfully, took %s (%d votes/second)",
+		c.nvotes, time.Since(startTime), int(float64(c.nvotes)/time.Since(startTime).Seconds()))
```
altergui: i believe you can also skip the votes part. just set up the election (READY) and then simply try to interrupt it.

cmd/end2endtest/interruptible.go
```diff
@@ -0,0 +1,105 @@
...
+
+	status := randomProcessStatus()
```
altergui: why not test all the possibilities (4), instead of a random one per run? that makes the test more robust.

cmd/end2endtest/lifecycle.go
```diff
@@ -0,0 +1,97 @@
...
+		ed.d.ElectionType.Autostart = true
+		ed.d.ElectionType.Interruptible = ed.interruptible
+		ed.d.Census = vapi.CensusTypeDescription{Type: vapi.CensusTypeWeighted}
+
+		go func(index int, desc *vapi.ElectionDescription) {
+			if err := t.e2eElections[index].setupElection(desc); err != nil {
+				log.Errorf("failed to setup election %d: %v", index, err)
+			}
+
+			defer wg.Done()
```
altergui: no need to `defer` if it's the last line of the func
you'd normally use `defer wg.Done()` in the very first line of the func.

cmd/end2endtest/lifecycle.go
```diff
@@ -0,0 +1,97 @@
...
+		}(i, ed.d)
+
+		wg.Wait()
```
altergui: this also sounds misplaced to me, shouldn't it be outside of the for loop?

cmd/end2endtest/lifecycle.go
```diff
@@ -0,0 +1,97 @@
...
+
+var _ VochainTest = (*E2ELifecycleElection)(nil)
+
+type e2eElections []e2eElection
+
+type E2ELifecycleElection struct {
+	e2eElections
+}
+
+func (t *E2ELifecycleElection) Setup(api *apiclient.HTTPclient, c *config) error {
```
altergui, line 31: in general i feel this whole Setup is very complex: new structs and types are used, and a whole goroutine architecture to parallelize creating 2 elections (which is something the outer framework already solves)

i'd suggest to keep it simple: make lifecycle.go create a single election with interruptible=false. why do you want to handle two elections at the same time?

altergui (commented):
is this ready for review? i read the code and made some comments, then thought maybe it was not meant to be reviewed yet. sorry in that case for reviewing too early.
posting the comments here anyway in case it helps development :smile_cat:

cmd/end2endtest/lifecycle.go
mariajdab: @p4u suggested it in a early comment handle the two elections in one end2end test named lifecycle

cmd/end2endtest/lifecycle.go
mariajdab: Yep, the defer is not needed in this case

cmd/end2endtest/lifecycle.go
mariajdab: Yeah right!

cmd/end2endtest/helpers.go
```diff
@@ -564,3 +573,16 @@ func votesToBigInt(votes []uint64) []*types.BigInt {
...
 }
+
+func isExpectedError(err error, status string) bool {
+	switch status {
+	case "READY":
+		return strings.Contains(err.Error(), "cannot set process status from READY to ready")
+	case "PROCESS_UNKNOWN":
+		return strings.Contains(err.Error(), "process status PROCESS_UNKNOWN unknown")
+	case "RESULTS":
+		return strings.Contains(err.Error(), "not authoritzed")
```
altergui: this looked like a typo,
but then i checked and the typo (`z`) is actually on pau's code :man_facepalming: 
feel free to correct both sides (vochain/transaction/election_tx.go:200)

cmd/end2endtest/lifecycle.go
```diff
@@ -0,0 +1,137 @@
...
+func (t *E2ELifecycleElection) Teardown() error {
+	// nothing to do here
+	return nil
+}
+
+func (t *E2ELifecycleElection) Run() error {
+	var wg sync.WaitGroup
+	errChan := make(chan error, 2)
+
+	wg.Add(2)
```
altergui: tiny nit: do `wg.Add(1)` before each `go func()`, 

with `wg.Add(2)` i was expecting you'd run 2 instances of the same goroutine

cmd/end2endtest/helpers.go
```diff
@@ -564,3 +573,16 @@ func votesToBigInt(votes []uint64) []*types.BigInt {
 	}
 	return vBigInt
 }
+
+func isExpectedError(err error, status string) bool {
```
altergui: this name is misleading since it sounds like a general thing but it's actually specific to the Uninterruptible test.

so please move this into `lifecycle.go`

you could even include this inside the first `go func`, either as a var 
`isExpectedError := func(err error, status string) bool { ... }`
or not even a func, a simple struct like you did for `transitions`

cmd/end2endtest/lifecycle.go
```diff
@@ -0,0 +1,137 @@
...
+	go func() {
+		defer wg.Done()
+
+		api := t.e2eElections[0].api
+		electionID := t.e2eElections[0].election.ElectionID
+
```
altergui, line 85: you could put `isExpectedError := func(err error, status string) bool { ... }` here

altergui (approved):
i proposed a couple of changes but in general LGTM, feel free to merge it

cmd/end2endtest/lifecycle.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+}
+
+var _ VochainTest = (*E2ELifecycleElection)(nil)
+
+type e2eElections []e2eElection
+
+type E2ELifecycleElection struct {
+	e2eElections
+}
```
altergui: no need for the first type, just do

```
type E2ELifecycleElection struct {
	e2eElections []e2eElection
}
```

altergui: or even a shorter name
```
type E2ELifecycleElection struct {
	elections []e2eElection
}
```

cmd/end2endtest/helpers.go
```diff
@@ -64,6 +64,15 @@ func newTestElectionDescription() *vapi.ElectionDescription {
...
 
+func newE2EElections(api *apiclient.HTTPclient, c *config, numElections int) []e2eElection {
+	elections := make([]e2eElection, numElections)
+	for i := range elections {
+		elections[i].api = api
+		elections[i].config = c
+	}
+	return elections
+}
+
```
altergui: i still have the feeling that it would have been more straightforward to make two separate tests. (lifecycle.go and uninterruptible.go) since there's no code or logic shared between these two tests. i bet pau didn't have the implementation details in mind, but anyway now it's done, and it looks clean enough to merge it as-is.

but i wouldn't do this approach again, so this func newE2EElections would stay specific to lifecycle.go, please move it there. 

i believe this could even be just a simple snippet:
 ```
	t.e2eElections = append(t.e2eElections, e2eElection{api: api, config: c})
	t.e2eElections = append(t.e2eElections, e2eElection{api: api, config: c})
 ```

cmd/end2endtest/helpers.go
mariajdab: Yeah, is a typo. Fixed in both sides

altergui (approved):
perfect

mariajdab:
It said `Merging is blocked.` I think @p4u needs to also approve due he requested changes before

p4u (approved):

PR #936, opened 2023-05-22
testsuite: fix zkCensus artifacts cache path

altergui:
point the bind-mount to /tmp to avoid owner/permissions errors
inside git repo (dockerized node creates the files as root:root)


lucasmenendez (approved):

altergui:
just for the record, this change prevents a ~125MB download on each testsuite run, which might go unnoticed in CI servers but in my laptop it took around a minute, with a pretty standard wifi speed (20mbps throughput)

now the download is done only once and subsequent runs start instantly (unless  host /tmp/ cache is cleared, of course)

PR #939, opened 2023-05-23
integrate dynamic election price

p4u:
New elections (created by NewProcessTx), have currently a dynamic price computed by the `election price calculator`.

This calculator requires some factors and parameters:

- maxCensusSize, electionDuration, etc... are the election parameters used to compute the price of the election
- k1...k7 are static factors that determine the weight of each parameter
- network capacity (defined in the Genesis) is used to make elections expensive/cheaper and refers to the number of votes per block that the network can achieve
- base cost is a minimum cost for the new election

See the `electionpricecalc` package for a detailed explanation.

This is a breaking change and will reset the blockchain.


p4u:
Please, do not base the review on the "formula" for the election price. This is an on-going discussion that takes place in another channel. We will modify the formula or parameters accordingly.  Instead base your review in the code around it :)

PR #944, opened 2023-05-23
F/generate swaggo docs

selankon:
Related to https://github.com/vocdoni/interoperability/issues/65

Implement swagger info docs and document some endpoint.


altergui:
https://github.com/vocdoni/vocdoni-node/actions/runs/5058108768/jobs/9077758364?pr=944
* [x] Error: api/docs/docs.go:5:8: no required module provides package github.com/swaggo/swag; to add it:
  go get github.com/swaggo/swag

https://github.com/vocdoni/vocdoni-node/actions/runs/5058108768/jobs/9077758577?pr=944
* [x] please run 'go mod tidy' and include those changes in a commit

* [x] el error de docker-compose lo puedes ignorar. ("transaction never mined")

altergui:
ahora queda nomas hacer un `go generate ./...` (que hace un `swag fmt`) y commitear esos whitespace fixes pero esa diversion te la dejo a ti

chilcano:
Podemos mover los mismos comandos arriba mencionados a un github workflow y generar el swagger yaml file en alguna carpeta o exportarlo como github artifact ?

Quiero ejecutar [api_security_review.yaml](https://github.com/vocdoni/developer-portal/blob/DOPS-400-review-the-security-vocdoni-api/.github/workflows/api_security_review.yaml) para validar si el swagger resultante sigue las recomendaciones de seguridad de OpenAPI y OWASP.

Creo que este github workflow también deberia ejecutarse cuando el swagger yaml se genere en vocdoni-node repo.
Por el momento está como [PR-7](https://github.com/vocdoni/developer-portal/pull/7) en developer-portal repo.

altergui:
i reviewed this more carefully and finally opted for the option @chilcano proposed, of committing the resulting .yaml in the repo after all.
so it's just a oneliner 079f467

BUT i'm still not happy with the approach because, the dev needs to run `go generate ./...` after each change, and either commit those changes in the same commit (which makes the commit less "clean" IMO) or make a separate commit (which leaves the original commit clean but makes the whole git history dirtier)
as i commented, this is already the case with protobuf, but in that case it's a requirement since the generated files NEED to be in this same repo since it's code.
`swagger.yml` doesn't really need to be in this repo. i'll think about it overnight and propose something tomorrow

selankon:
Well, @altergui the swagger does not need to be on this repo. We can do the same approach as on the SDK: when merging to main (or deploying a new version), the swagger is generated and pushed on the developer portal with a metadata tag with the version that is referring. 

For example, if version xxxxx is being deployed, a parallel action runs generating the swagger, adding the version inside, and creating a PR to the developer portal with the changes. 

So, for every new deployed version of the API the only git history that is going to track this is a specific branch on the developer portal. 

Clean enough imho. 

(And then run the security tests actions there ?)

chilcano:
@altergui - that is right, the generated swagger file should not be part of code base, it is outcome of extract info from API golang code. For that, that generated swagger should be published as a github artifact and/or be included when building a package.

The most important here is this process should be automated from a new github actions/workflow. This will be triggered if there are changes in `vocdoni-node/api/**.go`. In this way, we can get versioned swagger published as artifact and can be used to feed the DeveLoper Portal, even using the same github action workflow. This process is similar to explained by @selankon above, but in this case is triggered from `vocdoni-node` repo.

wdyt?

altergui:
fixed the conflict with c752fa51
and also cleaned up git history a bit

api/chain.go
```diff
@@ -184,10 +184,14 @@ func (a *API) enableChainHandlers() error {
...
-//	@Summary		List organizations
-//	@Description	List the existing organizations
-//	@Success		200	{object}	object
-//	@Router			/chain/organizations/page/{page} [get]
+//		@Summary		List organizations
+//		@Description.markdown	organizationListHandler
+//		@Tags			Chain
+//		@Accept			json
+//		@Produce		json
+//	 	@Param			page	path		int		true	"Define de page number"
```
altergui: s/de/the/

api/chain.go
```diff
@@ -541,10 +595,14 @@ func (a *API) chainTxByIndexHandler(msg *apirest.APIdata, ctx *httprouter.HTTPCo
 
 // chainTxByHeightHandler
 //
-//	@Summary		Returns the list of transactions for a given block
+//	@Summary		Block transactions
```
altergui: may i suggest `Transactions in a block`?

it's just a suggestion because maybe `Block transactions` does fit in better in the actual context (i don't know), but without context it sounds to me like blocking a transaction.

api/docs/descriptions/accountSetHandler.md
```diff
@@ -0,0 +1,5 @@
+It register an account to the blockchain including IPFS metadata upload. 
+
+- Requires a signed protobuff transaction on the body to perform the operation.
```
altergui: `protobuf`

altergui (approved):
LGTM, i did not proofread thoroughly all the contents.

api/api_types.go
```diff
@@ -13,20 +13,20 @@ import (
 )
 
 type Organization struct {
-	OrganizationID types.HexBytes      `json:"organizationID,omitempty"`
+	OrganizationID types.HexBytes      `json:"organizationID,omitempty" swaggertype:"string"`
```
altergui: did you try `--overridesFile` instead?
https://github.com/swaggo/swag/#use-global-overrides-to-support-a-custom-type

apparently it would suffice with
`replace types.HexBytes string`
or maybe
`replace api/types.HexBytes string`

altergui: (avoiding the need to specify `swaggertype:"string"`, shaving 38 changes off the PR)

selankon: Before I discarded this option because I didn't get this working for `[]byte` or `[]Key`. However, it worked for `time.Time` and for `types.HexBytes` so I am going to add the changes on the next commit.

selankon: 0ad7a1a5

api/docs/descriptions/accountSetHandler.md
selankon: 0ad7a1a5

api/chain.go
selankon: 0ad7a1a5

api/chain.go
selankon: 0ad7a1a5

PR #945, opened 2023-05-23
ci: refactor github workflows 

altergui:
* make docker-release a reusable workflow
* call docker-release from main.yml when:
  * push to dev, stage, master or release**
  * will publish corresponding docker tags
* call docker-release from deploy-dev when:
  * push to master
  * master can be pushed cleanly to dev
  * will publish resulting image to dev docker tag

### DIFF
--- a/.github/workflows/deploy-dev.yml
+++ b/.github/workflows/deploy-dev.yml
@@ -1,4 +1,5 @@
-name: Deploy dev env
+---
+name: Deploy dev
 
 on:
   push:
@@ -6,8 +7,8 @@ on:
       - master
 
 jobs:
-  build:
-    name: Deploy main branch to dev env
+  push-to-dev:
+    name: Git push main -> dev
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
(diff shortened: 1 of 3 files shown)


altergui:
i kept all the current behaviour as-is (but with less `if:`s, much simpler yamls)
and added the "push master to dev" feature in a way that doesn't duplicate code (reusable workflow)

also, it fulfills @p4u idea:
* on each push to master:
  * if `dev` is an ancestor of master, then:
    * the branch is pointed to current master.
    * image is built and pushed to both docker `master` tag and `dev` tag
  * if `dev` is not an ancestor of master (it was pointed somewhere else manually, to test something)
    * `dev` is left untouched, until manually re-pointed.
    * `master` is of course built and updated, independently.
    * that commit in master is going to show :x: ("failed" check) but i believe this is a good idea. if that looks ugly, whenever `dev` tag is pointed back to `master`, the failed check can be re-run and it will now pass, marking the commit as :heavy_check_mark: 
    
* on each manual push to `dev`:
  * an image is built and pushed to docker `dev` tag

p4u:
Looks like a good solution, thanks @altergui

p4u (approved):

p4u (approved):

PR #948, opened 2023-05-24
ci: fixes and improvements for start_test.sh

altergui:
fix the check_gw_is_up loop that has been broken for a long time but no-one noticed, it waited 40 seconds unnecessarily

### DIFF
--- a/dockerfiles/testsuite/start_test.sh
+++ b/dockerfiles/testsuite/start_test.sh
@@ -163,21 +163,29 @@ for test in ${tests_to_run[@]}; do
 done
 
 log "### Waiting for tests to finish ###"
-wait
 
-failed=""
-for test in ${tests_to_run[@]} ; do
-	RET=$(cat $results/$test.retval)
-	if [ "$RET" == "0" ] ; then
-		log "### Vochain test $test finished correctly! ###"
-	else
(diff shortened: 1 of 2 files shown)


altergui:
also, abort all tests when one of them fails. not only saves time, it also allows inspecting the relevant logs.

bash is horrible so i did my best to implement this idea

dockerfiles/testsuite/start_test.sh
```diff
@@ -165,42 +163,36 @@ for test in ${tests_to_run[@]}; do
...
-	else
-		log "### Vochain test $test failed! ###"
-		log "### Post run logs ###"
-		$COMPOSE_CMD logs --tail 1000
-		failed="$test"
-		break
-	fi
-done
+wait_jobs() {
+	while [ -n "$(jobs)" ]; do
```
lucasmenendez, line 168: just to learn, where the variable `jobs` is defined?

lucasmenendez (approved):

p4u:
please rebase

dockerfiles/testsuite/start_test.sh
altergui: i can't recommend you to learn bash :see_no_evil:
but here we go: `jobs` is not a var but a command

PR #966, opened 2023-05-31
enable mempool recheck and remove mempool TTL

p4u:
Instead of the mempool TTL, enable recheck but only execute the transactions every 12 blocks.

This is safer, since a valid mempool transaction won't be removed by TTL.

And doing recheck every 12 blocks, probably not much more costly than keeping track of all mempool transactions and runing a garbage collector process.


altergui (approved):
absolutely approved

PR #968, opened 2023-05-31
vocone: call InitChain app method

p4u:
Before, the network capacity was not set and
the price calculator was not created.

Now both are properly set and we use the default
app method InitChain with a dumy genesis data.
It improves compatibility with the default vochain. Around this change, some small fixes.


emmdim (approved):
:crossed_fingers:

PR #974, opened 2023-06-01
ci: refactor e2etest sendvote

altergui:
* api: 'mempool is full' now produces a '503 Service Unavailable'
    
* api: new ErrVochainOverloaded that returns http.StatusServiceUnavailable (503)

* apiclient: Request now retries 3 times if mempool is full
* apiclient: WaitUntilHeight doesn't panic anymore when ChainInfo fails
* apiclient: simplify WaitUntilNextBlock

* e2etest: refactor t.sendVotes and apiclient.Vote
    
    * apiclient: Clone() now uses SetAccount() so zkAddr is properly init'd
    * apiclient: VoteData now has VoterAccount, so Vote() can use it if set
    * apiclient: Election now checks that electionID is not nil
    * e2etest: overwriteVote now uses sendVotes()
    * e2etest: plaintext now uses sendVotes()
    * e2etest: censusize replace t.sendVote with t.api.Vote

### DIFF
--- a/api/wallet.go
+++ b/api/wallet.go
@@ -117,16 +117,9 @@ func (a *API) walletSignAndSendTx(stx *models.SignedTx, wallet *ethereum.SignKey
 		return nil, err
 	}
 
-	resp, err := a.vocapp.SendTx(txData)
+	resp, err := a.sendTx(txData)
 	if err != nil {
 		return nil, err
-	}
-
-	if resp == nil {
-		return nil, ErrVochainEmptyReply
-	}
-	if resp.Code != 0 {
-		return nil, ErrVochainReturnedErrorCode.Withf("(%d) %s", resp.Code, string(resp.Data))
 	}
 
 	return &Transaction{
--- a/apiclient/client.go
+++ b/apiclient/client.go
@@ -2,6 +2,7 @@ package apiclient
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -29,6 +30,10 @@ const (
 	HTTPDELETE = "DELETE"
 
 	errCodeNot200 = "API error"
+
+	// This enables Request() to handle the situation where the server replies
+	// "mempool is full", it will wait for next block and retry sending the tx
+	DefaultRetries = 3
 )
 
 // HTTPclient is the Vocdoni API HTTP client.
@@ -40,6 +45,7 @@ type HTTPclient struct {
 	chainID string
 	circuit circuit.ZkCircuitConfig
 	zkAddr  *zk.ZkAddress
+	retries int
 }
 
 // NewHTTPclient creates a new HTTP(s) API Vocdoni client.
@@ -51,9 +57,10 @@ func NewHTTPclient(addr *url.URL, bearerToken *uuid.UUID) (*HTTPclient, error) {
 		ReadBufferSize:     1 * 1024 * 1024, // 1 MiB
 	}
 	c := &HTTPclient{
-		c:     &http.Client{Transport: tr, Timeout: time.Second * 8},
-		token: bearerToken,
-		addr:  addr,
+		c:       &http.Client{Transport: tr, Timeout: time.Second * 8},
+		token:   bearerToken,
+		addr:    addr,
+		retries: DefaultRetries,
 	}
 	data, status, err := c.Request(HTTPGET, nil, "chain", "info")
 	if err != nil {
@@ -141,6 +148,10 @@ func (c *HTTPclient) SetHostAddr(addr *url.URL) error {
 	return nil
 }
 
+func (c *HTTPclient) SetRetries(n int) {
+	c.retries = n
+}
+
 // Request performs a `method` type raw request to the endpoint specified in urlPath parameter.
 // Method is either GET or POST. If POST, a JSON struct should be attached.  Returns the response,
 // the status code and an error.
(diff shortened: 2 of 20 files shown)


altergui:
...aaand, it's working!
i mixed up here several random commits that should be in a separate PR. i'll clean things up and only then ask for review

altergui:
some of the tests are still calling deprecated sendVote() and i want to include those in this PR as well, so i'm marking as draft and will extend it.

but as it is now, the PR is ready for review and i appreciate feedback.

apiclient/client.go
```diff
@@ -164,17 +164,26 @@ func (c *HTTPclient) Request(method string, jsonBody any, urlPath ...string) ([]
...
-		Header: headers,
-		Body: func() io.ReadCloser {
-			if jsonBody == nil {
-				return nil
-			}
-			return io.NopCloser(bytes.NewBuffer(body))
-		}(),
-	})
+	var resp *http.Response
+	for i := 1; i <= 3; i++ {
```
p4u: We might need to config this value (the number of iterations). 

Three is a magic number that could be not enough when performing benchmarks. 

You might make it an exported variable of the HTTPclient type and set it to 3 (from a documented constant please) on the type constructor.

apiclient/helpers.go
```diff
@@ -19,9 +19,9 @@ import (
 )
 
 const (
-	TimeBetweenBlocks = 8 * time.Second
-	WaitTimeout       = 3 * TimeBetweenBlocks
-	PollInterval      = TimeBetweenBlocks / 2
+	BlockInterval = 8 * time.Second
```
p4u: Call it "DefaultBlockInterval"

apiclient/helpers.go
```diff
@@ -65,31 +65,40 @@ func (c *HTTPclient) SignAndSendTx(stx *models.SignedTx) (types.HexBytes, []byte
...
-
 // WaitUntilHeight waits until the given height is reached and returns nil.
 //
 // If ctx.Done() is reached, returns ctx.Err() instead.
 func (c *HTTPclient) WaitUntilHeight(ctx context.Context, height uint32) error {
 	for {
 		info, err := c.ChainInfo()
 		if err != nil {
-			log.Warn(err)
+			log.Error(err)
```
p4u: Errorw() will look nicer :)

p4u (approved):
Just some nits, approve.

apiclient/vote.go
```diff
@@ -41,12 +43,22 @@ type VoteData struct {
...
 
 // Vote sends a vote to the Vochain. The vote is a VoteData struct,
-// which contains the electionID, the choices and the proof. The
-// return value is the voteID (nullifier).
-func (c *HTTPclient) Vote(v *VoteData) (types.HexBytes, error) {
+// which contains the electionID, the choices and the proof.
+// if VoterAccount is set, it's used to sign the vote, else it defaults
+// to signing with the account set in HTTPclient.
+// The return value is the voteID (nullifier).
+func (cl *HTTPclient) Vote(v *VoteData) (types.HexBytes, error) {
```
mariajdab, line 57: To be consistent with the other methods in this file, I think all methods should keep using `func (c *HTTPclient)` What do you think?

apiclient/helpers.go
```diff
@@ -65,31 +65,40 @@ func (c *HTTPclient) SignAndSendTx(stx *models.SignedTx) (types.HexBytes, []byte
...
 }
 
-func (c *HTTPclient) WaitUntilNBlocks(ctx context.Context, n uint32) {
+// WaitUntilNextBlock waits until next block.
+// It uses a context.WithTimeout(24s) before giving up and returning anyway.
+func (c *HTTPclient) WaitUntilNextBlock() {
+	var cancel context.CancelFunc
+	ctx, cancel := context.WithTimeout(context.Background(), WaitTimeout)
+	defer cancel()
+	c.WaitUntilNBlocks(ctx, 1)
```
mariajdab: `c.WaitUntilNBlocks` could return an error. I wonder if it's worth to handle it?

api/wallet.go
```diff
@@ -118,14 +120,14 @@ func (a *API) walletSignAndSendTx(stx *models.SignedTx, wallet *ethereum.SignKey
...
+		return nil, ErrVochainOverloaded.WithErr(err)
+	case err != nil:
+		return nil, ErrVochainSendTxFailed.WithErr(err)
+	case resp == nil:
 		return nil, ErrVochainEmptyReply
-	}
-	if resp.Code != 0 {
+	case resp.Code != 0:
 		return nil, ErrVochainReturnedErrorCode.Withf("(%d) %s", resp.Code, string(resp.Data))
 	}
```
mariajdab: Seems repeated code for `vote, wallet, elections, chain, account`

apiclient/client.go
altergui: 3 is definitely the most magical number ever

i opted for a c.SetRetries instead, to keep the style of the package which has no exported vars, and uses Set methods, like c.SetAuthToken

apiclient/helpers.go
altergui: right now not really, looking at all the current callers of WaitUntilNextBlock

but i took the suggestion and made WaitUntilNextBlock return the error (and ignore it in all current callers). maybe some future caller will want to handle the error.

apiclient/vote.go
altergui: that would mean changing many other lines in the func, i believe it's more simple this way

api/wallet.go
altergui: totally agree, got that feeling while coding this. i deduplicated this into a new a.sendTx() wrapper and now the commit is much cleaner, thanks!

altergui:
addressed all the comments. if you want to make a last review go ahead, else i'm ready for merge

mariajdab (approved):

PR #977, opened 2023-06-02
ballot: remove maxValue overflow check

p4u:
While maxCount should be limited (the number of questions or options available for a voting) in order to avoid spam attacks, maxValue might not.

Allowing any maxvalue (up to uint32) brings the possibility to perform votings such as quadratic using Ethereum balances, which usually are very big numbers.

### DIFF
--- a/vochain/transaction/election_tx.go
+++ b/vochain/transaction/election_tx.go
@@ -36,10 +36,10 @@ func (t *TransactionHandler) NewProcessTxCheck(vtx *vochaintx.Tx,
 		return nil, ethereum.Address{}, fmt.Errorf("missing vote maxCount parameter")
 	}
 	// check for maxCount/maxValue overflows
-	if tx.Process.VoteOptions.MaxCount > results.MaxQuestions || tx.Process.VoteOptions.MaxValue > results.MaxOptions {
+	if tx.Process.VoteOptions.MaxCount > results.MaxQuestions {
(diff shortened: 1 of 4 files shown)


altergui (approved):
going with faith, always

vochain/transaction/election_tx.go
```diff
@@ -36,10 +36,10 @@ func (t *TransactionHandler) NewProcessTxCheck(vtx *vochaintx.Tx,
 		return nil, ethereum.Address{}, fmt.Errorf("missing vote maxCount parameter")
 	}
 	// check for maxCount/maxValue overflows
-	if tx.Process.VoteOptions.MaxCount > results.MaxQuestions || tx.Process.VoteOptions.MaxValue > results.MaxOptions {
+	if tx.Process.VoteOptions.MaxCount > results.MaxQuestions {
 		return nil, ethereum.Address{},
-			fmt.Errorf("maxCount or maxValue overflows hardcoded maximums (%d, %d). Received (%d, %d)",
-				results.MaxQuestions, results.MaxOptions, tx.Process.VoteOptions.MaxCount, tx.Process.VoteOptions.MaxValue)
+			fmt.Errorf("maxCount overflows (%d, %d)",
+				results.MaxQuestions, tx.Process.VoteOptions.MaxCount)
```
altergui, line 42: this could be "maxCount overflows (%d > %d)" but it's just a detail

PR #982, opened 2023-06-05
db: replace ReadTx with a Reader interface

### LINKED ISSUE #326: PebbleDB.ReadTx uses a pebble.Batch
The implementation of Database with Pebble currently uses an IndexedBatch for a ReadTx:
https://github.com/vocdoni/vocdoni-node/blob/626c8984df6dd0bd78787d37ab27ad62f7df76e8/db/pebbledb/pebledb.go#L108

This is a bit strange because an IndexedBatch has some extra logic to handle writes and read from those written keys, but we're only reading here.  Also, a Batch is not a Transaction, so we don't get a view at a particular point in time.
Why not just store a pointer to the db in the ReadTx to avoid the overhead of creating an IndexedBatch?

mvdan:
(see commit message)

Fixes #326.


mvdan:
FYI @ed255, finally got around to this :)

p4u (approved):

PR #986, opened 2023-06-12
Dockerfile: cleanup and implement best practices

altergui:
implement https://docs.docker.com/develop/develop-images/dockerfile_best-practices/
* replace COPY with --mount=type=bind
* first step `go mod download -x` so this layer is only invalidated if go.mod changes
* image `node:lts-bullseye-slim` is no longer needed since
  commit 325b9247 "remove javascript files (unused)"
* use a single COPY for binaries, reducing intermediate layers

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,22 +7,25 @@ ARG BUILDARGS
 # Build all the binaries at once, so that the final targets don't require having
 # Go installed to build each of them.
 WORKDIR /src
-COPY . .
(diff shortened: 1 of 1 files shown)


chilcano (approved):
Looks good, although, we should consider the dependencies vulnerabilities in docker image [reported by Trivy](https://github.com/vocdoni/vocdoni-node/actions/runs/5244049880). These vulns were not introduced with this PR.

PR #987, opened 2023-06-12
log: print stack trace on Fatal

altergui:
since "switch to Zerolog" (0ff4b19a) we lost the useful feature that
    log.Error and log.Fatal did include a stack trace.
    
    Fix that regression in a simple way, only for Fatal.
    Error stays untouched, to avoid a performance hit in case of a flood of Error().
    Also, it's not as pretty as in zerolog because there's no SkipCallerCount,
    but i prefer to make this one-liner fix instead of something more complex.

### DIFF
--- a/log/log.go
+++ b/log/log.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"os"
 	"path"
+	"runtime/debug"
 	"time"
 
 	"github.com/rs/zerolog"
@@ -205,7 +206,7 @@ func Error(args ...interface{}) {
 
 // Fatal sends a fatal level log message
 func Fatal(args ...interface{}) {
-	log.Fatal().Msg(fmt.Sprint(args...))
+	log.Fatal().Msg(fmt.Sprint(args...) + "\n" + string(debug.Stack()))
 	// We don't support log levels lower than "fatal". Help analyzers like
 	// staticcheck see that, in this package, Fatal will always exit the
 	// entire program.
@@ -243,7 +244,7 @@ func Errorf(template string, args ...interface{}) {
 
 // Fatalf sends a formatted fatal level log message
 func Fatalf(template string, args ...interface{}) {
-	Logger().Fatal().Msgf(template, args...)
+	Logger().Fatal().Msgf(template+"\n"+string(debug.Stack()), args...)
 }
 
 // Debugw sends a debug level log message with key-value pairs.


log/log.go
```diff
@@ -265,3 +268,31 @@ func Warnw(msg string, keyvalues ...interface{}) {
 func Errorw(err error, msg string) {
 	Logger().Error().Err(err).Msg(msg)
 }
+
+// prettyStack returns a pretty stack trace
+//
+// if disableStackTrace is true (env LOG_DISABLE_STACKTRACE=true), it returns immediately
+func prettyStack() string {
+	fmt.Println("WTF", disableStackTrace)
```
p4u: ummm..

log/log.go
```diff
@@ -265,3 +268,31 @@ func Warnw(msg string, keyvalues ...interface{}) {
 func Errorw(err error, msg string) {
 	Logger().Error().Err(err).Msg(msg)
 }
+
+// prettyStack returns a pretty stack trace
+//
+// if disableStackTrace is true (env LOG_DISABLE_STACKTRACE=true), it returns immediately
+func prettyStack() string {
```
p4u: @mvdan can you review this func?

p4u (changes requested):

log/log.go
altergui: :facepalm:

log/log.go
```diff
@@ -265,3 +268,30 @@ func Warnw(msg string, keyvalues ...interface{}) {
 func Errorw(err error, msg string) {
 	Logger().Error().Err(err).Msg(msg)
```
mvdan: what about Errorw?

log/log.go
mvdan: @altergui did you see https://pkg.go.dev/runtime/debug#Stack? I think that's basically what we want. It doesn't skip any callers, but honestly I don't think that's a big deal, particularly if you can avoid `func prettyStack` to begin with.

mvdan (commented):
note that this will have a significant performance impact on every error log call, which is only OK if we assume we won't have lots of these

log/log.go
altergui: yeah :eyes: i'll just KISS and print stack on panic.

log/log.go
altergui: that was indeed my initial one-liner approach https://github.com/vocdoni/vocdoni-node/commit/1edb30acee08e7edb0525e26df8f547e75149a6b
but the log_test failed because of the new output, so i did the prettyStack to skip the `0xfff` thinking that would give me a reproduceable, testable output. only to THEN notice that the paths printed also depend on the host. :facepalm: 

happy to go back to the one-liner trivial-to-review PR :grin:

p4u (approved):

PR #993, opened 2023-06-14
vochain/indexer: start indexing blocks

mvdan:


jordipainan (approved):
lgtm :)

vochain/indexer/indexer.go
```diff
@@ -224,7 +216,7 @@ func (idx *Indexer) retrieveCounts() (map[uint8]uint64, error) {
 // still open (live) and updates all temporary data (current voting weight and live results
 // if unecrypted). This method might be called on a goroutine after initializing the Indexer.
 // TO-DO: refactor and use blockHeight for reusing existing live results
-func (idx *Indexer) AfterSyncBootstrap() {
+func (idx *Indexer) AfterSyncBootstrap(inTest bool) {
```
p4u, line 219: Wouldn't it make more sense to have AfterSyncBootstrapTest() and AfterSyncBootstrap()
Then an `_innerAfterSyncBootStrap()` function with the actual code?

vochain/state/state_test.go
```diff
@@ -175,6 +176,7 @@ type Listener struct {
 
 func (l *Listener) OnVote(vote *Vote, txIndex int32)                                             {}
 func (l *Listener) OnNewTx(tx *vochaintx.Tx, blockHeight uint32, txIndex int32)                  {}
+func (l *Listener) OnBeginBlock(req abcitypes.RequestBeginBlock)                                 {}
```
p4u: Umm, I was trying to abstract the listeners from all the tendermint messy types. That's why all event handler functions contain only our types and usefull information.

So instead of passing the entire abcitypes.RequestBeginBlock it would be better to pass only the relevant parameters that we need.: req.Header.Height, req.Header.Time, req.Header.DataHash and detach the listeners from potential tendermint API/types future changes.

vochain/indexer/indexer.go
mvdan: I don't think that would really help - more lines of code and two more funcs, and we would never be calling the inner func directly either. I personally think one bool parameter and a conditional is pretty easy to follow.

vochain/state/state_test.go
mvdan: ack, I wondered about this but wasn't sure.

mvdan: done

PR #997, opened 2023-06-16
end2endtest/csp

mariajdab:
- #863 

- add csp election end2end test
- update `setupElectionRaw` to support csp election that not include setup the census
- add new method to setup account
- add `generateCSPProofs` in helpers

### DIFF
--- a/cmd/end2endtest/helpers.go
+++ b/cmd/end2endtest/helpers.go
@@ -20,6 +20,7 @@ import (
 	"go.vocdoni.io/dvote/types"
 	"go.vocdoni.io/dvote/util"
 	"go.vocdoni.io/proto/build/go/models"
+	"google.golang.org/protobuf/proto"
 )
 
 const (
@@ -154,7 +155,7 @@ func (t *e2eElection) waitUntilElectionStarts(electionID types.HexBytes) (*vapi.
 	return election, nil
 }
 
-func (t *e2eElection) generateProofs(root types.HexBytes, isAnonymousVoting bool) map[string]*apiclient.CensusProof {
+func (t *e2eElection) generateProofs(root types.HexBytes, isAnonymousVoting bool, csp *ethereum.SignKeys) map[string]*apiclient.CensusProof {
 	type voterProof struct {
 		proof   *apiclient.CensusProof
 		address string
@@ -191,10 +192,17 @@ func (t *e2eElection) generateProofs(root types.HexBytes, isAnonymousVoting bool
 				voterKey = t.api.MyZkAddress().Bytes()
 			}
 
-			pr, err := t.api.CensusGenProof(root, voterKey)
-			if isAnonymousVoting {
-				apiClientMtx.Unlock()
+			var pr *apiclient.CensusProof
+			var err error
+			if csp != nil {
+				pr, err = cspGenProof(t.election.ElectionID, voterKey, csp)
+			} else {
+				pr, err = t.api.CensusGenProof(root, voterKey)
+				if isAnonymousVoting {
+					apiClientMtx.Unlock()
+				}
 			}
+
 			if err != nil {
 				log.Fatal(err)
 			}
@@ -221,17 +229,16 @@ func (t *e2eElection) generateProofs(root types.HexBytes, isAnonymousVoting bool
 	}
 
 	wg.Wait()
-	time.Sleep(time.Second) // wait a grace time for the last proof to be added
 	log.Debugf("%d/%d voting proofs generated successfully", len(proofs), len(t.voterAccounts))
 	stopProofs <- true
 
 	return proofs
 }
 
-func (t *e2eElection) setupCensus(censusType string) (types.HexBytes, error) {
+func (t *e2eElection) setupAccount() error {
 	// Set the account in the API client, so we can sign transactions
 	if err := t.api.SetAccount(t.config.accountPrivKeys[0]); err != nil {
-		return nil, err
+		return err
 	}
 
 	// If the account does not exist, create a new one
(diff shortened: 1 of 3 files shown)


p4u:
@altergui please review and feel free to merge

cmd/end2endtest/helpers.go
```diff
@@ -228,10 +229,66 @@ func (t *e2eElection) generateProofs(root types.HexBytes, isAnonymousVoting bool
 	return proofs
 }
 
-func (t *e2eElection) setupCensus(censusType string) (types.HexBytes, error) {
+func (t *e2eElection) generateCSPProofs(pid []byte, csp *ethereum.SignKeys) map[string]*apiclient.CensusProof {
```
altergui: 90% of this func is duplicated code from generateProofs
please refactor both into one

altergui: i'm leaving some comments for you to take into account while refactoring, that most likely also apply to the current code in generateProofs

cmd/end2endtest/helpers.go
```diff
@@ -228,10 +229,66 @@ func (t *e2eElection) generateProofs(root types.HexBytes, isAnonymousVoting bool
...
+		end := i + pcount
+		if end > len(t.voterAccounts) {
+			end = len(t.voterAccounts)
+		}
+		wg.Add(1)
+		go addNaccounts(t.voterAccounts[i:end], &wg)
+	}
+
+	wg.Wait()
+	time.Sleep(time.Second) // wait a grace time for the last proof to be added
```
altergui: a time.Sleep after a wg.Wait pops up to my eye (before even reading the previous code) as a mistake in goroutine architecture or misunderstanding of how wg.Wait works. if it's not, document clearly what are you waiting for. `last proof to be added`... where? in the vochain?

cmd/end2endtest/helpers.go
```diff
@@ -228,10 +229,66 @@ func (t *e2eElection) generateProofs(root types.HexBytes, isAnonymousVoting bool
...
 
-func (t *e2eElection) setupCensus(censusType string) (types.HexBytes, error) {
+func (t *e2eElection) generateCSPProofs(pid []byte, csp *ethereum.SignKeys) map[string]*apiclient.CensusProof {
+	type voterProof struct {
+		proof   *apiclient.CensusProof
+		address string
+	}
+	proofs := make(map[string]*apiclient.CensusProof, t.config.nvotes)
+	proofCh := make(chan *voterProof)
+	stopProofs := make(chan bool)
```
altergui: instead of the stopProofs chan you could simply close(proofCh) 
https://gobyexample.com/closing-channels

altergui: but this is just a small general remark for you to take in mind in the future. i'll suggest to really rearchitect this whole func

cmd/end2endtest/helpers.go
altergui: i general i see: (let's say `pcount=4`)
* t.voterAccounts being sliced in 4 equal parts, and those slices being passed to...
* 4 "workers" that sequentially generate proofs, sending them over `proofCh`
* 1 "receiver" goroutine that reads proofCh and fills in a map
* an stopProofs chan that could be replaced by a `close(proofCh)`

i believe the code could be much simpler by:
* make `proofs` a `sync.Map` (so you don't need the "receiver" goroutine)
* push all the voterAccounts into a chan (say, `queue`), and close() it to signal the queue is over
* spawn 4 (`pcount`) goroutines that read that `queue` chan, call genCSPproof and fill in `proofs` (which is now concurrent safe, being a sync.Map)

but TBH i have no idea if any (or all) of these suggestions are actually a bad idea in terms of performance or memory usage. @mvdan?

altergui: also, (but this is a more general remark of the whole end2endtest and not of this particular csp test):
we could merge t.voterAccounts and t.proofs into a t.voters, a map indexed by key.PublicKey that contains both the keys and the proofs of each voter. that would make this generateProofs code slightly more simple.

altergui (changes requested):

cmd/end2endtest/helpers.go
mariajdab: Excuse me @altergui this is the code we have been used I did not change it, instead I adapted it to use csp proof

cmd/end2endtest/helpers.go
mariajdab: Excuse @altergui, just to confirm. Do you mean doing a refactor and not only adapt the csp proof to the general `generateProofs`?

cmd/end2endtest/helpers.go
mariajdab: Get it, I think a  rearchitect this whole func should be in other PR, so if you agree I could adapt CSP proof in the general func generateCSP and in other PR consider your suggest.  I think that is the approach that we have been following until now.

cmd/end2endtest/helpers.go
mariajdab: Certainly, there is old code to improve but it should be done in other PR

cmd/end2endtest/helpers.go
mariajdab: Updated to use the `generateProofs`. Happy to work in the refactor version for that method, I just need to confirm is that should be in this PR or in a new one? Any comment of @mvdan about @altergui suggestion would be useful.

altergui: mmh to be honest i was thinking about already doing the refactor. but thanks to your question i gave it a second thought and you're right, it should really be another PR.

cmd/end2endtest/helpers.go
altergui: haha right! git blame points to pau or lucas. luckily not to me (happens all the time, when i say "omg who wrote this buggy code??" it turns out it was... me.

anyway, yes, i agree with your proposal of taking this PR as-is and working on refactor in a new one

dockerfiles/testsuite/start_test.sh
```diff
@@ -10,6 +10,7 @@
 #  e2etest_censusizelection: run max census size test
 #  e2etest_ballotelection: run ballot test
 #  e2etest_lifecyclelection: run lifecycle test
+#  e2etest_csplelection: run csp test
```
altergui: typo l

altergui (approved):

dockerfiles/testsuite/start_test.sh
mariajdab: Thanks!

Issue #1000, opened 2023-06-19
bug: some error codes have changed without notice + some have incorrect prefixes


elboletaire:
**Describe the bug**
The "election not found" error code has changed from 4046 to 4045 without notice in this commit 42823fb20d61096ddd32ca4e0c27db54bd2862a4 by @jordipainan 

Also, the new error introduced as 4046 seems more like a query validation error, than a not found error. For that reason I recommend using the `412` error prefix instead (precondition failed) for this specific validation/query error, or `400` (bad request). 

So two issues here:
- [x] election not found error has changed
- [ ] some errors have weird error prefixes, like the error for "election end date cannot be in the past", or "election end date must be after start date", "wallet private key already exists"... and probably more, would fit better in other error codes than in the current 404, like 400 or 412.

**To Reproduce**

Just take a look at commit 42823fb20d61096ddd32ca4e0c27db54bd2862a4, the file `api/errors.go` changed the error code of the "election not found" there.

**Current behavior**
Implementations using the old election not found code 4046 are now incorrectly parsing it. 

**Expected behavior**
To not change error codes unless it's strictly necessary

**Additional context**
- Bad request 400: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400
- Precondition failed 412: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/412

altergui:
since this error code change never landed in stage yet, only in dev, i'd suggest to revert the change (thus AGAIN changing the codes, but this time back to what they were supposed to be always) so that when stage is redeployed, is like nothing happened

elboletaire:
I don't know the implications of the rest of the commit if we revert only the error codes, plus I think we were planning switching dev to stg today, not sure if it would be on time (?)

altergui:
also i'll add a header notice in error.go to instruct to only append errors, even if this leads to a messy order (i see that @jordipainan had the best intention, putting the ErrMissingParameter right next to the other ErrParam*, which would be the right thing to do in any other case)
it was my mistake to not include this notice in the original error.go

altergui:
> I don't know the implications of the rest of the commit if we revert only the error codes, plus I think we were planning switching dev to stg today, not sure if it would be on time (?)

i believe the revert #1002 is trivial to review+merge, the other comments about the HTTP status codes i'd discuss them separately / another PR to allow this hotfix to land on time

altergui:
[...]

PR #1002, opened 2023-06-20
api: revert error code change done in commit fix/api/filters (#850) (42823fb2)

altergui:
error codes must never be changed. the change never landed on stage yet, so we'll change the codes again back to what they were before.


altergui:
fix our lucky issue #1000

elboletaire:
> fix our lucky issue #1000

Beware that the issue is double. Reverting would fix one of the two issues, but the "weird prefixes" issue will still be there after mergin this PR.

Issue #975, opened 2023-06-02
ist: cannot commit results for election: maxCount and/or maxValue is zero


altergui:
i just saw this in the wild, posting logs here in case it's a bug
```
vocdoninode_1  | 2023-06-01T16:28:33.168Z DBG vochain/app.go:302 > deliver tx hash=5c29ab0ed7029abbbb96436f8a1c8d876860d41c0c58db10d9c35e394fe67b1e height=8580 tx={"Payload":{"SetAccount":{"account":"hPnYYQpVsVCFmCOEJhDwXPpGtLM=","faucetPackage":{"payload":"CK28ju3U3Ku2QxIUhPnYYQpVsVCFmCOEJhDwXPpGtLMY6Ac=","signature":"C8XggQQAgaSWp41XnCqgJLccrLZ2LamgtcO66wXEi9cKBSP/Ij6uKvqUT07OQPNaX8ekGTa+Rk+5nizjrENDRAE="},"infoURI":"ipfs://bafybeigv4mfmn3tyusbwmskfn6xhoyvv2i2svdq7pns6dphdivefr62k3q","nonce":0,"txtype":23}}} type=setAccount
vocdoninode_1  | 2023-06-01T16:28:33.171Z DBG state/account.go:173 > creating account 0x84F9d8610A55B150859823842610F05CfA46b4b3 with infoURI ipfs://bafybeigv4mfmn3tyusbwmskfn6xhoyvv2i2svdq7pns6dphdivefr62k3q balance 0 and delegates []
vocdoninode_1  | 2023-06-01T16:28:33.171Z DBG state/account.go:188 > setAccount: address 0x84F9d8610A55B150859823842610F05CfA46b4b3, nonce 0, infoURI ipfs://bafybeigv4mfmn3tyusbwmskfn6xhoyvv2i2svdq7pns6dphdivefr62k3q, balance: 0, delegates: [], processIndex: 0
vocdoninode_1  | 2023-06-01T16:28:33.171Z DBG state/balances.go:243 > transferring balance amount=1 from=c7c6e17059801b6962cc144a374ecc3ba1b8a9e0 to=ffffffffffffffffffffffffffffffffffffffff
vocdoninode_1  | 2023-06-01T16:28:33.171Z DBG state/account.go:188 > setAccount: address 0xC7C6E17059801b6962cc144a374eCc3ba1b8A9e0, nonce 6, infoURI , balance: 9712712, delegates: [], processIndex: 0
[...]

PR #1004, opened 2023-06-21
vochain: more idiomatic handling of err

altergui:
this doesn't change the behaviour since the code is currently not "wrong"
but make it tidy to avoid future issues


altergui:
came across this while investigating https://github.com/vocdoni/vocdoni-node/issues/975, i was particularly suspicious about the last `return results, err`, took me a while to understand that err is always `nil`.

p4u (approved):

PR #1005, opened 2023-06-21
api: /chain/dateToBlock fix overflow for big timestamps

jordipainan:
Seems that we found a bug on /chain/dateToBlock that happens when big timestamps are used. 
Since the difftime is multiplied by 1000 this can lead to overflows because of the use of uint32.
This fix makes the proper changes for using uint64 and also that checks for overflows.


jordipainan:
I think that this is one possible solution, the other ones is to use int64 and increase the threshold for future times but if we do not limit values, same thing will happen.

p4u (approved):

PR #1006, opened 2023-06-22
end2endtest: Dynamic Census

mariajdab:
- #863 

- apiclient: add TransactionSetCensus
- end2end: Dynamicensus test

### DIFF
--- a/cmd/end2endtest/dynamicensus.go
+++ b/cmd/end2endtest/dynamicensus.go
@@ -1,0 +1,295 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"strings"
+	"sync"
+	"time"
+
+	vapi "go.vocdoni.io/dvote/api"
+	"go.vocdoni.io/dvote/apiclient"
+	"go.vocdoni.io/dvote/log"
+	"go.vocdoni.io/dvote/types"
+)
+
+func init() {
+	ops["dynamicensuselection"] = operation{
+		test:        &E2EDynamicensusElection{},
+		description: "Publish an election with the dynamic census flag set to true and another election with dynamic census flag set to false",
+		example:     os.Args[0] + " --operation=dynamicensuselection --votes=1000",
+	}
+}
+
+var _ VochainTest = (*E2EDynamicensusElection)(nil)
+
+type E2EDynamicensusElection struct {
+	elections []e2eElection
+}
+
+func (t *E2EDynamicensusElection) Setup(api *apiclient.HTTPclient, c *config) error {
+	t.elections = append(t.elections, e2eElection{api: api, config: c})
+	t.elections = append(t.elections, e2eElection{api: api, config: c})
+
+	eDescriptions := []struct {
+		d             *vapi.ElectionDescription
+		dynamicCensus bool
+	}{
+		{d: newTestElectionDescription(), dynamicCensus: true},
+		{d: newTestElectionDescription(), dynamicCensus: false},
+	}
+
+	for i, ed := range eDescriptions {
+		ed.d.ElectionType.Autostart = true
+		ed.d.ElectionType.Interruptible = true
+		ed.d.ElectionType.DynamicCensus = ed.dynamicCensus
+		ed.d.Census = vapi.CensusTypeDescription{Type: vapi.CensusTypeWeighted}
+
+		// create a census with 2 voterAccounts less than the nvotes passed, that will allow to create another
+		// census with the missing nvotes
+		censusRoot, censusURI, err := t.elections[i].setupCensus(vapi.CensusTypeWeighted, t.elections[i].config.nvotes-2)
+		if err != nil {
+			return err
+		}
+
+		// add the censusRoot and censusURI to the election description
+		ed.d.Census = vapi.CensusTypeDescription{
+			Type:     vapi.CensusTypeWeighted,
+			RootHash: censusRoot,
+			URL:      censusURI,
+			Size:     uint64(t.elections[1].config.nvotes),
+		}
+
+		// set up the election with the custom census created
+		if err := t.elections[i].setupElection(ed.d); err != nil {
+			return err
+		}
+		log.Debugf("election detail: %+v", *t.elections[i].election)
+	}
+
+	return nil
+}
+
+func (t *E2EDynamicensusElection) Teardown() error {
+	// nothing to do here
+	return nil
+}
+
+func (t *E2EDynamicensusElection) Run() error {
+	var wg sync.WaitGroup
+	errCh := make(chan error, 2)
+
+	setupNewCensusAndVote := func(election e2eElection) (types.HexBytes, apiclient.VoteData, error) {
+		api := election.api
+		nvotes := election.config.nvotes
+
+		censusRoot2, censusURI2, err := election.setupCensus(vapi.CensusTypeWeighted, 2)
+		if err != nil {
+			return nil, apiclient.VoteData{}, err
+		}
+		log.Infof("root : %x URI: %s  err: %s", censusRoot2, censusURI2, err)
+
+		voterKey := election.voterAccounts[nvotes-1].Address().Bytes()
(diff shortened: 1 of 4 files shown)


mariajdab:
I'm not sure if it's worth sending votes for this test. Any suggestions are welcome, @altergui

p4u:
> I'm not sure if it's worth sending votes for this test. Any suggestions are welcome, @altergui

Yes its worth. I think you should test the following:

1. Create census A 
2. Create census B (with different voting keys)
3. Create election with censusRoot A
4. Cast a vote from census A (should work)
5. Cast a vote from census B (should fail)
6. Set the censusRoot for the election to census B
7. Cast a vote from census B (should work)
8. Cast a vote from census A (should not work)
9. End the election
10. Check results are the expected ones

It would be great to also test that if you create an election with dynamicCensus disabled, you cannot change the census root.

altergui (commented):
regarding your question: sending the votes is definitely part of the test. in fact, the current code looks good to me as "MVP" but i'd request you to extend it to cover a few more cases. right now you create the election already with the "correct" census, then vote, then update the census with a TransactionSetCensus (with a mock, empty census, AFAIU?) to just check the tx doesn't fail.

i'd suggest you to push a "real" census, that for example includes some new voters (not in the original census) and also excludes some voter (that was allowed to vote in the original census but after the update shouldn't be able to). then try to vote with those accounts and check the results are as expected.

altergui:
more examples:
what happens if you exclude a voter that already voted?
as in, alice is part of the original census. alice sends her vote (and is commited in a block). in the next block, census is updated (let's say "v2") and alice is not included. election ends. will her vote be counted in the results?

i could suggest even more esoteric cases, like alice trying to overwrite her vote after being excluded from census. you might uncover funny bugs like you already did with the other tests. but that would greatly increase the scope of the test and not be just a "dynamic census test", more like a "crazy interactions test" :rofl:

p4u:
@mariajdab maybe you found this issue? https://github.com/vocdoni/vocdoni-node/pull/1026/files

The indexer was not properly updating the new census.

mariajdab:
> @mariajdab maybe you found this issue? https://github.com/vocdoni/vocdoni-node/pull/1026/files
> 
> The indexer was not properly updating the new census.

Yeah, that's why even after using `api.WaitUntilTxIsMined`, the election was still returning the old `censusRoot` and `censusURL`. They were only updated at the end when the election was finalized. Now is working as expected.

mariajdab:
I left a comment about `CensusURL`used with `&models.SetProcessTx{Txtype: models.TxType_SET_PROCESS_CENSUS}` in the first election, the code allow to set any `censusURL` and not only the `censusURL`  returned from the `api.CensusPublish`. Is that a bug? @altergui @p4u

cmd/end2endtest/dynamicensus.go
```diff
@@ -0,0 +1,320 @@
...
+			errCh <- fmt.Errorf("unexpected err when retrieve election details, %s", err)
+			return
+		}
+
+		log.Debugf("election details after: %s %s %x", t.elections[0].election.Census.CensusOrigin, t.elections[0].election.Census.CensusURL, t.elections[0].election.Census.CensusRoot)
+
+		// try again to vote with the vote data from the last census created
+		if _, err := api.Vote(&v); err != nil {
+			// should not fail
+			errCh <- fmt.Errorf("unexpected error when voting with the new cesnsus vote %s", err)
```
altergui: census

cmd/end2endtest/dynamicensus.go
```diff
@@ -0,0 +1,320 @@
...
+			log.Infof("verified %d/%d votes", count, c.nvotes)
+			if time.Since(startTime) > c.timeout {
+				log.Fatalf("timeout waiting for votes to be registered")
+			}
+		}
+
+		log.Infof("%d votes registered successfully, took %s (%d votes/second)",
+			c.nvotes, time.Since(startTime), int(float64(c.nvotes)/time.Since(startTime).Seconds()))
+
+		// Set the account back to the organization account
```
altergui: you'll want to use the new t.endElectionAndFetchResults

cmd/end2endtest/helpers.go
```diff
@@ -26,6 +26,7 @@ import (
 const (
 	nextBlock = "nextBlock"
 	sameBlock = "sameBlock"
+	weight10  = 10
```
altergui: s/weight10/defaultWeight/

cmd/end2endtest/helpers.go
```diff
@@ -127,14 +136,15 @@ func (t *e2eElection) addParticipantsCensus(censusType string, censusID types.He
 	return nil
 }
 
-func (t *e2eElection) isCensusSizeValid(censusID types.HexBytes) bool {
+// isCensusSizeValid validate the size of the census, in case like dynamic census test the size could be different to nvotes
+func (t *e2eElection) isCensusSizeValid(censusID types.HexBytes, sizeExpected uint64) bool {
```
altergui: then you probably want to rename the func, say for example `censusSizeEquals`

dockerfiles/testsuite/start_test.sh
```diff
@@ -11,6 +11,7 @@
 #  e2etest_ballotelection: run ballot test
 #  e2etest_lifecyclelection: run lifecycle test
 #  e2etest_cspelection: run csp test
+#  e2etest_dynamicensustelection: run dynamic census test
```
altergui: why the `t`?
i think `dynamicensuselection` would be more consistent with the rest

cmd/end2endtest/dynamicensus.go
```diff
@@ -0,0 +1,320 @@
...
+		}
+
+		log.Debugf("election details before: %s %s %x", t.elections[1].election.Census.CensusOrigin, t.elections[1].election.Census.CensusURL, t.elections[1].election.Census.CensusRoot)
+
+		if _, err := api.TransactionSetCensus(electionID, vapi.ElectionCensus{
+			CensusOrigin: "OFF_CHAIN_TREE_WEIGHTED",
+			CensusRoot:   censusRoot2,
+			CensusURL:    "http://test/census",
+		}); err != nil {
+			log.Debug("expected error from election without dynamic census enabled", err)
```
altergui: please check for the specific error you're expecting. if the error is not what you expect, fail the test

altergui (commented):
i see no strong reason to bundle both tests into one, it makes the code much more complex. we had this tradeoff in the lifecycle test, and i accepted it. but in this case i believe it's not worth it (i'm open to discussing it, like i did in lifecycle, of course)

i'd make two separate ops, the current dynamicensuselection but simple, without any goroutines or chans, that just tests dynamicCensus: true
(and stays true to its description "Publish an election with the dynamic census flag set to true")

and another op, say dynamicensuserror, simple, without any goroutines or chans, that tests dynamicCensus: false

dockerfiles/testsuite/start_test.sh
mariajdab: yep, it was a typo

mariajdab:
> i see no strong reason to bundle both tests into one, it makes the code much more complex. we had this tradeoff in the lifecycle test, and i accepted it. but in this case i believe it's not worth it (i'm open to discussing it, like i did in lifecycle, of course)
> 
> i'd make two separate ops, the current dynamicensuselection but simple, without any goroutines or chans, that just tests dynamicCensus: true (and stays true to its description "Publish an election with the dynamic census flag set to true")
> 
> and another op, say dynamicensuserror, simple, without any goroutines or chans, that tests dynamicCensus: false

As previously agreed, we will proceed with this approach for now to cover more cases in the test, and we will perform a refactor to improve handling of various cases in end2end

mariajdab:
ready for review @altergui :see_no_evil:

dockerfiles/testsuite/start_test.sh
altergui: comment still has the typo. (the others are fixed, nice)

altergui (approved):
LGTM to rebase and merge

PR #1012, opened 2023-06-23
Feature/new anonymous circuit

lucasmenendez:
As Vocdoni, we want to be able to have fully anonymous voting available in the protocol/SDK/UIs to allow the voters to have full privacy. 

**Protocol/Backend**: Implement `Secret Identity Key (SIK)` in the Vochain
- [x] Create the new transactions for SIK: Update and create needed protobuffer definitions (more info [here](https://hackmd.io/8QtTh6FGTn6FlYgiQ1Ubxg?view#Tech-specification)) --> PR [here](https://github.com/vocdoni/dvote-protobuf/pull/63).
- [x] Keeping track of the starting blocks of each open election storing into a no-state database the `startBlock` of every started election. If a election is finished, delete it from the database. If a election is paused, the database will store the current block (height) (more info [here](https://hackmd.io/8QtTh6FGTn6FlYgiQ1Ubxg#SIK-reset)).
- [x] Create new Merkle Tree in the Vochain state ([here](https://github.com/vocdoni/vocdoni-node/blob/main/vochain/state/trees.go#L120)) with similar logic that the `RollingCensus` but using `Poseidon` hash ([current example](https://github.com/vocdoni/vocdoni-node/blob/main/vochain/state/rollingcensus.go))
- [x] Store `SIK` roots for hysteresis into statedb (nostate)
- [x] Implement the anonymous voting transaction flow with root hysteresis.
  - [x] Support to store multiple `electionID`s by `blockNumber` on ProcessBlockRegistry.
    - [x] Unit testing for ProcessBlockRegistry.
  - [x] Check validity of sikRoot.
- [x] API endpoints and APIClient methods to manage `SIK`'s:
  - [x] Update account metadata setter to handle `SIK` during creation and update the account ([API](https://github.com/vocdoni/vocdoni-node/blob/ed06c3517c0560312eac3ed105f827b00b1efc9f/api/accounts.go#L139) & [APIClient](https://github.com/vocdoni/vocdoni-node/blob/ed06c3517c0560312eac3ed105f827b00b1efc9f/apiclient/account.go#L155))
  - [x] Create API endpoint and APIClient method to delete account `SIK`.
  - [x] Update API endpoints and APIClient methods to retrieve not just the census proof (root + siblings) for an election, but also the `SIK` proof (root + siblings).
- [x] Review leaf insertion in both trees (`SIK` and census) to ensure that follows something similar to this [example](https://github.com/vocdoni/zk-franchise-proof-circuit/blob/9a2c24e5c0fdddc77f65cac16a9e411dabeb1725/internal/helpers.go#L39).
- [x] Update nullifier formula and implementation to include user `SIK` 
- [x] Deprecate and remove the ZkAddress.
- [x] Update circuit reference and artifacts URL's.
- [x] Adapt end2end tests.
- [x] Expose the list of valid SIK roots on a new API endpoint.
- [x] Split the sikproof from the censusproof API endpoint and move everything related to the SIK to a new path (`/sik`) (read more [here](https://hackmd.io/_tGmlKE-SluYsuP10Ts33g)).
- [x] Truncate the sik signature to 64 bytes discarding the byte number 65 (for recovery) that produces differences between the signatures generated with go and javascript.
- [x] Create a free tx to allow to non registered users that are in the census of a created election to register a sik without create a vochain account. This tx will have no cost.

**Important TODO's before merge**
- [x] Set new tx's costs on staging
- [x] Change resource of the API endpoints related to the SIK to plural (`/sik` => `/siks`).
- [x] Change the hysteresis end2end test to unit test to allow to set the chain height manually and optimize the test
- [x] Update the API endpoint to get account info to include if the account has a valid SIK registered.
- [x] Keep track of the number of SIKs registered with RegisterSIKTx by process to prevent reaching the MaxCensusSize number. Remove the record when the election ends.


vochain/state/sik.go
```diff
@@ -0,0 +1,128 @@
...
+	"math/big"
+
+	"github.com/ethereum/go-ethereum/common"
+	"go.vocdoni.io/dvote/log"
+	"go.vocdoni.io/dvote/tree/arbo"
+)
+
+// encodedHeightLen constant is the number of bytes of the encoded hysteresis
+// that contains the hysteresis height value, starting from the last byte
+const encodedHeightLen = 4
```
p4u: Join both consts in a single statement

vochain/state/sik.go
```diff
@@ -0,0 +1,128 @@
...
+//   - It checks if already exists a valid SIK for the provided address and if
+//     so it returns an error.
+//   - If no SIK exists for the provided address, it will create one with the
+//     value provided.
+//   - If exists a logically deleted SIK for the address provided, checks the
+//     hysteresis height for that address before update its SIK.
+//   - If the hysteresis height is not reached yet, it returns an error.
+//   - If the hysteresis height is reached, it updates the value of the sik with
+//     the provided one.
+func (v *State) SetSIK(address common.Address, newSik []byte) error {
```
p4u: I think SetSik should just set the sik. All the logic remaining hysteresis and the conditions to reset an existing sik should be part of the APP logic, not the state. Specifically part of the transaction handler.

vochain/state/sik.go
```diff
@@ -0,0 +1,128 @@
...
+	// return an error
+	if hysteresisHeight <= v.CurrentHeight() {
+		return ErrInvalidHysteresis
+	}
+	v.Tx.Lock()
+	defer v.Tx.Unlock()
+	return v.Tx.DeepSet(address.Bytes(), encodeHysteresis(hysteresisHeight), StateTreeCfg(TreeSIK))
+}
+
+// encodeHysteresis funtion returns the encoded value of the height hysteresis
```
p4u: This approach is not correct. It does not avoid double voting.

Instead of an hysteresis height, we should store the the current maximum endBlock for a process.

p4u (changes requested):

vochain/state/errors.go
```diff
@@ -12,4 +12,8 @@ var (
 	ErrAccountBalanceZero   = fmt.Errorf("zero balance account not valid")
 	ErrAccountAlreadyExists = fmt.Errorf("account already exists")
 	ErrInvalidURILength     = fmt.Errorf("invalid URI length")
+	ErrRegisteredValidSIK   = fmt.Errorf("address already has a valid sik")
+	ErrSIKAlreadyInvalid    = fmt.Errorf("sik is already invalidated")
+	ErrHysteresisNotReached = fmt.Errorf("hysteresis height not reached")
+	ErrInvalidHysteresis    = fmt.Errorf("hysteresis must be greater than the curren height")
```
altergui: s/curren/current/

vochain/state/sik.go
```diff
@@ -0,0 +1,122 @@
...
+	"github.com/ethereum/go-ethereum/common"
+	"go.vocdoni.io/dvote/log"
+	"go.vocdoni.io/dvote/tree/arbo"
+)
+
+// encodedHeightLen constant is the number of bytes of the encoded hysteresis
+// that contains the hysteresis height value, starting from the last byte
+const encodedHeightLen = 4
+
+// hysteresisLen constant contains the number of bytes that the encode
```
altergui: maybe `s/encode/encoded/` ?

vochain/state/sik.go
```diff
@@ -0,0 +1,122 @@
...
+	sik, err := v.mainTreeViewer(false).DeepGet(address.Bytes(), StateTreeCfg(TreeSIK))
+	if errors.Is(err, arbo.ErrKeyNotFound) {
+		// if not exists create it
+		log.Debugw("setSIK (create)",
+			"address", address.String(),
+			"sik", hex.EncodeToString(sik))
+		v.Tx.Lock()
+		defer v.Tx.Unlock()
+		return v.Tx.DeepAdd(address.Bytes(), newSik, StateTreeCfg(TreeSIK))
+	} else if err != nil {
```
altergui: no need for "else", you can simply do two separate ifs

vochain/state/sik.go
```diff
@@ -0,0 +1,122 @@
...
+// It will have 32 bytes:
+//   - The initial 28 bytes must be zero.
+//   - The remaining 4 bytes must contain the height encoded in LittleEndian
+func encodeHeight(height uint32) []byte {
+	bHeight := big.NewInt(int64(height)).Bytes()
+	// fill with zeros until reach the encoded height length
+	for len(bHeight) < encodedHeightLen {
+		bHeight = append([]byte{0}, bHeight...)
+	}
+	// create the hysteresis with the rigth number of zeros
```
altergui: right

vochain/state/sik.go
```diff
@@ -0,0 +1,122 @@
...
+	// copy the height bytes swapping endianness in the last bytes
+	for i := encodedHeightLen - 1; i >= 0; i-- {
+		hysteresis = append(hysteresis, bHeight[i])
+	}
+	return hysteresis
+}
+
+// decodeHeigh funtion returns the decoded height uint32 from the leaf
+// value that contains the encoded hysteresis height.
+func decodeHeigh(leafValue []byte) uint32 {
```
altergui: s/decodeHeigh/decodeHeight/

altergui: (also in func doc comment, and in callers)

altergui (commented):
reviewed but i'm not familiar with the concepts/idea so i'll leave the approval to p4u, commenting here just some typos

vochain/state/sik.go
lucasmenendez: Both are used independently

lucasmenendez:
@p4u i have solved all comments regarding hysteresis because they comes from my misunderstanding of the concept, i have just reimplemented the logic about SIK including methods to keep on track of the valid SIKs roots.

vochain/state/process.go
```diff
@@ -161,6 +164,52 @@ func (v *State) CountProcesses(committed bool) (uint64, error) {
...
+	v.Tx.Lock()
+	defer v.Tx.Unlock()
+	processesTree, err := v.Tx.DeepSubTree(StateTreeCfg(TreeProcess))
+	if err != nil {
+		return err
+	}
+	return processesTree.NoState().Delete(pid)
+}
+
+// MinOnGoingStartBlock returns the minimun start block of the current on going
```
altergui: `ongoing` is a word, not two.
so this would be MinOngoingStartBlock

also, `minimum` is the correct spelling

vochain/state/process.go
```diff
@@ -211,9 +260,28 @@ func (v *State) UpdateProcess(p *models.Process, pid []byte) error {
...
 		return ErrProcessNotFound
 	}
+	// update the process
 	v.Tx.Lock()
-	defer v.Tx.Unlock()
-	return updateProcess(&v.Tx, p, pid)
+	if err := updateProcess(&v.Tx, p, pid); err != nil {
+		return err
+	}
+	v.Tx.Unlock()
```
altergui: if `err != nil` happens, Unlock will never happen. add it before the `return err` , or use defer, if the `try to update startBlocks database` doesn't take long

altergui: or even simpler
```go
 	v.Tx.Lock()
	err := updateProcess(&v.Tx, p, pid)
    v.Tx.Unlock()
	if err != nil {
		return err
	}
```

lucasmenendez: this code is old, but very good point, it will be fixed in the next commit

vochain/state/process.go
lucasmenendez: Sorry about my english ;(

vochain/state/process.go
```diff
@@ -211,9 +264,28 @@ func (v *State) UpdateProcess(p *models.Process, pid []byte) error {
...
+	// try to update startBlocks database
```
lucasmenendez, line 181: @p4u I dont know if this file is the best place for this code or if it needs to be moved to `transaction/transaction.go` because its not about the state at all.

api/census_test.go
```diff
@@ -225,9 +225,10 @@ func TestCensusProof(t *testing.T) {
 		&models.Proof{
 			Payload: &models.Proof_Arbo{
 				Arbo: &models.ProofArbo{
-					Type:       models.ProofArbo_BLAKE2B,
-					Siblings:   censusData.Proof,
-					LeafWeight: censusData.Value,
+					Type:            models.ProofArbo_BLAKE2B,
+					Siblings:        censusData.Proof,
+					AvalaibleWeight: censusData.Value,
```
altergui: typo here: should be `Available`

lucasmenendez:
**⚠️ ALERT ⚠️**

This branch is working progress and no tested yet, it needs to be deeply reviewed and tested.

api/accounts.go
```diff
@@ -407,3 +425,30 @@ func (a *API) tokenTransfersHandler(msg *apirest.APIdata, ctx *httprouter.HTTPCo
...
 	return ctx.Send(data, apirest.HTTPstatusOK)
 }
+
+// setSikHandler
+//
+//	@Summary		Updates the Secret Identity Key of an Address
+//	@Description	Updates the Secret Identity Key of an Address
+//	@Success		200	{object}	SikSet
+//	@Router			/accounts/sik [post]
+func (a *API) setSikHandler(msg *apirest.APIdata, ctx *httprouter.HTTPContext) error {
```
p4u: I'm not sure of the need of this API method. It does nothing different from `/chain/transaction`.

We do have specific endpoints for some transactions that require special logic. Such as new election, which requires the Gateway to publish the metadata into IPFS. But for the rest of transactions, we use the generic endpoint which just relays the transaction.

api/api_types.go
```diff
@@ -228,18 +228,26 @@ type AccountSet struct {
 	MetadataURL string         `json:"metadataURL" swaggertype:"string"`
 }
 
+type SikSet struct {
```
p4u: No need either.

api/accounts.go
lucasmenendez: Perfect, it makes sense

apiclient/account.go
```diff
@@ -107,6 +107,16 @@ func (c *HTTPclient) AccountBootstrap(faucetPkg *models.FaucetPackage, metadata
 		metadataURI = "ipfs://" + ipfs.CalculateCIDv1json(metadataBytes)
 	}
 
+	if sik == nil {
+		sikSign, err := c.account.SignEthereum([]byte(DefaultSIKContent))
```
p4u: I think the crypto/ethereum package could be extended to have a method that returns and abstracts the SIK generation. So we can do `c.account.Sik()`.

apiclient/account.go
```diff
@@ -227,3 +238,95 @@ func (c *HTTPclient) GetTransfers(from common.Address, page, pageSize int) ([]*i
 	}
 	return transfers, nil
 }
+
+// SetSik function allows to update the Secret Identity Key for the current
+// HTTPClient account. To do that, the function requires a secret user input.
+func (c *HTTPclient) SetSik(secret []byte) (types.HexBytes, error) {
```
p4u: secret is not used :/

I'd export two methods, this might be useful for testing purposes:

- SetSik()
- SetCustomSik(sik []byte)

The first can call the second.

apiclient/helpers.go
```diff
@@ -314,5 +320,49 @@ func UnmarshalFaucetPackage(data []byte) (*models.FaucetPackage, error) {
...
 	}, nil
+}
 
+// GenerateSik function generates the Secret Identity Key for the current client
+// account with the signature and the secret provided following the definition:
+//
+//	SIK = poseidon(address, signature, secret*)
+//
+// *The secret is optional.
+func (c *HTTPclient) GenerateSik(sign, secret []byte) ([]byte, error) {
```
p4u: I don't think we need this HTTPclient method. 

The SIK generation should be only part of the crypto/ethereum package. Let's abstract it.

vochain/state/process.go
```diff
@@ -161,6 +162,92 @@ func (v *State) CountProcesses(committed bool) (uint64, error) {
 	return count, nil
 }
 
+// RegisterStartBlock function creates a new record on the key-value database
```
p4u: This is a method, not a function. 

Please, specify this is being stored in the no-state database

vochain/state/process.go
```diff
@@ -161,6 +162,92 @@ func (v *State) CountProcesses(committed bool) (uint64, error) {
...
+// associated to the processes subtree for the process ID and the startBlock
+// number provided.
+func (v *State) RegisterStartBlock(pid []byte, startBlock uint32) error {
```
p4u: I think the name of the method is not self-descriptive. I'd be more explicit: `ProcessRegisterStartBlockNoState()`

p4u: Actually I think we might make a better structure to clearly differentiate this methods, since its hard to understand what is the actual purpose.

To this end I'd propose to create a new type: `ProcessBlockRegistry`

On the `State` constructor, we can create a new instance for this type:

`State.ProcessBlockRegistry = ProcessBlockRegistry{ treeRW: processTree.NoState(), treeRead: ....,  treeLock: v.Tx }`

Then whenever you want to use it, just:

`state.ProcessBlockRegistry.SetStartBlock(pid, startBlock)`

So all methods are contained inside the ProcessBlockRegistry.

vochain/state/process.go
p4u: LGTM

vochain/state/sik.go
```diff
@@ -0,0 +1,247 @@
...
+	sikLeafValueLen = 32
+)
+
+// SIK type abstracts a slice of bytes that contains the Secret Identity Key
+// value of a user
+type SIK []byte
+
+// SIKByAddress function return the current SIK value associated to the provided
+// address.
+func (v *State) SIKByAddress(address common.Address) (SIK, error) {
```
p4u: SIKfromAddress()

vochain/state/sik.go
```diff
@@ -0,0 +1,247 @@
...
+
+// SetSIK function creates or update the SIK of the provided address in the
+// state. It covers the following cases:
+//   - It checks if already exists a valid SIK for the provided address and if
+//     so it returns an error.
+//   - If no SIK exists for the provided address, it will create one with the
+//     value provided.
+//   - If it exists but it is not valid, overwrite the stored value with the
+//     provided one.
+func (v *State) SetSIK(address common.Address, newSik SIK) error {
```
p4u: SetAddressSik()

vochain/state/sik.go
```diff
@@ -0,0 +1,247 @@
...
+	// copy the height bytes swapping endianness in the last bytes
+	for i := encodedHeightLen - 1; i >= 0; i-- {
+		s = append(s, bHeight[i])
+	}
+	return s
+}
+
+// DecodeInvalidatedHeight funtion returns the decoded height uint32 from the
+// leaf value that contains an invalidated SIK.
+func (s SIK) DecodeInvalidatedHeight() uint32 {
```
p4u, line 396: Move this methods next to the type definition

vochain/state/sik.go
```diff
@@ -0,0 +1,247 @@
...
+// voting to already registered users when an election is on going and new users
+// are registered. When a new sikRoot is generated, the sikRoot’s from an older
+// block than the current block minus the hysteresis blocks will be deleted:
+//   - If exists a sikRoot for the minimun hysteresis block number
+//     (currentBlock - hysteresis), just remove all the roots with a lower block
+//     number.
+//   - If it does not exist, remove all roots with a lower block number except
+//     for the next lower sikRoot. It is becouse it still being validate for a
+//     period.
+func (v *State) UpdateSIKRoots() error {
```
p4u: I wonder if for SIKs we can follow the same approach as in ProcessBlockRegistry. 
My feeling is that the State package is growing too much and some hierarchy would be appreciated.

vochain/state/sik.go
```diff
@@ -0,0 +1,247 @@
...
+	v.Tx.Lock()
+	defer v.Tx.Unlock()
+	siksTree, err := v.Tx.DeepSubTree(StateTreeCfg(TreeProcess))
+	if err != nil {
+		return err
+	}
+	sikRootsDB := siksTree.NoState()
+	// calculate the current minimun block to purge useless sik roots
+	currentBlock := v.CurrentHeight()
+	if currentBlock > SIKROOT_HYSTERESIS_BLOCKS {
```
p4u: `if currentBlock := v.CurrentHeight(); currentBlock > SIKROOT_HySTERESIS_BLOCS {`

vochain/transaction/proof.go
```diff
@@ -115,21 +115,21 @@ func VerifyProofOffChainTree(process *models.Process, proof *models.Proof,
...
 		}
 		// Legacy: support p.LeafWeight == nil, assume then value=1
-		if p.LeafWeight == nil {
+		if p.AvailableWeight == nil {
 			return true, bigOne, nil
 		}
 
-		factoryWeight := arbo.BytesToBigInt(p.LeafWeight)
-		if p.VotingWeight == nil {
+		factoryWeight := arbo.BytesToBigInt(p.AvailableWeight)
```
p4u: Maybe it is a good time to change factoryWeight by the new naming

vochain/transaction/account_tx.go
```diff
@@ -239,3 +239,42 @@ func (t *TransactionHandler) SetAccountInfoTxCheck(vtx *vochaintx.Tx) error {
...
+func (t *TransactionHandler) DelSikTxCheck(vtx *vochaintx.Tx) error {
+	if vtx == nil || vtx.Signature == nil || vtx.SignedBody == nil || vtx.Tx == nil {
+		return ErrNilTx
+	}
+	tx := vtx.Tx.GetSetSik()
+	if tx == nil {
+		return fmt.Errorf("invalid transaction")
+	}
+
+	bAddress := tx.GetAddress()
```
p4u: This is wrong, you cannot trust the Address available in the Protobuf model is correct.

Actually the address should not be a field, instead you should extract the address from the Signature.

Please check initializeSignedVote() function. You need to do something similar.

p4u: Let's ensure no other TxCheck method works like this... either we have a huge security hole.

vochain/transaction/transaction.go
```diff
@@ -434,6 +439,41 @@ func (t *TransactionHandler) CheckTx(vtx *vochaintx.Tx, forCommit bool) (*Transa
 			return response, nil
 		}
 
+	case *models.Tx_SetSik:
+		err := t.SetSikTxCheck(vtx)
```
p4u: this method might return the address and the newSik, so you don't need to recalculate/refetch it again.

Dockerfile
```diff
@@ -7,6 +7,7 @@ ARG BUILDARGS
 # Build all the binaries at once, so that the final targets don't require having
 # Go installed to build each of them.
 WORKDIR /src
+# COPY . .
```
altergui: ?

do note that instead of doing the COPY, a bind mount is used in line 17
`--mount=type=bind,target=.`

api/api_types.go
```diff
@@ -222,25 +222,27 @@ type Account struct {
...
 type AccountSet struct {
-	TxPayload   []byte         `json:"txPayload,omitempty"`
-	Metadata    []byte         `json:"metadata,omitempty"`
-	TxHash      types.HexBytes `json:"txHash"`
-	MetadataURL string         `json:"metadataURL"`
-	SIK         []byte         `json:"sik,omitempty"`
+	TxPayload   []byte         `json:"txPayload,omitempty" swaggerignore:"true"`
+	Metadata    []byte         `json:"metadata,omitempty" swaggerignore:"true"`
+	TxHash      types.HexBytes `json:"txHash" `
+	MetadataURL string         `json:"metadataURL" swaggertype:"string"`
```
altergui: why do you need `swaggertype:"string"` if the type is already a string?

apiclient/helpers.go
```diff
@@ -343,3 +345,24 @@ func (c *HTTPclient) GenerateSik(sign, secret []byte) ([]byte, error) {
...
 	return arbo.BigIntToBytes(arbo.HashFunctionPoseidon.Len(), hash), nil
 }
+
+// Nullifier returns snark friendly nullifier based on the provided signature,
+// the user secret and electionId. The nullifier is calculated following its
+// definition:
+//
+//	nullifier = poseidon(signature, secret*, sha256(electionId))
+//
+// *The secret is optional.
```
altergui: nit: the `*` has a special meaning in golang so i'd avoid using it for other meanings. you can simply mention "secret", it will be evident you're referring to the parameter, no need for `*`

what does "optional" mean? parameters are never optional like in other languages

my suggestion

```go
//  nullifier = poseidon(signature, secret, sha256(electionId))
//
// secret can be nil, in that case blah will happen
```

crypto/zk/utils.go
```diff
@@ -20,6 +20,19 @@ const (
 	publicSigLen = 7
 )
 
+var modulus, _ = new(big.Int).SetString("21888242871839275222246405745257275088548364400416034343698204186575808495617", 10)
```
altergui: seeing this number made me 21888242871839275222246405745257275088548364400416034343698204186575808495617 times curious to know where does it come from. please add a comment?

altergui (commented):
i'm still completely lost on the actual code, so i'm just contributing some trivial comments based on "blind" proofreading

Dockerfile
lucasmenendez: Do you know if this works with docker-compose too? or only with docker?

crypto/zk/utils.go
lucasmenendez: hahaha, for sure, thanks a lot!

Dockerfile
altergui: both. docker-compose uses docker to build images

vochain/state/sik.go
lucasmenendez: it can't be, it is used later

vochain/transaction/proof.go
lucasmenendez: done, i will check the rest of the code to ensure that we use the same naming

crypto/zk/prover/prover.go
```diff
@@ -105,7 +105,7 @@ func (p *Proof) Nullifier() (*big.Int, error) {
 		return nil, ErrPublicSignalFormat
 	}
 	// Get the nullifier from the fourth public signal of the proof
-	strNullifier := p.PubSignals[3]
+	strNullifier := p.PubSignals[2]
```
altergui: fourth? or third?

cmd/end2endtest/hysteresis.go
```diff
@@ -0,0 +1,129 @@
...
+
+	log.Infow("votes submitted successfully",
+		"n", validVotes, "time", time.Since(startTime),
+		"vps", int(float64(validVotes)/time.Since(startTime).Seconds()))
+
+	if err := t.verifyVoteCount(validVotes); err != nil {
+		return err
+	}
+
+	log.Info("wating to reach hysteresis to create new account and force to delete old sikroots")
```
mariajdab: typo `wating`

cmd/end2endtest/hysteresis.go
```diff
@@ -0,0 +1,129 @@
...
+		return err
+	}
+
+	log.Info("create second account to force to update sik roots")
+	testAccount = ethereum.NewSignKeys()
+	if err := testAccount.Generate(); err != nil {
+		return err
+	}
+	testAccountPrivKey = testAccount.PrivateKey()
+	if _, _, err := t.createAccount(testAccountPrivKey.String()); err != nil {
```
mariajdab: I would suggest due this code is repeated twice, you could use a anonymous function: 
```
	forceUpdateSikRoots := func() error {
		log.Info("wating to reach hysteresis to create new account and force to delete old sikroots")

		ctx, cancel := context.WithTimeout(context.Background(), t.config.timeout)
		defer cancel()
		if err := t.api.WaitUntilNBlocks(ctx, state.SIKROOT_HYSTERESIS_BLOCKS); err != nil {
			return err
		}
		log.Info("create to account to force to update sik roots")
		testAccount := ethereum.NewSignKeys()
		if err := testAccount.Generate(); err != nil {
			return err
		}
		testAccountPrivKey := testAccount.PrivateKey()
		if _, _, err := t.createAccount(testAccountPrivKey.String()); err != nil {
			return err
		}
		return nil
	}
```
It's not worth putting it in the helpers since it won't be used again, I think.

vochain/state/processblockregistry.go
```diff
@@ -0,0 +1,83 @@
...
+	"go.vocdoni.io/dvote/statedb"
+)
+
+type ProcessBlockRegistry struct {
+	startBlocksDB statedb.Updater
+	dbLock        *treeTxWithMutex
+	mainTree      statedb.TreeViewer
+}
+
+// RegisterStartBlock method creates a new record on the key-value database
```
mariajdab: Just a quick remember to keep the comments starting whit the function o method name

vochain/state/processblockregistry.go
```diff
@@ -0,0 +1,83 @@
...
+
+// DeleteStartBlock method deletes the record on the key-value database
+// associated to the processes subtree for the process ID provided.
+func (pbr *ProcessBlockRegistry) DeleteStartBlock(pid []byte) error {
+	pbr.dbLock.Lock()
+	defer pbr.dbLock.Unlock()
+	return pbr.startBlocksDB.Delete(pid)
+}
+
+// MinReadyProcessStartBlock returns the minimun start block of the current on going
```
mariajdab: I think this needs to be updated too

vochain/state/sik.go
```diff
@@ -0,0 +1,274 @@
...
+	// check if exists a registered sik for the provided address, query also for
+	// no commited tree version
+	v.Tx.Lock()
+	rawSik, err := siksTree.Get(address.Bytes())
+	v.Tx.Unlock()
+	if errors.Is(err, arbo.ErrKeyNotFound) {
+		// if not exists create it
+		log.Debugw("setSIK (create)",
+			"address", address.String(),
+			"sik", SIK(newSik).String())
```
mariajdab: If I am not wrong `newSik` is already a SIK type (type of the argument function)

crypto/ethereum/ethereum.go
```diff
@@ -182,6 +188,18 @@ func (k *SignKeys) SignVocdoniMsg(message []byte) ([]byte, error) {
 	return signature, nil
 }
 
+// SignVocdoniSik signs the default vocdoni sik payload. It envolves the
+// SignEthereum method, using the DefaultSikPayload and discarding the last
+// byte of the signature (used for recovery), different that the same byte of a
+// signature generated with javascript.
+func (k *SignKeys) SignVocdoniSik() ([]byte, error) {
```
p4u: I'd call it something such as `SIKsignature()`

crypto/ethereum/ethereum.go
```diff
@@ -196,6 +214,66 @@ func (k *SignKeys) VerifySender(message, signature []byte) (bool, ethcommon.Addr
...
 }
 
+// CustomSik method generates the Secret Identity Key for the current SignKeys
+// with the signature of the DefaultSikPayload and the user secret (if it is
+// provided) following the definition:
+//
+//	SIK = poseidon(address, signature, secret)
+//
+// The secret could be nil.
+func (k *SignKeys) CustomSik(secret []byte) ([]byte, error) {
```
p4u: For clarity, I'd rather call it `AccountSIK(secret []byte)` and remove the Sik() method bellow. 
The caller can just specify secret=nil.

crypto/ethereum/ethereum.go
```diff
@@ -196,6 +214,66 @@ func (k *SignKeys) VerifySender(message, signature []byte) (bool, ethcommon.Addr
...
+
+// Sik method returns the calculated SIK for the current SignKeys with any
+// password or secret as input.
+func (k *SignKeys) Sik() ([]byte, error) {
+	return k.CustomSik([]byte{})
+}
+
+// Nullifier method composes the nullifier of the current SignKeys for the
+// desired election id and the secret provided.
+func (k *SignKeys) Nullifier(electionId, secret []byte) ([]byte, error) {
```
p4u: We should specify this is actually the SIK nullifier for the current account.  So I'd propose `AccountSIKnullifier()`

crypto/zk/circuit/inputs.go
```diff
@@ -12,45 +14,71 @@ import (
...
 // GenerateCircuitInput receives the required parameters to encode them
 // correctly to return the expected CircuitInputs. This function uses the
 // ZkAddress to get the private key and generates the nullifier. Also encodes
 // the census root, the election id and the weight provided, and includes the
 // census siblings provided into the result.
-func GenerateCircuitInput(zkAddr *zk.ZkAddress, censusRoot, electionId []byte,
-	factoryWeight, votingWeight *big.Int, censusSiblings []string) (*CircuitInputs, error) {
-	if zkAddr == nil || censusRoot == nil || electionId == nil ||
-		factoryWeight == nil || votingWeight == nil || len(censusSiblings) == 0 {
+func GenerateCircuitInput(account *ethereum.SignKeys, password, electionId,
```
p4u: Create a struct out of these parameters.

vochain/state/process.go
```diff
@@ -211,9 +211,28 @@ func (v *State) UpdateProcess(p *models.Process, pid []byte) error {
...
 		return ErrProcessNotFound
 	}
+	// update the process
 	v.Tx.Lock()
-	defer v.Tx.Unlock()
-	return updateProcess(&v.Tx, p, pid)
+	if err := updateProcess(&v.Tx, p, pid); err != nil {
+		return err
+	}
+	v.Tx.Unlock()
```
p4u: The Unlock should be deferred as it was before, else in case of an error on updateProcess, the mutex would be locked forever.

vochain/state/processblockregistry.go
```diff
@@ -0,0 +1,83 @@
+package state
+
+import (
+	"encoding/binary"
+
+	"go.vocdoni.io/dvote/statedb"
+)
+
+type ProcessBlockRegistry struct {
```
p4u, line 14: Please document what the ProcessBlockregistry does and its purpose

vochain/state/sik.go
```diff
@@ -0,0 +1,279 @@
...
+	sikLeafValueLen = 32
+)
+
+// SIK type abstracts a slice of bytes that contains the Secret Identity Key
+// value of a user
+type SIK []byte
+
+// SIKFromAddress function return the current SIK value associated to the provided
+// address.
+func (v *State) SIKFromAddress(address common.Address) (SIK, error) {
```
p4u, line 44: SIKfromAddress according to camelCase rules

vochain/state/sik.go
```diff
@@ -0,0 +1,279 @@
...
+	if err != nil {
+		return err
+	}
+	return v.UpdateSIKRoots()
+}
+
+// ValidSIKRoots returns the list of current valid roots from the SIK's merkle
+// tree. It reads the roots from the key-value database associated to the SIK's
+// subtree.
+func (v *State) ValidSIKRoots() ([][]byte, error) {
```
p4u: This looks like a query that we might perform several times on the same block, in case there are several votes attempting the same election. 

Same as we do with Height, we might store the ValidSIKRoots() slice into memory on app.BeginBlock(). Then any query from the application or transactionHandler should use the cached version of `app.SIKvalidRoots()`. Specially the ones checking the validity of a vote.

Take this into consideration.

vochain/state/sik.go
```diff
@@ -0,0 +1,279 @@
...
+		return ErrSIKAlreadyInvalid
+	}
+	v.Tx.Lock()
+	invalidatedSIK := make(SIK, sikLeafValueLen).InvalidateAt(v.CurrentHeight())
+	err = v.Tx.DeepSet(address.Bytes(), invalidatedSIK, StateTreeCfg(TreeSIK))
+	v.Tx.Unlock()
+	if err != nil {
+		return err
+	}
+	return v.UpdateSIKRoots()
```
p4u, line 128: Review all namings that do not follow the camel case rules, please :)

https://capitalizemytitle.com/camel-case/

vochain/state/sik.go
```diff
@@ -0,0 +1,279 @@
...
+	siksTree.NoState().Iterate(nil, func(_, root []byte) bool {
+		validRoots = append(validRoots, root)
+		return true
+	})
+	return validRoots, nil
+}
+
+// ExpiredSik returns if the provided siksRoot is still valid or not, checking
+// if it is included into the list of current valid sik roots.
+func (v *State) ExpiredSik(candidateRoot []byte) (bool, error) {
```
p4u: We should always use capital letters for `SIK`

vochain/state/sik.go
```diff
@@ -0,0 +1,279 @@
...
+//   - If it does not exist, remove all roots with a lower block number except
+//     for the next lower sikRoot. It is becouse it still being validate for a
+//     period.
+func (v *State) UpdateSIKRoots() error {
+	// get sik roots key-value database associated to the siks tree
+	v.Tx.Lock()
+	defer v.Tx.Unlock()
+	siksTree, err := v.Tx.DeepSubTree(StateTreeCfg(TreeSIK))
+	if err != nil {
+		return err
```
p4u: Please, try to wrap the errors (done in my proposed function)

vochain/state/sik.go
```diff
@@ -0,0 +1,279 @@
...
+// voting to already registered users when an election is on going and new users
+// are registered. When a new sikRoot is generated, the sikRoot’s from an older
+// block than the current block minus the hysteresis blocks will be deleted:
+//   - If exists a sikRoot for the minimun hysteresis block number
+//     (currentBlock - hysteresis), just remove all the roots with a lower block
+//     number.
+//   - If it does not exist, remove all roots with a lower block number except
+//     for the next lower sikRoot. It is becouse it still being validate for a
+//     period.
+func (v *State) UpdateSIKRoots() error {
```
p4u, line 179: You can simplify this function by doing a single iteration.

```
func (v *State) UpdateSIKRoots() error {
	// get sik roots key-value database associated to the siks tree
	v.Tx.Lock()
	defer v.Tx.Unlock()

	siksTree, err := v.Tx.DeepSubTree(StateTreeCfg(TreeSIK))
	if err != nil {
		return fmt.Errorf("error getting deep sub tree: %w", err)
	}
	newSikRoot, err := siksTree.Root()
	if err != nil {
		return fmt.Errorf("error getting siksTree root: %w", err)
	}

	sikRootsDB := siksTree.NoState()

	currentBlock := v.CurrentHeight()
	if currentBlock > SIKROOT_HYSTERESIS_BLOCKS {
		minBlock := currentBlock - SIKROOT_HYSTERESIS_BLOCKS
		minBlockKey := make([]byte, 32)
		binary.LittleEndian.PutUint32(minBlockKey, minBlock)

		var toPurge [][]byte
		var nearestLowerBlock uint32

		sikRootsDB.Iterate(nil, func(key, value []byte) bool {
			candidateKey := bytes.Clone(key)
			blockNumber := binary.LittleEndian.Uint32(candidateKey)
			if blockNumber < minBlock {
				if _, err := sikRootsDB.Get(minBlockKey); err == nil || blockNumber > nearestLowerBlock {
					toPurge = append(toPurge, candidateKey)
					nearestLowerBlock = blockNumber
				}
			}
			return true
		})

		for _, blockToDelete := range toPurge {
			if err := sikRootsDB.Delete(blockToDelete); err != nil {
				return fmt.Errorf("error deleting block: %w", err)
			}
			log.Debugw("updateSIKRoots (deleted)",
				"blockNumber", binary.LittleEndian.Uint32(blockToDelete))
		}
	}

	log.Debugw("updateSIKRoots (created)",
		"newSikRoot", hex.EncodeToString(newSikRoot),
		"blockNumber", currentBlock)

	blockKey := make([]byte, 32)
	binary.LittleEndian.PutUint32(blockKey, currentBlock)

	if err := sikRootsDB.Set(blockKey, newSikRoot); err != nil {
		return fmt.Errorf("error setting new sikRoot: %w", err)
	}

	return nil
}
```

vochain/transaction/account_tx.go
```diff
@@ -15,95 +15,96 @@ import (
 )
 
 // CreateAccountTxCheck checks if an account creation tx is valid
```
p4u: I'm uncertain if it is worth it to return the SIK, the function does not perform any operation on it, so there won't be extra cost if tx.GetSIK() is executed by the caller (instead of the function).

vochain/transaction/account_tx.go
```diff
@@ -239,3 +240,133 @@ func (t *TransactionHandler) SetAccountInfoTxCheck(vtx *vochaintx.Tx) error {
 	}
 	return nil
 }
+
+// DelSikTxCheck checks if a delete sik tx is valid
+func (t *TransactionHandler) DelSikTxCheck(vtx *vochaintx.Tx) (common.Address, error) {
```
p4u: Shouldn't we check if the SIK can actually be deleted?  So it must exist already and not be invalidated.

vochain/transaction/account_tx.go
```diff
@@ -239,3 +240,133 @@ func (t *TransactionHandler) SetAccountInfoTxCheck(vtx *vochaintx.Tx) error {
...
+	}
+
+	newSik := vtx.Tx.GetSetSik().GetSik()
+	if newSik == nil {
+		return common.Address{}, nil, fmt.Errorf("no sik value provided")
+	}
+	// check if the address already has invalidated sik to ensure that it is
+	// not updated after reach the correct height to avoid double voting
+	if currentSik, err := t.state.SIKFromAddress(txAddress); err == nil {
+		maxEndBlock, err := t.state.ProcessBlockRegistry.MaxEndBlock(t.state.CurrentHeight(), false)
```
p4u: Please use app.Height() instead

vochain/transaction/account_tx.go
```diff
@@ -239,3 +240,133 @@ func (t *TransactionHandler) SetAccountInfoTxCheck(vtx *vochaintx.Tx) error {
...
+		return common.Address{}, nil, fmt.Errorf("cannot get tx account: %w", err)
+	}
+
+	newSik := vtx.Tx.GetSetSik().GetSik()
+	if newSik == nil {
+		return common.Address{}, nil, fmt.Errorf("no sik value provided")
+	}
+	// check if the address already has invalidated sik to ensure that it is
+	// not updated after reach the correct height to avoid double voting
+	if currentSik, err := t.state.SIKFromAddress(txAddress); err == nil {
```
p4u: You should check that err != nil and different to arbo.KeyNotFound

vochain/transaction/account_tx.go
```diff
@@ -239,3 +240,133 @@ func (t *TransactionHandler) SetAccountInfoTxCheck(vtx *vochaintx.Tx) error {
...
+		return common.Address{}, fmt.Errorf("cannot extract address from public key: %w", err)
+	}
+
+	if _, err := t.state.GetAccount(txAddress, false); err != nil {
+		return common.Address{}, fmt.Errorf("cannot get tx account: %w", err)
+	}
+	return txAddress, nil
+}
+
+// SetSikTxCheck checks if a set sik tx is valid
```
p4u: Explain how this function allows or discards a potential sik change.

vochain/transaction/account_tx.go
```diff
@@ -239,3 +240,133 @@ func (t *TransactionHandler) SetAccountInfoTxCheck(vtx *vochaintx.Tx) error {
...
+	}
+	txAddress, err := ethereum.AddrFromPublicKey(pubKey)
+	if err != nil {
+		return common.Address{}, nil, fmt.Errorf("cannot extract address from public key: %w", err)
+	}
+	if _, err := t.state.GetAccount(txAddress, false); err != nil {
+		return common.Address{}, nil, fmt.Errorf("cannot get the account for the tx address: %w", err)
+	}
+	newSik := tx.GetSik()
+	if newSik == nil {
```
p4u: Do the "cheap" checks before the expensive.

vochain/transaction/account_tx.go
```diff
@@ -239,3 +240,133 @@ func (t *TransactionHandler) SetAccountInfoTxCheck(vtx *vochaintx.Tx) error {
...
+		return common.Address{}, nil, fmt.Errorf("no sik value provided")
+	}
+	// check if the address already has invalidated sik to ensure that it is
+	// not updated after reach the correct height to avoid double voting
+	if currentSik, err := t.state.SIKFromAddress(txAddress); err == nil {
```
p4u: I'd not allow to change the SIK on this function, to simplify and reduce potential attack vectors. I'd just allow register if the SIK does not exist at all (not even invalidated).

vochain/transaction/account_tx.go
```diff
@@ -239,3 +240,133 @@ func (t *TransactionHandler) SetAccountInfoTxCheck(vtx *vochaintx.Tx) error {
...
+		}
+		if height := currentSik.DecodeInvalidatedHeight(); height >= maxEndBlock {
+			return common.Address{}, nil, fmt.Errorf("the sik could not be changed yet")
+		}
+	}
+	// check if the address already has a valid sik
+	if _, err := t.state.SIKFromAddress(txAddress); err == nil {
+		return txAddress, nil, fmt.Errorf("this address already has a valid sik")
+	}
+	// get census proof and election information provided
```
p4u: I'd do all the gets and basic checks (such as the electionID field exist) at the beginning of this function. Since all of these are cheap checks.

vochain/transaction/account_tx.go
```diff
@@ -239,3 +240,133 @@ func (t *TransactionHandler) SetAccountInfoTxCheck(vtx *vochaintx.Tx) error {
...
+	censusProof := tx.GetCensusProof()
+	if censusProof == nil {
+		return common.Address{}, nil, fmt.Errorf("no proof provided")
+	}
+	electionId := tx.GetElectionId()
+	if electionId == nil {
+		return common.Address{}, nil, fmt.Errorf("no proof provided")
+	}
+	election, err := t.state.Process(electionId, false)
```
p4u: Warning! You should check the election is in READY state

vochain/transaction/transaction.go
```diff
@@ -195,22 +195,22 @@ func (t *TransactionHandler) CheckTx(vtx *vochaintx.Tx, forCommit bool) (*Transa
```
p4u: Pau: keep review from here

p4u (changes requested):

vochain/state/sik.go
lucasmenendez: According with this page and others that I've seen, the correct form is `UpdateSIKRoots` or `UpdateSikRoots` (but the last one brakes our *rule* of keep SIK in uppercase)

vochain/state/sik.go
lucasmenendez: Also on protobuf messages?

vochain/transaction/account_tx.go
lucasmenendez: It seems that i have not access to the app here... But probably i am wrong

vochain/transaction/vote_tx.go
```diff
@@ -129,40 +129,46 @@ func (t *TransactionHandler) VoteTxCheck(vtx *vochaintx.Tx, forCommit bool) (*vs
...
-		// Get vote weight from proof publicSignals
+		// get sikroot from the proof
+		proofSIKRoot, err := proof.SIKRoot()
+		if err != nil {
+			return nil, fmt.Errorf("failed getting sik root from the proof: %w", err)
+		}
+		// check if it is expired
+		if expired, err := t.state.ExpiredSIK(proofSIKRoot); err != nil {
+			return nil, fmt.Errorf("error checking sik root expiration")
+		} else if expired {
```
p4u: you can avoid the else

vochain/transaction/account_tx.go
lucasmenendez: In fact, `CheckTx` function does not need the SIK unless the Tx is for commit, so we can call `tx.GetSIK()` just in this case.

p4u (approved):

vochain/state/sik.go
```diff
@@ -323,6 +323,63 @@ func (v *State) CountRegisterSIK(pid []byte) (uint32, error) {
 	return binary.LittleEndian.Uint32(rawCount), nil
 }
 
+// SIKAssingTo function persists the relation between a created SIK (without
```
altergui: func name is outdated

tree/arbo/tree.go
```diff
@@ -1008,21 +1008,32 @@ func PackSiblings(hashFunc HashFunction, siblings [][]byte) ([]byte, error) {
 
 // UnpackSiblings unpacks the siblings from a byte array.
 func UnpackSiblings(hashFunc HashFunction, b []byte) ([][]byte, error) {
-	if len(b) < 2 {
+	// to prevent runtime slice out of bounds error check if the lenght of the
+	// rest of the slice is at least equal to the encoded full length value
+	if len(b) < 4 {
 		return nil, fmt.Errorf("no enought size of packed siblings provided")
```
altergui: the `t` moved a loooong way: should be "not enough size"

but maybe it could be even `no packed siblings provided`?

tree/arbo/tree.go
```diff
@@ -1008,21 +1008,32 @@ func PackSiblings(hashFunc HashFunction, siblings [][]byte) ([]byte, error) {
 
 // UnpackSiblings unpacks the siblings from a byte array.
 func UnpackSiblings(hashFunc HashFunction, b []byte) ([][]byte, error) {
-	if len(b) < 2 {
+	// to prevent runtime slice out of bounds error check if the lenght of the
```
altergui: `length` is misspelled in all of the comments

vochain/state/processblockregistry.go
```diff
@@ -33,7 +33,7 @@ func (pbr *ProcessBlockRegistry) DeleteStartBlock(pid []byte) error {
 
 // MinStartBlock returns the minimun start block of the current on going
 // processes from the no-state db associated to the process sub tree.
-func (pbr *ProcessBlockRegistry) MinStartBlock(fromBlock uint32, committed bool) (uint32, error) {
+func (pbr *ProcessBlockRegistry) MinStartBlock(fromBlock uint32, _ bool) (uint32, error) {
```
altergui: given there's no interface involved, this unused parameter sounds like a bug or an incomplete feature. same with MaxEndBlock

PR #1013, opened 2023-06-25
vochain/indexer: implement process times via block JOINs too

mvdan:
(see commit message)

### DIFF
--- a/vochain/app.go
+++ b/vochain/app.go
@@ -492,6 +492,8 @@ func (app *BaseApplication) Height() uint32 {
 	return app.height.Load()
 }
 
+// TODO(mvdan): both of these methods are unused now, and perhaps TimestampFromBlock too
+
 // Timestamp returns the last block end timestamp
 func (app *BaseApplication) Timestamp() int64 {
 	return app.endBlockTimestamp.Load()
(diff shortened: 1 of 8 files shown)


mvdan:
Hm, the e2e test fails because it sets StartBlock to an estimated block given a timestamp, and perhaps that block isn't created or indexed yet.

Maybe this PR is a bad idea. I assumed that CreationTime was basically the block timestamp from StartBlock, but it looks like it's the timestamp from the block that _created_ the process, not the block that _started or will start_ the process.

p4u:
@mvdan I need help to understand what you are trying to do here.

mvdan:
Just like we removed the creation timestamps from the votes table, I wanted to do the same with the processes table. It's not needed due to performance like it was with votes, but I thought it would be another nice deduplication of data. That much I think was fine, even if it wasn't high priority.

However, I assumed that a process's creation time was the block time of its StartBlock, but it's not, because StartBlock could be in the future. That's why the e2e tests fail.

I see two options:

1) Leave the code the way it was.
2) Add a `CreateBlock` column to processes, and use that to JOIN with the blocks table to retrieve the "creation time".

They both seem fine, so if you're fine with the existing code, I think going with 1 and closing is fine as well.

p4u:
But on option 2, you would still have the same issue. If startBlock is in the future, the JOIN with block tables won't work.

So, I'm fine with 1, but you are welcome to find a more creative way to solve this issue :)

Issue #1020, opened 2023-06-29
Fix deepsource issues


p4u:
Start working on resolving the issues reported by deepsource:

https://app.deepsource.com/gh/vocdoni/vocdoni-node/issues/?category=all&page=1

altergui:
a couple more deepsource PRs #1046 #1057 were merged.

there are still a few outstanding issues reported by deepsource, most of them are low impact but it would be nice to fix them all

PR #1022, opened 2023-06-30
fix: deepsource issues

mariajdab:
- #1020 

- nested if can be replaced with else if  `CRT-A0011`

- switch with single case can be rewritten as if or if-else `CRT-A0014`

- simplify slice expression to sliced value itself  `CRT-A0016`

- call to os.Exit or log.Fatal and friends made in function using defer `CRT-D0011`

- fix comments `GO-D5001, GO-D5001, GO-D5003`

- potential slowloris attack (add timeout to the server) `GO-S2112`

-  fix typos, remove unnecessary type conversion


p4u (approved):

mariajdab:
I added a few more changes, could you please check again @p4u? Thanks

PR #1025, opened 2023-07-02
index vote packages, small sqlite breaking change cleanups

mvdan:


p4u:
Nice trick with the two databases.

p4u (approved):

PR #1028, opened 2023-07-03
test: refactor to allow benchmark Vote (WIP, MVP)

### LINKED ISSUE #965: vochain high level benchmark
We need a high level benchmark for testing the vochain performance, as pointed by issue #963.

This benchmark should be as realistic as possible (without requiring a network) and execute the whole stack: app, state, indexer and api.

A good starting point is this test: https://github.com/vocdoni/vocdoni-node/blob/master/test/api_test.go#L25

We might transform this test into a benchmark where the number of votes and parallel voting routines can be specified.

The benchmark should return the tempos required for performing an election of the specified size (from the creation to the result's scrutiny).

Note that the benchmarks on `benchmarks/` directory are outdated, since all of them use the old RPC interface. However they can be used as example for the new benchmark.

altergui:
started working on #965 and opted for a refactor that will allow me reuse most of the code


vochain/indexer/vote.go
```diff
@@ -41,6 +41,7 @@ func (idx *Indexer) GetEnvelope(nullifier []byte) (*indexertypes.EnvelopePackage
 		OverwriteCount:       uint32(voteRef.OverwriteCount),
 		Date:                 voteRef.BlockTime,
 		Meta: indexertypes.EnvelopeMetadata{
+			VoterID:   voteRef.VoterID.Address(),
```
altergui: this is actually a hotfix of an issue i found out while working on this. will push a separate PR for this

test/api_test.go
```diff
@@ -138,51 +186,54 @@ func TestAPIcensusAndVote(t *testing.T) {
...
+		v2 := &api.Vote{}
+		err := json.Unmarshal(resp, v2)
+		qt.Assert(t, err, qt.IsNil)
+		qt.Assert(t, v2.VoteID.String(), qt.Equals, voter.vote.VoteID.String())
+		qt.Assert(t, v2.BlockHeight, qt.Equals, uint32(2))
+		qt.Assert(t, v2.VoterID.String(), qt.Equals, fmt.Sprintf("%x", voter.key.Address().Bytes()))
+	}
 }
 
 func TestAPIaccount(t *testing.T) {
```
altergui, line 258: there's a lot of code in TestAPIaccount that could be deduplicated now, after the proposed refactoring. i'd include this in the same PR

altergui:
how does it look so far? am i on the right track?

test/benchmark_vochain_test.go
```diff
@@ -0,0 +1,26 @@
+package test
```
p4u, line 1: @mvdan can you check the basics of this benchmark and maybe propose improvements? The idea is to have a full benchmark to test performance of state+app+indexer+api

mvdan: I think the logic looks okay. Two suggestions come to mind:

* What kind of election is this - encrypted or not? We should probably benchmark the one that is most common. If both are common, then I would suggest that the benchmark do _both_ side by side. We could similarly think about other kinds of votes, like overwrites.

* You use a sequential benchmark via `b.N`, which is fine, but not very realistic - in reality, people won't vote in a big election in a big line :) For that reason, I would use a parallel benchmark via https://pkg.go.dev/testing#B.RunParallel. You can still use `b.N` as the total number of iterations to set up the census and anything else you need, but the looping should use https://pkg.go.dev/testing#PB.Next instead.

mvdan: Also, should the benchmark advance votes while iterating? Because again, in reality, you won't have an entire election get all of its votes within a single block. Plus, the indexer does a significant amount of its work in the `Commit` method, so you should advance at least one block while the benchmark timer is on.

PR #1030, opened 2023-07-05
indexer: GetEnvelope now includes VoterID in returned envelope

altergui:
Until now, the returned envelope was missing this field,
causing the endpoint `GET /votes/{voteId}` to return a nil VoterID


altergui:
i dug into this while working on #1028

https://github.com/vocdoni/vocdoni-node/blob/af85c0571e2834cf85a1906e1ece99c12f8939a2/test/api_test.go#L184-L185

i'm not uncommenting that line on this PR (now it works) since i'm refactoring the whole thing on #1028

jordipainan (approved):

PR #1032, opened 2023-07-06
fix/api/err transaction not found

jordipainan:
This PR fixes some errors and what I think are not desired behaviours, related to 204 HTTP status code and ErrTransactionNotFound:

- The ErrTransactionNotFound message changes from: "transaction hash not found" to "transaction not found" because it is used for other transaction API calls that are not limited to work with hashes.

- I've found that 204 http error code was used just in the ErrTransactionNotFound specific error case and it does make sense to keep using 404 errors for any kind of error like this. With this reasoning this commit changes the error code from 204 to 404.
Adding @marcvelmer to the conversation as probably affects the SDK.

- Finally, the Send() method of the httprouter was not writing nicely a void data stream required by the HTTP standard when handling 204 errors, aka not calling the Writer.Write() function in the server code.
This raised some errors in clients interacting with the API, returing a `HTTP/2 stream 1 was not closed cleanly: PROTOCOL_ERROR (err 1)` error. Hopefully this will be avoided in the future with the changes.

Please do not squash. See commit messages.

cc: @selankon @mvdan @elboletaire

### DIFF
--- a/api/errors.go
+++ b/api/errors.go
@@ -33,7 +33,7 @@ var (
 	ErrAccountAlreadyExists             = apirest.APIerror{Code: 4004, HTTPstatus: apirest.HTTPstatusBadRequest, Err: fmt.Errorf("account already exists")}
 	ErrTreasurerNotFound                = apirest.APIerror{Code: 4005, HTTPstatus: apirest.HTTPstatusNotFound, Err: fmt.Errorf("treasurer account not found")}
 	ErrOrgNotFound                      = apirest.APIerror{Code: 4006, HTTPstatus: apirest.HTTPstatusNotFound, Err: fmt.Errorf("organization not found")}
-	ErrTransactionNotFound              = apirest.APIerror{Code: 4007, HTTPstatus: apirest.HTTPstatusNoContent, Err: fmt.Errorf("transaction hash not found")}
+	ErrTransactionNotFound              = apirest.APIerror{Code: 4007, HTTPstatus: apirest.HTTPstatusNoContent, Err: fmt.Errorf("transaction not found")}
 	ErrBlockNotFound                    = apirest.APIerror{Code: 4008, HTTPstatus: apirest.HTTPstatusNotFound, Err: fmt.Errorf("block not found")}
 	ErrMetadataProvidedButNoURI         = apirest.APIerror{Code: 4009, HTTPstatus: apirest.HTTPstatusBadRequest, Err: fmt.Errorf("metadata provided but no metadata URI found in transaction")}
 	ErrMetadataURINotMatchContent       = apirest.APIerror{Code: 4010, HTTPstatus: apirest.HTTPstatusBadRequest, Err: fmt.Errorf("metadata URI does not match metadata content")}
--- a/httprouter/message.go
+++ b/httprouter/message.go
@@ -49,17 +49,21 @@ func (h *HTTPContext) Send(msg []byte, httpStatusCode int) error {
 		// The connection was closed, so don't try to write to it.
 		return fmt.Errorf("connection is closed")
 	}
-	h.Writer.Header().Set("Content-Length", fmt.Sprintf("%d", len(msg)+1))
 	h.Writer.Header().Set("Content-Type", "application/json")
+
+	if httpStatusCode == http.StatusNoContent {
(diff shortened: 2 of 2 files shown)


p4u:
The 204 (No Content success) return coded was added in purpose (after some discussion) to avoid raising browser error messages (404). 

What's now the reasoning for regressing to 404 @jordipainan ?

jordipainan:
> The 204 (No Content success) return coded was added in purpose (after some discussion) to avoid raising browser error messages (404).
> 
> What's now the reasoning for regressing to 404 @jordipainan ?

As commented in the PR message: I've found that 204 http error code was used just in the ErrTransactionNotFound specific error case and it does make sense to keep using 404 errors for any kind of error like this. With this reasoning this commit changes the error code from 204 to 404.

If this code was added in purpose why is just used in one specific error an not on others that are of the same type? Shouldnt be the API using extensively 204 instead of 404 ? I tried to be pragmatic here and just change the discordant error code.

Note that we still have support for 204.

altergui:
i suspect the `HTTP/2 stream 1 was not closed cleanly: PROTOCOL_ERROR (err 1) error` is due to the fact that `204 No content` sends no body at all, and i guess our code (in vocdoni-node) is buggy in that sense. i can dig into this

jordipainan:
> i suspect the `HTTP/2 stream 1 was not closed cleanly: PROTOCOL_ERROR (err 1) error` is due to the fact that `204 No content` sends no body at all, and i guess our code (in vocdoni-node) is buggy in that sense. i can dig into this

AFAIK It shouldnt be sending any body and just a void stream as indicated in my previous comment. I think the changes I've made solve this situation. See: 5717c988c7770a604e5208051dfac93cf58380ce

jordipainan:
@p4u reverted 204 change

api/errors.go
```diff
@@ -33,7 +33,7 @@ var (
 	ErrAccountAlreadyExists             = apirest.APIerror{Code: 4004, HTTPstatus: apirest.HTTPstatusBadRequest, Err: fmt.Errorf("account already exists")}
 	ErrTreasurerNotFound                = apirest.APIerror{Code: 4005, HTTPstatus: apirest.HTTPstatusNotFound, Err: fmt.Errorf("treasurer account not found")}
 	ErrOrgNotFound                      = apirest.APIerror{Code: 4006, HTTPstatus: apirest.HTTPstatusNotFound, Err: fmt.Errorf("organization not found")}
-	ErrTransactionNotFound              = apirest.APIerror{Code: 4007, HTTPstatus: apirest.HTTPstatusNoContent, Err: fmt.Errorf("transaction hash not found")}
+	ErrTransactionNotFound              = apirest.APIerror{Code: 4007, HTTPstatus: apirest.HTTPstatusNoContent, Err: fmt.Errorf("transaction not found")}
```
altergui, line 36: do note that this string is never sent to the client, it just appears on our logs. but the change LGTM, the word `hash` is indeed unneeded

altergui (commented):
i believe the code would be cleaner / more idiomatic if you handle the case `if httpStatusCode == http.StatusNoContent` first instead, so the "default" code flow is deintended (and there's no else clause)

but it's just a belief and maybe i'm wrong, i didn't try to write it. if it's cleaner this way, it's OK, LGTM

api/errors.go
jordipainan: yup, noticed! I changed it also because of the httpstatus change, but finally wont be applied. 
The word hash will be removed anyways so we can have cleaner logs

jordipainan:
> i believe the code would be cleaner / more idiomatic if you handle the case `if httpStatusCode == http.StatusNoContent` first instead, so the "default" code flow is deintended (and there's no else clause)
> 
> but it's just a belief and maybe i'm wrong, i didn't try to write it. if it's cleaner this way, it's OK, LGTM

applied

jordipainan:
Lets squash now

altergui (approved):

altergui:
> This PR modifies the autogenerated vocdoni-api.yaml: Check PR [vocdoni/developer-portal#98](https://github.com/vocdoni/developer-portal/pull/98)

(haha  so untrue :rofl:  obviously ignore. my fault, i will fix this)

p4u (approved):

PR #1033, opened 2023-07-07
fix deepsource issues

mariajdab:
Continue fixing the  reported deepsourse issues, plus another fixes
- #1020

### DIFF
--- a/cmd/end2endtest/account.go
+++ b/cmd/end2endtest/account.go
@@ -44,19 +44,18 @@ func (t *E2ETokenTxs) Setup(api *apiclient.HTTPclient, config *config) error {
 
 	// create alice signer
 	t.alice = ethereum.NewSignKeys()
-	err := t.alice.Generate()
-	if err != nil {
-		return err
+	if err := t.alice.Generate(); err != nil {
+		return fmt.Errorf("error in alice.Generate: %w", err)
 	}
 
 	// create bob signer
 	t.bob = ethereum.NewSignKeys()
-	err = t.bob.Generate()
-	if err != nil {
-		return err
+	if err := t.bob.Generate(); err != nil {
+		return fmt.Errorf("error in bob.Generate: %w", err)
 	}
 
 	// get faucet package for alice
+	var err error
 	t.aliceFP, err = faucetPackage(t.config.faucet, t.config.faucetAuthToken, t.alice.Address().Hex())
 	if err != nil {
 		return err
@@ -64,12 +63,12 @@ func (t *E2ETokenTxs) Setup(api *apiclient.HTTPclient, config *config) error {
 
 	// check transaction cost
 	if err := testGetTxCost(t.api); err != nil {
-		log.Fatal(err)
+		return fmt.Errorf("error in testGetTxCost: %w", err)
 	}
 
 	// check create and set account
 	if err := testCreateAndSetAccount(t.api, t.aliceFP, t.alice, t.bob); err != nil {
-		log.Fatal(err)
+		return fmt.Errorf("error in testCreateAndSetAccount: %w", err)
 	}
 
 	return nil
@@ -83,7 +82,7 @@ func (t *E2ETokenTxs) Teardown() error {
 func (t *E2ETokenTxs) Run() error {
 	// check send tokens
 	if err := testSendTokens(t.api, t.alice, t.bob); err != nil {
-		log.Fatal(err)
+		return fmt.Errorf("error in testSendTokens: %w", err)
 	}
 
 	return nil
@@ -110,7 +109,7 @@ func testCreateAndSetAccount(api *apiclient.HTTPclient, fp *models.FaucetPackage
 	// create account with faucet package
 	aliceAcc, err := ensureAccountExists(api.Clone(hex.EncodeToString(alice.PrivateKey())), fp)
 	if err != nil {
-		return err
+		return fmt.Errorf("error in ensureAccountExists(alice) %w", err)
 	}
 	if aliceAcc == nil {
 		return state.ErrAccountNotExist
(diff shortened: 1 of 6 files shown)


cmd/node/main.go
```diff
@@ -377,15 +376,22 @@ func main() {
 			if globalCfg.PprofPort == 0 {
 				globalCfg.PprofPort = int(time.Now().Unix()%100) + 61000
 			}
-			ln, err := net.Listen("tcp", fmt.Sprintf("0.0.0.0:%d", globalCfg.PprofPort))
-			if err != nil {
+
+			server := &http.Server{
```
p4u: This is the pprof endpoint for debuging purposes. We don't want to have a timeout of 10 seconds, some times it requires more than 30 minutes waiting time in order to recover a big enough trace for analyzing the cpu, memory, etc... profiles. 

This endpoint is only enabled in dev scenarios, so no need to protect it against potential slowloris attacks.

httprouter/httprouter.go
```diff
@@ -234,7 +234,14 @@ func (r *HTTProuter) AddRawHTTPHandler(pattern, HTTPmethod string, handler http.
 func (r *HTTProuter) routerHandler(namespaceID string, accessType AuthAccessType,
 	handlerFunc RouterHandlerFn) func(w http.ResponseWriter, req *http.Request) {
 	return func(w http.ResponseWriter, req *http.Request) {
-		defer req.Body.Close()
+		var err error
+
+		defer func() {
+			if err := req.Body.Close(); err != nil {
+				http.Error(w, err.Error(), http.StatusInternalServerError)
```
p4u: If we cannot close the body (because there is some error) I don't think we can write an error as reply. You might just ignore this error, no need to print it either since it might be the client disconnecting from the socket, and that's nothing we can solve.

httprouter/message.go
```diff
@@ -40,7 +40,12 @@ func (h *HTTPContext) Send(msg []byte, httpStatusCode int) error {
 		}
 	}()
 	defer close(h.sent)
-	defer h.Request.Body.Close()
+
+	defer func() {
+		if err := h.Request.Body.Close(); err != nil {
+			log.Warnf("error closing body req %v", err)
```
p4u: Same here, we don't want to print this error.

cmd/end2endtest/account.go
```diff
@@ -64,12 +63,12 @@ func (t *E2ETokenTxs) Setup(api *apiclient.HTTPclient, config *config) error {
 
 	// check transaction cost
 	if err := testGetTxCost(t.api); err != nil {
-		log.Fatal(err)
+		return err
```
altergui: please wrap all these err, as in `return fmt.Errorf("testGetTxCost: %w", err)`.

else, it becomes very hard to debug where the error happened (with log.Fatal, the actual line is reported. with return, that info is lost, that's why i'm asking to wrap with a unique string)

cmd/end2endtest/account.go
```diff
@@ -64,12 +63,12 @@ func (t *E2ETokenTxs) Setup(api *apiclient.HTTPclient, config *config) error {
...
 	}
 
 	// check create and set account
 	if err := testCreateAndSetAccount(t.api, t.aliceFP, t.alice, t.bob); err != nil {
-		log.Fatal(err)
+		return err
```
altergui: same here, and all the rest where you replace log.Fatal with a naked `return err`

crypto/zk/circuit/circuit.go
```diff
@@ -211,7 +211,13 @@ func downloadFile(ctx context.Context, fileUrl string) ([]byte, error) {
...
 		return nil, err
 	}
-	defer res.Body.Close()
+
+	defer func() {
+		if err := res.Body.Close(); err != nil {
+			log.Warnf("error closing body response %v", err)
+		}
+	}()
+
```
mvdan, line 220: we're just reading from the http body, so I'm pretty sure we can safely ignore the close error. if the download started and then failed for any reason, `io.ReadAll` would catch it.

mvdan (approved):

crypto/zk/circuit/circuit.go
mvdan: as a general rule of thumb, `Close` methods should have their errors checked if you're writing to the file/object, since in some cases the "close" is what triggers a flush or some last checks. When reading, `Close` is just about releasing resources, because you're not actually sending any data that you need to ensure got received/stored properly.

mariajdab: Thanks! @mvdan

cmd/end2endtest/account.go
```diff
@@ -46,14 +46,13 @@ func (t *E2ETokenTxs) Setup(api *apiclient.HTTPclient, config *config) error {
...
+		return fmt.Errorf("error in generate new alice keys: %w", err)
 	}
 
 	// create bob signer
 	t.bob = ethereum.NewSignKeys()
-	err = t.bob.Generate()
-	if err != nil {
-		return err
+	if err := t.bob.Generate(); err != nil {
+		return fmt.Errorf("cannot generate new bob keys: %w", err)
```
altergui: keep the consistency:

* either inline `if blah ; err != nil { }` (as in bob) , or two separate lines (as in alice).

* keep the error strings as consistent as possible. why is alice `error in generate` and bob `cannot generate`?
* since this is a test tool and those strings are never exposed to any enduser, i'd even suggest to use "code-related" strings, as in `fmt.Errorf("bob.Generate: %w", err)` and so on, like you did with testGetTxCost. 
* if you want to be more verbose than that (for example `fmt.Errorf("error in bob.Generate: %w", err)`, please keep the consistency, always use `error in` instead of `cannot retrieve`, `failed to`, and all the variations. The string is already unique because of the keyword (testGetTxCost, etc).

cmd/end2endtest/account.go
```diff
@@ -110,17 +109,16 @@ func testCreateAndSetAccount(api *apiclient.HTTPclient, fp *models.FaucetPackage
 	// create account with faucet package
 	aliceAcc, err := ensureAccountExists(api.Clone(hex.EncodeToString(alice.PrivateKey())), fp)
 	if err != nil {
-		return err
+		return fmt.Errorf("cannot not ensure if aliceAcc exits %w", err)
```
altergui: `s/cannot not/cannot/`

`s/exits/exists/`

or simply `"ensureAccountExists(alice): %w"`

altergui (approved):

PR #1037, opened 2023-07-10
end2endtest: remove repeated code when ending election

mariajdab:
This will also help to make the code easier to read when adding more test cases on each end2end

### DIFF
--- a/cmd/end2endtest/helpers.go
+++ b/cmd/end2endtest/helpers.go
@@ -631,3 +631,53 @@ func cspGenProof(pid, voterKey []byte, csp *ethereum.SignKeys) (*apiclient.Censu
 
 	return &apiclient.CensusProof{Proof: caProofBytes}, nil
 }
+
+func (t *e2eElection) verifyVoteCount(nvotesExpected int) error {
+	startTime := time.Now()
+
+	// Wait for all the votes to be verified
+	for {
+		count, err := t.api.ElectionVoteCount(t.election.ElectionID)
+		if err != nil {
+			log.Warn(err)
+		}
+		if count == uint32(nvotesExpected) {
+			break
+		}
+
+		if err := t.api.WaitUntilNextBlock(); err != nil {
+			return fmt.Errorf("timeout waiting for next block")
+		}
+
+		log.Infof("verified %d/%d votes", count, nvotesExpected)
+		if time.Since(startTime) > t.config.timeout {
+			log.Fatalf("timeout waiting for votes to be registered")
+		}
+	}
+
+	log.Infof("%d votes registered successfully, took %s (%d votes/second)",
+		t.config.nvotes, time.Since(startTime), int(float64(nvotesExpected)/time.Since(startTime).Seconds()))
+
+	return nil
+}
+
+func (t *e2eElection) endElectionAndFetchResults() (*vapi.ElectionResults, error) {
+	// Set the account back to the organization account
+	api := t.api.Clone(t.config.accountPrivKeys[0])
+
+	// End the election by setting the status to ENDED
+	log.Infof("ending election...")
+	if _, err := api.SetElectionStatus(t.election.ElectionID, "ENDED"); err != nil {
+		return nil, fmt.Errorf("cannot set election status to ENDED %w", err)
+	}
+
+	// Wait for the election to be in RESULTS state
+	ctx, cancel := context.WithTimeout(context.Background(), t.config.timeout*3)
+	defer cancel()
+
+	results, err := api.WaitUntilElectionResults(ctx, t.election.ElectionID)
+	if err != nil {
+		return nil, fmt.Errorf("error waiting for election publish final results %w", err)
+	}
+	return results, nil
+}
--- a/cmd/end2endtest/ballot.go
+++ b/cmd/end2endtest/ballot.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"context"
 	"fmt"
 	"os"
 	"time"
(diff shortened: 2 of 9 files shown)


cmd/end2endtest/helpers.go
```diff
@@ -631,3 +631,53 @@ func cspGenProof(pid, voterKey []byte, csp *ethereum.SignKeys) (*apiclient.Censu
...
+	log.Infof("waiting for all the votes to be registered...")
+	for {
+		count, err := t.api.ElectionVoteCount(t.election.ElectionID)
+		if err != nil {
+			log.Warn(err)
+		}
+		if count == uint32(nvotesExpected) {
+			break
+		}
+		time.Sleep(time.Second * 5)
```
altergui: avoid time.Sleep(). do you want to WaitUntilNextBlock?

cmd/end2endtest/helpers.go
```diff
@@ -631,3 +631,53 @@ func cspGenProof(pid, voterKey []byte, csp *ethereum.SignKeys) (*apiclient.Censu
...
+
+	log.Infof("%d votes registered successfully, took %s (%d votes/second)",
+		t.config.nvotes, time.Since(startTime), int(float64(nvotesExpected)/time.Since(startTime).Seconds()))
+
+	return nil
+}
+
+func (t *e2eElection) setElectionEnd() error {
+	// Set the account back to the organization account
+	if err := t.api.SetAccount(t.config.accountPrivKeys[0]); err != nil {
```
altergui: this method modifies the receiver `t` and that may be unexpected for future callers of setElectionEnd. use t.api.Clone() instead

cmd/end2endtest/helpers.go
```diff
@@ -631,3 +631,53 @@ func cspGenProof(pid, voterKey []byte, csp *ethereum.SignKeys) (*apiclient.Censu
 
 	return &apiclient.CensusProof{Proof: caProofBytes}, nil
 }
+
+func (t *e2eElection) verifyVoteCount(startTime time.Time, nvotesExpected int) error {
```
altergui: i'd not pass startTime and instead set 	`startTime := time.Now()` inside verifyVoteCount. it does change the current behaviour but i believe it's reasonable to have each step have it's own timeout. step sendVotes, step verifyVoteCount, step endElectionAndFetchResults.

to be honest the current `c.timeout` is pretty useless (5 minutes, in start_test.sh never happens since there's a "wrapper" timeout that happens before) so we could refactor it:
```
	flag.DurationVar(&c.timeout, "timeout", apiclient.WaitTimeout, "timeout duration of each step")
```

cmd/end2endtest/helpers.go
```diff
@@ -631,3 +631,53 @@ func cspGenProof(pid, voterKey []byte, csp *ethereum.SignKeys) (*apiclient.Censu
...
+		}
+	}
+
+	log.Infof("%d votes registered successfully, took %s (%d votes/second)",
+		t.config.nvotes, time.Since(startTime), int(float64(nvotesExpected)/time.Since(startTime).Seconds()))
+
+	return nil
+}
+
+func (t *e2eElection) setElectionEnd() error {
```
altergui: is there any case where you want to end an election and not waitUntilElectionResults?

if not, i'd merge both funcs into a `endElectionAndFetchResults`

cmd/end2endtest/helpers.go
```diff
@@ -631,3 +631,53 @@ func cspGenProof(pid, voterKey []byte, csp *ethereum.SignKeys) (*apiclient.Censu
...
+	log.Infof("ending election...")
+	if _, err := t.api.SetElectionStatus(t.election.ElectionID, "ENDED"); err != nil {
+		return fmt.Errorf("cannot set election status to ENDED %w", err)
+	}
+	return nil
+}
+
+func (t *e2eElection) waitUntilElectionResults() (*vapi.ElectionResults, error) {
+	// Wait for the election to be in RESULTS state
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*100)
```
altergui: if you take the suggestion of refactoring `c.timeout` flag, you can replace this arbitrary `time.Second*100` with `c.timeout` (which would default to apiclient.WaitTimeout = 3 blocks = 24s)

altergui (approved):
this is such a great initiative, i love it :clap: 

i requested a couple of simple changes ("avoid time.Sleep()" and "use t.api.Clone"). the rest are suggestions of further refactoring or improvements that you might take or leave.

cmd/end2endtest/helpers.go
mariajdab: Using `c.timeout` (which would default to apiclient.WaitTimeout = 3 blocks = 24s) seems not enough time for election publish final results, so I update to use `t.config.timeout*3`

cmd/end2endtest/helpers.go
```diff
@@ -631,3 +631,53 @@ func cspGenProof(pid, voterKey []byte, csp *ethereum.SignKeys) (*apiclient.Censu
...
+
+	log.Infof("%d votes registered successfully, took %s (%d votes/second)",
+		t.config.nvotes, time.Since(startTime), int(float64(nvotesExpected)/time.Since(startTime).Seconds()))
+
+	return nil
+}
+
+func (t *e2eElection) endElectionAndFetchResults() (*vapi.ElectionResults, error) {
+	// Set the account back to the organization account
+	t.api = t.api.Clone(t.config.accountPrivKeys[0])
```
altergui: setting `t.api` you're also modifying the receiver :sweat_smile: 

try `api := t.api.Clone(t.config.accountPrivKeys[0])` and use `api`

altergui (approved):
LGTM, except for the endElectionAndFetchResults detail

p4u:
> LGTM, except for the endElectionAndFetchResults detail

feel free to merge it once ready

cmd/end2endtest/helpers.go
mariajdab: Yep, my bad   :sweat_smile:

PR #1046, opened 2023-07-18
fix: deepsource issues

mariajdab:
### Issues: 

- Functions prefixed with Get should return a value RVV-A0006

Example
```
func (vi *VochainInfo) getMetrics() {

	VochainHeight.Set(float64(vi.Height()))

	VochainMempool.Set(float64(vi.MempoolSize()))

	p, v, vxm := vi.TreeSizes()
```

- Method modifies receiver RVV-B0006

Example:
```
func (e APIerror) With(s string) APIerror {

	e.Err = fmt.Errorf("%w: %v", e.Err, s)

	return e

}
```

- Append possibly assigns to a wrong variable CRT-D0001 (when is possible keep the pattern to append to the same slice)

- Exported function returning value of unexported type RVV-B0011

Example:
```
func (v *State) ListSnapshots() []diskSnapshotInfo {
...
}
```

- Unused parameter in function RVV-B0012 

Note: many recommendation for this issue was omitted in cases like:  
`func (a *API) tokenTransfersHandler(msg *apirest.APIdata, ctx *httprouter.HTTPContext) error {} `  msg is not used

And in cases like:
`func (idx *Indexer) OnProcessStatusChange(pid []byte, status models.ProcessStatus, txIndex int32) {}`

```
func (l *Listener) OnCancel(pid []byte, txIndex int32)                                           {}

func (l *Listener) OnProcessKeys(pid []byte, encryptionPub string, txIndex int32)                {}

func (l *Listener) OnRevealKeys(pid []byte, encryptionPriv string, txIndex int32)                {}

func (l *Listener) OnProcessResults(pid []byte, results *models.ProcessResult, txIndex int32) {

}
```

### DIFF
--- a/data/ipfs/metrics.go
+++ b/data/ipfs/metrics.go
@@ -38,8 +38,8 @@ func (i *Handler) registerMetrics(ma *metrics.Agent) {
 	ma.Register(FilePins)
 }
 
-// GetMetrics to be called as a loop and grab metrics
-func (i *Handler) getMetrics(ctx context.Context) error {
+// setMetrics to be called as a loop and grab metrics
+func (i *Handler) setMetrics(ctx context.Context) error {
 	peers, err := i.CoreAPI.Swarm().Peers(ctx)
 	if err != nil {
 		return err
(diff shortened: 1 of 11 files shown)


p4u:
LGTM

Please @mvdan take a look

mvdan (approved):

altergui:
> Note: many recommendation for this issue was omitted in cases like:
> `func (a *API) tokenTransfersHandler(msg *apirest.APIdata, ctx *httprouter.HTTPContext) error {} ` msg is not used

just as a side comment, i believe in that case you can use `_` to make it clear you are ignoring that parameter, as in:

`func (a *API) tokenTransfersHandler(_ *apirest.APIdata, ctx *httprouter.HTTPContext) error {}`

anyway the PR is merged, congratulations and thanks for all the work!!

PR #1053, opened 2023-07-21
indexer: set endblock to the actual end block for an election

p4u:
If the election is ended manually, the indexer stores as endblock the current height.

Furthermore, add a new boolean value to let the API caller know that the election was ended manually.

### DIFF
--- a/api/api_types.go
+++ b/api/api_types.go
@@ -33,6 +33,7 @@ type ElectionSummary struct {
 	VoteCount      uint64            `json:"voteCount"`
 	FinalResults   bool              `json:"finalResults"`
 	Results        [][]*types.BigInt `json:"result,omitempty"`
+	ManuallyEnded  bool              `json:"manuallyEnded"`
 }
 
 // ElectionResults is the struct used to wrap the results of an election
--- a/vochain/indexer/process.go
+++ b/vochain/indexer/process.go
@@ -191,6 +191,11 @@ func (idx *Indexer) newEmptyProcess(pid []byte) error {
 	// Get the block time from the Header
 	currentBlockTime := time.Unix(idx.App.TimestampStartBlock(), 0)
 
+	// If startBlock is zero, then we use the current block height (zero means start right now)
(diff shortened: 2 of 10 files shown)


p4u:
ping @mvdan

api/api_types.go
```diff
@@ -25,14 +25,15 @@ type OrganizationList struct {
...
-	Results        [][]*types.BigInt `json:"result,omitempty"`
+	ElectionID      types.HexBytes    `json:"electionId" `
+	OrganizationID  types.HexBytes    `json:"organizationId" `
+	Status          string            `json:"status"`
+	StartDate       time.Time         `json:"startDate"`
+	EndDate         time.Time         `json:"endDate"`
+	VoteCount       uint64            `json:"voteCount"`
+	FinalResults    bool              `json:"finalResults"`
+	Results         [][]*types.BigInt `json:"result,omitempty"`
+	ManuallyStopped bool              `json:"manuallyStopped"`
```
mvdan: ManuallyEnded, to be consistent with EndDate and EndBlock?

vochain/indexer/process.go
```diff
@@ -202,6 +207,7 @@ func (idx *Indexer) newEmptyProcess(pid []byte) error {
 		EntityID:          nonNullBytes(eid),
 		StartBlock:        int64(p.StartBlock),
 		EndBlock:          int64(p.BlockCount + p.StartBlock),
```
mvdan, line 209: should we still insert EndBlock when creating a new process? it's StartBlock+BlockCount, so until the process is actually ended (either automatically or manually), it's not really useful at all.

perhaps it's best to always have it there for consistency, but wanted to flag this anyway.

mvdan (approved):

vochain/indexer/process.go
p4u: yes, for consistency, it makes sense to set it up when creating the process. Thanks for flagging.

PR #1054, opened 2023-07-22
api: consolidate how we build ElectionSummaries

mvdan:
(see commit message)

### DIFF
--- a/api/accounts.go
+++ b/api/accounts.go
@@ -316,9 +316,14 @@ func (a *API) electionListHandler(_ *apirest.APIdata, ctx *httprouter.HTTPContex
 		return ErrParamStatusMissing
 	}
 
-	elections, err := a.electionSummaryList(pids...)
-	if err != nil {
-		return err
+	elections := []*ElectionSummary{}
+	for _, pid := range pids {
+		procInfo, err := a.indexer.ProcessInfo(pid)
+		if err != nil {
+			return ErrCantFetchElection.WithErr(err)
+		}
+		summary := a.electionSummary(procInfo)
+		elections = append(elections, &summary)
 	}
 	data, err := json.Marshal(&Organization{
 		Elections: elections,
--- a/api/helpers.go
+++ b/api/helpers.go
@@ -4,10 +4,9 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"errors"
-	"fmt" // required for evm encoding
+	"fmt"
 	"math/big"
 	"reflect"
-	"strings"
 
 	cmtpool "github.com/cometbft/cometbft/mempool"
 	ctypes "github.com/cometbft/cometbft/rpc/core/types"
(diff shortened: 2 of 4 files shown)


mvdan:
Please look at the XXX comments; we should make decisions on those two, since the deduplicated bits of code did something different - I think by mistake. That's the sort of thing that happens with copy pasting.

api/accounts.go
```diff
@@ -316,12 +316,17 @@ func (a *API) electionListHandler(msg *apirest.APIdata, ctx *httprouter.HTTPCont
 		return ErrParamStatusMissing
 	}
 
-	elections, err := a.electionSummaryList(pids...)
-	if err != nil {
-		return err
+	processes := []*ElectionSummary{}
```
altergui: any reason against `elections`?

api/helpers.go
```diff
@@ -4,42 +4,34 @@ import (
...
-			StartDate:      procInfo.CreationTime,
-			EndDate:        a.vocinfo.HeightTime(int64(procInfo.EndBlock)),
-			FinalResults:   procInfo.FinalResults,
-			VoteCount:      procInfo.VoteCount,
-		})
+func (a *API) electionSummary(pi *indexertypes.Process) ElectionSummary {
+	return ElectionSummary{
+		ElectionID:     pi.ID,
+		OrganizationID: pi.EntityID,
+		// XXX: some callers used strings.ToLower on Status; what should we consolidate on?
```
p4u: I think we should consolidate it to the name defined in the protobuf file. So it if the enum is using capital letters, we should return capital letters. The consumers of the API use the same protobuf file.

api/helpers.go
```diff
@@ -4,42 +4,34 @@ import (
...
+		Status: models.ProcessStatus_name[pi.Status],
+		// XXX: some callers used pi.CreationTime instead; that seemed wrong?
```
p4u: Yes, using HeightTime() is more accurate.

api/helpers.go
```diff
@@ -4,42 +4,34 @@ import (
...
-			EndDate:        a.vocinfo.HeightTime(int64(procInfo.EndBlock)),
-			FinalResults:   procInfo.FinalResults,
-			VoteCount:      procInfo.VoteCount,
-		})
+func (a *API) electionSummary(pi *indexertypes.Process) ElectionSummary {
+	return ElectionSummary{
+		ElectionID:     pi.ID,
+		OrganizationID: pi.EntityID,
+		// XXX: some callers used strings.ToLower on Status; what should we consolidate on?
+		Status: models.ProcessStatus_name[pi.Status],
```
altergui: i checked this and it's a tough decision. ideally i'd love to have consistency between internal code and what's exposed in the API, but it doesn't seem possible. using lowercase in our code looks like a no-go. 

and uppercase in API would look funny in `/accounts/{organizationID}/elections/status/{status}/page/{page}` or in the reply to that query.

so i vote for strings.ToLower , so that API replies are lowercase and all of our internal code is kept uppercase

our public docs are not uppercase
https://developer.vocdoni.io/protocol/overview#election-lifecycle-states

api/accounts.go
mvdan: I can change it. the other method used processes. they're pretty much synonyms.

api/helpers.go
mvdan: That's what I thought, thanks.

api/helpers.go
mvdan: Gui appears to have started a new thread about this, so to consolidate the two, I'll copy your response there and respond to both there.

api/helpers.go
mvdan: @p4u said:

> I think we should consolidate it to the name defined in the protobuf file. So it if the enum is using capital letters, we should return capital letters. The consumers of the API use the same protobuf file.

I tend to think the same - unless there's a really good reason to avoid uppercase, we should use the same casing everywhere, including protobuf.

Unless the existing clients already assume lowercase in lots of places and changing that could be problematic. Going all consistent on lowercase would also be fine by me as long as everyone is happy on the client side.

A third option is to always return the canonical form (uppercase, like protobuf), but queries could accept either uppercase or lowercase forms. This is probably what breaks the fewest clients right now, but it's also the trickiest to deal with in the long term, because it could lead to subtle bugs.

altergui: SGTM :v:

CAPS FTW

(everywhere, and fix the few places that might be using lowercase ATM)

api/helpers.go
p4u: we should remove this comment before merging

api/helpers.go
mvdan: Just removing the XXX comment then, since we want to return the uppercase protobuf casing directly :)

mvdan:
ready for approval and merging, just coordinate with the client/sdk people so they are not surprised by this.

altergui (approved):
LGTM

altergui:
we added today deepsource to the PR checks and still fine-tuning it.

this warning given in https://app.deepsource.com/gh/vocdoni/vocdoni-node/run/9f1bb5d3-bb97-405a-97b7-4fd8c67d3294/go/GO-W1027?listindex=0

"Empty slice literal used to declare a variable GO-W1027" is based on this https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices and i believe the usage in this particular case is precisely a valid exception according to that same doc:

> Note that there are limited circumstances where a non-nil but zero-length slice is preferred, such as when encoding JSON objects (a nil slice encodes to null, while []string{} encodes to the JSON array []).

am i understanding this right @mvdan?
in that case i think it would be worth it to disable that whole check completely (GO-W1027), since we have quite a few of these JSON encodings

mvdan:
@altergui for the particular case you mention, I'm not saying the suggestion is wrong, but it's also very nitpicky :) I would honestly be against using deepsource with that kind of aggressive rule. I'm also not sure about enforcing changes like the recent ones to forbid unused receivers or parameters, for example. Replacing them with empty or underscore names doesn't really make the code any measurably better, it just adds git noise.

altergui:
thanks for your input as always!!
i disabled that check ("Empty slice literal used to declare a variable GO-W1027") across the whole repo

regarding the unused receivers or parameters, the git noise is already done :see_no_evil: reverting it would only make it worse, AFAIU.
regarding the future, i'm happy to discuss whether to enforce it or not, when a PR fails that check.

PR #1055, opened 2023-07-24
api(swagger): split server url into schemes, host and BasePath

altergui:
that's how it's meant to be parsed by swaggo

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -40,7 +40,10 @@ import (
 //	@tag.name			Wallet
 //	@tag.description	Operations for the wallets managed backend side
 
-//	@BasePath					https://api-dev.vocdoni.net/v2/
+//	@schemes	https
+//	@host		api-dev.vocdoni.net
+//	@BasePath	/v2
+
 //	@securityDefinitions.basic	BasicAuth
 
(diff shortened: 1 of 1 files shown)


altergui:
this was just a test, never intended to be merged, just to produce an automatic PR
https://github.com/vocdoni/developer-portal/pull/104

altergui:
this is the resulting diff in vocdoni-api.yaml 

image

elboletaire (approved):

api/api.go
```diff
@@ -40,7 +40,10 @@ import (
 //	@tag.name			Wallet
 //	@tag.description	Operations for the wallets managed backend side
 
-//	@BasePath					https://api-dev.vocdoni.net/v2/
+//	@schemes	https
+//	@host		api-dev.vocdoni.net
+//	@BasePath	/v2
```
elboletaire, line 45: Should be `basePath` instead of `BasePath`.

elboletaire (changes requested):

api/api.go
elboletaire: Ok now I see it's something (terrible) from swaggo https://github.com/swaggo/swag#declarative-comments-format

elboletaire (approved):

PR #1056, opened 2023-07-25
end2end: refactor generateProof

mariajdab:
We want to make `generateProof` more simple, so in this approach is used a sync.map `t.voters`, a map indexed by `key.PublicKey` that contains acctProof struct

```
type acctProof struct {
	account  *ethereum.SignKeys
	proof    *apiclient.CensusProof
	proofSIK *apiclient.CensusProof
}
```

Relevant changes: 
* use sync.map (key:acc.Public, value: acctProof struct)
* remove voterAccounts, proofs, sikproofs from e2eElection
* remove generateSIKProofs to use the general approach generateProofs
* add publishCheckCensus to separate the feature of publish and check census
* setupElection now has an extra parameter to specify the number of accounts wanted for the census. It allows to remove code and simplify the logic for censusize end2endtest

### DIFF
--- a/cmd/end2endtest/dynamicensus.go
+++ b/cmd/end2endtest/dynamicensus.go
@@ -46,23 +46,8 @@ func (t *E2EDynamicensusElection) Setup(api *apiclient.HTTPclient, c *config) er
 		ed.d.ElectionType.DynamicCensus = ed.dynamicCensus
 		ed.d.Census = vapi.CensusTypeDescription{Type: vapi.CensusTypeWeighted}
 
-		// create a census with 2 voterAccounts less than the nvotes passed, that will allow to create another
-		// census with the missing nvotes
-		censusRoot, censusURI, err := t.elections[i].setupCensus(vapi.CensusTypeWeighted, t.elections[i].config.nvotes-2, false)
-		if err != nil {
-			return err
-		}
-
-		// add the censusRoot and censusURI to the election description
-		ed.d.Census = vapi.CensusTypeDescription{
-			Type:     vapi.CensusTypeWeighted,
-			RootHash: censusRoot,
-			URL:      censusURI,
-			Size:     uint64(t.elections[1].config.nvotes),
-		}
-
 		// set up the election with the custom census created
-		if err := t.elections[i].setupElection(ed.d); err != nil {
+		if err := t.elections[i].setupElection(ed.d, t.elections[i].config.nvotes-1); err != nil {
 			return err
 		}
 		log.Debugf("election detail: %+v", *t.elections[i].election)
@@ -82,16 +67,19 @@ func (t *E2EDynamicensusElection) Run() error {
 
 	setupNewCensusAndVote := func(election e2eElection) (types.HexBytes, apiclient.VoteData, error) {
 		api := election.api
-		nvotes := election.config.nvotes
 
-		censusRoot2, censusURI2, err := election.setupCensus(vapi.CensusTypeWeighted, 2, false)
+		censusID, vAccts, err := election.setupCensus(vapi.CensusTypeWeighted, 1, false)
+		if err != nil {
+			return nil, apiclient.VoteData{}, err
+		}
+
+		censusRoot2, censusURI2, err := election.publishCheckCensus(censusID, 1)
 		if err != nil {
 			return nil, apiclient.VoteData{}, err
 		}
 		log.Infof("root : %x URI: %s  err: %s", censusRoot2, censusURI2, err)
 
-		voterKey := election.voterAccounts[nvotes-1].Address().Bytes()
-		proof, err := api.CensusGenProof(censusRoot2, voterKey)
+		proof, err := api.CensusGenProof(censusRoot2, vAccts[0].Address().Bytes())
 		if err != nil {
 			return nil, apiclient.VoteData{}, err
 		}
@@ -99,7 +87,7 @@ func (t *E2EDynamicensusElection) Run() error {
 		v := apiclient.VoteData{
 			ElectionID:   election.election.ElectionID,
 			ProofMkTree:  proof,
-			VoterAccount: election.voterAccounts[nvotes-1],
+			VoterAccount: vAccts[0],
 			Choices:      []int{1},
 		}
 		if _, err := api.Vote(&v); err != nil {
@@ -123,26 +111,28 @@ func (t *E2EDynamicensusElection) Run() error {
 
 		// Send the votes (parallelized)
 		startTime := time.Now()
-
-		log.Infof("enqueuing %d votes", len(t.elections[0].voterAccounts[1:]))
 		votes := []*apiclient.VoteData{}
-		for _, acct := range t.elections[0].voterAccounts[1:] {
-			votes = append(votes, &apiclient.VoteData{
-				ElectionID:   electionID,
-				ProofMkTree:  t.elections[0].proofs[acct.Address().Hex()],
-				Choices:      []int{0},
-				VoterAccount: acct,
-			})
-		}
-		errs := t.elections[0].sendVotes(votes)
+
+		t.elections[0].voters.Range(func(key, value any) bool {
+			if acctp, ok := value.(acctProof); ok {
+				votes = append(votes, &apiclient.VoteData{
+					ElectionID:   t.elections[0].election.ElectionID,
+					ProofMkTree:  acctp.proof,
+					Choices:      []int{0},
+					VoterAccount: acctp.account,
+				})
+			}
+			return true
+		})
+		errs := t.elections[0].sendVotes(votes[1:])
 		if len(errs) > 0 {
 			errCh <- fmt.Errorf("error from electionID: %s, %+v", electionID, errs)
 			return
 		}
 
 		log.Infow("votes submitted successfully",
-			"n", len(t.elections[0].voterAccounts[1:]), "time", time.Since(startTime),
-			"vps", int(float64(len(t.elections[0].voterAccounts[1:]))/time.Since(startTime).Seconds()))
+			"n", len(votes[1:]), "time", time.Since(startTime),
+			"vps", int(float64(len(votes[1:]))/time.Since(startTime).Seconds()))
 
 		var (
 			censusRoot2 types.HexBytes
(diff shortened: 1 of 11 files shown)


mariajdab:
@altergui, @mvdan  could you please take a look when you have chance? :raised_hands: Is a proposal, suggestions are welcome

cmd/end2endtest/helpers.go
```diff
@@ -265,66 +228,72 @@ func (t *e2eElection) setupAccount() error {
...
 
-	log.Infof("census published with root %x", censusRoot)
+	log.Infof("census published with root %x", cd.RootHash)
+
+	//t.proofs = t.generateProofs(ed.Census.RootHash, ed.ElectionType.Anonymous, nil)
 
 	// Check census size (of the published census)
 	if !t.censusSizeEquals(censusID, uint64(nAcct)) {
-		return nil, "", fmt.Errorf("failed census size invalid, root %x", censusRoot)
+		return vapi.CensusTypeDescription{}, nil, fmt.Errorf("failed census size invalid, root %x", cd.RootHash)
```
altergui: it sounds like there's a mistake in "failed census size invalid"
maybe it's just "census size invalid"?

cmd/end2endtest/helpers.go
```diff
@@ -265,66 +228,72 @@ func (t *e2eElection) setupAccount() error {
...
 	}
 
-	return censusRoot, censusURI, nil
+	if cd.Size == 0 {
+		cd.Size = func() uint64 {
+			if t.config.nvotes == 0 {
+				return 1 // allows to test with no voters
+			}
+			return uint64(t.config.nvotes)
+		}()
```
altergui: i see you using anonymous funcs in many places,
why not a simple `if else`?
```go
	if cd.Size == 0 {
		if t.config.nvotes == 0 {
			cd.Size = 1 // allows to test with no voters
		} else {
		    cd.Size = uint64(t.config.nvotes)
		}
	}
```

cmd/end2endtest/helpers.go
```diff
@@ -165,84 +161,51 @@ func (t *e2eElection) waitUntilElectionStarts(electionID types.HexBytes) (*vapi.
...
-				if isAnonymousVoting {
-					apiClientMtx.Unlock()
+			pr, err := func() (*apiclient.CensusProof, error) {
+				if csp != nil {
+					return cspGenProof(t.election.ElectionID, voterKey, csp)
 				}
-			}
-
+				return t.api.CensusGenProof(t.election.Census.CensusRoot, voterKey)
+			}()
```
altergui: same here, you can replace this anon func with a simpler `if else`

mariajdab: Yeah, I was using `if else` but is more redundant code, I think. But I can reverse the change

cmd/end2endtest/helpers.go
mariajdab: This was added by @p4u at some point. I just move to use it inside `setupCensus`

cmd/end2endtest/helpers.go
mariajdab: Yeah, I will update it.

cmd/end2endtest/dynamicensus.go
```diff
@@ -28,6 +28,7 @@ type E2EDynamicensusElection struct {
 	elections []e2eElection
 }
 
+// TODO: adapt to the new approach
```
mvdan: is this TODO for a future PR, or this one?

cmd/end2endtest/main.go
```diff
@@ -35,10 +36,11 @@ type VochainTest interface {
...
 
 type e2eElection struct {
-	api    *apiclient.HTTPclient
-	config *config
-
-	election      *vapi.Election
+	api      *apiclient.HTTPclient
+	config   *config
+	election *vapi.Election
+	voters   sync.Map
```
mvdan: see the `go vet` warnings; you need to pass around `e2eElection` as a pointer now, otherwise you're copying the `sync.Map` with a mutex, which is probably wrong.

cmd/end2endtest/plaintext.go
```diff
@@ -54,16 +54,24 @@ func (t *E2EPlaintextElection) Run() error {
 	// Send the votes (parallelized)
 	startTime := time.Now()
 
-	log.Infof("enqueuing %d votes", len(t.voterAccounts))
+	//log.Infof("enqueuing %d votes", len(t.voterAccounts))
```
mvdan: I'd personally delete a log line if it's no longer useful. It's always in the git history.

cmd/end2endtest/helpers.go
```diff
@@ -265,66 +228,72 @@ func (t *e2eElection) setupAccount() error {
...
+		cd.Size = func() uint64 {
+			if t.config.nvotes == 0 {
+				return 1 // allows to test with no voters
+			}
+			return uint64(t.config.nvotes)
+		}()
+	}
+	cd.Type = censusType
+
+	//t.generateProofs(nil, voterAccounts)
```
mvdan: is this meant to be removed?

cmd/end2endtest/helpers.go
```diff
@@ -165,84 +161,51 @@ func (t *e2eElection) waitUntilElectionStarts(electionID types.HexBytes) (*vapi.
...
 			}
 
-			if !isAnonymousVoting {
-				pr.KeyType = models.ProofArbo_ADDRESS
-			}
-			proofCh <- &voterProof{
-				proof:   pr,
-				address: acc.Address().Hex(),
-			}
+			t.voters.Store(acc.Public, acctProof{account: acc, proof: pr})
```
mvdan: I'm starting to think - why do you need a `sync.Map` at all? presumably you always set up the election before running it, so the voters map is only written to first, and then later only read from. A regular Go map is fine for that. You only need a `sync.Map` if the writes are concurrent with other reads or writes, but if that were to happen, the test would likely be wrong - you would be starting the election before the setup has finished, for example.

cmd/end2endtest/dynamicensus.go
mariajdab: For this one

cmd/end2endtest/main.go
mariajdab: Yeah, this is included in the TODO

cmd/end2endtest/helpers.go
mariajdab: Thanks @mvdan

The idea of use `sync.Map` comes from a previous suggestion in a [PR comment](https://github.com/vocdoni/vocdoni-node/pull/997#discussion_r1234184159) for the vote proof refactor. I was not sure about using it due the documentation mention 2 cases: 
```
The Map type is optimized for two common use cases: (1) when the entry for a given
key is only ever written once but read many times, as in caches that only grow,
 or (2) when multiple goroutines read, write, and overwrite entries for disjoint
 sets of keys.
```
But, I wanted to confirm if using `sync.Map` could be a option or not, so I open a PR as proposal.

altergui: the whole anon func is called concurrently (as long as pcount > 1, default 4) 

`go addNaccounts(voterAccts[i:end], &wg)`

altergui:
is this ready for re-review?

mariajdab:
> is this ready for re-review?

ready and updated for anonymous proof too :raised_hands:

p4u:
@CodiumAI-Agent /review

mariajdab:
@CodiumAI-Agent /review

mariajdab:
Solved all relevant comments

cmd/end2endtest/helpers.go
```diff
@@ -183,123 +181,65 @@ func (t *e2eElection) waitUntilElectionStarts(electionID types.HexBytes) (*vapi.
...
-			}
-		}
-	}()
-
+func (t *e2eElection) generateProofs(csp *ethereum.SignKeys, voterAccts []*ethereum.SignKeys) error {
+	var (
+		wg     sync.WaitGroup
+		vcount int32
+	)
+	errorChan := make(chan error, 1)
```
altergui: this buffer with size 1 will deadlock as soon as 1 error is pushed into it
you want to `make(chan error, len(voterAccts)` instead

cmd/end2endtest/helpers.go
```diff
@@ -183,123 +181,65 @@ func (t *e2eElection) waitUntilElectionStarts(electionID types.HexBytes) (*vapi.
...
+	t.voters = new(sync.Map)
 	addNaccounts := func(accounts []*ethereum.SignKeys, wg *sync.WaitGroup) {
```
altergui: no need to pass `wg` as an argument anymore, since you moved the `var wg     sync.WaitGroup` to the top

i.e. just do
```
addNaccounts := func(accounts []*ethereum.SignKeys) {
```

cmd/end2endtest/helpers.go
```diff
@@ -183,123 +181,65 @@ func (t *e2eElection) waitUntilElectionStarts(electionID types.HexBytes) (*vapi.
...
+		go addNaccounts(voterAccts[i:end], &wg)
 	}
 
-	wg.Wait()
-	log.Debugf("%d/%d census proofs generated successfully", len(proofs), len(t.voterAccounts))
-	stopProofs <- true
+	go func() { // avoid blocking the main goroutine
+		wg.Wait()
+		close(errorChan) // close the error channel after all goroutines have finished
+	}()
```
altergui, line 233: wait... what?

oh that's an interesting idea. the `for err := range errorChan` will stay blocked until the close(errorChan). but then you can ignore my previous comment about the buffered chan (you don't even need the `1`, it can be simply unbuffered)

i like it. but i would describe the mechanism in a comment (and make the errorChan unbuffered, otherwise it looks misleading)

altergui (approved):
i loved this PR as soon as i started reading it, loved the elegant `e.voters.Range(func())`, and as i continued it got better and better, all in all simply excellent work, congrats maria!!

cmd/end2endtest/helpers.go
altergui: or just `make(chan error)`, now that i saw the whole approach

cmd/end2endtest/helpers.go
altergui: like you did in https://github.com/vocdoni/vocdoni-node/pull/1103/files#diff-e41fced2d6ee22c899abb1543729ef493417c53d17c00c7a644be97e6095b3faR355 , that simple comment is enough

cmd/end2endtest/helpers.go
mariajdab: I think is not necessary for this approach, in the code when an error is encountered, it is returned. Also, I intentionally made `generateProofs` fail to test the behavior with an error.

cmd/end2endtest/helpers.go
mariajdab: Sounds good! :raised_hands:

cmd/end2endtest/helpers.go
mariajdab: Fixed, thanks!

PR #1065, opened 2023-07-28
unify noState access behavior and refactor

p4u:
Instead of using different NoState databases, one per each state tree, use a single one at the database root level.

Create a specific type state.NoState that can used for interacting with the nostate database. This new type is safe wit the state.Tx transactions.

Extend the on-going state-snapshot feature, now including the noState database.

Finally, simplify the ISTC, do the results computation sync. The last benchmarks show that the computation is fast enough to be part of the block processing.

As bonus, remove some Rolling Census related code, since its going to be deprecated soon in favor of the new SIK merkle-tree approach.


vochain/ist/results.go
```diff
@@ -1,113 +1,24 @@
...
-	if err != nil {
-		return fmt.Errorf("cannot encode results: %w", err)
-	}
-	if err := c.storeToNoState(dbResultsIndex(electionID), rEncoded); err != nil {
-		return fmt.Errorf("cannot store results: %w", err)
-	}
-	return nil
-}
-
 // commitResults commits the results to the state. If the `r` parameter is nil, it will
```
lucasmenendez: Please update the comment for this function. It still tries to get the results from the no-state?

lucasmenendez: If does not, probably the function must check that the `r` parameter is not `nil`

vochain/offchaindatahandler/offchaindatahandler.go
```diff
@@ -140,24 +135,8 @@ func (d *OffChainDataHandler) OnCensusUpdate(pid, censusRoot []byte, censusURI s
 
 }
 
-// OnProcessesStart is triggered when a process starts. It checks if the process contains a rolling census.
+// OnProcessesStart is triggered when a process starts. Does nothing.
 func (d *OffChainDataHandler) OnProcessesStart(pids [][]byte) {
```
lucasmenendez: It could be like this:
```go
func (*OffChainDataHandler) OnProcessesStart(_ [][]byte) {}
```

vochain/state/state.go
```diff
@@ -73,15 +71,15 @@ type State struct {
...
+	store          *statedb.StateDB
 	eventListeners []EventListener
 	// Tx must always be accessed via mutex because there will be
 	// concurrent operations on it.  In particular, while the Tx is being
 	// written serially via Tendermint DeliverTx (which updates the StateDB
 	// by processing Vochain transactions), parallel calls of Tendermint
 	// CheckTx happen (which read the temporary state kept in the Tx to
 	// validate Vochain transactions).
-	Tx                treeTxWithMutex
+	tx                treeTxWithMutex
```
lucasmenendez, line 82: Why is `tx` now private and how do we get access to the subtrees?

lucasmenendez (changes requested):

vochain/state/state.go
p4u: so we can be sure no one is using the internal methods of statedb nor tree packages. State should have the API to cover all needs of the vochain application.

PR #1066, opened 2023-07-29
end2end: update sendVotes to avoid infinite for loop

mariajdab:
Currently in the approach for send a batch of votes `sendVotes` is possible to experiment a infinite for loop, due in some cases the error is handled with a log using `log.Warn(err)` or it is just ignored and continue, in order to prevent the infinite loop a couple of variables was added to count each of those errors and if the count reach the max count allowed, that error is added to the slice of error returned by the method. 

Also, currently the errs returned by the `sendVotes` is not been checked in each test, it was also included in this change. 

And in the other hand `overwriteVote` was simplified to not use `sendVotes` as intermediary method, instead just use `api.Vote` and check the error once.

### DIFF
--- a/cmd/end2endtest/helpers.go
+++ b/cmd/end2endtest/helpers.go
@@ -26,6 +26,7 @@ const (
 	nextBlock     = "nextBlock"
 	sameBlock     = "sameBlock"
 	defaultWeight = 10
+	retriesSend   = retries / 2
 )
 
 type ballotData struct {
@@ -474,11 +475,10 @@ func (t *e2eElection) overwriteVote(choices []int, indexAcct int, waitType strin
 			Choices:      []int{choices[i]}}})
 		for _, err := range errs {
 			// check the error expected for overwrite with waitUntilNextBlock
-			if strings.Contains(err.Error(), "overwrite count reached") {
-				log.Debug("error expected: ", err.Error())
-			} else {
+			if !strings.Contains(err.Error(), "overwrite count reached") {
 				return fmt.Errorf("unexpected overwrite error: %w", err)
 			}
+			log.Debug("error expected: ", err.Error())
 		}
 		switch waitType {
 		case sameBlock:
@@ -561,11 +561,16 @@ func ballotVotes(b ballotData, nvotes int) ([][]int, [][]*types.BigInt) {
 
 // sendVotes sends a batch of votes concurrently
 // (number of goroutines defined in t.config.parallelCount)
-func (t *e2eElection) sendVotes(votes []*apiclient.VoteData) (errs map[int]error) {
-	errs = make(map[int]error)
-	var timeouts atomic.Uint32
+// sendVotes sends a batch of votes concurrently
+// (number of goroutines defined in t.config.parallelCount)
+func (t *e2eElection) sendVotes(votes []*apiclient.VoteData) map[int]error {
+	var errs = make(map[int]error)
+	// used to avoid infinite for loop
+	var timeoutsRetry, mempoolRetry, warnRetry atomic.Uint32
 	var wg sync.WaitGroup
 	var queues []map[int]*apiclient.VoteData
+	var mutex sync.Mutex
+
 	for p := 0; p < t.config.parallelCount; p++ {
 		queues = append(queues, make(map[int]*apiclient.VoteData, len(votes)))
 	}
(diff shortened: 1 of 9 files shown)


p4u:
@altergui please

cmd/end2endtest/helpers.go
```diff
@@ -508,11 +507,13 @@ func ballotVotes(b ballotData, nvotes int) ([][]int, [][]*types.BigInt) {
...
-func (t *e2eElection) sendVotes(votes []*apiclient.VoteData) (errs map[int]error) {
-	errs = make(map[int]error)
-	var timeouts atomic.Uint32
+func (t *e2eElection) sendVotes(votes []*apiclient.VoteData) []error {
+	// used to avoid infinite for loop
+	var timeoutsRetry, waitRetry, warnRetry atomic.Uint32
 	var wg sync.WaitGroup
 	var queues []map[int]*apiclient.VoteData
+
+	errs := []error{}
```
altergui: slice errs is not concurrent safe (and you're reading/writing it in concurrent goroutines)

cmd/end2endtest/helpers.go
```diff
@@ -536,28 +537,43 @@ func (t *e2eElection) sendVotes(votes []*apiclient.VoteData) (errs map[int]error
 						delete(queue, i)
 					case errors.Is(err, context.DeadlineExceeded) || os.IsTimeout(err):
 						// if the context deadline is reached, no need to print it, just retry
-						timeouts.Add(1)
+						if timeoutsRetry.Load() > 5 {
+							errs = append(errs, err)
+							wg.Done()
```
altergui: why this `wg.Done()`? there's already a `defer wg.Done()`.
i guess you wanted to `return` instead?

cmd/end2endtest/helpers.go
```diff
@@ -508,11 +507,13 @@ func ballotVotes(b ballotData, nvotes int) ([][]int, [][]*types.BigInt) {
 
 // sendVotes sends a batch of votes concurrently
 // (number of goroutines defined in t.config.parallelCount)
-func (t *e2eElection) sendVotes(votes []*apiclient.VoteData) (errs map[int]error) {
-	errs = make(map[int]error)
```
altergui, line 565: the point of errs being a map was for the caller to be able to see which vote produced which error.

why are you discarding that feature?

cmd/end2endtest/helpers.go
altergui: i must say that my code used a map, that is also not concurrent safe. so the current code is buggy as well, sorry about that.

cmd/end2endtest/helpers.go
```diff
@@ -536,28 +537,43 @@ func (t *e2eElection) sendVotes(votes []*apiclient.VoteData) (errs map[int]error
 						delete(queue, i)
 					case errors.Is(err, context.DeadlineExceeded) || os.IsTimeout(err):
 						// if the context deadline is reached, no need to print it, just retry
-						timeouts.Add(1)
+						if timeoutsRetry.Load() > 5 {
```
altergui: make 5 a `const`

cmd/end2endtest/helpers.go
```diff
@@ -536,28 +537,43 @@ func (t *e2eElection) sendVotes(votes []*apiclient.VoteData) (errs map[int]error
...
+							wg.Done()
+						}
+						timeoutsRetry.Add(1)
 					case strings.Contains(err.Error(), "mempool is full"):
 						log.Warn(err)
 						// wait and retry
-						_ = t.api.WaitUntilNextBlock()
-					case strings.Contains(err.Error(), "already exists") ||
-						strings.Contains(err.Error(), "overwrite count reached"):
+						waitErr := t.api.WaitUntilNextBlock()
```
altergui, line 607: WaitUntilNextBlock will only fail (AFAIU) if the network is not producing blocks, a rare edge case. if the mempool is persisntently full, your code will still be stuck in an infinite loop. maybe you want to skip that `if waitErr != nil {` make waitErr count how many "mempool is full" errors happened? (possibly rename it to mempoolRetry then?)

cmd/end2endtest/helpers.go
```diff
@@ -536,28 +537,43 @@ func (t *e2eElection) sendVotes(votes []*apiclient.VoteData) (errs map[int]error
...
+						warnRetry.Add(1)
 					}
 				}
 			}
 		}(queues[p])
 	}
 	wg.Wait()
 	log.Infow("sent votes",
-		"n", len(votes), "timeouts", timeouts.Load(), "failed", len(errs))
+		"n", len(votes), "timeouts", timeoutsRetry.Load(), "failed", len(errs))
```
altergui: errs is not anymore a count of "failures" (in the sense of votes that failed), and more of "errors". it's perfectly possible to, for example, try to send 100 votes, have some errors in the process (a timeout or mempool full), retry them, and have them all ultimately succeed (so none of them "failed").

i'd update the string to "errors"

cmd/end2endtest/helpers.go
```diff
@@ -414,18 +414,17 @@ func (t *e2eElection) checkElectionPrice(ed *vapi.ElectionDescription) error {
 func (t *e2eElection) overwriteVote(choices []int, indexAcct int, waitType string) error {
 	for i := 0; i < len(choices); i++ {
 		// assign the choices wanted for each overwrite vote
-		errs := t.sendVotes([]*apiclient.VoteData{{
+		if _, err := t.api.Vote(&apiclient.VoteData{
```
altergui: do note that t.api.Vote doesn't handle the "mempool is full" situation, so the overwritetest will fail in that case.

cmd/end2endtest/helpers.go
mariajdab: Yeah, in fact the code below I am counting how many "mempool is full" errors happened

`
						if waitErr != nil {
							if waitRetry.Load() > 5 {
								errs = append(errs, err)
								wg.Done()
							}
							waitRetry.Add(1)
}`

cmd/end2endtest/helpers.go
mariajdab: Yeah, agree. I think it's worthy.

cmd/end2endtest/helpers.go
mariajdab: I didn't see the usefulness before, due is possible that the votes will not keep the order, to look for it when we are debugging. But, maybe I am not seeing the approach correctly.

cmd/end2endtest/helpers.go
mariajdab: Yeah, thanks

cmd/end2endtest/helpers.go
mariajdab: updated

cmd/end2endtest/helpers.go
mariajdab: updated to `mempoolRetry`

mariajdab:
Ready for re-review @altergui

altergui (approved):
LGTM

PR #1067, opened 2023-07-29
vochain/indexer: drop all custom encodings in sqlite in favor of json

mvdan:


mvdan:
For values that we want to use in queries often, regular columns will still be better and faster than JSON objects. However, for these we don't need to query on them directly, so it's fine.

For the native JSON support, see https://www.sqlite.org/json1.html. Note that we don't use any of this yet, but we can whenever we want to.

mvdan:
Deepsource is wrong. The code compiles fine. Why is this in CI?

p4u:
> Deepsource is wrong. The code compiles fine. Why is this in CI?

we are testing it

p4u (approved):

PR #1072, opened 2023-08-04
end2end: extent ballot test

mariajdab:
Update ballot test to allow test different voting types: `ranked, quadratic, range and approval`
update helpers for ballotVotes and add newTestProcess to use as standard `models.Process` struct on tests and avoid repeat code. 

Also was added `newE2EElections` to allow to setup elections with different `accountPrivKeys`, in order to avoid nonce update when manage more than one election

Note: I think the code associated with Run each voting case could be refactor to not repeat code, but I think we could open another PR to work that improvement.

- #863

### DIFF
--- a/cmd/end2endtest/ballot.go
+++ b/cmd/end2endtest/ballot.go
@@ -2,112 +2,367 @@ package main
 
 import (
 	"fmt"
+	"math"
 	"os"
 	"time"
 
+	vapi "go.vocdoni.io/dvote/api"
 	"go.vocdoni.io/dvote/apiclient"
 	"go.vocdoni.io/dvote/log"
+	"go.vocdoni.io/dvote/types"
 	"go.vocdoni.io/proto/build/go/models"
 )
 
+const (
+	rankedVote    = "ranked"
+	quadraticVote = "quadratic"
+	rangeVote     = "range"
+	approvalVote  = "approval"
+)
+
 func init() {
-	ops["ballotelection"] = operation{
-		test:        &E2EBallotElection{},
-		description: "ballot election with unique values, maxCount, maxValue, maxTotalCost to test different ballotProtocol configurations",
-		example:     os.Args[0] + " --operation=ballotelection --votes=1000",
+	ops["ballotRanked"] = operation{
+		test:        &E2EBallotRanked{},
+		description: "ballot election to test ranked voting",
+		example:     os.Args[0] + " --operation=ballotRanked --votes=1000",
+	}
+	ops["ballotQuadratic"] = operation{
+		test:        &E2EBallotQuadratic{},
+		description: "ballot election to test quadratic voting",
+		example:     os.Args[0] + " --operation=ballotQuadratic --votes=1000",
+	}
+	ops["ballotRange"] = operation{
+		test:        &E2EBallotRange{},
+		description: "ballot election to test range voting",
+		example:     os.Args[0] + " --operation=ballotRange --votes=1000",
+	}
+	ops["ballotApproval"] = operation{
+		test:        &E2EBallotApproval{},
+		description: "ballot election to test approval voting",
+		example:     os.Args[0] + " --operation=ballotApproval --votes=1000",
 	}
 }
 
-var _ VochainTest = (*E2EBallotElection)(nil)
+var _ VochainTest = (*E2EBallotRanked)(nil)
+var _ VochainTest = (*E2EBallotQuadratic)(nil)
+var _ VochainTest = (*E2EBallotRange)(nil)
+var _ VochainTest = (*E2EBallotApproval)(nil)
 
-type E2EBallotElection struct {
-	e2eElection
-}
+type E2EBallotRanked struct{ e2eElection }
+type E2EBallotQuadratic struct{ e2eElection }
+type E2EBallotRange struct{ e2eElection }
+type E2EBallotApproval struct{ e2eElection }
 
-func (t *E2EBallotElection) Setup(api *apiclient.HTTPclient, c *config) error {
+func (t *E2EBallotRanked) Setup(api *apiclient.HTTPclient, c *config) error {
 	t.api = api
 	t.config = c
 
-	p := &models.Process{
-		StartBlock: 0,
-		BlockCount: 100,
-		Status:     models.ProcessStatus_READY,
-		EnvelopeType: &models.EnvelopeType{
-			EncryptedVotes: false,
-			UniqueValues:   true},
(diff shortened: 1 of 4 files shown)


cmd/end2endtest/ballot.go
```diff
@@ -63,48 +103,186 @@ func (*E2EBallotElection) Teardown() error {
...
-	}
+	nvotes := t.elections[0].config.nvotes
+	var wg sync.WaitGroup
+	errCh := make(chan error, 4)
 
-	choices, expectedResults := ballotVotes(bdata, len(t.voterAccounts))
+	choicesE1, expectedResults := ballotVotes(t.elections[0].election.TallyMode, nvotes, "ranked", true)
+	choicesE2, expectedResults2 := ballotVotes(t.elections[1].election.TallyMode, nvotes, "quadratic", false)
+	choicesE3, expectedResults3 := ballotVotes(t.elections[2].election.TallyMode, nvotes, "range", true)
+	choicesE4, expectedResults4 := ballotVotes(t.elections[3].election.TallyMode, nvotes, "approval", false)
```
altergui: please avoid hardcoded strings, ballotVotes is an unexported func so you control both caller and func, and can replace those strings by some global `const` that are much easier to debug when refactoring

altergui: this is a general suggestion for you to keep in mind when coding.

in this particular case i'll propose (in the discussion with pau) to refactor the whole PR, that would also make these 4 strings disappear, so i wouldn't worry about turning them into `const`s until we can have that discussion

altergui (commented):
when i reached the end of the PR and saw that in newE2EElections you're basically creating new accounts for each process, exactly what the "outer" structure does, i really see no point in making a test (an "op") handle more than a single election.
my vote goes for holding this PR, and revisiting that discussion with @p4u now that we have your experience dealing with this idea and architecture

altergui:
AFAIU, this PR is blocked until we reach consensus on the approach,
but if i remember correctly, you also wanted to propose yet another approach, would you push it here for discussion/review? (or a new PR? however you feel is more appropiate)

mariajdab:
> AFAIU, this PR is blocked until we reach consensus on the approach, but if i remember correctly, you also wanted to propose yet another approach, would you push it here for discussion/review? (or a new PR? however you feel is more appropiate)

I updated the code to remove some repeated code and help whit the readability of the proposal. The approach remain with the same logic

cmd/end2endtest/ballot.go
```diff
@@ -13,48 +15,87 @@ import (
...
+	// update uniqueValues to false
+	p2.EnvelopeType.UniqueValues = false
+	p2.VoteOptions = &models.ProcessVoteOptions{
+		MaxCount:     4,
+		MaxValue:     3,
+		MaxTotalCost: 12,
+		CostExponent: 2,
+	}
+
+	if err := t.elections[1].setupElectionRaw(p2); err != nil {
```
altergui: maybe make p an array so you can reference `p[1]` which would look less confusing

this would also let you iterate and deduplicate, as in
 ```
 // not real code
 for i := 0; i<4; i++ {  t.elections[i].setupElectionRaw(p[i])  }
 ```

altergui: you can also populate the p array with a `for` loop, deduplicating the 4 lines of newTestProcess()

cmd/end2endtest/helpers.go
```diff
@@ -438,71 +455,112 @@ func (t *e2eElection) overwriteVote(choices []int, indexAcct int, waitType strin
...
+	case "approval":
+		v = [][]int{
+			{1, 0, 0, 0}, {2, 1, 0, 1}, {0, 1, 2, 3}, {0, 0, 0, 1},
+			{0, 0, 1, 1}, {2, 2, 1, 0}, {1, 0, 1, 1}, {0, 0, 0, 0},
+			{1, 1, 0, 3}, {1, 1, 1, 1},
+		}
+		resultsFields = append(resultsFields, votesToBigInt(30, 30, 0, 0),
+			votesToBigInt(50, 10, 0, 0), votesToBigInt(30, 30, 0, 0),
+			votesToBigInt(20, 40, 0, 0))
 	}
```
altergui: i have the feeling this block would fit better in ballot.go, 
i.e. make Run() evaluate `b.ballotType` and pass `v` and `resultsFields` to `ballotVotes`

altergui (commented):
i see you're still creating different accounts and basically totally independent elections, so i'm still in favor of doing 4 separate tests (grouped in the same file ballot.go) but that discussion is out of scope
so, i re-reviewed the PR as-is and left some remarks

cmd/end2endtest/helpers.go
mariajdab: uhm, maybe adding two more parameters to `ballotVotes` function that currently has 4 parameters could complicate the things instead of simplify :thinking:

mariajdab:
ready for re-review @altergui

cmd/end2endtest/helpers.go
```diff
@@ -492,71 +509,112 @@ func (t *e2eElection) overwriteVote(choices []int, indexAcct int, waitType strin
 }
 
 // ballotVotes from a default list of 10 vote values that exceed the max value, max total cost and is not unique will
-func ballotVotes(b ballotData, nvotes int) ([][]int, [][]*types.BigInt) {
+func ballotVotes(vop vapi.TallyMode, nvotes int, ballotType string, expectUniqueV bool) ([][]int, [][]*types.BigInt) {
```
altergui: please move this whole ballotVotes func to ballot.go since it's only used by that test

cmd/end2endtest/ballot.go
```diff
@@ -7,104 +7,248 @@ import (
...
+	if err := sendAndValidateVotes(t.e2eElection, choices, expectedResults); err != nil {
+		return err
 	}
 
-	choices, expectedResults := ballotVotes(bdata, len(t.voterAccounts))
+	return nil
+}
 
-	// Send the votes (parallelized)
+func sendAndValidateVotes(e e2eElection, choices [][]int, expectedResults [][]*types.BigInt) error {
```
altergui, line 220: would this func be useful to other tests? i.e. is it generic enough to be reused elsewhere?
if so, please move it to helpers.go (even if not used right now by any other tests than the ones in ballot.go)

altergui (approved):
LGTM, great work

if you can move the funcs i commented that would be great, but if you want to merge it as-is it's also fine for me (helpers.go is already a mess)

PR #1074, opened 2023-08-07
end2endtest/ballot test as separate ops

altergui:
this a proposal (wrote directly in code, to show how it would look like) to be discussed as an alternative to the approach of #1072

(intentionally not finished, copypasted the code from just 1 test, left the other 3 as TODOs)

### DIFF
--- a/cmd/end2endtest/helpers.go
+++ b/cmd/end2endtest/helpers.go
@@ -28,13 +28,6 @@ const (
 	defaultWeight = 10
 )
 
-type ballotData struct {
-	maxValue     uint32
-	maxTotalCost uint32
-	costExponent uint32
-	maxCount     uint32
-}
-
 func newTestElectionDescription() *vapi.ElectionDescription {
 	return &vapi.ElectionDescription{
 		Title:       map[string]string{"default": fmt.Sprintf("Test election %s", util.RandomHex(8))},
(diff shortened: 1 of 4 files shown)


cmd/end2endtest/helpers.go
```diff
@@ -438,71 +455,112 @@ func (t *e2eElection) overwriteVote(choices []int, indexAcct int, waitType strin
...
+	case "ranked":
+		v = [][]int{
+			{0, 5, 0, 2}, {3, 1, 0, 2}, {3, 2, 0, 1}, {1, 0, 3, 2},
+			{2, 3, 1, 1}, {4, 1, 1, 0}, {0, 0, 2, 1}, {0, 2, 1, 3},
+			{1, 1, 1, 4}, {0, 3, 2, 1},
+		}
+		// default results on field 1,2,3,4 for 10 votes
+		resultsFields = append(resultsFields, votesToBigInt(20, 10, 0, 20),
+			votesToBigInt(10, 10, 20, 10), votesToBigInt(20, 10, 10, 10),
+			votesToBigInt(0, 20, 20, 10))
```
altergui, line 474: i'd also refactor this, so that it's part of the Run (or Setup?) of the relevant test

altergui: also, i'd move the whole ballotVotes func to ballot.go, and only leave in helpers.go funcs that are used by more than one test group (such as endElectionAndFetchResults)

but right now, helpers.go is full of funcs that are only used by one specific test. so i'd move each of those funcs to its corresponding test as well (including ballotVotes), in a PR specific to do that

altergui:
closing this proposal, it was incorporated in #1072

PR #1080, opened 2023-08-10
feature/delete_temp_siks_when_election_ends

lucasmenendez:
Now the elections have a flag during creation (`tempSIKs`:boolean) which if set to true, the temporary SIKs (created using `RegisterSIKTx`) must be purged when the election ends.


lucasmenendez (commented):
review done!

p4u:
Finally... everything works as expected :)

@lucasmenendez please make a new review

p4u:
should not be squashed, the commits are correctly ordered and with correct comments

tree/arbo/batch.go
```diff
@@ -0,0 +1,325 @@
...
+		return nil, err
+	}
+	if nLeafs > t.thresholdNLeafs {
+		return t.addBatchInDisk(wTx, keys, values)
+	}
+	return t.addBatchInMemory(wTx, keys, values)
+}
+
+func (t *Tree) addBatchInDisk(wTx db.WriteTx, keys, values [][]byte) ([]Invalid, error) {
+	nCPU := flp2(runtime.NumCPU())
```
lucasmenendez, line 61: Why do you do this?

vochain/state/sik_test.go
```diff
@@ -159,13 +159,14 @@ func TestAssignSIKToElectionAndPurge(t *testing.T) {
...
 	c.Assert(err, qt.IsNil)
+
 	// purge siks and check
 	c.Assert(s.PurgeSIKsByElection(pid), qt.IsNil)
 	c.Assert(err, qt.IsNil)
 	_, err = s.NoState(true).Get(key)
 	c.Assert(err, qt.IsNotNil)
-	_, err = s.SIKFromAddress(testAccount.Address())
-	c.Assert(err, qt.IsNil)
+	sik, err = s.SIKFromAddress(testAccount.Address())
```
lucasmenendez: here `sik` var should be an underscore `_` because it is not used anymore.

tree/arbo/tree.go
```diff
@@ -138,22 +139,27 @@ func NewTreeWithTx(wTx db.WriteTx, cfg Config) (*Tree, error) {
...
+	t := Tree{
+		db:              cfg.Database,
+		maxLevels:       cfg.MaxLevels,
+		thresholdNLeafs: cfg.ThresholdNLeafs,
+		hashFunction:    cfg.HashFunction,
+	}
+
 	t.emptyHash = make([]byte, t.hashFunction.Len()) // empty
+	var err error
+	t.emptyNode, _, err = t.newIntermediate(t.emptyHash, t.emptyHash)
```
lucasmenendez, line 152: Why the empty node is required?

tree/arbo/batch.go
p4u: I did not change anything related the Add Batch.

tree/arbo/tree.go
p4u: The empty node is used to compare in the up() function if both childs are empty. Precomputing and comparing is faster.

p4u:
Is this ready @lucasmenendez? lets merge it!

p4u:
I just created an issue with a couple of optimizations that can be done in the future: https://github.com/vocdoni/vocdoni-node/issues/1086

PR #1085, opened 2023-09-04
e2etest fixes

altergui:
* apiclient: fix default CostExponent in NewElection
* remove unnecessary use of time.Sleep()
* elections can now be created with more than 2 choices 
* overwrite now votes different options, so now it's clear from results
 which vote was taken into account in case something unexpected happens
* avoid a panic in matchResults

### DIFF
--- a/cmd/end2endtest/helpers.go
+++ b/cmd/end2endtest/helpers.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"math/big"
 	"os"
+	"reflect"
 	"strings"
 	"sync"
 	"sync/atomic"
@@ -28,7 +29,18 @@ const (
 	retriesSend   = retries / 2
 )
 
-func newTestElectionDescription() *vapi.ElectionDescription {
+func newTestElectionDescription(numChoices int) *vapi.ElectionDescription {
+	choices := []vapi.ChoiceMetadata{}
+	if numChoices < 2 {
+		numChoices = 2
+	}
+	for i := 0; i < numChoices; i++ {
+		choices = append(choices, vapi.ChoiceMetadata{
+			Title: map[string]string{"default": fmt.Sprintf("Choice number %d", i)},
+			Value: uint32(i),
+		})
+	}
+
 	return &vapi.ElectionDescription{
 		Title:       map[string]string{"default": fmt.Sprintf("Test election %s", util.RandomHex(8))},
 		Description: map[string]string{"default": "Test election description"},
@@ -38,16 +50,7 @@ func newTestElectionDescription() *vapi.ElectionDescription {
 			{
 				Title:       map[string]string{"default": "Test question 1"},
 				Description: map[string]string{"default": "Test question 1 description"},
-				Choices: []vapi.ChoiceMetadata{
-					{
-						Title: map[string]string{"default": "Yes"},
-						Value: 0,
-					},
-					{
-						Title: map[string]string{"default": "No"},
-						Value: 1,
-					},
-				},
+				Choices:     choices,
 			},
 		},
 	}
@@ -520,11 +523,7 @@ func (t *e2eElection) overwriteVote(choices []int, indexAcct int, waitType strin
 			}
 			log.Debug("error expected: ", err.Error())
 		}
-		switch waitType {
-		case sameBlock:
-			time.Sleep(time.Second * 5)
-
-		case nextBlock:
+		if waitType == nextBlock {
 			_ = t.api.WaitUntilNextBlock()
 		}
 	}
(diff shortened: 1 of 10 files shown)


cmd/end2endtest/overwrite.go
```diff
@@ -75,19 +75,17 @@ func (t *E2EOverwriteElection) Run() error {
 
 	// overwrite the previous vote (choice 0) associated with account of index 0, using enough time to do it in the nextBlock
 	// try to make 3 overwrites (number of choices passed to the method). The last overwrite should fail due the maxVoteOverwrite constrain
-	err := t.overwriteVote([]int{0, 1, 0}, 0, nextBlock)
+	err := t.overwriteVote([]int{0, 1, 2}, 0, nextBlock)
```
mariajdab: In `newTestElectionDescription` we have declared only 2 possible choices allowed (`0` and `1`), so I think this value of `2` will not be taken into account

altergui: that's so true! good catch <3

i refactored newTestElectionDescription to solve this

p4u:
fix conflics please @altergui

mariajdab:
There is a particular case when we expect for example a results like: `[[30 30 0 0] [50 10 0 0] [30 30 0 0] [20 40 0 0]]` all the fields only have the first 2 positions with values, so the election results will return a 4x2 matrix `[[30 30] [50 10] [30 30] [20 40]]` instead of a 4x4 matrix

altergui:
> There is a particular case when we expect for example a results like: `[[30 30 0 0] [50 10 0 0] [30 30 0 0] [20 40 0 0]]` all the fields only have the first 2 positions with values, so the election results will return a 4x2 matrix `[[30 30] [50 10] [30 30] [20 40]]` instead of a 4x4 matrix

oh, i see, thanks for the tip. i jumped in and found this way around, WDYT? https://github.com/vocdoni/vocdoni-node/pull/1085/commits/16cae7f754cb03bdedaaa1724569b1a740f78d14

p4u:
Let's approve or request changes, @mariajdab :)

cmd/end2endtest/helpers.go
```diff
@@ -602,15 +601,7 @@ func faucetPackage(faucet, faucetAuthToken, myAddress string) (*models.FaucetPac
...
-	// iterate over each question to check if the results match with the expected results
-	for i, result := range results {
-		for q, r := range result {
-			if !(expectedResults[i][q].String() == r.String()) {
-				return false
-			}
-		}
-	}
-	return true
+	return reflect.DeepEqual(results, expectedResults)
```
mariajdab, line 628: Just to confirm, using reflect is not slower?

mariajdab:
> > There is a particular case when we expect for example a results like: `[[30 30 0 0] [50 10 0 0] [30 30 0 0] [20 40 0 0]]` all the fields only have the first 2 positions with values, so the election results will return a 4x2 matrix `[[30 30] [50 10] [30 30] [20 40]]` instead of a 4x4 matrix
> 
> oh, i see, thanks for the tip. i jumped in and found this way around, WDYT? [16cae7f](https://github.com/vocdoni/vocdoni-node/commit/16cae7f754cb03bdedaaa1724569b1a740f78d14)

mariajdab:
> > There is a particular case when we expect for example a results like: `[[30 30 0 0] [50 10 0 0] [30 30 0 0] [20 40 0 0]]` all the fields only have the first 2 positions with values, so the election results will return a 4x2 matrix `[[30 30] [50 10] [30 30] [20 40]]` instead of a 4x4 matrix
> 
> oh, i see, thanks for the tip. i jumped in and found this way around, WDYT? [16cae7f](https://github.com/vocdoni/vocdoni-node/commit/16cae7f754cb03bdedaaa1724569b1a740f78d14)

Looks good :100:

cmd/end2endtest/helpers.go
altergui: i don't know. but in any case it's executed just once per test :shrug:

mariajdab (approved):

altergui:
looking at the code coverage of this PR i realize how easy it is to write totally untested code :rofl: 

https://coveralls.io/builds/62571779/source?filename=cmd%2Fend2endtest%2Fballot.go#L338

image

altergui:
but (unknowingly) merging untested code is what we've been doing until now, all the time, so i'm just going to merge this anyway for now

mariajdab:
> what we've been doing until now, all the time, so i'm just going to merge this anyway for now

But, I think this function `ballotVotes` is not the case, because we are only building the results to compare with the elections results from the api, that is the reason we don't need use `else`

PR #1088, opened 2023-09-04
bug: maxCensusSize for anonymous envelope cannot be bigger than the number of levels of the circuit

### LINKED ISSUE #1087: bug: maxCensusSize for anonymous envelope cannot be bigger than the number of levels of the circuit
**Describe the bug**
Trying to create an anonymous census3 process, I get an error saying `maxCensusSize for anonymous envelope cannot be bigger than the number of levels of the circuit (160)`, but based on what @lucasmenendez said to me, it should be `2^160` instead, making the check invalid.

**To Reproduce**
- Create an anonymous process defining its max census size to a high number, like 500

**Current behavior**
It returns the aforementioned issue, saying it cannot be bigger than the levels of the circuit

**Expected behavior**
Check should be for `2^levelsOfCircuit`

lucasmenendez:
Trying to fix https://github.com/vocdoni/vocdoni-node/issues/1087

### DIFF
--- a/crypto/zk/circuit/config.go
+++ b/crypto/zk/circuit/config.go
@@ -48,7 +48,7 @@ type ZkCircuitConfig struct {
 
 // KeySize returns the maximum number of bytes of a leaf key according to the
 // number of levels of the current circuit (nBytes = nLevels / 8).
-func (conf ZkCircuitConfig) KeySize() int {
+func (conf *ZkCircuitConfig) KeySize() int {
 	return conf.Levels / 8
 }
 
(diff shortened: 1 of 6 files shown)


crypto/zk/circuit/config.go
```diff
@@ -47,6 +47,18 @@ func (config ZkCircuitConfig) KeySize() int {
 	return config.Levels / 8
 }
 
+// SupportsCensusSize returns if the provided censusSize is supported by the
+// current circuit configuration. It ensures that the provided value is lower
+// than 2^config.Levels.
+func (config ZkCircuitConfig) SupportsCensusSize(maxCensusSize uint64) bool {
+	return true
```
p4u: Wow, I don't really understand the `true` :1st_place_medal: 

I think we should make this (and everything related to the number of levels of the tree), dynamic.

So in the ZkCircuitConfig constructor I'd pre-compute the number and store it `config.circuitMaxCensusSize = new(big.Int).Exp(big.NewInt(2), big.NewInt(int64(config.Levels)), nil)` then SupportsCensusSize just performs a bigint comparison: `return config.circuitMaxCensusSize.Cmp(new(big.Int).SetUint64(maxCensusSize)) != 1`

lucasmenendez: I think it is solved now!

p4u (approved):

crypto/zk/circuit/config.go
```diff
@@ -39,19 +40,42 @@ type ZkCircuitConfig struct {
...
+func (conf *ZkCircuitConfig) KeySize() int {
+	return conf.Levels / 8
+}
+
+// MaxCensusSize returns the maximum number of keys that circuit merkle tree
+// for the census supports. The method checks if it is already precalculated
+// or not. If it is not precalculated, it will calculate and initialise it. In
+// any case, the value is returned as big.Int.
+func (conf *ZkCircuitConfig) MaxCensusSize() *big.Int {
+	circuitMaxCensusSize := new(big.Int).Exp(big.NewInt(2), big.NewInt(int64(conf.Levels)), nil)
```
elboletaire, line 60: I just came here to see if this `Exp` (aka pow) was as expected 👌🏼

PR #1089, opened 2023-09-05
add mongodb implementation for the db package

p4u:
The db interface has now an implementation for mongodb. So instead of pebble a user or consumer of the vocdoni-node API can choose to use mongodb. Its not faster than local KV, but useful when the host runing the database has no persistent storage available, so a remote storage is required.

The mongodb url must be set on a env var, such as:

MONGODB_URL="mongodb://root:vocdoni@localhost:27017/" go run ./cmd/node -t mongodb -c dev


db/mongodb/mongodb.go
```diff
@@ -0,0 +1,221 @@
...
+	"go.mongodb.org/mongo-driver/mongo/options"
+	"go.mongodb.org/mongo-driver/mongo/readpref"
+	"go.vocdoni.io/dvote/db"
+	"go.vocdoni.io/dvote/log"
+)
+
+const (
+	MONGODB_TIMEOUT_CONNECT = 10 * time.Second
+	MONGODB_TIMEOUT_COMMIT  = 10 * time.Second
+	MONGODB_TIMEOUT_QUERY   = 4 * time.Second
```
mvdan: FYI there's no convention in Go to use ALL_CAPS constants, this is from C/Java

db/mongodb/mongodb.go
```diff
@@ -0,0 +1,221 @@
...
+	client, err := mongo.Connect(ctx, clientOptions)
+	if err != nil {
+		return nil, err
+	}
+	database := client.Database(collection)
+
+	// Shutdown database connection when SIGTERM received
+	go func() {
+		c := make(chan os.Signal, 1)
+		signal.Notify(c, os.Interrupt, syscall.SIGTERM)
```
mvdan: this is a bit hacky - a library package like db/mongodb shouldn't listen for a process signal, it should have a Shutdown method. it should only be the main package that does OS process stuff like listening for signals.

db/mongodb/mongodb.go
```diff
@@ -0,0 +1,221 @@
...
+func (d *MongoDB) Close() error {
+	ctx, cancel := context.WithTimeout(context.Background(), 20*time.Second)
+	defer cancel()
+	return d.db.Client().Disconnect(ctx)
+}
+
+func (d *MongoDB) WriteTx() db.WriteTx {
+	return &WriteTx{
+		db:         d.db,
+		batch:      make([]mongo.WriteModel, 0),
```
mvdan: remove this - unlike a map, you don't need to "make" a slice for apends to work.

db/mongodb/mongodb.go
```diff
@@ -0,0 +1,221 @@
...
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), MONGODB_TIMEOUT_COMMIT)
+	defer cancel()
+	collection := tx.db.Collection(tx.collection)
+	_, err := collection.BulkWrite(ctx, tx.batch)
+	return err
+}
+
+func (tx *WriteTx) Discard() {
+	tx.batch = make([]mongo.WriteModel, 0)
```
mvdan: similar to before, this should be `= nil`.

db/mongodb/mongodb.go
```diff
@@ -0,0 +1,221 @@
...
+	ctx, cancel := context.WithTimeout(context.Background(), 20*time.Second)
+	defer cancel()
+	return d.db.Client().Disconnect(ctx)
+}
+
+func (d *MongoDB) WriteTx() db.WriteTx {
+	return &WriteTx{
+		db:         d.db,
+		batch:      make([]mongo.WriteModel, 0),
+		inMem:      make(map[string][]byte),
```
mvdan, line 75: you route all Get calls through WriteTx, which is fine since it's only a wrapper, but then this allocation is always a waste for any get call. I would keep `inMem` nil initially, which is fine because any read operation on a nil map works the same as a non-nil empty map. you only need to check if the map is nil, and make it, before you insert an entry in Set.

db/mongodb/mongodb.go
```diff
@@ -0,0 +1,221 @@
...
+			}
+		}
+	}
+
+	// Now, handle database keys, but skip those already handled from in-memory map
+	collection := tx.db.Collection(tx.collection)
+
+	filter := bson.M{}
+	if len(prefix) > 0 {
+		filter = bson.M{"_id": bson.M{"$regex": primitive.Regex{Pattern: "^" + string(prefix)}}}
```
mvdan: I wonder how slow this might be...

db/mongodb/mongodb.go
```diff
@@ -0,0 +1,221 @@
...
+	return nil
+}
+
+func (tx *WriteTx) Set(k, v []byte) error {
+	update := bson.D{
+		primitive.E{Key: "$set", Value: bson.D{
+			primitive.E{Key: "value", Value: v},
+		}},
+	}
+	model := mongo.NewUpdateOneModel().SetFilter(bson.D{primitive.E{Key: "_id", Value: string(k)}}).SetUpsert(true).SetUpdate(update)
```
mvdan: the point of a WriteTx is to bulk/batch writes; why are you using NewUpdateOneModel instead of NewUpdateManyModel? note that I'm not familiar with how mongo works at all, just wondering outloud.

mvdan (approved):
SGTM, though I'm not familiar with mongo, and I have no idea how performant this might be. If it's over the network, compared to pebble or leveldb on disk, we might see significantly worse performance just due to the round-trip nature of the network.

db/mongodb/mongodb.go
p4u: I hope mongo people made it efficient enough :/ But there is not other way, I have been looking for a better (non string regex bassed) mechanism, but there is nothing we can use.

db/mongodb/mongodb.go
p4u: So the strategy is to accumulate single model updates in the memory batch. 

Then on commit, I use collection.BulkWrite() that takes as parameter a list of models to apply. 

This is the way it usually works for mongo, at least what I found it in the official docs.

Issue #1040, opened 2023-07-12
feature: improve the metrics exported


altergui:
see

https://github.com/vocdoni/vocdoni-node/blob/main/vochain/vochaininfo/metrics.go
https://github.com/vocdoni/vocdoni-node/blob/main/data/ipfs/metrics.go

altergui:
for example, metric regarding the latency on the DHT node discovery
https://github.com/vocdoni/vocdoni-node/pull/1014

altergui:
i can think of a thousand metrics more to add, but closing this for now

PR #1014, opened 2023-06-26
f/optimistic provide

altergui:
fix #989

PR #1090, opened 2023-09-06
metrics: refactor to collect metrics on demand, and add measure DHT latency

altergui:
related to #1040 and #1014


p4u (approved):

altergui:
WIP, don't merge

PR #1094, opened 2023-09-07
F/index transactions cost

p4u:
More info https://github.com/vocdoni/interoperability/issues/84


vochain/indexer/indexer.go
```diff
@@ -587,3 +588,69 @@ func (idx *Indexer) GetTokenTransfersByFromAccount(from []byte, offset, maxItems
...
+func (idx *Indexer) OnSpendTokens(address []byte, txType models.TxType, cost uint64, reference string) {
+	idx.blockMu.Lock()
+	defer idx.blockMu.Unlock()
+	queries := idx.blockTxQueries()
+	if _, err := queries.CreateTokenSpending(context.TODO(), indexerdb.CreateTokenSpendingParams{
+		FromAccount: address,
+		TxType:      strings.ToLower(txType.String()),
+		Cost:        int64(cost),
+		Reference:   reference,
+		SpendTime:   time.Now(),
```
mvdan: FYI: perhaps grab time.Now before you lock the mutex, since it might take a few seconds or potentially longer, skewing the timestamp unnecessarily. we might have the same minor issue elsewhere.

mvdan (approved):

PR #1098, opened 2023-09-11
ci: send code coverage to coveralls.io

altergui:
code coverage was enabled in
https://github.com/vocdoni/vocdoni-node/pull/853
and temporarily disabled in
https://github.com/vocdoni/vocdoni-node/pull/901

now i'm enabling it again in a way that is an independent job that will never affect the result (:heavy_check_mark: or :x: ) of job_compose_test and job_go_checks


chilcano (approved):
Reviewed

PR #1099, opened 2023-09-12
ci: improvements to code coverage handling

altergui:
Most notably, now any failure during coverage jobs flow will trigger a warning BUT NOT mark the CI as failed

 * send code coverage to DeepSource 
 * generate coverage in binary fmt
 * update .gitignore
 * publish textfmt artifacts ready to be applied in VSCode
 * job_go_test: covermode=count
 * job_compose_test: now builds with `-race`
     this makes the integration test only slightly slower (around +10%)
     unlike the abismal effect in unit test (10x)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -81,31 +81,39 @@ jobs:
     env:
       LOG_PANIC_ON_INVALIDCHARS: true # check that log lines contains no invalid chars (evidence of format mismatch)
     steps:
-      - name: Checkout code
-        uses: actions/checkout@v3
-      - name: Set up Go environment
-        uses: actions/setup-go@v4
+      - uses: actions/checkout@v3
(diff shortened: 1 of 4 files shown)


p4u:
conflicts @altergui

mariajdab:
Looks good! :rocket:  
Just one suggest I would add the flag `--no-install-recommends` in `apt-get install ` of the `Dockerfile`. I already used it and the installation time decrease also could help with the space, in some cases my container reached 30Gb. What do you think @altergui ?

mariajdab (approved):

mariajdab:
> Looks good! 🚀 Just one suggest I would add the flag `--no-install-recommends` in `apt-get install ` of the `Dockerfile`. I already used it and the installation time decrease also could help with the space, in some cases my container reached 30Gb. What do you think @altergui ?

Maybe open another PR is more clean, due this is for code coverage handling.  #1145

PR #1103, opened 2023-09-13
end2end: fix errorChan in SetupCensus

mariajdab:
After try intentionally fail the go routine and send an error to `errorChan` the code didn't return the expected error.  
Now if an error is sent to `errorChan` it will not remain waiting infinitely and will return the error.

### DIFF
--- a/cmd/end2endtest/helpers.go
+++ b/cmd/end2endtest/helpers.go
@@ -290,7 +290,7 @@ func (t *e2eElection) setupCensus(censusType string, nAcct int, createAccounts b
 
 	// Register the accounts in the vochain if is required
 	if censusType == vapi.CensusTypeZKWeighted && createAccounts {
-		errorChan := make(chan error, 1) // Create a buffered channel to prevent deadlock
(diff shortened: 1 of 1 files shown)


mariajdab:
@altergui not sure why the coverage decreased

altergui:
coverage decreased because errorChan is empty and then it doesn't even enter the for loop. but that's ok, ignore the notice in this case

altergui (approved):
i'd ask you to change L329 as well,
```
		errorChan := make(chan error, 1) // Create a buffered channel to prevent deadlock
```
since that line was buggy anyway (size 1 only prevented a deadlock if there was only 1 error, but a 2nd error caused a deadlock anyway) 
now you're preventing the deadlock in a different way, you can simply remove the comment and make it unbuffered
```
errorChan := make(chan error)
```

altergui:
and as i said in https://github.com/vocdoni/vocdoni-node/pull/1056#pullrequestreview-1635451964 if you could, please add a comment describing that the `for range` will stay blocked until wg.Done and close() happen

altergui:
perfect, thanks! i'll let you merge it yourself

PR #1107, opened 2023-09-19
f/ci random fixes

altergui:
- testsuite: start_test.sh respect passed BUILDARGS
- ci: job_compose_test now builds with `-race`
- e2etest: fix trivial race
- e2etest: refactor generateProofs, use buffered chan


altergui:
absolutely unnecessary, completely superseded by #1056 , sorry for the noise

Issue #1083, opened 2023-08-30
feature: modify anonymous end-to-end test


lucasmenendez:
The goal of this issue is to update the current anonymous end-to-end test to include the new uses cases.

Until now, only users with account in the vochain (and a valid SIK) can cast anonymous votes. Now, a non registered user can cast a vote registering its SIK previously using the RegisterSIKTx proving that he or she is already included into the election census (census proof).

The new test version should include this use case, using 5 registered accounts to vote but also 5 new identities that are not registered in the vochain.

This issue requires the merge of #1080. In this PR, the e2e helpers had been modified to prevent the account creation if it is not desired. Also includes some new APIClient methods to register a temporal SIK.

PR #1117, opened 2023-09-26
en2endtest: extend annonelection

mariajdab:
Adding a test with `TempSIKs` flag set to true to vote with half of the accounts that are not registered using `RegisterSIKForVote`, and the remaining half voting with registered annon accounts.
#1083

### DIFF
--- a/cmd/end2endtest/helpers.go
+++ b/cmd/end2endtest/helpers.go
@@ -290,39 +290,7 @@ func (t *e2eElection) setupCensus(censusType string, nAcct int, createAccounts b
 
 	// Register the accounts in the vochain if is required
 	if censusType == vapi.CensusTypeZKWeighted && createAccounts {
-		errorChan := make(chan error) // Create a buffered channel to prevent deadlock
-		wg := &sync.WaitGroup{}
-
-		for i, acc := range voterAccounts {
-			if i%10 == 0 {
-				// Print some information about progress on large censuses
-				log.Infof("creating %d anonymous census accounts...", len(voterAccounts))
-			}
-
-			wg.Add(1)
-			go func(i int, acc *ethereum.SignKeys) {
-				defer wg.Done()
-				pKey := acc.PrivateKey()
-				if _, _, err := t.createAccount(pKey.String()); err != nil &&
-					!strings.Contains(err.Error(), "createAccountTx: account already exists") {
-					errorChan <- err
-				}
-				log.Infow("anonymous census account created", "index", i, "address", acc.AddressString())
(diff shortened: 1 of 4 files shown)


cmd/end2endtest/helpers.go
```diff
@@ -655,3 +667,38 @@ func (t *e2eElection) endElectionAndFetchResults() (*vapi.ElectionResults, error
...
+			if _, _, err := t.createAccount(pKey.String()); err != nil &&
+				!strings.Contains(err.Error(), "createAccountTx: account already exists") {
+				errorChan <- err
+			}
+			log.Infow("anonymous census account created", "index", i, "address", acc.AddressString())
+		}(i, acc) // Pass the acc variable as a parameter to avoid data race
+	}
+
+	wg.Wait()
+	close(errorChan) // Close the error channel after all goroutines have finished
```
altergui: this needs to be in a goroutine

cmd/end2endtest/helpers.go
```diff
@@ -655,3 +667,38 @@ func (t *e2eElection) endElectionAndFetchResults() (*vapi.ElectionResults, error
 	}
 	return results, nil
 }
+
+func (t *e2eElection) registerAnnonAccts(voterAccounts []*ethereum.SignKeys) error {
+	errorChan := make(chan error) // Create a buffered channel to prevent deadlock
```
altergui: this is unbuffered, delete comment

altergui (commented):
please correct Annon -> Anon

cmd/end2endtest/helpers.go
mariajdab: Yeah, but the PR #1103 resolved that

mariajdab:
ready to review @altergui  :raised_hands:

lucasmenendez (approved):

cmd/end2endtest/zkweighted.go
```diff
@@ -17,13 +17,18 @@ func init() {
 		description: "Performs a complete test of anonymous election, from creating a census to voting and validating votes",
 		example:     os.Args[0] + " --operation=anonelection --votes=1000",
 	}
+	ops["anonelectionTempSIKS"] = operation{
```
altergui: i'd rather keep the case consistent, TempSIKs instead of TempSIKS
(there are more ocurrences of this)

altergui (approved):
LGTM :+1: merge as-is (or fix the case and merge, no need to re-review)

cmd/end2endtest/zkweighted.go
mariajdab: added the suggestion :raised_hands:

PR #1121, opened 2023-09-27
vochain: order transactions when creating new blocks

p4u:
The main objective of the PR was to add transaction ordering to the block production; however, it ended up being a bit more complex, since I found several things that required attention.

1. Transactions are now ordered by sender address and nonce during the Propose block phase. Those transactions which are discarded (because they are invalid) are sent back to the mempool and will be proposed in a further block.

2. The Propose Block method is now checking transactions against a mutable, ongoing state. This means that to consider a list of transactions valid, we need to construct a temporary State which is modified with the transaction's execution. Once the proposal block finishes, it calls State.Rollback() in order to remove the changes.

3. The Process Proposal method checks the validity of transactions and discards the block if some is invalid. This assures we are not including invalid transactions into blocks.

4. There is a new internal TTL mechanism for mempool transactions. After a certain number of blocks, if the transaction did not succeed, it is removed. 

5. The Update SIK roots process now only happens once per block. This is a cleaner approach and prevent consensus issues.

6. The tokens test has been modified to send transactions with reversed nonce, this way we are also testing the ordering of the proposal block creation.


p4u:
@mariajdab could you review the end2end test?

@mvdan I'd love to have your input on this method, specially on the ordering algorithm https://github.com/vocdoni/vocdoni-node/pull/1121/files#diff-088e50f432c8393ca86f7bb7c1054fbd1867eef3d5e25e8773d0dbeabf7bc8adR497

vochain/app.go
```diff
@@ -33,7 +35,10 @@ import (
 const (
 	// recheckTxHeightInterval is the number of blocks after which the mempool is
 	// checked for transactions to be rechecked.
-	recheckTxHeightInterval = 12
+	recheckTxHeightInterval = 6 * 5 // 5 minutes
+	// transactionBlocksTTL is the number of blocks after which a transaction is
+	// removed from the mempool.
+	transactionBlocksTTL = 6 * 10 // 10 minutes
```
jordipainan, line 35: Are these values related to the consesus block time? If so, I would like to suggest to calculate the values on runtime and depending on the block time related parameters.

vochain/app.go
```diff
@@ -474,25 +496,75 @@ func (*BaseApplication) Query(_ context.Context,
...
+			Addr:      senderAddr,
+			Nonce:     nonce,
+			DecodedTx: vtx,
 		})
-		if err != nil || resp.Code != 0 {
+	}
+
+	// Sort the transactions based on the sender's address and nonce
+	sort.Slice(validTxInfos, func(i, j int) bool {
+		if validTxInfos[i].Addr == nil && validTxInfos[j].Addr != nil {
```
jordipainan, line 532: Before the sort is executed, all the transactions included in `validTxInfos` are going to have a non `nil` address. I think the nil checks on the sort can be avoided.

cmd/end2endtest/account.go
```diff
@@ -181,23 +184,48 @@ func testSendTokens(api *apiclient.HTTPclient, aliceKeys, bobKeys *ethereum.Sign
...
+	// send a couple of token txs to increase the nonce, without waiting for them to be mined
+	// this tests that the mempool transactions are properly ordered.
+	wg := sync.WaitGroup{}
+	wg.Add(1)
+	go func() {
+		// send 1 token to burn address with nonce + 1 (should be mined after the other txs)
+		_, err = alice.TransferWithNonce(state.BurnAddress, 1, aliceAcc.Nonce+1)
+		if err != nil {
+			log.Fatalf("cannot burn tokens: %v", err)
+		}
```
mariajdab:
```
if err := alice.TransferWithNonce(state.BurnAddress, 1, aliceAcc.Nonce+1); err != nil {
}
```

Looks good the changes on end2end, maybe avoid use `log.Fatal`, but I can take care of it :raised_hands:

vochain/app.go
```diff
@@ -74,6 +81,9 @@ type BaseApplication struct {
 	dataDir             string
 	genesisInfo         *tmtypes.GenesisDoc
 
+	// prepareProposalLock is used to avoid concurrent calls between PrepareProposal and FinalizeBlock
```
jordipainan: Given that is also used in ProcessProposal, is it right to understand that we try to avoid concurrency issues between PrepareProposal, ProcessProposal and FinalizeBlock ?

vochain/app.go
p4u: I'm not sure about this approach. Because the same way someone defines a shorter block period to have a faster blockchain, the TTL, Recheck and so on should also happen quickly. So I prefer to base everything on the number of blocks and not on the time between blocks.

vochain/app.go
p4u: yes

vochain/transaction/nonce.go
```diff
@@ -0,0 +1,112 @@
...
+			break
+		}
+		txNonce = txWithNonce(ptx)
+	case *models.Tx_SetProcess:
+		ptx := payload.SetProcess
+		if ptx == nil {
+			err = fmt.Errorf("set process payload is nil")
+			break
+		}
+		txNonce = txWithNonce(ptx)
```
altergui: i see a lot of deduplication possible in this switch block, what about:
```go
func (t *TransactionHandler) ExtractNonceAndSender(vtx *vochaintx.Tx) (*common.Address, uint32, error) {
	var ptx interface {
		GetNonce() uint32
	}

	switch payload := vtx.Tx.Payload.(type) {
	case *models.Tx_NewProcess:
		ptx = payload.NewProcess
	case *models.Tx_SetProcess:
		ptx = payload.SetProcess
	case *models.Tx_SendTokens:
		ptx = payload.SendTokens
	case *models.Tx_SetAccount:
		if payload.SetAccount.Txtype == models.TxType_CREATE_ACCOUNT {
			// create account tx is a special case where the nonce is not relevant
			return nil, 0, nil
		}
		ptx = payload.SetAccount
	case *models.Tx_CollectFaucet:
		ptx = payload.CollectFaucet
	case *models.Tx_Vote, *models.Tx_Admin, *models.Tx_MintTokens, *models.Tx_SetKeykeeper,
		*models.Tx_SetTransactionCosts, *models.Tx_DelSIK, *models.Tx_RegisterKey, *models.Tx_SetSIK,
		*models.Tx_RegisterSIK:
		// these tx does not have incremental nonce
		return nil, 0, nil
	default:
		// force panic to detect new tx types
		panic(fmt.Sprintf("unknown payload type on extract nonce: %T", payload))
	}

	if ptx == nil {
		return nil, 0, fmt.Errorf("payload is nil")
	}

	pubKey, err := ethereum.PubKeyFromSignature(vtx.SignedBody, vtx.Signature)
	if err != nil {
		return nil, 0, fmt.Errorf("cannot extract public key from vtx.Signature: %w", err)
	}
	addr, err := ethereum.AddrFromPublicKey(pubKey)
	if err != nil {
		return nil, 0, fmt.Errorf("cannot extract address from public key: %w", err)
	}

	return &addr, ptx.GetNonce(), nil
}

```

vochain/transaction/nonce.go
```diff
@@ -0,0 +1,112 @@
...
+	case *models.Tx_Vote, *models.Tx_Admin, *models.Tx_MintTokens, *models.Tx_SetKeykeeper,
+		*models.Tx_SetTransactionCosts, *models.Tx_DelSIK, *models.Tx_RegisterKey, *models.Tx_SetSIK,
+		*models.Tx_RegisterSIK:
+		// these tx does not have incremental nonce
+		return nil, 0, nil
+	default:
+		// force panic to detect new tx types
+		panic(fmt.Sprintf("unknown payload type on extract nonce: %T", payload))
```
altergui: is this a debugging leftover?
then maybe just `log.Warnf`

vochain/app.go
```diff
@@ -506,8 +579,42 @@ func (app *BaseApplication) PrepareProposal(ctx context.Context,
 // Application SHOULD accept a prepared proposal passed via ProcessProposal, even if a part of the proposal
 // is invalid (e.g., an invalid transaction); the Application can ignore the invalid part of the prepared
 // proposal at block execution time. The logic in ProcessProposal MUST be deterministic.
```
altergui: this comment says the opposite of what happens inside the func (returns ResponseProcessProposal_REJECT in case a transaction was invalid)

i trust your code, but then the comment needs updating (or an explanation why are you deviating from this SHOULD)

vochain/transaction/nonce.go
p4u: It's in purpose (as the comment says) to detect new tx types which have not been added here. So it actas as a protection for future new txs.

vochain/app.go
```diff
@@ -257,11 +268,23 @@ func (app *BaseApplication) InitChain(_ context.Context,
 // CheckTx unmarshals req.Tx and checks its validity
 func (app *BaseApplication) CheckTx(_ context.Context,
 	req *abcitypes.RequestCheckTx) (*abcitypes.ResponseCheckTx, error) {
+	txReference := vochaintx.TxKey(req.Tx)
+	// store the initial height of the tx
+	initialTTLheight, isReferenced := app.txTTLReferences.LoadOrStore(txReference, app.Height())
+	// check if the tx is referenced by a previous block and the TTL has expired
+	if isReferenced && app.Height() > initialTTLheight.(uint32)+transactionBlocksTTL {
```
altergui: the `isReferenced` here is redundant, since `app.Height() > initialTTLheight.(uint32)+transactionBlocksTTL` will never be true unless the value was loaded (not stored)
you can simply do

```go
	// retrieve the height when the tx was first checked, or else store the current height
	initialTTLheight, _ := app.txTTLReferences.LoadOrStore(txReference, app.Height())
	// if the tx was checked in a previous block and the TTL has expired
	if app.Height() > initialTTLheight.(uint32)+transactionBlocksTTL {
		// remove tx reference and return checkTx error
```

altergui (approved):
LGTM,
the only real issue i saw is the `panic` on unknown payloads,

the rest are comments on code readability, that doesn't change behaviour at all

vochain/transaction/nonce.go
altergui: but that means if someone broadcasts a transaction of a new type, the whole network is immediately brought down with a panic?

vochain/app.go
```diff
@@ -474,25 +497,75 @@ func (*BaseApplication) Query(_ context.Context,
 // the ResponsePrepareProposal call. The logic modifying the raw proposal MAY be non-deterministic.
 func (app *BaseApplication) PrepareProposal(ctx context.Context,
```
altergui: deepsource says `ctx` is now unused, suggests to name it `_` then

vochain/app.go
p4u: I guess this comment is for the safety of the non-experneiced users. 

Here more details: https://github.com/cometbft/cometbft/blob/main/spec/abci/abci++_methods.md#processproposal

> Moreover, application implementors SHOULD always set ResponseProcessProposal.status to ACCEPT, unless they really know what the potential liveness implications of returning REJECT are.

So, if we accept any block proposed by a validator, we might run into the risk of including invalid transactions to blocks, which won't be executed since deliverTx() will fail, but the invalid tx will remain as part of the blockchain forever. Thus, a misbehaving validator could potentially fill the blockchain with garbage.

In the other side, if we reject the proposal, we force the validators to act honestly. The downside is that a block will take longer to be produced since a new validator needs to start over the transaction set for a new block.

I think, the way we do it should be safe since on prepare proposal, we are executing all tx against an ongoing state (reverted with Rollback() at the end) and on process proposal we do exactly the same. If the state is deterministic (which must be) the outcome should always be the same (either ACCEPT or REJECT).

vochain/app.go
```diff
@@ -257,11 +268,23 @@ func (app *BaseApplication) InitChain(_ context.Context,
...
+	txReference := vochaintx.TxKey(req.Tx)
+	// store the initial height of the tx
+	initialTTLheight, isReferenced := app.txTTLReferences.LoadOrStore(txReference, app.Height())
+	// check if the tx is referenced by a previous block and the TTL has expired
+	if isReferenced && app.Height() > initialTTLheight.(uint32)+transactionBlocksTTL {
+		// remove tx reference and return checkTx error
+		log.Debugw("pruning expired tx from mempool", "height", app.Height(), "hash", fmt.Sprintf("%x", txReference))
+		app.txTTLReferences.Delete(txReference)
+		return &abcitypes.ResponseCheckTx{Code: 1, Data: []byte(fmt.Sprintf("tx expired %x", txReference))}, nil
+	}
```
altergui: also, these lines are not covered by tests: https://coveralls.io/builds/62999336/source?filename=vochain%2Fapp.go#L276

it would be great if you could add a test or extend one of the existing ones. 

AFAIU, pushing a tx with a very high nonce would do it, right?
and then wait until it's pruned, then check somehow the result

i can also self-assign this little task as well if you prefer and push a commit to PR

vochain/transaction/nonce.go
p4u: That's a good point. But I believe Protobuf won't allow the unmarshal or would set it to default value which is models.Tx_UNKNOWN. It is something to check, for sure.

If the protobuf models contain the txType because it is a new introduced tx type, then the tests should fail.

altergui: if protobuf "protects" us from a DDOS, that's very nice, but in any case, we should really avoid panics in the codepath that processes input from outside, i bet it would raise questions from any security auditor

p4u: I have moved the Panic to log.Error(), thanks for the input @altergui

tree/arbo/tree.go
```diff
@@ -620,6 +625,8 @@ func (t *Tree) SetRootWithTx(wTx db.WriteTx, root []byte) error {
 // The provided root must be a valid existing intermediate node in the tree.
 // The list of roots for a level can be obtained using tree.RootsFromLevel().
 func (t *Tree) Snapshot(fromRoot []byte) (*Tree, error) {
+	t.Lock()
+	defer t.Unlock()
```
mvdan, line 629: I assume these were added while you were debugging the pebble panic, and should now be removed?

vochain/cometbft.go
```diff
@@ -0,0 +1,452 @@
...
+		validTxInfos = append(validTxInfos, txInfo{
+			Data:      tx,
+			Addr:      senderAddr,
+			Nonce:     nonce,
+			DecodedTx: vtx,
+		})
+	}
+
+	// Sort the transactions based on the sender's address and nonce
+	sort.Slice(validTxInfos, func(i, j int) bool {
```
mvdan, line 328: FYI now that we depend on Go 1.21 you can use better generics, e.g. https://pkg.go.dev/slices#SortFunc

vochain/cometbft.go
```diff
@@ -0,0 +1,452 @@
...
+			return true
+		}
+		if validTxInfos[i].Addr != nil && validTxInfos[j].Addr == nil {
+			return false
+		}
+		if validTxInfos[i].Addr != nil && validTxInfos[j].Addr != nil {
+			if validTxInfos[i].Addr.String() == validTxInfos[j].Addr.String() {
+				return validTxInfos[i].Nonce < validTxInfos[j].Nonce
+			}
+			return validTxInfos[i].Addr.String() < validTxInfos[j].Addr.String()
```
mvdan: nit: reuse the calls to String, since they might not be free

mvdan: or even better, can we compare the addresses without turning them to strings? presumably the addresses are bytes, so we can use `bytes.Compare` without needing to turn them to e.g. hex or base64 strings.

tree/arbo/tree.go
```diff
@@ -84,7 +84,7 @@ var (
 
 // Tree defines the struct that implements the MerkleTree functionalities
 type Tree struct {
-	sync.Mutex
+	sync.RWMutex
```
mvdan: remember that RWMutex is more complex and expensive than Mutex, it is only worth it if we're going to be doing mostly reads. if there is a roughly equal mix of reads and writes, a regular Mutex will perform much better.

vochain/app.go
p4u: please, feel free to add the missing test :)

tree/arbo/tree.go
p4u: Right, going back to Mutex. It was part of my previous testing.

vochain/cometbft.go
p4u: I don't feel ready yet to use Go Generics in a comfortable way. I think I'll let this improvement for the future, or feel free if you want to take it.

tree/arbo/tree.go
mvdan: not sure what happened here - I still think these added locks/unlocks were only there while you debugged pebble

p4u: Yes, you are right. But while I was reading the Arbo code, I found some missing locks. Which might not be critical, but I prefer to make it secure.

Issue #1126, opened 2023-10-02
bug: Vocone can not download from IPFS censuses created by census3 


lucasmenendez:
### Describe the bug

Using the [deploy-snippets](https://github.com/vocdoni/deploy-snippets) repo I found an error during the creation of elections using a *"Token-based census"* from Census3.

Census3 syncs, creates and publishes successfully the census to IPFS but the Vocone can not download it, so the census never is created and the voters can cast any vote.

To validate the published census, you can use the following [gist](https://gist.github.com/lucasmenendez/36fc09785c9efbc60e92d51b2c84a3ce). Example:
```sh
go run main.go -uri ifps://...
```

### To Reproduce (please complete the following information)

- Run the whole stack using [deploy-snippets](https://github.com/vocdoni/deploy-snippets). Be sure that you defines a valid goerli web3 endpoint for census3.
- Create a test token:
```sh
curl -X POST \
  'http://localhost:8083/api/token' \
  --header 'Accept: */*' \
  --header 'Content-Type: application/json' \
  --data-raw '{
    "id": "0xf530280176385af31177d78bbfd5ea3f6d07488a",
    "type": "erc20",
    "tag": "test2",
    "chainID": 5
}'
```
- Wait to be synced. Check it with the following `curl`:
```sh
curl  -X GET \
  'http://localhost:8083/api/token/0xf530280176385af31177d78bbfd5ea3f6d07488a' \
  --header 'Accept: */*'
```
- The go to `localhost:8080` and create an election with a token-based census using the `Token Monkeys (MON)`.
- Look at the vocone logs.

altergui:
so, upon investigation, this is due to a bug in `voconed`, that doesn't init the offchaindatadownloader properly
it just does:

https://github.com/vocdoni/vocdoni-node/blob/ddea83d2a0fd2b463475d00be1d37ffaac9beb3e/vocone/vocone.go#L129-L136

crucially, it's missing `vs.DataDownloader.Start()` that `node` does during init

https://github.com/vocdoni/vocdoni-node/blob/61101a8e4f32103c24a8e28fee2d6715fd9a7af6/service/offchaindata.go#L16-L37

this is a simple fix

```diff
diff --git a/vocone/vocone.go b/vocone/vocone.go
index 568e0444..6200ea99 100644
--- a/vocone/vocone.go
+++ b/vocone/vocone.go
[...]

PR #1129, opened 2023-10-02
vocone: refactor Vocone struct as an extension of service.VocdoniService

altergui:
this way, the whole init phase is much similar to what `node` does, avoiding bugs
and making it easier to maintain in the future when features are added
    
this refactor was originated by a bug introduced when OffchainDataDownloader feature
was merged on `node` but not fully properly in `voconed`,
so `voconed` was not doing vs.DataDownloader.Start() that `node` does during init
    
the best solution was to make voconed call the same init function that node calls
srv.OffChainDataHandler()

### DIFF
--- a/cmd/node/main.go
+++ b/cmd/node/main.go
@@ -387,7 +387,6 @@ func main() {
 	}
 
 	var err error
-	var vochainKeykeeper *keykeeper.KeyKeeper
 	srv := service.VocdoniService{Config: globalCfg.Vochain}
 
 	if globalCfg.Mode == types.ModeGateway {
@@ -503,7 +502,7 @@ func main() {
 		} else {
 			// start keykeeper service (if key index specified)
 			if validator.KeyIndex > 0 {
-				vochainKeykeeper, err = keykeeper.NewKeyKeeper(
+				srv.KeyKeeper, err = keykeeper.NewKeyKeeper(
 					path.Join(globalCfg.Vochain.DataDir, "keykeeper"),
 					srv.App,
 					&signer,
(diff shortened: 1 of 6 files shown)


altergui:
i'm pushing this trivial fix, so that it can be used (or tested) already.

but i'd not merge this PR as-is, i'll try first to make a nicer refactor of voconed init and push that instead

altergui:
there you go, nicer refactor of voconed, that as a side effect fixes #1126 and prevents future issues like this when features are implemented

altergui:
i need to fix TestVocone but it's probably something trivial,
the PR can still be reviewed as is

i also thought of replacing that TestVocone with an actual integration test, since voconed binary (and docker image) is totally untested (and this created problems in the past for vocdoni-sdk CI)

lucasmenendez:
> there you go, nicer refactor of voconed, that as a side effect fixes #1126 and prevents future issues like this when features are implemented

I was testing with this branch with two different tokens: 
 * `0xBFAbdE619ed5C4311811cF422562709710DB587d` (mainnet) -> `ipfs://bafybeiehw3bldmap6buiscwbsys3bdcahpqt2xzugieu6ai2yq4ohijghq`
 * `0xF530280176385AF31177D78BbFD5eA3f6D07488A` (goerli) -> `ipfs://bafybeiezzvgxwv6nykjkbcmqfd55g7smpilvj7nedx2crynyhtp2u2tx5a`

And for me, the problem continues 🙁. The file is in the downloader queue but the census is not created.

altergui:
thanks for the report, 
i'd say it's related to the error that also happens in TestVocone
>           e"missing modules attached for enabling handler censuses"
i'm looking into it

p4u (approved):
Makes sense, thx

lucasmenendez (commented):
The changes are OK for me, but testing the latest version it still fails to download censuses from IPFS, even with the same IPFS connect key in both services (census3 and vocone).

PR #1135, opened 2023-10-09
indexer and api: track accounts, balances 

mariajdab:
### Note: 
The proposal approach use upsert operation to insert or update an account, due the currently approach calls `OnSetAccount` Listener when an account is created and also when the balance is updated. This approach was tested in the indexer and integration test. Suggestions are welcome.

- [x] SQL changes
- [x] Query for fetching the paginated list of accounts
- [x] Query for fetching the list of token transfers by `to_account`
- [x] Query to count the number of accounts
- [x] Implement a query to count the number of transfers where the `to_account` is the provided address and where the `from_account` is the provided address
- [x] IAPI endpoint and APIclient method  `/accounts/page/:num`
- [x] API endpoint and APIclient method  `/accounts/count`
- [x] API endpoint  `/accounts/{accountID}/transfers/count` to provide the total number of sent and received transactions
- [x] Current API endpoint `/accounts/{accountID}/transfers/page/:page` extended so it contains the `received` and `sent` transfers.
- [x] unit tests of the indexer new queries
- [x] integration tests into `test/api_test.go`


mariajdab:
@mvdan could you please review this PR when you have a chance? thanks :raised_hands:

apiclient/account.go
```diff
@@ -260,20 +260,22 @@ func (c *HTTPclient) AccountSetMetadata(metadata *api.AccountMetadata) (types.He
...
 	resp, code, err := c.Request(HTTPGET, nil, "accounts", from.Hex(), "transfers", "page", strconv.Itoa(page))
 	if err != nil {
 		return nil, err
 	}
 	if code != apirest.HTTPstatusOK {
 		return nil, fmt.Errorf("%s: %d (%s)", errCodeNot200, code, resp)
 	}
-	var transfers []*indexertypes.TokenTransferMeta
-	if err := json.Unmarshal(resp, &transfers); err != nil {
+	trnsfs := new(struct {
```
altergui: why not `transfers`?

apiclient/account.go
```diff
@@ -260,20 +260,22 @@ func (c *HTTPclient) AccountSetMetadata(metadata *api.AccountMetadata) (types.He
 	return accv.TxHash, nil
 }
 
-// GetTransfers returns the list of token transfers associated with an account
-func (c *HTTPclient) GetTransfers(from common.Address, page int) ([]*indexertypes.TokenTransferMeta, error) {
+// GetTransfers returns the list of sent and received token transfers associated with an account
+func (c *HTTPclient) GetTransfers(from common.Address, page int) (map[string][]*indexertypes.TokenTransferMeta, error) {
```
altergui: i'd rename `from` to something less misleading, like `account`

altergui: also i'd name the method ListTransfers or even ListTokenTransfers to be consistent with ListAccounts and CountTokenTransfers

apiclient/account.go
```diff
@@ -417,3 +419,61 @@ func (c *HTTPclient) RegisterSIKForVote(electionId types.HexBytes, proof *Census
...
+
+	if err := json.Unmarshal(resp, &accts); err != nil {
+		return nil, err
+	}
+
+	return accts.Accounts, nil
+}
+
+// AccountCount returns the total count of exiting accounts
+func (c *HTTPclient) AccountCount() (uint64, error) {
```
altergui: i'd name this CountAccounts to be consistent with CountTokenTransfers  and ListAccounts

api/accounts.go
```diff
@@ -441,3 +492,77 @@ func (a *API) tokenFeesHandler(_ *apirest.APIdata, ctx *httprouter.HTTPContext)
...
+//
+//	@Summary		List of the existing accounts
+//	@Description	Returns information (address, balance and nonce) of the existing accounts
+//	@Tags			Accounts
+//	@Accept			json
+//	@Produce		json
+//	@Param			page	path		string	true	"Paginator page"
+//	@Success		200		{object}	object{accounts=[]indexertypes.Account}
+//	@Router			/accounts/page/{page} [get]
+func (a *API) accountsListHandler(_ *apirest.APIdata, ctx *httprouter.HTTPContext) error {
```
altergui: i'd suggest `accountListHandler` ,to be consistent with other methods

api/accounts.go
```diff
@@ -351,7 +402,7 @@ func (a *API) electionCountHandler(_ *apirest.APIdata, ctx *httprouter.HTTPConte
 
 // tokenTransfersHandler
```
altergui: you could take the chance to rename this to tokenTransfersListHandler , for extra consistency

vochain/indexer/indexer.go
```diff
@@ -705,3 +715,78 @@ func (idx *Indexer) GetTokenFeesByType(txType string, offset, maxItems int32) ([
 	}
 	return tt, nil
 }
+
+// GetTokenTransfersByToAccount returns all the token transfers made to a given account
+// from the database, ordered by timestamp and paginated by maxItems and offset
+func (idx *Indexer) GetTokenTransfersByToAccount(to []byte, offset, maxItems int32) ([]*indexertypes.TokenTransferMeta, error) {
```
altergui, line 743: what about dropping the "By", could it be GetTokenTransfersToAccount ?
that would make also GetTokenTransfersFromAccount
and the very explicit GetTokenTransfersFromOrToAccount

altergui (commented):
i didn't finish the review yet but leaving here some thought about names

vochain/indexer/indexer.go
mariajdab: In the indexer methods the "By" is currently used in: `GetProcessIDsByFinalResults, GetTokenFeesByFromAccount, UpdateProcessResultByID`. In order to keep the pattern I think we can also use it here.

altergui: perfect, SGTM

vochain/indexer/indexer.go
```diff
@@ -699,3 +709,78 @@ func (idx *Indexer) GetTokenFeesByType(txType string, offset, maxItems int32) ([
...
+			TxHash:    t.TxHash,
+			Timestamp: t.TransferTime,
+		})
+	}
+	return tt, nil
+}
+
+// GetTokenTransfersAccount returns all the token transfers made to and from a given account
+// from the database, ordered by timestamp and paginated by maxItems and offset
+func (idx *Indexer) GetTokenTransfersAccount(acc []byte, offset, maxItems int32) (map[string][]*indexertypes.TokenTransferMeta, error) {
```
altergui: i believe this should be `GetTokenTransfersByAccount`

vochain/indexer/indexer.go
```diff
@@ -699,3 +709,78 @@ func (idx *Indexer) GetTokenFeesByType(txType string, offset, maxItems int32) ([
...
+
+	return map[string][]*indexertypes.TokenTransferMeta{
+		"Received": transfersTo,
+		"Sent":     transfersFrom,
+	}, nil
+
+}
+
+// CountTokenTransfersAccount returns the count all the token transfers made from a given account
+func (idx *Indexer) CountTokenTransfersAccount(acc []byte) (uint64, error) {
```
altergui: i believe this should be CountTokenTransfersByAccount

vochain/indexer/queries/token_transfers.sql
```diff
@@ -18,4 +18,18 @@ WHERE from_account = sqlc.arg(from_account)
...
+-- name: GetTokenTransfersByToAccount :many
+SELECT * FROM token_transfers
+WHERE to_account = sqlc.arg(to_account)
+ORDER BY transfer_time DESC
+LIMIT sqlc.arg(limit)
+OFFSET sqlc.arg(offset)
+;
+
+
+-- name: CountTokenTransfersAccount :one
```
altergui: same here: should it be CountTokenTransfersByAccount ?

altergui (commented):
a couple more comments about names consistency, since i see in your commit message:
   > add new queries in token_transfers: GetTokenTransfersByToAccount, GetTokenTransfersByAccount, CountTokenTransfersByAccount

which sounds excellent and consistent,

yet in the code i see you missed the "By" in GetTokenTransfersAccount, CountTokenTransfersAccount. i pointed out the ones i spotted, maybe there are more

p4u:
@mariajdab could you resolve and close the pending comments please?

mariajdab:
> @mariajdab could you resolve and close the pending comments please?

Done :raised_hands:

p4u:
It LGTM, good job @mariajdab

p4u:
@mvdan it would be nice if you can take a look, however if that's not possible, we'll merge this into main, since other changes to the indexer are pending and would conflict with this one.

vochain/indexer/queries/account.sql
```diff
@@ -0,0 +1,20 @@
+-- name: UpsertAccount :execresult
+INSERT INTO accounts(
+  account, balance, nonce
+) VALUES (
+    ?, ?, ?
+) ON CONFLICT(account)
+  DO UPDATE SET
+    balance = excluded.balance,
+    nonce = excluded.nonce
+;
```
mvdan, line 5: this should just be REPLACE INTO; see queries/votes.sql for an example

vochain/indexer/indexer.go
```diff
@@ -727,3 +737,78 @@ func (idx *Indexer) GetTokenFeesByType(txType string, offset, maxItems int32) ([
...
+	}
+	transfersFrom, err := idx.GetTokenTransfersByFromAccount(acc, offset, maxItems)
+	if err != nil {
+		return nil, err
+	}
+
+	return map[string][]*indexertypes.TokenTransferMeta{
+		"Received": transfersTo,
+		"Sent":     transfersFrom,
+	}, nil
```
mvdan: personally I wouldn't use a map with always the same two string keys - a struct will work much better

mvdan (commented):
SGTM

vochain/indexer/queries/account.sql
mariajdab: I was reading about `REPLACE` and `UPSERT` approach, seems that last is faster due the `REPLACE` deletes the row and then inserts a new one with the same primary key, and the `UPSERT` only update the row. Or maybe I am missing something and there is another reason to use REPLACE?

mvdan: Have you benchmarked your approach to actually be noticeably faster, or is this explicitly documented upstream? as far as I can tell both methods will work fine, and REPLACE is shorter and simpler, so that's what I have been using

p4u: Let's apply our current standard which is REPLACE, and let's check in the future if it is worth it to change everything to UPSERT or not. Thanks @mariajdab for taking the time to consider a potential faster alternative.

mariajdab: Sure, sounds good! :raised_hands: It would be worth confirming with the benchmark test in the future

mariajdab:
The last step is: open the issue in Interoperability announcing the changes to the API, so the SDK and documentation can incorporate them. Just to confirm that needs to be done before or after merge this PR ? @p4u

p4u:
> The last step is: open the issue in Interoperability announcing the changes to the API, so the SDK and documentation can incorporate them. Just to confirm that needs to be done before or after merge this PR ? @p4u

You can do it after. Just document the new endpoints so at some point they can be incorporated into the SDK https://github.com/vocdoni/interoperability/issues

p4u:
I will merge this PR. Thanks for your work @mariajdab

PR #1136, opened 2023-10-09
fix consensus failure

altergui:
When we migrated to cometBFT 0.37, the new API requires to do the block execution process in two steps: 1) execute 2) save

Our statedb was not designed to work in two steps, but one. So there was a bug on the root hash returned by the execute step, which was different from the one returned by save. That created a consensus failure if the node was stop when processing transactions.


altergui:
i'm worried about mixing this important hotfix in a feature PR so i cherry-picked it from @p4u to be reviewed and merged separately

altergui:
(still solving conflicts)

altergui:
so AFAIU this is the fix (i untangled several unrelated changes from your original commit, mainly related to tx reordering)
https://github.com/vocdoni/vocdoni-node/pull/1136/commits/f1c82d136daf5d1febd2dac1128798304102909e

if we merge this on main, you'll likely need to solve conflicts when rebasing your branch f/order_transactions but i think it's really worth it

altergui:
fixed by 19c6d68

PR #1144, opened 2023-10-16
stage hotfix election price

altergui:
hotfix(stage): election price calculation
    
* to avoid resetting the chain vocdoni-stage-8, soft fork at block 221000

* adapt TestAPIElectionCost for soft fork (vocdoni-stage-8 block 221000)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -89,12 +89,12 @@ jobs:
           go-version: '1.21'
       - name: Run Go test (and collect code coverage)
         if: github.event_name == 'pull_request' || github.ref == 'refs/heads/dev' # quicker, non-race test in case it's a PR or push to dev
-        run: go test -coverprofile=unit.covdata.txt ./...
+        run: go test -coverprofile=unit.covdata.txt -coverpkg=./... ./...
       - name: Run Go test -race (and collect code coverage)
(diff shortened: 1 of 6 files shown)


altergui:
so here is my proposal, in essence it's just

https://github.com/vocdoni/vocdoni-node/blob/f2584cb1cc74b3a72ea0101be323b634e370f695/vochain/transaction/election_tx.go#L258-L271

the rest is an integration test to confirm the behavior is expected

altergui:
before applying the hotfix, the cherry-picked test from main fails (just as expected)
https://github.com/vocdoni/vocdoni-node/actions/runs/6532485302/job/17735749457

after applying the fix it passes OK
https://github.com/vocdoni/vocdoni-node/actions/runs/6532487742/job/17735756812

and finally i adapted the test to check what happens with elections created on `vocdoni-stage-8` before the fork and after the fork
https://github.com/vocdoni/vocdoni-node/actions/runs/6532881636/job/17736916742

altergui:
updated to fork at block 221000 (estimated to happen on wed 2023-10-18T10:25:29.318924937Z)

altergui:
i believe at some point we'll want to implement something more structured, like go-ethereum does, having a package `params` specifying which chains (stage, production) fork at which points.

https://github.com/ethereum/go-ethereum/blob/db9afae2eab509d01805a75440390bd000aa3eb4/params/config.go#L39-L49

yet that proposal is out of scope for this PR. it would, if anything, be developed on top of main

PR #1145, opened 2023-10-16
use the flag --no-install-recommends

mariajdab:
use --no-install-recommends for installation time decrease and could help with the container space.


altergui (approved):
love it!

tiny detail, if you agree: i'd leave the verb `install` first, as in `apt-get install --no-install-recommends -y ....`

altergui:
also, you could cherry-pick this into https://github.com/vocdoni/census3 and https://github.com/vocdoni/vocfaucet

PR #1161, opened 2023-10-23
vochain: remove treasurer logic

p4u:


jordipainan (approved):
LGTM but not tested :)

PR #1169, opened 2023-10-28
add an end2end benchmark test for census database

p4u:


altergui (approved):
LGTM

regarding the `t.batchSize = 250 // TODO: make this configurable` i believe at some point it would make sense to implement https://github.com/spf13/cobra , refactor all the current `VochainTest`s as `cobra.Command`  (current method Setup would become PreRun, very small changes in fact)

but totally out of scope for this PR

mariajdab (approved):

Issue #1170, opened 2023-10-30
indexer: better support for archive


p4u:
In order to support the transparent process archive (https://github.com/vocdoni/vocdoni-node/issues/1155) we need to perform some changes on the indexer.

- [x] vouteCount should be a static integer, not a COUNT
- [x] voteCount should be marshaled by JSON 
- [x] we need a startDate and a endDate (currently computed dynamically by the API using the block count, but this does not work for archive). Both of them can be easily set by the indexer (OnProcessStart  + OnProcessStatusChange{ENDED, CANCELED}).
- [x] we need a new boolean (fromArchive) which is used to indicate the process was added from an archive
- [x] nice to have: a chainID string for the process to indicate on which chain the process has held

p4u:
Regarding ResultsWeight() I don't find the way to incorporate it into the database. Can you confirm @mvdan whether there is a way or not to do so?

p4u:
https://github.com/vocdoni/vocdoni-node/pull/1185

PR #1172, opened 2023-10-30
F/integrated archive

### LINKED ISSUE #1155: Integrate process archive in the API backend
When an election is queried, the API backend should:

1. Check if the chainID of the electionID matches its own, if so, return it from the current state/indexer
2. If its another chainID, then try to fetch it from the process archive

Things to consider:

- the process archive is currently an IPNS directory
- there could be more than 1 IPNS directory (multiple archives)
- we need somewhere to store the IPNS archives, maybe a remote HTTP file (could be a github json manifest) that is fetched by the gateway nodes
- the gateway should allow enabling/disabling the archive retrieval and specify a custom URL
- consider that the IPNS directories can be changed over time
- index locally the archive elections and use a goroutine to keep them updated
- when an election is returned and it comes from archive, there should be a boolean in the reply specifying it
- probably the indexer can take care of this task, it makes sense
- check our IPFS package, it contains most of the functions required for dealing with IPNS

p4u:
Closes #1155


p4u:
needs https://github.com/vocdoni/vocdoni-node/issues/1170 to finish

p4u:
I'll merge this PR with as a partial solution. We still need to add some extra fields to indexer in order to make it fully transparent for the UI.

data/ipfs/ipfs.go
```diff
@@ -313,10 +329,29 @@ func (i *Handler) Retrieve(ctx context.Context, path string, maxSize int64) ([]b
...
+	fsize, err := file.Size()
+	if err != nil {
+		return nil, err
+	}
+
+	if fsize > MaxFileSizeBytes {
+		return nil, fmt.Errorf("file too big: %d", fsize)
+	}
+
+	return io.ReadAll(file)
```
mvdan: use `io.LimitReader` with `MaxFileSizeBytes` too; otherwise if `Size` is buggy you might still read gigabytes here.

vochain/indexer/archive.go
```diff
@@ -0,0 +1,138 @@
...
+	EndDate     *time.Time            `json:"endDate,omitempty"`
+}
+
+// ImportArchive imports an archive list of processes into the indexer database.
+func (idx *Indexer) ImportArchive(archive []*ArchiveProcess) error {
+	tx, err := idx.readWriteDB.Begin()
+	if err != nil {
+		panic(err) // shouldn't happen, use an error return if it ever does
+	}
+	queries := idx.blockQueries.WithTx(tx)
```
mvdan: don't use `blockQueries`, that's only for `Commit` and is protected by `blockMu`. you can always add `importQueries` if you want to keep the prepared queries around for this as well.

vochain/indexer/archive.go
```diff
@@ -0,0 +1,138 @@
...
+	StartDate   *time.Time            `json:"startDate,omitempty"`
+	EndDate     *time.Time            `json:"endDate,omitempty"`
+}
+
+// ImportArchive imports an archive list of processes into the indexer database.
+func (idx *Indexer) ImportArchive(archive []*ArchiveProcess) error {
+	tx, err := idx.readWriteDB.Begin()
+	if err != nil {
+		panic(err) // shouldn't happen, use an error return if it ever does
+	}
```
mvdan, line 39: you're missing a `defer tx.Discard()` here, otherwise if this func fails you leak the transaction without committing

mvdan (approved):

vochain/indexer/archive.go
p4u: Can you extend a bit more on that please? What you mean by "add importQueries" ?

vochain/indexer/archive.go
p4u: Do you mean tx.Rollback()

mvdan: yes sorry

vochain/indexer/archive.go
mvdan: easiest to not use prepared statements for now. we can always add them later without a data race..

PR #1174, opened 2023-10-31
metrics: use VictoriaMetrics instead of prometheus/client_golang

altergui:
Most notably:
* the syntax is much simpler (no Register, less verbose calls)
* now we can Get metrics (concurrent-safe, along with Set), so:
  * data/ipfs/metrics.go now uses Set and Get instead of atomic.Float64
  * vochaininfo also uses Set and Get for most metrics, out of vi.lock.Lock()

Additional changes:
* drop metric vochain_vote_tree_increase_last_minute
* drop package `go.vocdoni.io/dvote/metrics` altogether
* unify types of vochaininfo to uint64
* drop vs.MetricsAgent field


altergui:
i triple-checked that there are no regressions (manually, using curl, cut, sort, uniq, meld), i.e. the exposed metrics stays the same.

i don't think it's worth it to write a test for such a regression, so you'll just have to take my word for it

p4u (approved):
LGTM

PR #1176, opened 2023-11-02
stashed WalkDir cleanup, update pebble

mvdan:


p4u:
Is this backwards compatible with previous pebble databases?

mvdan:
I haven't checked. They are v0 and don't do releases, so they don't specify. The commits are just over 200: https://github.com/cockroachdb/pebble/compare/06034ff014e0...aa077af62593

FormatDefault in that diff hasn't changed, meaning the DB format should still be the same.

Issue #1154, opened 2023-10-17
bug: flake in e2etest_overwritelection 


altergui:
i just came across this CI failure:
https://github.com/vocdoni/vocdoni-node/actions/runs/6549290779/job/17785876915?pr=1153
```
2023-10-17 15:22:29+00:00 ### Logs from failed test (e2etest_overwritelection) ###
[...]
2023-10-17T15:19:25.613Z INF end2endtest/helpers.go:648 > 10 votes registered successfully, took 6.087378ms (1642 votes/second)
2023-10-17T15:19:25.613Z INF end2endtest/helpers.go:659 > ending election...
2023-10-17T15:19:25.613Z DBG apiclient/client.go:160 > http request body=null path=/v2/accounts/0xe6Cb303036BC55Ea41C8F91991B95f580d7cDC1a type=GET
2023-10-17T15:19:25.618Z DBG apiclient/client.go:160 > http request body={"payload":"CioiKAgCEAEaIMXSRgGG9+bLMDA2vFXqQcj5GZG5X1gNfNwaAgAAAAAAIAISQYEv6MzVIJ4bqaIacrBSRZwGQln7dzvZWYHVPGQ6inajbVrERQxWuqv7S5nTy82uRKJ9W4XKj7i5dAAtwTUROMMA"} path=/v2/chain/transactions type=POST
2023-10-17T15:19:25.627Z INF apiclient/helpers.go:169 > waiting for election c5d2460186f7e6cb303036bc55ea41c8f91991b95f580d7cdc1a020000000000 to publish final results
2023-10-17T15:19:25.628Z DBG apiclient/client.go:160 > http request body=null path=/v2/elections/c5d2460186f7e6cb303036bc55ea41c8f91991b95f580d7cdc1a020000000000/scrutiny type=GET
2023-10-17T15:19:29.631Z DBG apiclient/client.go:160 > http request body=null path=/v2/elections/c5d2460186f7e6cb303036bc55ea41c8f91991b95f580d7cdc1a020000000000/scrutiny type=GET
2023-10-17T15:19:33.634Z DBG apiclient/client.go:160 > http request body=null path=/v2/elections/c5d2460186f7e6cb303036bc55ea41c8f91991b95f580d7cdc1a020000000000/scrutiny type=GET
[...]

PR #1181, opened 2023-11-06
vochain: ensure transaction order is deterministic when Nonce and Addr are equal

altergui:
while digging into https://github.com/vocdoni/vocdoni-node/issues/1154 i realised two txs can have the same Nonce and Addr (two Votes from the same voter, that are vote overwrites) and in that case the order is not deterministic, but the resulting state (election result) will be different, depending on which tx is accepted and which one is discarded (as invalid, due to MaxVoteOverwrite)


p4u:
But this should not be relevant for the consensus, since the one ordering is the proposer. The rest of the nodes only need to execute and verify the execution. The order of two transactions with the same nonce (the second would be invalid and must be discarted) should not matter.

altergui:
> one ordering is the proposer. The rest of the nodes only need to execute and verify the execution

aaaaah makes total sense. sorry for the noise

PR #1183, opened 2023-11-07
end2end: allow to run all tests and refactor

mariajdab:
- allow to run all tests with `-- operation all`
- refactor to improve the readability

### DIFF
--- a/cmd/end2endtest/main.go
+++ b/cmd/end2endtest/main.go
@@ -23,6 +23,10 @@ import (
 // * how many times to retry opening a connection to an endpoint before giving up
 const retries = 10
 
+// specifying the flag --operation all will allow to run all tests available in
+// the ops map
+const allTests = "all"
+
 // VochainTest is the interface that tests should comply with.
 // Duration of each step will be measured and reported.
 //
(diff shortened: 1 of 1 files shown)


mariajdab:
@altergui if we use goroutine to rull the tests, it will return the error:  `faucet request failed: 402 Payment Required` Do you have any idea about the reason?

altergui (approved):
LGTM

about the faucet error, here's my hunch (although i didn't look at the code at all):
it's because you're using the same accountPrivKeys for all the tests, and the faucet might have a rate limit per account.

if you run all tests concurrently but using different accountPrivKeys, faucet doesn't complain
or if you use the same accountPrivKeys but don't ask for tokens all at the same time (not concurrently) that's also OK for the faucet

mariajdab:
> LGTM
> 
> about the faucet error, here's my hunch (although i didn't look at the code at all): it's because you're using the same accountPrivKeys for all the tests, and the faucet might have a rate limit per account.
> 
> if you run all tests concurrently but using different accountPrivKeys, faucet doesn't complain or if you use the same accountPrivKeys but don't ask for tokens all at the same time (not concurrently) that's also OK for the faucet

I will check it. Thanks @altergui

mariajdab:
> > LGTM
> > about the faucet error, here's my hunch (although i didn't look at the code at all): it's because you're using the same accountPrivKeys for all the tests, and the faucet might have a rate limit per account.
> > if you run all tests concurrently but using different accountPrivKeys, faucet doesn't complain or if you use the same accountPrivKeys but don't ask for tokens all at the same time (not concurrently) that's also OK for the faucet
> 
> I will check it. Thanks @altergui

Updated to run concurrently

cmd/end2endtest/main.go
```diff
@@ -172,20 +172,61 @@ func NewAPIclient(host string) (*apiclient.HTTPclient, error) {
...
+				cfgCopy := *c
+				cfgCopy.accountPrivKeys = []string{util.RandomHex(32)}
+				if err := createSignKeys(&cfgCopy); err != nil {
+					log.Fatal(err)
+				}
+				c = &cfgCopy
+				if err := setupAndRun(op, api, c); err != nil {
+					log.Fatal(err)
+				}
+			}(op) // Pass the op variable
```
altergui: you're mixing here two styles, https://go.dev/doc/faq#closures_and_goroutines
* passing op as an arg
* accessing c directly

i'd stick to one style, namely accessing directly, since it's the most common in our codebase.

so just 

```
for _, op := range ops {
  wg.Add(1)
  op := op
  go func() {
    // use op
  }()
```

i believe you can also copy `c := c` (along with `op:=op`, before the `go func`) and avoid the new name "cfgCopy" as well as `c = &cfgCopy`, but since it's a pointer i'm not 100% sure

mariajdab: Yeah, agree. I think in that way is cleaner. :raised_hands:

mariajdab: Updated

PR #1185, opened 2023-11-08
indexer: extend to improve archive support

p4u:
- add startTime and endTime to indexer, this way we do not need the block time estimation.

- add the chainID into the indexer database, so we can identify the origin chain

- add fromArchive flag to inform the user about the process being added by the archive

### DIFF
--- a/vochain/indexer/archive.go
+++ b/vochain/indexer/archive.go
@@ -40,7 +40,6 @@ func (idx *Indexer) ImportArchive(archive []*ArchiveProcess) ([]*ArchiveProcess,
 		return nil, err
 	}
 	defer tx.Rollback()
-	height := idx.App.State.CurrentHeight()
 	queries := indexerdb.New(tx)
 	added := []*ArchiveProcess{}
 	for _, p := range archive {
@@ -61,17 +60,33 @@ func (idx *Indexer) ImportArchive(archive []*ArchiveProcess) ([]*ArchiveProcess,
 		} else {
 			continue
 		}
-		creationTime := time.Now()
-		if p.StartDate != nil {
-			creationTime = *p.StartDate
+
+		// For backward compatibility, we try to fetch the start/end date from multiple sources.
+		// If not found, we calculate them from the block count and the default block time.
+		startDate := p.ProcessInfo.StartDate
+		if startDate.IsZero() {
+			if p.StartDate != nil {
+				startDate = *p.StartDate
+			} else {
+				// Calculate startDate equal to time.Now() minus defaultBlockTime*p.ProcessInfo.BlockCount
+				startDate = time.Now().Add(-types.DefaultBlockTimeSeconds * time.Duration(p.ProcessInfo.BlockCount))
+			}
 		}
+		endDate := p.ProcessInfo.EndDate
+		if endDate.IsZero() {
+			// Calculate endDate equal to startDate plus defaultBlockTime*p.ProcessInfo.BlockCount
+			endDate = startDate.Add(types.DefaultBlockTimeSeconds * time.Duration(p.ProcessInfo.BlockCount))
+		}
+
 		// Create and store process in the indexer database
 		procParams := indexerdb.CreateProcessParams{
 			ID:                nonNullBytes(p.ProcessInfo.ID),
 			EntityID:          nonNullBytes(p.ProcessInfo.EntityID),
-			StartBlock:        int64(height),
-			EndBlock:          int64(height + 1),
-			BlockCount:        int64(1),
+			StartBlock:        int64(p.ProcessInfo.StartBlock),
+			StartDate:         startDate,
+			EndBlock:          int64(p.ProcessInfo.EndBlock),
+			EndDate:           endDate,
+			BlockCount:        int64(p.ProcessInfo.BlockCount),
 			HaveResults:       p.ProcessInfo.HaveResults,
 			FinalResults:      p.ProcessInfo.FinalResults,
 			CensusRoot:        nonNullBytes(p.ProcessInfo.CensusRoot),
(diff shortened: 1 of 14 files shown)


vochain/indexer/archive.go
```diff
@@ -61,17 +61,33 @@ func (idx *Indexer) ImportArchive(archive []*ArchiveProcess) ([]*ArchiveProcess,
...
+			}
+		}
+		endDate := p.ProcessInfo.EndDate
+		if endDate.IsZero() {
+			// Calculate endDate equal to startDate plus defaultBlockTime*p.ProcessInfo.BlockCount
+			endDate = startDate.Add(defaultBlockTime * time.Duration(p.ProcessInfo.BlockCount))
 		}
+
 		// Create and store process in the indexer database
 		procParams := indexerdb.CreateProcessParams{
```
altergui, line 82: why are we not storing the ManuallyEnded bool anywhere?

altergui: ah, now i see, ManuallyEnded is calculated based on `pi.EndBlock < pi.StartBlock+pi.BlockCount` , sorry for the noise

vochain/indexer/archive.go
```diff
@@ -61,17 +61,33 @@ func (idx *Indexer) ImportArchive(archive []*ArchiveProcess) ([]*ArchiveProcess,
...
 			ID:                nonNullBytes(p.ProcessInfo.ID),
 			EntityID:          nonNullBytes(p.ProcessInfo.EntityID),
-			StartBlock:        int64(height),
-			EndBlock:          int64(height + 1),
-			BlockCount:        int64(1),
+			StartBlock:        int64(p.ProcessInfo.StartBlock),
+			StartDate:         startDate,
+			EndBlock:          int64(p.ProcessInfo.EndBlock),
+			EndDate:           endDate,
```
altergui, line 88: to me it looks somewhat confusing, prone to error (or redundant?), that we're storing both Block and Date... blocks already have a very definite date

p4u: The whole point for this PR is to be able to store processes from previous (dead) chains. This is the `archive`.
So we don't have a way to estimate the start/endblock on those processes.

vochain/indexer/archive.go
```diff
@@ -85,13 +101,16 @@ func (idx *Indexer) ImportArchive(archive []*ArchiveProcess) ([]*ArchiveProcess,
...
 			Metadata:          p.ProcessInfo.Metadata,
 			ResultsVotes:      indexertypes.EncodeJSON(p.Results.Votes),
 			VoteCount:         int64(p.ProcessInfo.VoteCount),
+			ChainID:           p.ChainID,
+			FromArchive:       true,
 		}
+
 		if _, err := queries.CreateProcess(context.TODO(), procParams); err != nil {
 			return nil, fmt.Errorf("create archive process: %w", err)
 		}
```
altergui, line 115: you could seize the opportunity to fix the typo in the next func name `StartArchiveRetrival` -> `StartArchiveRetrieval`

vochain/vochaininfo/vochaininfo.go
```diff
@@ -14,6 +14,10 @@ import (
 	"go.vocdoni.io/dvote/vochain/state"
 )
 
+const (
+	defaultBlockTime = 10000 // 10 seconds
```
altergui: why is this different from the other defaultBlockTime (12s)?

vochain/vochaininfo/vochaininfo.go
```diff
@@ -68,14 +72,17 @@ func (vi *VochainInfo) updateCounters() {
 	mempoolSize.Set(uint64(vi.vnode.MempoolSize()))
 }
 
-// Height returns the current number of blocks of the blockchain
+// Height returns the current number of blocks of the blockchain.
```
altergui, line 72: ???? :rofl: if you really want to lint this, make it on a separate commit linting the thousand comments with missing dot

vochain/vochaininfo/vochaininfo.go
```diff
@@ -68,14 +72,17 @@ func (vi *VochainInfo) updateCounters() {
...
 func (vi *VochainInfo) Height() uint64 {
 	return height.Get()
 }
 
 // BlockTimes returns the average block time in milliseconds for 1, 10, 60, 360 and 1440 minutes.
-// Value 0 means there is not yet an average
+// Value 0 means there is not yet an average.
 func (vi *VochainInfo) BlockTimes() *[5]uint64 {
+	if vi.vnode.IsSynchronizing() {
+		return &[5]uint64{defaultBlockTime, 0, 0, 0, 0}
```
altergui: i'd avoid changing BlockTimes, which is called in many places. 
in fact there's already a fallback in HeightTime and EstimateBlockHeight, so i pushed a commit with a better fix

altergui (approved):
in general LGTM so approving with the usual faith, although i'm a bit lost on why are we storing Date and Block, and ImportArchive is duplicating the code for calculating startDate and endDate, instead of using one of the already existing func HeightTime

vochain/vochaininfo/vochaininfo.go
p4u: Actually we should unify this into type types package

p4u:
> in general LGTM so approving with the usual faith, although i'm a bit lost on why are we storing Date and Block, and ImportArchive is duplicating the code for calculating startDate and endDate, instead of using one of the already existing func HeightTime

We don't have the height time from a Chain which is not the existing one, @altergui, because we don't have the blocks neither the headers. What is confusing you about this?

I.e in this scenario, our current chain could be at block 1000 but the startBlock of the archive process is 500000. We don't want to return startDate=future while the process is actually finished and was closed months ago. We need to have the exact startDate at which the process was created on the previous chain.

vochain/vochaininfo/vochaininfo.go
p4u: Umm, but if the chain is synchronizing the blocktimes will have completely absurd values (such as 0.005). So having a fallback for such scenario that returns a kind of correct value, makes sense IMHO.

PR #1186, opened 2023-11-08
fix: not allow to send tokens with "to" and "from" address equals

mariajdab:
If we transfer from an account to the exact same account the cost is charged but the balance increase, for example, if an account have a balance of `1000` and send 89 tokens to the own same account with a cost of 10, the final balance will be 1079. This fix will prevent send token when address "from" and "to" are the same, in order to avoid the problem that.

Also the genesis dev was updated

### DIFF
--- a/vochain/genesis/genesis.go
+++ b/vochain/genesis/genesis.go
@@ -41,8 +41,8 @@ var Genesis = map[string]Vochain{
 }
 
 var devGenesis = Doc{
-	GenesisTime: time.Date(2023, time.October, 31, 1, 0, 0, 0, time.UTC),
-	ChainID:     "vocdoni/DEV/28",
+	GenesisTime: time.Date(2023, time.November, 8, 1, 0, 0, 0, time.UTC),
+	ChainID:     "vocdoni/DEV/29",
 	ConsensusParams: &ConsensusParams{
 		Block: BlockParams{
 			MaxBytes: 2097152,
--- a/vochain/transaction/tokens_tx.go
+++ b/vochain/transaction/tokens_tx.go
@@ -1,6 +1,7 @@
 package transaction
 
 import (
+	"bytes"
 	"encoding/binary"
 	"fmt"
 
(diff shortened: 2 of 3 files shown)


vochain/genesis/genesis.go
```diff
@@ -122,8 +122,8 @@ var devGenesis = Doc{
 }
 
 var stageGenesis = Doc{
-	GenesisTime: time.Date(2023, time.November, 7, 1, 0, 0, 0, time.UTC),
-	ChainID:     "vocdoni/STAGE/9",
+	GenesisTime: time.Date(2023, time.November, 8, 1, 0, 0, 0, time.UTC),
+	ChainID:     "vocdoni/STAGE/10",
```
p4u: I think you want to modify DEV, not Stage.

vochain/transaction/tokens_tx.go
```diff
@@ -30,6 +30,10 @@ func (t *TransactionHandler) SendTokensTxCheck(vtx *vochaintx.Tx) error {
 	if len(tx.To) == 0 {
 		return fmt.Errorf("invalid to address")
 	}
+	if string(tx.From) == string(tx.To) {
```
p4u: Please use bytes.Equl()

p4u (changes requested):

vochain/transaction/tokens_tx.go
```diff
@@ -30,6 +31,10 @@ func (t *TransactionHandler) SendTokensTxCheck(vtx *vochaintx.Tx) error {
 	if len(tx.To) == 0 {
 		return fmt.Errorf("invalid to address")
 	}
+	if bytes.Equal(tx.From, tx.To) {
```
p4u: Actually I think the check should go to line 54 and compare txFromAddress.Bytes() and txToAddress.Bytes().

This is because an attacker could add extra bytes on tx.From or tx.To, which would make the current comparison not match. However, when the bytes are converted to Ethereum Addresses, the bytes are cropped, thus the resulting addres could be the same.

p4u (changes requested):

vochain/transaction/tokens_tx.go
mariajdab: Makes sense, I will add the change

p4u (approved):

Issue #1179, opened 2023-11-06
tempSIK end2end test does not check if the siks are deleted


lucasmenendez:
**Describe the feature**
The vochain must remove temporal SIK's created for a voting process when it is ended and the results are calculated.

**Proposal**
I think that will be enough with just wait for the results and then use `HTTPclient.ValidSIK()` method with every voter account address to ensure that it is removed.

PR #1193, opened 2023-11-15
end2endtest: check siks are deleted when use TempSIKs

mariajdab:
For the `anonelectionTempSIKsValidate` test, at the end of the election the half of the accounts should not have a valid SIK, due they are using temporal SIK, and the other half of accounts should have it.

#1179

### DIFF
--- a/cmd/end2endtest/zkweighted.go
+++ b/cmd/end2endtest/zkweighted.go
@@ -4,6 +4,9 @@ import (
 	"fmt"
 	"math/big"
 	"os"
+	"strings"
+	"sync"
+	"sync/atomic"
 	"time"
 
 	vapi "go.vocdoni.io/dvote/api"
@@ -113,6 +116,7 @@ func (t *E2EAnonElectionTempSIKs) Setup(api *apiclient.HTTPclient, c *config) er
 	}
 	ed.VoteType = vapi.VoteType{MaxVoteOverwrites: 1}
 	ed.Census = vapi.CensusTypeDescription{Type: vapi.CensusTypeZKWeighted}
+	// use temporal siks
 	ed.TempSIKs = true
 
 	if err := t.setupElection(ed, t.config.nvotes); err != nil {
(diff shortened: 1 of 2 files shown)


cmd/end2endtest/zkweighted.go
```diff
@@ -163,8 +167,58 @@ func (t *E2EAnonElectionTempSIKs) Run() error {
...
 	return nil
 }
+
+// removedSiks get the number of accounts without a valid SIK
+func removedSiks(api *apiclient.HTTPclient, votes []*apiclient.VoteData) (accts int32, err error) {
+	wg := &sync.WaitGroup{}
+	errorChan := make(chan error)
+	var siksRemoved int32
+
+	for _, v := range votes {
```
altergui, line 195: mmh this spawns `len(votes)` concurrent goroutines, i'm worried about a memory spike. why not limit the number of parallel processes?

cmd/end2endtest/zkweighted.go
```diff
@@ -163,8 +167,64 @@ func (t *E2EAnonElectionTempSIKs) Run() error {
...
+
+	// worker pool channel with a maximum of parallelCount workers
+	workerPool := make(chan struct{}, parallelCount)
+
+	for _, v := range votes {
+		workerPool <- struct{}{} // reserves a slot in the worker pool
+		wg.Add(1)
+		go func(v *apiclient.VoteData) {
+			defer wg.Done()
+			defer func() { <-workerPool }() // make available for another goroutine to use
```
altergui: i believe you should first do wg.Add(1), then reserve the slot, otherwise the wg.Wait could conceptually happen before you expect it to. in practice i think this will never happen, but if you reorder then AFAIU it can definitely never happen, conceptually

the order of defers is OK because the defer stack is LIFO, so the <-workerPool will happen before wg.Done (which is perfect)

cmd/end2endtest/zkweighted.go
```diff
@@ -163,8 +167,64 @@ func (t *E2EAnonElectionTempSIKs) Run() error {
 		return err
 	}
 
+	// the half of accounts should not have valid sik, due they are using temporal sik
+	halfAccts := len(votes) / 2
```
altergui: the name `halfAccts` sounds strange to me, as if it held accounts, but actually is just an int. it's used just in two places and the len() / 2 is very cheap so i'd just remove this var and just put `len(votes)/2` in both places where it's used, i think it's clearer

same with `accs`, in that case i'd call it `count` or `removedCount` (also in the returned args of the func)

altergui (approved):
i commented just minor details, really nitpicking, so LGTM

cmd/end2endtest/zkweighted.go
mariajdab: Sounds good, agree. I updated the code to reserve the slot after the wg.Add(1)

PR #1204, opened 2023-11-27
add endpoint for calculating the next electionId of an organization

### LINKED ISSUE #1203: API: Return the next election ID
POST /elections/id

JSON BODY must contain: 

 - organization: hex address of the wallet creating the process
 - census origin: kind of census used for the process (protobuf models, numeric)
 - envelope type: ballot properties configuration (protobuf models, numeric)
 
Then the backend needs to fetch the `processCount` for the Organization (from state).
And the Chain ID for the current network.

And compute the resulting electionID using the following package: https://github.com/vocdoni/vocdoni-node/blob/main/vochain/processid/process_id.go

jordipainan:
closes #1203

### DIFF
--- a/api/api_types.go
+++ b/api/api_types.go
@@ -270,6 +270,18 @@ type Validator struct {
 	Score            uint32         `json:"score"`
 }
 
+type NextElectionID struct {
+	OrganizationID types.HexBytes `json:"organizationId"`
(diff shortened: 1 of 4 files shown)


jordipainan:
@p4u  solved encoding, ready to review

p4u (approved):
Looks good, please open the interoperability issue with an example of Body and response.

PR #1208, opened 2023-11-30
vochain: improve blocktime estimation

p4u:
Use blocktimes from the block headers. This provides a more accurate way to estimate election times.

### DIFF
--- a/vochain/vochaininfo/vochaininfo.go
+++ b/vochain/vochaininfo/vochaininfo.go
@@ -18,13 +18,9 @@ import (
 // VochainInfo stores some metrics and information regarding the Vochain Blockchain
 // Avg1/10/60/360 are the block time average for 1 minute, 10 minutes, 1 hour and 6 hours
 type VochainInfo struct {
-	sync           bool
 	votesPerMinute uint64
-	avg1           uint64
-	avg10          uint64
-	avg60          uint64
-	avg360         uint64
-	avg1440        uint64
+	blockTimes     [5]uint64
+	blocksMinute   float64
 	vnode          *vochain.BaseApplication
 	close          chan bool
 	lock           sync.RWMutex
@@ -67,6 +63,8 @@ func (vi *VochainInfo) updateCounters() {
 
 	voteCacheSize.Set(uint64(vi.vnode.State.CacheSize()))
 	mempoolSize.Set(uint64(vi.vnode.MempoolSize()))
+	blockPeriodMinute.Set(vi.BlockTimes()[0])
+	blocksSyncLastMinute.Set(uint64(vi.BlocksLastMinute()))
 }
 
 // Height returns the current number of blocks of the blockchain.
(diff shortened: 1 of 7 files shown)


vochain/vochaininfo/vochaininfo.go
```diff
@@ -67,22 +63,58 @@ func (vi *VochainInfo) updateCounters() {
...
 
 // Height returns the current number of blocks of the blockchain.
 func (vi *VochainInfo) Height() uint64 {
 	return height.Get()
 }
 
+// averageBlockTime calculates the average block time for the last intervalBlocks blocks.
+// The timestamp information is taken from the block headers.
+func (vi *VochainInfo) averageBlockTime(intervalBlocks int64) float64 {
+	currentHeight := int64(vi.vnode.Height())
```
jordipainan, line 81:
```suggestion
        if intervalBlocks == 0 {
                return 0  // avoid later division by 0
        }
        currentHeight := int64(vi.vnode.Height())
```

vochain/vochaininfo/vochaininfo.go
```diff
@@ -67,22 +63,58 @@ func (vi *VochainInfo) updateCounters() {
...
+	// Calculate the time frame in seconds
+	timeFrameSeconds := currentTime.Sub(*startTime).Seconds()
+
+	// Adjust the average block time based on the actual time frame
+	return timeFrameSeconds / float64(intervalBlocks)
+}
+
+func (vi *VochainInfo) updateBlockTimes() {
+	vi.blockTimes = [5]uint64{
+		uint64(1000 * vi.averageBlockTime(60/int64(types.DefaultBlockTime.Seconds())*1)),
```
jordipainan, line 107: can `types.DefaultBlockTime.Seconds()` return 0  at some point ? i.e setting 0 to comet config for using default values ? If so, check 0 division

p4u: It cannot, this is a constant.

vochain/vochaininfo/vochaininfo.go
```diff
@@ -234,86 +266,49 @@ func (vi *VochainInfo) NPeers() int {
...
+func (vi *VochainInfo) BlocksLastMinute() float64 {
+	vi.lock.RLock()
+	defer vi.lock.RUnlock()
+	return vi.blocksMinute
+}
+
 // Start initializes the Vochain statistics recollection.
-// TODO: use time.Duration instead of int64
 func (vi *VochainInfo) Start(sleepSecs uint64) {
 	log.Infof("starting vochain info service every %d seconds", sleepSecs)
```
jordipainan, line 284:
```suggestion
        if (sleepSecs == 0) {
                sleepSecs = 1
        }
	log.Infof("starting vochain info service every %d seconds", sleepSecs)
```

jordipainan (commented):
lgtm, maybe we can add the extra checks suggested for 0 divisions.

vochain/vochaininfo/vochaininfo.go
p4u: Umm, I don't think sleepSecs=0 should be a valid value. It is ok if it panics cause this is wrong.

vochain/vochaininfo/vochaininfo.go
p4u: I don't think intervalBlocks=0 is a valid value, it is ok to panic

vochain/vochaininfo/vochaininfo.go
p4u: Maybe we can just set a panic("sleepSecs is zero") to better control the crash

jordipainan (approved):

PR #1217, opened 2023-12-12
vocone: fix dateToBlock estimation endpoint

altergui:
vocone: fix dateToBlock estimation endpoint

* vocinfo.BlockTimes() now returns [5]time.Duration
* during startup, endpoint /chain/info now returns BlockTimeTarget instead of `[0, 0, 0, 0, 0]`
* vochain/app now has BlockTimeTarget and SetBlockTimeTarget
* types.DefaultBlockTime fixed (10s instead of 12s)
* vi.EstimateBlockHeight and vi.HeightTime refactored and fixed
* vocone now keeps track of blockTimestamps

### DIFF
--- a/api/chain.go
+++ b/api/chain.go
@@ -300,9 +300,17 @@ func (a *API) chainInfoHandler(_ *apirest.APIdata, ctx *httprouter.HTTPContext) 
 		return err
 	}
 
+	var blockTimesInMs [5]uint64
+	for i, v := range a.vocinfo.BlockTimes() {
+		blockTimesInMs[i] = uint64(v.Milliseconds())
+	}
+	if blockTimesInMs[0] == 0 {
+		blockTimesInMs[0] = uint64(a.vocapp.BlockTimeTarget().Milliseconds())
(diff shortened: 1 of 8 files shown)


altergui:
this refactor went wild so i'm splitting this down into an easier-to-review vocone fix in #1218

config/config.go
```diff
@@ -122,8 +124,8 @@ type VochainCfg struct {
 	SkipPreviousOffchainData bool
 	// Enable Prometheus metrics from tendermint
 	TendermintMetrics bool
-	// Target block time in seconds (only for miners)
-	MinerTargetBlockTimeSeconds int
+	// Target block time (only for miners)
+	MinerTargetBlockTime time.Duration
```
p4u: This change might be breaking. We should check node configurations out there that needs to be changed.

p4u (approved):

config/config.go
altergui: i've removed this change, and some others that were not essential to the fix.

PR #1218, opened 2023-12-13
vocone: fix dateToBlock estimation endpoint

altergui:
vocone: fix dateToBlock estimation endpoint

* vocinfo.BlockTimes() now returns [5]time.Duration
* during startup, endpoint /chain/info now returns BlockTimeTarget instead of `[0, 0, 0, 0, 0]`
* vochain/app now has BlockTimeTarget and SetBlockTimeTarget
* types.DefaultBlockTime fixed (10s instead of 12s)
* vi.EstimateBlockHeight and vi.HeightTime refactored and fixed
* vocone now keeps track of blockTimestamps

### DIFF
--- a/api/chain.go
+++ b/api/chain.go
@@ -300,9 +300,17 @@ func (a *API) chainInfoHandler(_ *apirest.APIdata, ctx *httprouter.HTTPContext) 
 		return err
 	}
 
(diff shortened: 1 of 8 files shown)


altergui:
closed as dup

PR #1220, opened 2023-12-13
upgrade cometbft to version v1.0.0-alpha.1

altergui:
```sh
go get github.com/cometbft/cometbft@v1.0.0-alpha.1
go get github.com/cometbft/cometbft/api
go get go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracehttp@v1.21.0
go get go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc@v1.21.0
go mod tidy
```

 * use os.Stat instead of tmos.FileExists

 * github.com/cometbft/cometbft/proto/tendermint/types -> github.com/cometbft/cometbft/api/cometbft/types/v1


altergui:
broke the build, WIP

altergui:
MVP is done, i.e. it builds and passes the tests. ready to be pushed to `dev`

i'm leaving a lot of tech debt tho, because i see a big refactor could be done, now that cometbft exposes an api (that we're mostly not using)

one example among many:
https://github.com/vocdoni/vocdoni-node/blob/1df4860044d00cf7a7e44e7cfc56c071c7313461/api/api_types.go#L334

here, this `tmtypes.Block` that currently comes from "github.com/cometbft/cometbft/types"

https://github.com/vocdoni/vocdoni-node/blob/1df4860044d00cf7a7e44e7cfc56c071c7313461/api/api_types.go#L7

should be replaced with `github.com/cometbft/cometbft/api/cometbft/types/v1`

altergui:
other cases are not so trivial, 

https://github.com/vocdoni/vocdoni-node/blob/be524658e09e054d2ecd67dc6d867c5851146db1/api/helpers.go#L11-L12

https://github.com/vocdoni/vocdoni-node/blob/be524658e09e054d2ecd67dc6d867c5851146db1/api/helpers.go#L50

i don't find ErrMempoolIsFull anywhere inside `cometbft/api`, maybe even the whole rpc interaction model is deprecated and we should reimplement it using api methods? (just a hunch, no idea really)

p4u:
> other cases are not so trivial,
> 
> https://github.com/vocdoni/vocdoni-node/blob/be524658e09e054d2ecd67dc6d867c5851146db1/api/helpers.go#L11-L12
> 
> https://github.com/vocdoni/vocdoni-node/blob/be524658e09e054d2ecd67dc6d867c5851146db1/api/helpers.go#L50
> 
> i don't find ErrMempoolIsFull anywhere inside `cometbft/api`, maybe even the whole rpc interaction model is deprecated and we should reimplement it using api methods? (just a hunch, no idea really)

https://github.com/cometbft/cometbft/blame/8f38d634822490949aa13df6d8136cde360c920e/mempool/errors.go#L27C6-L27C22

altergui:
yes the error is still there, in `cometbft/mempool` but that's outside `cometbft/api` so maybe there's now an "api way" of doing this, that we have yet to discover.

anyway as i said, in any case this would just be tech debt: the PR works perfectly and is ready to merge as-is

altergui:
so, i double-checked upstream docs and code and what we're doing it's fine, we'll continue using `rpc/client/local` since:

> Local is a Client implementation that directly executes the RPC
functions on a given node, without going through HTTP.

> This implementation is useful for:

>   - Running tests against a node in-process without the overhead
    of going through an HTTP server
>   - Communication between an ABCI app and CometBFT when they
    are compiled in process.

i thought maybe we were supposed to use ONLY things inside `cometbft/api/`, but that's not the case, instead, SOME things that we used were moved to `cometbft/api` and that's ok, since i updated the needed paths.

so this PR is ready to approve and merge

p4u (approved):

PR #1225, opened 2023-12-15
avoid code duplication with pflag and viper

mvdan:

### DIFF
--- a/cmd/node/main.go
+++ b/cmd/node/main.go
@@ -616,42 +616,40 @@ func main() {
 	//
 	// Gateway API and RPC
 	//
-	if conf.Mode == types.ModeGateway {
+	if conf.Mode == types.ModeGateway && conf.EnableAPI {
 		// HTTP API REST service
-		if conf.EnableAPI {
-			log.Info("enabling API")
-			uAPI, err := urlapi.NewAPI(srv.Router, "/v2", conf.DataDir, conf.Vochain.DBType)
-			if err != nil {
-				log.Fatal(err)
-			}
-			uAPI.Attach(
-				srv.App,
-				srv.Stats,
-				srv.Indexer,
-				srv.Storage,
-				srv.CensusDB,
-			)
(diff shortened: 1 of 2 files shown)


p4u:
I think this should be reviewed by @altergui

p4u (approved):

cmd/node/main.go
```diff
@@ -172,152 +229,28 @@ func newConfig() (*config.Config, config.Error) {
...
-	if err = viper.BindPFlag("logErrorFile", flag.Lookup("logErrorFile")); err != nil {
-		log.Fatalf("failed to bind logErrorFile flag to viper: %v", err)
-	}
-	if err = viper.BindPFlag("logOutput", flag.Lookup("logOutput")); err != nil {
-		log.Fatalf("failed to bind logOutput flag to viper: %v", err)
-	}
-	if err = viper.BindPFlag("saveConfig", flag.Lookup("saveConfig")); err != nil {
-		log.Fatalf("failed to bind saveConfig flag to viper: %v", err)
-	}
-	if err = viper.BindPFlag("signingKey", flag.Lookup("signingKey")); err != nil {
```
p4u, line 227: So this actually removes the need to bind a flag with a viper config option (that we use mainly for ENV vars and yaml config file)? I don't really understand how it works but it that's the case is a very nice improvement.

mvdan: We still bind, but now effectively in a loop rather than one by one. See these bits of code:

https://github.com/vocdoni/vocdoni-node/pull/1225/files#diff-41988d47676978b9c3637b55fba6588aed77dab7cff1aadfb41b119964785392R232

https://github.com/vocdoni/vocdoni-node/pull/1225/files#diff-41988d47676978b9c3637b55fba6588aed77dab7cff1aadfb41b119964785392R74-R105

The old code bound e.g. the `vochainPeers` flag as `vochain.Peers` on viper for the config. We now do this via the CutPrefix string handling.

PR #1233, opened 2024-01-08
don't export vochaininfo metrics during blocksync

altergui:


p4u:
Shall we still merge this? @altergui

altergui:
sure, it avoid future noise in grafana

p4u (approved):

altergui:
for some reason job_compose_test times out. i thought it was something with the CI server but after so many tries it still times out so it looks something related to this PR (even though i don't understand how). will review when i find some time

PR #1237, opened 2024-01-12
F/timestamp based elections

p4u:


apiclient/election.go
```diff
@@ -95,28 +95,18 @@ func (c *HTTPclient) NewElection(description *api.ElectionDescription) (types.He
...
-		}
-		info, err := c.ChainInfo() // get current Height
-		if err != nil {
-			return nil, err
-		}
-		blockCount = endBlock - info.Height
+	var startTime, duration uint32
+	if description.StartDate.IsZero() {
+		// if start date is empty, start the election immediately.
+		startTime = 0
```
jordipainan, line 106: `startTime` is already `0` at this point since it is initialized at line 102 and not modified. You can delete `startTime = 0`

cmd/end2endtest/helpers.go
```diff
@@ -770,3 +771,24 @@ func (t *e2eElection) registerAnonAccts(voterAccounts []*ethereum.SignKeys) erro
 	}
 	return nil
 }
+
+// logElection prints the election description in the log.
+func logElection(e *vapi.Election) {
+	log.Debugw("election description",
```
jordipainan, line 805: IMO the `election description` message is misleading as one can expect some actual description of the election.
I suggest to change it for `election data`. Change the comment accordingly too.

cmd/node/main.go
```diff
@@ -165,7 +165,7 @@ func loadConfig() *config.Config {
 		"enable IPFS group synchronization using the given secret key")
 	flag.StringSlice("ipfsConnectPeers", []string{},
 		"use custom ipfsconnect peers/bootnodes for accessing the DHT (comma-separated)")
-	flag.String("archiveURL", types.ArchiveURL, "enable archive retrival from the given IPNS url")
+	flag.String("archiveURL", types.ArchiveURL, "enable archive retrival from the given IPNS url (set \"none\" to disable)")
```
jordipainan, line 168: Why not `leave it unset to disable` ?

dockerfiles/testsuite/env.gateway0
```diff
@@ -12,3 +12,4 @@ VOCDONI_METRICS_ENABLED=True
 VOCDONI_METRICS_REFRESHINTERVAL=5
 VOCDONI_CHAIN=dev
 VOCDONI_SIGNINGKEY=e0f1412b86d6ca9f2b318f1d243ef50be23d315a2e6c1c3035bc72d44c8b2f90 # 0x88a499cEf9D1330111b41360173967c9C1bf703f
+VOCDONI_ARCHIVEURL=none
```
jordipainan, line 15: Cannot use an empty string ?

util/helpers.go
```diff
@@ -67,3 +68,8 @@ func BuildURL(parts ...string) (string, error) {
 	base.Path = fullPath
 	return base.String(), nil
 }
+
+// TimestampToTime converts a uint32 timestamp to a time.Time object.
+func TimestampToTime(timestamp uint32) time.Time {
```
jordipainan, line 73: Since `timestamp` will be passed at execution time, we need to check  for a possible overflow here. If a big enough value is passed it will be truncated and will produce an unexpected time.

vochain/indexer/db/models.go
```diff
@@ -25,11 +26,10 @@ type Block struct {
 type Process struct {
 	ID                 types.ProcessID
 	EntityID           types.EntityID
-	StartBlock         int64
-	EndBlock           int64
 	StartDate          time.Time
 	EndDate            time.Time
-	BlockCount         int64
+	ManuallyEnded      bool
+	BlockCount         sql.NullInt64
```
jordipainan: Do we really need to use this type ?

vochain/indexer/process.go
```diff
@@ -144,21 +144,14 @@ func (idx *Indexer) newEmptyProcess(pid []byte) error {
...
-	}
 
 	// Create and store process in the indexer database
 	procParams := indexerdb.CreateProcessParams{
 		ID:                pid,
 		EntityID:          nonNullBytes(eid),
-		StartBlock:        int64(p.StartBlock),
-		EndBlock:          int64(p.BlockCount + p.StartBlock),
-		BlockCount:        int64(p.BlockCount),
+		StartDate:         time.Unix(int64(p.StartTime), 0),
```
jordipainan, line 152: Why not using `TimestampToTime` function recently created ?
Same for all similar operations.

vochain/state/electionprice/electionprice.go
```diff
@@ -164,6 +167,12 @@ func (p *Calculator) Price(params *ElectionParameters) uint64 {
 		return 0
 	}
 
+	// If the election duration is specified in seconds, convert it to blocks
+	// This is a temporary solution until the block duration support is removed.
+	if params.ElectionDurationSeconds > 0 {
```
jordipainan, line 172: Is this still required?, asking because new created elections will be working with time and not blocks.

vochain/state/state.go
```diff
@@ -420,6 +430,32 @@ func (v *State) SetHeight(height uint32) {
 	v.currentHeight.Store(height)
 }
 
+// SetTimestamp sets the timestamp for the current (not committed) block.
+func (v *State) SetTimestamp(timestamp uint32) error {
```
jordipainan, line 434: We should probably check for uint32 overflows here too

jordipainan (commented):
Thank you for the PR! 

Changing from block calculation to timestamps is going make the vochain better and we will get rid of annoying block to time calculations that have been a pain in the past ^^ 

Left some comments :)

I also want to mention that seems that this PR does not only changes the way elections work (from blocks to timestamps), but it also includes changes on IST and other small things. Please include changes like the ones commented in other PRs in the future, that way the reviewers can focus on the scope of the PR and can be more effective.

cmd/node/main.go
p4u: Because if uset by flag but present on the YML file or somewhere else, it does not work.

dockerfiles/testsuite/env.gateway0
p4u: nop

vochain/indexer/db/models.go
p4u: You are probably seeing the diff changes from the feature branch to the main. This change is not from this PR.

vochain/state/electionprice/electionprice.go
p4u: We still calculate the price in blocks because the formula is based on blocks. Change to seconds would require changing the full formula, and I let this for a future PR (since it requires also simulations and so on).

util/helpers.go
p4u: time.Unix() is int64 on Golang, what do you mean by check possible overflow?

vochain/state/state.go
p4u: By overflow you mean calling SetTimestamp with a uint32 value higher than  2^32 -1 ? If that is what you mean, there is no way to check the overflow here, the uint32 will already have an "overflow" value at this point. I.e imagine I call `SetTimestamp(-1)` (underflow), the function will receive timestamp=2^32 -1 which is a valid uint32

vochain/indexer/db/models.go
jordipainan: Seems so, ignore then my last comment, regarding mixing changes. Sorry about that.

vochain/state/state.go
jordipainan: You're right. I was thinking about handling possible unwanted values, more of a design consideration.
The responsibility basically lies in ensuring that the function is used correctly.
My thoughts were on modifying the function to something like:

```golang
func TimestampToTime(timestamp uint32) (time.Time, error) {
    if timestamp < 0 || timestamp > math.MaxUint32 {
        // return error
    }
    return time.Unix(timestamp, 0), nil
}
```

But it does not make much sense, Thank  you for the clarifications.

p4u:
@mvdan I have fixed the conflicts with main. Now the only pending commit is the one that correctly adds the migrations.

PR #1241, opened 2024-01-24
keep archived process creation time, other tweaks

mvdan:

Fixes vocdoni/interoperability#130.


vochain/processarchive/processarchive.go
```diff
@@ -27,25 +27,17 @@ type ProcessArchive struct {
 	close      chan bool
 }
 
-type Process struct {
```
p4u, line 30: I think this will break backwards compatibility with the old archive processes. Ideally, we would skip the Process wrapper and just use the indexertypes.Process struct, however either we transform the old archive to the new structure or we keep this format for backwards compatibility. You can see the archive files that need to be imported at https://archive.vocdoni.io

I can provide a TAR file of the JSON archive files if you wish to make this parsing to the new format, then I can upload the new JSON files to the archive node (the one publishing the IPNS record). Let me know.

p4u:
here the archive containing the old elections that need migration
[archive.tar.gz](https://github.com/vocdoni/vocdoni-node/files/14049191/archive.tar.gz)

vochain/processarchive/processarchive.go
mvdan: This change is perfectly backwards compatible, as far as I can tell. The two types in the indexer and processarchive packages were identical, so deduplicating them has zero effect in terms of json encoding and decoding.

```
package indexer
[...]
type ArchiveProcess struct {
	ChainID     string                `json:"chainId,omitempty"` // Legacy
	ProcessInfo *indexertypes.Process `json:"process"`
	Results     *results.Results      `json:"results"`
	StartDate   *time.Time            `json:"startDate,omitempty"` // Legacy
	EndDate     *time.Time            `json:"endDate,omitempty"`   // Legacy
}
```
```
package processarchive
[...]
type Process struct {
	ChainID     string                `json:"chainId,omitempty"`
	ProcessInfo *indexertypes.Process `json:"process"`
	Results     *results.Results      `json:"results"`
	StartDate   *time.Time            `json:"startDate,omitempty"`
	EndDate     *time.Time            `json:"endDate,omitempty"`
}
```

p4u (approved):

PR #1242, opened 2024-01-28
vochain/indexer: minor speed-ups and tweaks

mvdan:


mvdan:
(the benchmark numbers jump up and down because I started being plugged in and later benchmarked on battery, so I had to get new numbers)

p4u (approved):

PR #1246, opened 2024-02-07
vochain: refactor proof verification logic

p4u:
Add a new interface that all proving systems must implement. Split the proving systems in different packages.
On zkProof validation, check for the missing inputs.

### DIFF
--- a/crypto/zk/circuit/inputs_test.go
+++ b/crypto/zk/circuit/inputs_test.go
@@ -33,7 +33,7 @@ func TestGenerateCircuitInput(t *testing.T) {
 	// mock the availableWeight
 	availableWeight := new(big.Int).SetUint64(10)
 	// calc vote hash
-	voteHash := util.BytesToArboStr(availableWeight.Bytes())
+	voteHash := util.BytesToArboSplitStr(availableWeight.Bytes())
 	// decode the test root
 	hexTestRoot, err := hex.DecodeString(testRoot)
 	c.Assert(err, qt.IsNil)
--- a/crypto/zk/prover/pubsignals.go
+++ b/crypto/zk/prover/pubsignals.go
@@ -1,0 +1,112 @@
+package prover
+
+import (
+	"math/big"
+
+	"go.vocdoni.io/dvote/crypto/zk/circuit"
+	"go.vocdoni.io/dvote/tree/arbo"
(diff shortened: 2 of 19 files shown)


crypto/zk/circuit/inputs_test.go
```diff
@@ -33,7 +33,7 @@ func TestGenerateCircuitInput(t *testing.T) {
 	// mock the availableWeight
 	availableWeight := new(big.Int).SetUint64(10)
 	// calc vote hash
-	voteHash := util.BytesToArboStr(availableWeight.Bytes())
+	voteHash := util.BytesToArboSplitStr(availableWeight.Bytes())
```
lucasmenendez, line 36: If you consider that it is helpful, add a TODO comment to update the voteHash.

crypto/zk/prover/pubsignals.go
lucasmenendez, line 1: It might be better to include some functions in this package to compare the received values with the publicsignals, without having to know how they are constructed. These functions could be helpful to compare e.g. the vote ID without calculating hashes or knowing which hash function to use. What do you think?

vochain/transaction/proofs/zkproof/zkproof.go
```diff
@@ -0,0 +1,83 @@
...
+	//hashedVotePackage := sha256.Sum256(envelope.VotePackage)
+	//proofVoteHash, err := proof.VoteHash()
+	//if err != nil {
+	//	return false, nil, fmt.Errorf("failed on parsing vote hash from public inputs provided: %w", err)
+	//}
+	//if !bytes.Equal(hashedVotePackage[:], proofVoteHash) {
+	//	return false, nil, fmt.Errorf("vote hash mismatch")
+	//}
+
+	// get vote weight from proof publicSignals
```
lucasmenendez, line 58: the weight is only returned if the proof is verified, so this can be moved after the verify check.

lucasmenendez (approved):
LGTM!

crypto/zk/prover/pubsignals.go
p4u: Yes, I thought the same. However, I'm not going to do it now because this refactor was not part of my initial task, and I need to move on... So if you like, open an issue and we tackle it in the future.

PR #1251, opened 2024-02-13
update to Go 1.22, update official github actions

mvdan:


p4u:
staticcheck fails @mvdan

mvdan:
Hm, it sounds like staticcheck isn't quite ready for 1.22.0 yet - left a comment in https://github.com/dominikh/go-tools/issues/1495#issuecomment-1943433182. Let's leave this PR open for a couple of days to see what happens.

mvdan:
I didn't want to wait much longer so now we build staticcheck from source, which is fine too. should be ready @p4u

PR #1235, opened 2024-01-09
vochain: implement StateSync

altergui:
this should use the snapshots introduced in
 https://github.com/vocdoni/vocdoni-node/pull/601
 
through cometbft state sync mechanism, superseding https://github.com/vocdoni/vocdoni-node/pull/604

PR #1260, opened 2024-03-04
all: ensure safety of slices on db.Iterate callbacks

altergui:
i included here some cosmetic changes (_ for unused vars, and using bytes.Clone instead of make()+copy()).

the real bugs are only 2 AFAIU, in ExportNoStateDB and FetchValidSIKRoots, but i'm not even sure about the FetchValidSIKRoots, if it's really a bug how didn't we notice this until now?

### DIFF
--- a/vochain/state/processblockregistry.go
+++ b/vochain/state/processblockregistry.go
@@ -1,6 +1,7 @@
 package state
 
 import (
+	"bytes"
 	"encoding/binary"
 )
 
@@ -52,7 +53,7 @@ func (pbr *ProcessBlockRegistry) MinStartBlock(fromBlock uint32) (uint32, error)
 func (pbr *ProcessBlockRegistry) MaxEndBlock(fromBlock uint32) (uint32, error) {
 	maxEndBlock := fromBlock
 	if err := pbr.db.Iterate(pbrDBPrefix, func(pid, _ []byte) bool {
-		p, err := pbr.state.Process(pid, false)
+		p, err := pbr.state.Process(bytes.Clone(pid), false)
 		if err != nil {
 			return false
 		}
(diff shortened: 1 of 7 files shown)


vochain/state/snapshot.go
```diff
@@ -447,7 +447,7 @@ type DBPair struct {
 func ExportNoStateDB(w io.Writer, reader *NoState) (uint32, error) {
 	pairs := []DBPair{}
 	err := reader.Iterate(nil, func(key []byte, value []byte) bool {
-		pairs = append(pairs, DBPair{Key: key, Value: value})
```
altergui, line 450: i came across this bug on this particular line while working on #1235 , i included this particular fix in that PR but i also checked other uses of db.Iterate so i made this into a separate PR for discussion

vochain/state/sik.go
```diff
@@ -149,7 +149,7 @@ func (v *State) ValidSIKRoots() [][]byte {
 func (v *State) FetchValidSIKRoots() error {
 	var validRoots [][]byte
 	if err := v.NoState(true).Iterate(sikDBPrefix, func(_, root []byte) bool {
-		validRoots = append(validRoots, root)
```
altergui, line 151: i believe this is bug? but why haven't we came across this before?

vochain/state/processblockregistry.go
```diff
@@ -52,7 +53,7 @@ func (pbr *ProcessBlockRegistry) MinStartBlock(fromBlock uint32) (uint32, error)
 func (pbr *ProcessBlockRegistry) MaxEndBlock(fromBlock uint32) (uint32, error) {
 	maxEndBlock := fromBlock
 	if err := pbr.db.Iterate(pbrDBPrefix, func(pid, _ []byte) bool {
-		p, err := pbr.state.Process(pid, false)
+		p, err := pbr.state.Process(bytes.Clone(pid), false)
```
altergui, line 56: AFAIU this is not REALLY needed, but it also doesn't hurt to be extra-safe, especially if Process method changes in the future.

p4u: Umm, the Clone is not necessary since this is a callback function and has its own context. Am I wrong @mvdan ?

vochain/state/sik.go
p4u: same here

p4u (approved):

PR #1264, opened 2024-03-05
keykeeper: do not schedule key add and reveal based on block

p4u:
Simplify keykeeper approach and do not use block to schedule actions (since now elections are based on timestamps).

Remove all storage layer wich is legacy and no longer necessary. Do not schedule, just send transactions when the event of new process or process status change, is executed.

Modify docker suit so it waits at least until height 8 to start. Else the keykeeper might not be yet ready.


vochain/appsetup.go
```diff
@@ -61,7 +66,9 @@ func (app *BaseApplication) SetDefaultMethods() {
 	})
 	app.SetFnMempoolPrune(app.fnMempoolRemoveTxTendermint)
 	app.SetFnSendTx(func(tx []byte) (*cometcoretypes.ResultBroadcastTx, error) {
-		result, err := app.NodeClient.BroadcastTxSync(context.Background(), tx)
+		ctx, cancel := context.WithTimeout(context.Background(), transactionSendTimeout)
+		defer cancel()
+		result, err := app.NodeClient.BroadcastTxSync(ctx, tx)
```
altergui, line 71: why is this change included in this commit? please split the changes of `vochain/appsetup.go` into a different commit, or if it's needed somehow for the keykeeper refactor, then add a comment maybe to transactionSendTimeout const? or at least to the commit message

dockerfiles/testsuite/env.miner2
```diff
@@ -1,6 +1,6 @@
 VOCDONI_DATADIR=/app/run
 VOCDONI_MODE=miner
-VOCDONI_LOGLEVEL=info
+VOCDONI_LOGLEVEL=debug
```
altergui: i'd leave both miners in info loglevel, 
else this clutters the CI log, when there's a failure

altergui (approved):
love the simplification <3

in general LGTM after reading the diff. didn't actually test it but YOLO :rocket:

vochain/appsetup.go
p4u: I have been stock on a deadlock for a while... Then I realized we do not have context on sendTx, which might be an issue. So I just added it.

PR #1267, opened 2024-03-07
statesync: populate parameters dynamically 

altergui:


p4u (approved):
I think DEV and STAGE should not have hardcoded checkpoints since these are development networks. But anyway, accept. We might change it in the future.

PR #1269, opened 2024-03-07
create a HTTPclient method to modify the requests timeout

lucasmenendez:


altergui (approved):
LGTM

apiclient/client.go
```diff
@@ -32,6 +32,8 @@ const (
 	// DefaultRetries this enables Request() to handle the situation where the server replies
 	// "mempool is full", it will wait for next block and retry sending the tx
 	DefaultRetries = 3
+	// DefaultTimeout is the default timeout for the HTTP client
+	DefaultTimeout = 10 * time.Second
```
altergui, line 36: where is this used?

lucasmenendez: ops, missing commit :P

apiclient/client.go
```diff
@@ -144,6 +146,11 @@ func (c *HTTPclient) SetRetries(n int) {
 	c.retries = n
 }
 
+func (c *HTTPclient) SetTimeout(d time.Duration) {
+	c.c.Timeout = d
+	c.c.Transport.(*http.Transport).ResponseHeaderTimeout = d
```
altergui: i have the feeling this can panic. better make an assertion and simply skip if it's not http.Transport

altergui:
```go
func (c *HTTPclient) SetTimeout(d time.Duration) {
	c.c.Timeout = d
	if transport, ok := c.c.Transport.(*http.Transport); ok {
		transport.ResponseHeaderTimeout = d
	}
}

```

altergui: i mean in our codebase it's always http.Transport. but it also doesn't hurt to avoid panics

PR #1272, opened 2024-03-07
ci: reduce testsuite block interval to 3 seconds

altergui:


p4u:
What's the motivation? I'm not sure if this solves something...

altergui:
the motivation was to see if it would reduce the time needed by compose_test in CI. it's hovering around 8mins on main :hankey: 

but right now it looks like the ci is overloaded and the times are wild, so i can't say if it helps or not

PR #1277, opened 2024-03-12
statesync: fetch the block before the snapshot, and use RPC instead of API

altergui:
so this is the weirdest hack ever.

turns out during statesync, cometbft needs the block exactly before the snapshot (i.e. if snapshot was done at height=100, then block 100) but a node bootstrapped via statesync will not have this block, because that block in particular is not saved to blockstore, it's just used to verify headers and then discarded. so BlockStore starts at height+1 (in my example above, at block 101)

so a node that was bootstrapped via statesync can't bootstrap another node. 
so i made this curious workaround to fetch that important block and save it to blockstore, and IT WORKS.
 but the code is scattered over weird places, mainly because i couldn't import apiclient from vochain, due to an import cycle. suggestions welcome.

### DIFF
--- a/apiclient/client.go
+++ b/apiclient/client.go
@@ -36,16 +36,6 @@ const (
 	DefaultTimeout = 10 * time.Second
 )
 
-// DefaultAPIUrls is a map of default API URLs for each network.
-var DefaultAPIUrls = map[string]string{
-	"dev":     "https://api-dev.vocdoni.net/v2/",
-	"develop": "https://api-dev.vocdoni.net/v2/",
-	"stg":     "https://api-stg.vocdoni.net/v2/",
-	"stage":   "https://api-stg.vocdoni.net/v2/",
-	"lts":     "https://api.vocdoni.io/v2/",
-	"prod":    "https://api.vocdoni.io/v2/",
-}
-
 // HTTPclient is the Vocdoni API HTTP client.
 type HTTPclient struct {
 	c       *http.Client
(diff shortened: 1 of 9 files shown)


apiclient/blockchain.go
```diff
@@ -45,7 +45,7 @@ func (c *HTTPclient) Block(height uint32) (*api.Block, error) {
 		return nil, fmt.Errorf("%s: %d (%s)", errCodeNot200, code, resp)
 	}
 	block := &api.Block{}
-	err = json.Unmarshal(resp, block)
+	err = json.Unmarshal(convertKeysFromCamel(resp), block)
```
altergui: Block method was buggy, the Unmarshal was just parsing correctly height, hash and version (i.e. all the fields that were not mangled by camelCase conversion)

altergui: nevermind, won't care about this

altergui:
this has been tested on dev and it works, please approve and merge without squashing

apiclient/client.go
```diff
@@ -36,16 +36,6 @@ const (
 	DefaultTimeout = 10 * time.Second
 )
 
-// DefaultAPIUrls is a map of default API URLs for each network.
```
p4u, line 39: Isn't this map used for the end2end test or something else?

p4u (approved):

apiclient/client.go
altergui: nope. i added this hack as part of my "statesync v2" and i'm happy to revert it TBH

PR #1281, opened 2024-03-17
vochain/indexer: move away from protojson

mvdan:
(see commit message)


p4u:
@altergui I understand this is superseded by #1282

altergui:
maybe. i cherrypicked over because i need some hacks to test this. but i guess i'd still review/approve mvdans work here, independently from my hacks

altergui (approved):
so i've tested this on dev and can confirm that the table contents are now deterministic, which is great.

the table sqlite_stat1 is still preventing the whole dump to be deterministic, so @mvdan will address that. maybe in a separate PR or in this one, in any case i'm approving this one for sure

PR #1282, opened 2024-03-18
feat/statesync deterministic

altergui:
- **vochain/indexer: move away from protojson**
- **abuse Format field to implement a simple version**


p4u:
Once this is merged, we need to coordinate a full re-sync (not snapshot) on the api server. Please @altergui confirm you are going to coordinate this task.

p4u (approved):

PR #1305, opened 2024-04-25
indexer: fix start_date and end_date (WIP)

altergui:
not intended to be merged, but instead to be applied as a one-shot on LTS


altergui:
done, applied, closing

PR #1311, opened 2024-05-07
api: support encrypted metadata

p4u:


altergui (approved):
LGTM

api/elections.go
```diff
@@ -449,21 +455,25 @@ func (a *API) electionCreateHandler(msg *apirest.APIdata, ctx *httprouter.HTTPCo
...
+			return "", false, err
 		}
 		if np := tx.GetNewProcess(); np != nil {
 			if p := np.GetProcess(); p != nil {
-				return p.GetMetadata(), nil
+				encryptedMeta := false
+				if p.GetMode() != nil {
+					encryptedMeta = p.GetMode().EncryptedMetaData
+				}
+				return p.GetMetadata(), encryptedMeta, nil
```
altergui, line 473: i'd simplify this to 

```go
			if p := np.GetProcess(); p != nil {
				if p.GetMode() != nil {
					 				return p.GetMetadata(), p.GetMode().EncryptedMetaData, nil
				}
				return p.GetMetadata(), false, nil

PR #1316, opened 2024-05-28
indexer/api: return envelope weight as string instead of JSON

p4u:
The indexer was doing EncodeJSON(weight) which is not needed, since weight is a big.Int and can be marshaled directly calling weight.String().

Until the database is not migrated to fix previous entries, we need to clean the backslashes and quote entries from the Weight at the api layer.


p4u:
Superseded by #1317

PR #1325, opened 2024-06-03
API Docs: Fix wrong API reponse types, add authorization header

NateWilliams2:


p4u:
@NateWilliams2 could you fix the issue with gofmt please?

NateWilliams2:
@p4u fixed with gofumpt, but I'm not sure why the go module check is failing. I ran go mod tidy and tried upgrading my go version.

PR #1329, opened 2024-06-10
indexer: index more details about blocks and transactions

### LINKED ISSUE #1321: Return correct date for old blocks
Either by indexing the block headers or by other methods, we should provide a correct output to calls such as:

`curl https://api.vocdoni.net/v2/chain/blockToDate/500000`

### LINKED ISSUE #1349: bug: cannot get firsts blocks information
**Describe the bug**

On STG I cannot retrieve the information of firsts blocks, for example 

```bash
➜ curl 'https://api-stg.vocdoni.net/v2/chain/blocks/1' -s | jq 
{
  "error": "block not found",
  "code": 4008
}
```

But yes for latests block

```bash
➜ curl 'https://api-stg.vocdoni.net/v2/chain/blocks/1119027' -s | jq
```

`Dev` seems working

### LINKED ISSUE #1358: feature: return tx subtype on tx summary
**Describe the feature**

Return tx subtype on a transaction summary on:

`chain/transactions/page/0`
`chain/blocks/1374876/transactions/page/0`

**Motivation**

When getting a list of txs on the explorer, a nice [feature requested](https://github.com/vocdoni/explorer/issues/78) is to show the tx subtype instead of the tx type.

**Constraints & Assumptions**

Maybe can be also implemented on fees endpoints  because they have tx info and the tx type. 

`/chain/fees/page/:page`
`/chain/fees/reference/:reference/page/:page`
`/chain/fees/type/:type/page/:page`
`/accounts/:accountID/fees/page/:page`

### LINKED ISSUE #1370: API: transactions return the account
The information returned by the transactions endpoint to get the details of a transaction, the API backend should add a new field "account" which points to the address signing the transaction.

Currently, we do return the signature, so we can extract the address.

Example with: https://api.vocdoni.io/v2/chain/transactions/2414592/0

```json
{
  "tx": {
    "collectFaucet": {
      "txType": "TX_UNKNOWN",
      "faucetPackage": {
        "payload": "CIeH3RkSFA09aBdi+e2pVB+ras+xfFACW2MLGGQ=",
        "signature": "0j0vNTSzJo7NBxDIt++asacf7bW15s6KtzOzHMd8E1M0IIBMVHBUg0AMdMPZhTMevSN2J7wN6gcYDI5TfvsogAE="
      },
      "nonce": 16
    }
  },
  "txInfo": {
    "transactionNumber": 33680,
    "transactionHash": "76057f3845cf9e34e3cc35c5c08d598b61f11d46d4845706bb67f7b3d2612444",
    "blockHeight": 2414592,
    "transactionIndex": 0,
    "transactionType": "collectFaucet"
  },
  "signature": "82bdce114c0399a051d69b9ada79d0ab63d2540bc90025b3fe09cc6d73a3ce4752f00f7e46c01d089327818d4ebf1fb96c692897fe50a7ded9525048c9c021fc1b",
  "account": "0x123456789..."
}
```

This way, the explorer (@selankon), on each transaction, can show and link the account creating it.

Consider that some transactions (vote) won't have this information.

altergui:
needed since we started pruning the blockstore

### DIFF
--- a/api/chain.go
+++ b/api/chain.go
@@ -13,12 +13,9 @@ import (
 	"go.vocdoni.io/dvote/crypto/zk/circuit"
 	"go.vocdoni.io/dvote/httprouter"
 	"go.vocdoni.io/dvote/httprouter/apirest"
-	"go.vocdoni.io/dvote/types"
 	"go.vocdoni.io/dvote/util"
-	"go.vocdoni.io/dvote/vochain"
 	"go.vocdoni.io/dvote/vochain/genesis"
 	"go.vocdoni.io/dvote/vochain/indexer"
-	"go.vocdoni.io/dvote/vochain/indexer/indexertypes"
 	"go.vocdoni.io/dvote/vochain/state"
 )
 
@@ -119,7 +116,7 @@ func (a *API) enableChainHandlers() error {
 		"/chain/transactions/{height}/{index}",
 		"GET",
 		apirest.MethodAccessTypePublic,
-		a.chainTxHandler,
+		a.chainTxByHeightAndIndexHandler,
 	); err != nil {
 		return err
 	}
@@ -136,6 +133,14 @@ func (a *API) enableChainHandlers() error {
 		"GET",
 		apirest.MethodAccessTypePublic,
 		a.chainTxListHandler,
+	); err != nil {
+		return err
+	}
+	if err := a.Endpoint.RegisterMethod(
+		"/chain/transactions/{hash}",
+		"GET",
+		apirest.MethodAccessTypePublic,
+		a.chainTxByHashHandler,
 	); err != nil {
 		return err
 	}
@@ -159,7 +164,7 @@ func (a *API) enableChainHandlers() error {
 		"/chain/blocks/{height}",
 		"GET",
 		apirest.MethodAccessTypePublic,
-		a.chainBlockHandler,
+		a.chainBlockByHeightHandler,
 	); err != nil {
 		return err
 	}
@@ -642,7 +647,7 @@ func (a *API) chainTxCostHandler(_ *apirest.APIdata, ctx *httprouter.HTTPContext
 //	@Success				204		"See [errors](vocdoni-api#errors) section"
 //	@Router					/chain/transactions/reference/{hash} [get]
 func (a *API) chainTxRefByHashHandler(_ *apirest.APIdata, ctx *httprouter.HTTPContext) error {
-	hash, err := hex.DecodeString(util.TrimHex(ctx.URLParam("hash")))
+	hash, err := hex.DecodeString(util.TrimHex(ctx.URLParam(ParamHash)))
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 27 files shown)


altergui:
@mvdan wdyt in general?

and in particular, TestRestoreBackupAndMigrate is failing, since i changed the db columns. what's the best fix? recreate the sqlite-backup-0009.sql.zst ?

i understand this would not happen if i avoided modifying 0006_create_table_blocks.sql and instead created a 0012_add_columns_to_table_blocks.sql , which i can also do.

api/helpers.go
```diff
@@ -51,6 +52,38 @@ func (a *API) sendTx(tx []byte) (*cometcoretypes.ResultBroadcastTx, error) {
 	return resp, nil
 }
 
+func (a *API) getBlockFromAppOrIndexer(height int64) (*Block, error) {
+	if tmblock := a.vocapp.GetBlockByHeight(height); tmblock != nil {
+		return &Block{
+			Block: comettypes.Block{
+				Header:     tmblock.Header,
+				Data:       tmblock.Data,
+				Evidence:   tmblock.Evidence,
```
p4u: Do we need evidence?

altergui: i haven't touched this, it's just the previous code, and i'd leave it as-is.

but better yet, i've simplified this and now just use tmblock altogether

altergui: https://github.com/vocdoni/vocdoni-node/compare/dbdd78ff508f48e60a450f43d5a5d139a50ef990..711f82e200237bdd406173d8d5874eb462b3a96a

mvdan:
Yes, please add a new migration step. As far as I can tell, and from what we agreed with @p4u, we are treating the database as stable and we should avoid any breaking changes that force us to delete it and re-index everything. In most cases, like adding new columns, it should be easy to do it as a new step anyway.

api/chain.go
```diff
@@ -422,9 +421,16 @@ func (a *API) chainEstimateDateHandler(_ *apirest.APIdata, ctx *httprouter.HTTPC
 		return err
 	}
 	timestamp := a.vocinfo.HeightTime(height)
+	if timestamp == nil {
```
mvdan: FWIW I would keep HeightTime to returning a non-pointer, and you can use the IsZero method to check if it's valid/present or not.

api/helpers.go
```diff
@@ -51,6 +52,33 @@ func (a *API) sendTx(tx []byte) (*cometcoretypes.ResultBroadcastTx, error) {
 	return resp, nil
 }
 
+func (a *API) getBlockFromAppOrIndexer(height int64) (*Block, error) {
```
mvdan: why is it useful to query either the app or the indexer? wouldn't it be enough to always just query one? if this is necessary at least I would document it

altergui: good point, maybe we could refactor this, but for now i just added a comment

altergui:
TestRestoreBackupAndMigrate is still failing, with the new migration 0013, complaining about a missing column. digging in :detective:

api/helpers.go
altergui: actually, in the end, i split up the crucial hotfix (#1339) and removed this "getBlockFromAppOrIndexer" completely

altergui:
> TestRestoreBackupAndMigrate is still failing, with the new migration 0013, complaining about a missing column. digging in 🕵️

this was failing due to my buggy migration 0013. fixed, all tests are passing now

p4u (approved):

PR #1331, opened 2024-06-11
vochain: allow set maxCensusSize during election

p4u:
Now the SET_PROCESS_CENSUS transaction allows a new field named "censusSize" which can be used to extend the current maxCensusSize of the elelction.

The cost for this transaction is the difference between the original cost and the new cost.


vochain/transaction/election_tx.go
```diff
@@ -301,3 +333,22 @@ func (t *TransactionHandler) txElectionCostFromProcess(process *models.Process)
 		MaxVoteOverwrite:        process.GetVoteOptions().MaxVoteOverwrites,
 	})
 }
+
+// txCostIncreaseCensusSize calculates the cost increase of a process based on the new census size.
+func (t *TransactionHandler) txCostIncreaseCensusSize(process *models.Process, newSize uint64) uint64 {
```
altergui, line 338: txCostIncreaseMaxCensusSize

vochain/transaction/election_tx.go
```diff
@@ -301,3 +333,22 @@ func (t *TransactionHandler) txElectionCostFromProcess(process *models.Process)
...
+	if newCost < oldCost {
+		log.Warnw("txCostIncreaseCensusSize: new cost is lower than the old cost", "oldCost", oldCost, "newCost", newCost)
+		cost, err := t.state.TxBaseCost(models.TxType_SET_PROCESS_CENSUS, false)
+		if err != nil {
+			log.Errorw(err, "txCostIncreaseCensusSize: cannot get transaction cost")
+			return 0
+		}
+		return cost
+	}
+	return newCost - oldCost
```
altergui: i believe you should also add TxBaseCost here.

altergui: i'd do this

```go
	cost, err := t.state.TxBaseCost(models.TxType_SET_PROCESS_CENSUS, false)
	if err != nil {
		log.Errorw(err, "txCostIncreaseCensusSize: cannot get transaction cost")
		return 0
	}

	costDiff := newCost - oldCost
	if costDiff < 0 {
		log.Warnw("txCostIncreaseCensusSize: new cost is lower than the old cost", "oldCost", oldCost, "newCost", newCost)
		costDiff = 0
	}

	return cost + costDiff

```

altergui (approved):
i'm suggesting changes but approving anyway

in general i feel some inconsistency with the name CensusSize vs MaxCensusSize, could we stick to MaxCensusSize?

vochain/transaction/election_tx.go
p4u: I think maxCensusSize naming was a bad decision, it has led to a lot of confusion.

In this case "censusSize" looks friendlier. I think we must s/maxCensusSize/censusSize but that would break compatibility. At some point we need to do a refactor on protobuf models naming, but not now.

So this is why I named it "censusSize", I know it is not consistent, but its the correct naming IMHO.

vochain/transaction/election_tx.go
p4u: looks good

altergui: ok will amend this and merge

p4u: Hahaha. Actually `SA4003: no value of type uint64 is less than 0 (staticcheck) go-golangci-lint`

PR #1337, opened 2024-06-13
all: cleanup code that supported forks

### LINKED ISSUE #1326: cleanup code that supported forks
for example this in beginBlock
https://github.com/vocdoni/vocdoni-node/blob/fce41728fd1cdedb4f9b1536ead460d6b6205da9/vochain/app.go#L292

and so on

altergui:
since we're using chain bumps now,
and the blockstore of LTS is pruned,
we don't need forks anymore.


altergui:
fixes #1326

p4u (approved):

PR #1338, opened 2024-06-13
indexer: backups are now a deterministic set of SQL statements

### LINKED ISSUE #1289: deterministic indexer backup so that all nodes produce the same bit-by-bit snapshot
right now the content of the indexer db is deterministic, but the backup is not since it's in binary format. to make it deterministic we should instead do something equivalent of a "sqlite3 .dump"
i.e. that the backup is a (compressed) bunch of SQL INSERT (etc) statements instead of native binary sqlite3 format

altergui:
this is much less performant that simply exporting/importing the raw binary database, but makes the backup deterministic, allowing StateSync to pull the snapshot from many nodes at the same time

### DIFF
--- a/vochain/indexer/indexer.go
+++ b/vochain/indexer/indexer.go
@@ -1,6 +1,7 @@
 package indexer
 
 import (
+	"bufio"
 	"bytes"
 	"context"
 	"database/sql"
@@ -18,6 +19,7 @@ import (
 	"sync"
 	"time"
 
+	"go.vocdoni.io/dvote/data/compressor"
 	"go.vocdoni.io/dvote/log"
 	"go.vocdoni.io/dvote/statedb"
 	"go.vocdoni.io/dvote/types"
@@ -31,6 +33,7 @@ import (
 
 	"github.com/pressly/goose/v3"
 
+	"github.com/schollz/sqlite3dump"
 	// modernc is a pure-Go version, but its errors have less useful info.
 	// We use mattn while developing and testing, and we can swap them later.
 	// _ "modernc.org/sqlite"
@@ -240,7 +243,8 @@ func (idx *Indexer) Close() error {
 	return nil
 }
 
-// BackupPath restores the database from a backup created via SaveBackup.
+// RestoreBackup restores the indexer by copying the file (raw binary format)
+// from the passed path.
 // Note that this must be called with ExpectBackupRestore set to true,
 // and before any indexing or queries happen.
 func (idx *Indexer) RestoreBackup(path string) error {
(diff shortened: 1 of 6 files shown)


vochain/indexer/indexer.go
```diff
@@ -249,37 +254,112 @@ func (idx *Indexer) RestoreBackup(path string) error {
...
+	defer db.Close()
+
+	scanner := bufio.NewScanner(r)
+	var statement strings.Builder
+	for scanner.Scan() {
+		line := scanner.Text()
+		statement.WriteString(line)
+		statement.WriteString("\n")
+
+		if strings.HasSuffix(line, ";") {
```
mvdan, line 322: are you sure that the entire text ends with `;`? what if it's like

```
foo;
bar;
baz
```

in that case you would ignore `baz`, so perhaps add a sanity check at the end that there's no remaining statement.

vochain/indexer/indexer.go
```diff
@@ -249,37 +254,112 @@ func (idx *Indexer) RestoreBackup(path string) error {
...
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+	defer cancel()
+
+	tmpDB, err := os.CreateTemp("", "indexer*.sqlite3")
+	if err != nil {
+		return fmt.Errorf("could not create tmpdb file: %w", err)
+	}
+	defer func() {
+		if err := os.Remove(tmpDB.Name()); err != nil {
+			log.Warnw("error removing tmpdb file", "path", tmpDB.Name(), "err", err)
```
mvdan: nit: don't log from a func that returns an error. you can return this error via a named error return.

vochain/indexer/indexer.go
```diff
@@ -249,37 +254,112 @@ func (idx *Indexer) RestoreBackup(path string) error {
...
+	if _, err := db.ExecContext(ctx, `DROP TABLE IF EXISTS sqlite_stat1;`); err != nil {
+		return fmt.Errorf("could not drop table sqlite_stat1: %w", err)
+	}
+
+	// make goose_db_version table deterministic
+	if _, err := db.ExecContext(ctx, `UPDATE goose_db_version SET tstamp = '1970-01-01 00:00:00';`); err != nil {
+		return fmt.Errorf("could not update goose_db_version: %w", err)
+	}
+
+	gzw := gzip.NewWriter(w)
```
mvdan: we already use zstd, so why not use it here as well? it's better than gzip in every way :)

mvdan (approved):
Thanks for doing this. I said I would, but haven't had the time. LGTM with some nits.

Did you consider a test? Just to make sure that we don't regress with any other non-determinism in SQL backups in the future.

Also, I wonder if we should rename the indexer sql "backup" feature to "snapshot", as the latter is perhaps more well understood to mean "deterministic".

vochain/indexer/indexer.go
p4u: yes, there is a compressor package already

p4u:
Could you explain why this is now deterministic? thanks

vochain/indexer/indexer.go
altergui: but it's a defer,
i'll do `defer os.Remove(tmpDB.Name())`

go-golangci-lint complains about
"Error return value of `os.Remove` is not checked (errcheck)"
but we have 1000s of such unchecked errors in defers anyway :sweat_smile:

vochain/indexer/indexer.go
altergui: nice! thanks both

compressor package didn't have NewReader and NewWriter so i added those wrappers

altergui:
it's deterministic because the SQL statements produced by each node are the same (bit-by-bit) . this was not the case with the raw db format.

PR #1339, opened 2024-06-17
api: /chain/blockToDate now fallbacks to fetching timestamp from indexer

### LINKED ISSUE #1321: Return correct date for old blocks

altergui:
hotfix cherrypicked from #1329

### DIFF
--- a/api/chain.go
+++ b/api/chain.go
@@ -422,6 +422,14 @@ func (a *API) chainEstimateDateHandler(_ *apirest.APIdata, ctx *httprouter.HTTPC
 		return err
 	}
 	timestamp := a.vocinfo.HeightTime(height)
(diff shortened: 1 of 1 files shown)


altergui:
before:
curl http://127.0.0.1:19090/v2/chain/blockToDate/2
{"date":"0001-01-01T00:00:00Z"}

after:
curl http://127.0.0.1:19090/v2/chain/blockToDate/2
{"date":"2023-10-24T14:10:55.831423637Z"}

p4u (approved):

PR #1344, opened 2024-06-19
vochain: add set_process_duration transaction

p4u:
With this transaction we can now modify the duration of an existing project (while in READY or PAUSE state).

If interruptible=false the duration can only be extended.

The cost of the transactions is computed dynamically by:
 max(nextPrice - previousPrice, txBaseCost)


altergui (approved):
LGTM

PR #1355, opened 2024-07-17
api: endpoint /chain/transactions/{height}/{index} now outputs fields as hex instead of base64

altergui:

PR #1356, opened 2024-07-22
Revert "api: endpoint /chain/transactions/{height}/{index} now outputs fields as hex instead of base64"

altergui:
Reverts vocdoni/vocdoni-node#1355

that PR just panics, shouldn't have been merged, sorry i didn't close it, @mvdan is working on a proper fix.

### DIFF
--- a/api/helpers.go
+++ b/api/helpers.go
@@ -6,7 +6,6 @@ import (
 	"errors"
 	"fmt"
 	"math/big"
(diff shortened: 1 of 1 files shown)


altergui:
fixed by https://github.com/vocdoni/vocdoni-node/commit/6caf6a54906c35ce7857fda88e467544e55af169

PR #1359, opened 2024-08-06
api: implement pagination and query params for /chain/transfers

altergui:
add a new endpoint, that includes `pagination` field in reply, and accepts QueryParams:
 * GET /chain/transfers
   * page
   * limit
   * accountId
   * accountIdFrom
   * accountIdTo

* mark all of these endpoints as deprecated on swagger docs:
  * /accounts/{accountId}/transfers/page/{page}
  * /accounts/{accountId}/fees/page/{page}

* api: add TransfersList struct

* indexer: replace GetTokenTransfersBy*Account methods with a new TokenTransfersList

### DIFF
--- a/test/apierror_test.go
+++ b/test/apierror_test.go
@@ -64,10 +64,6 @@ func TestAPIerror(t *testing.T) {
 			want: api.ErrFileSizeTooBig,
 		},
 		{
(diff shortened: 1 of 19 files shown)


elboletaire (approved):
LGTM but an additional review from a Go expert would be more than welcome.

test/apierror_test.go
elboletaire, line 1: 👍

PR #1371, opened 2024-08-27
feat/indexer drop tx id

altergui:
indexer: drop concept of transaction id, hash is now the primary key
    
     * api: remove endpoint /chain/transactions/reference/index/{index}


altergui:
this was included in #1329

PR #1373, opened 2024-08-28
api: endpoint /elections now filters by title and description (WIP)

altergui:


vochain/indexer/process.go
```diff
@@ -312,3 +318,29 @@ func (idx *Indexer) updateProcess(ctx context.Context, queries *indexerdb.Querie
...
+			log.Warnf("cannot get metadata from %s: %v", election.MetadataURL, err)
+		} else {
+			// if metadata exists and is not encrypted, add it to the indexed election
+			// if the metadata is not encrypted, unmarshal it, otherwise store it as bytes
+			if !election.ElectionMode.EncryptedMetaData {
+				electionMetadata := api.ElectionMetadata{}
+				if err := json.Unmarshal(metadataBytes, &electionMetadata); err != nil {
+					log.Warnf("cannot unmarshal metadata from %s: %v", election.MetadataURL, err)
+				}
+				election.Metadata = &electionMetadata
```
altergui, line 342: this doesn't work at all, doesn't even compile, it's just a copypaste from api/elections.go

vochain/offchaindatahandler/offchaindatahandler.go
```diff
@@ -26,6 +28,7 @@ type importItem struct {
 	itemType   int
 	uri        string
 	censusRoot string
+	pid        types.HexBytes
```
altergui, line 31: i added this so offchaindatahandler can keep track of which process metadata it is, and tell indexer to update that process

vochain/offchaindatahandler/offchaindatahandler.go
```diff
@@ -87,7 +92,7 @@ func (d *OffChainDataHandler) Commit(_ uint32) error {
 			go d.enqueueOffchainCensus(item.censusRoot, item.uri)
 		case itemTypeElectionMetadata, itemTypeAccountMetadata:
 			log.Infow("importing data", "type", itemTypesToString[item.itemType], "uri", item.uri)
-			go d.enqueueMetadata(item.uri)
+			go d.enqueueMetadata(item)
```
altergui, line 95: same here, i pass the whole item so i can have access to all the info about the item queued

vochain/offchaindatahandler/offchaindatahandler.go
```diff
@@ -42,6 +45,7 @@ var itemTypesToString = map[int]string{
 // offchain data (usually on IPFS).
 type OffChainDataHandler struct {
 	vochain       *vochain.BaseApplication
+	indexer       *indexer.Indexer
```
altergui, line 48: i don't like this, but i can't think of a better option right now to communicate offchaindatahandler and indexer

altergui: maybe a (cleaner?) alternative would be to add a new Event in EventListeners that is "MetadataReady" or something like this, that indexer can subscribe (and trigger the update in the db) and offchaindatahandler can trigger the event when the metadata is downloaded

PR #1377, opened 2024-08-30
feat/swagger openapi3

altergui:
ci: use swag v2.0.0 that now supports OpenAPI v3.1


altergui:
we can of course wait for the final version instead of rushing into the rc1

PR #1379, opened 2024-09-03
update to go 1.23.0

p4u:


p4u:
Any glue why this is happening @altergui?

altergui:
```
#16 0.063 go: go.mod requires go >= 1.23.0 (running go 1.22.6; GOTOOLCHAIN=local)
#16 ERROR: process "/bin/sh -c go mod download -x" did not complete successfully: exit code: 1
```
maybe something outdated at the runner? i'm taking a look

altergui:
ok so wait, actually the failing job is the scan-vuln-deps that we seldom maintain, i'd bet there's some action that needs to be bumped (happened to me in the past). but maybe we could simply drop that action altogether, due to lack of maintenance.

p4u:
I'd drop it @altergui

altergui (approved):

PR #1380, opened 2024-09-04
extend faucet package support for other transactions

p4u:
Faucet package can now be used on NewProcess, SetAccount, SetProcess and SetSik related transactions.


vochain/account_test.go
```diff
@@ -491,15 +496,49 @@ func TestSetAccountTx(t *testing.T) {
...
 	qt.Assert(t, signer1Account.InfoURI, qt.CmpEquals(), "")
+
+	// should work, try using a faucet package for a set account info tx
+	// first create the account (and check balance is 0, because tx cost 200)
+	faucetPkg, err = GenerateFaucetPackage(signers[0], signers[9].Address(), 200)
+	qt.Assert(t, err, qt.IsNil)
+	qt.Assert(t, testSetAccountTx(t,
+		signers[9], signers[9].Address(), faucetPkg, app, "", uint32(0), true),
+		qt.IsNil,
+	)
```
altergui, line 517: maybe it would be interesting to include as well 199, so the faucet would not produce enough balance to actually execute the transaction. this might trigger weird bugs, not now (due to the canPayForTx check :raised_hands:) but the test would catch future regressions

altergui: i added this test i mentioned, and also another corner case test, where none of the current balance OR the faucet are enough to cover the txcost, but combined they do.

altergui (approved):
LGTM, added a couple tests and a small refactor as `fixup` commits, please do `git rebase -i main` before merging

vochain/account_test.go
p4u: thanks

altergui:
found a small bug, fixing it

p4u:
ok, feel free to merge it yourself when ready

altergui:
i pushed the fix, i'm still extending and readapting the tests, will merge as soon as i finish

vochain/state/account.go
```diff
@@ -103,6 +103,20 @@ func (v *State) GetAccount(address common.Address, committed bool) (*Account, er
 	return &acc, acc.Unmarshal(raw)
 }
 
+// GetAccountBalance retrieves the Account.Balance for an address.
+// Returns 0 and no error if the account does not exist.
+// Committed is relative to the state on which the function is executed.
+func (v *State) GetAccountBalance(address common.Address, committed bool) (uint64, error) {
```
p4u: name it AccountBalance() please

PR #1383, opened 2024-09-09
feat/indexer hotfixes

altergui:
- **indexer: hotfix 0013_recreate_table_transactions**
- **indexer: optimize idx.BlockList performance**

PR #1389, opened 2024-09-12
indexer: optimize idx.VoteList, etc performance

altergui:
this is basically the same optimization done for BlockList (the most affected endpoint) https://github.com/vocdoni/vocdoni-node/pull/1383 but extended to all endpoints


altergui:
just in case, i've tested this with a new benchmark (BenchmarkElectionList) and confirmed a 10x improvement on that endpoint as well, no regressions

p4u (commented):
@altergui sorry for the late review, could you update the PR so we can merge it?

PR #1392, opened 2024-09-16
api: make census IPFS export async

p4u:
Add new method /censuses/export/ipfs/list to list all exported censuses (runtime).


altergui (approved):
LGTM, tested working

PR #1394, opened 2024-09-18
fix/downgrade go and ipfs

altergui:
- **Revert "update to go 1.23.0"**
- **Revert "go.mod: update ipfs/kubo to v0.29.0"**
- **Revert "all: use slices.Backward from Go 1.23"**


mvdan:
Do you have more details about this leak? Downgrading all of this is rather unfortunate.

PR #1395, opened 2024-09-20
support converting our rapidsnark proofs into gnark format (WIP)

altergui:
- **exploring how to convert our ProofToGnarkProof**
- **convertToGnarkProof seems to be working**


altergui:
ngmi, we took another approach

PR #1401, opened 2025-01-13
update cometBFT and retry on state-sync context fail

p4u:
If the statesync fails due a wrong or unaccessible RPC server, try again with the hardcoded list of seeds for the network.


altergui (approved):
LGTM

PR #1411, opened 2025-12-05
api: add support for custom voter weight when using CSP

altergui:


vochain/transaction/proofs/cspproof/cspproof.go
```diff
@@ -92,16 +89,13 @@ func (*ProofVerifierCSP) Verify(process *models.Process, envelope *models.VoteEn
...
-				envelope.ProcessId, rootPub.Bytes(), cspbundleDec.ProcessId, cspbundleDec.Address, signature.Bytes())
+				"pid %x with CSP key %x. CSP bundle {pid:%x, addr:%x, weight:%x}. CSP signature %x",
+				envelope.ProcessId, rootPub.Bytes(), p.Bundle.ProcessId, p.Bundle.Address, p.Bundle.VoteWeight,
+				signature.Bytes())
 		}
 	default:
 		return false, nil, fmt.Errorf("CSP proof %s type not supported", p.Type)
 	}
-	return true, bigOne, nil
+	return true, new(big.Int).SetBytes(p.Bundle.GetVoteWeight()), nil
```
p4u, line 103: For backgwards compatibility, Weight in the protobuf should be optional. And here we must check if GetVoteWeight() is not nil (then parse), else return bigOne.

p4u (changes requested):

vochain/transaction/proofs/cspproof/cspproof.go
altergui: great catch, addressed in 91b5147, please review

p4u (approved):

PR #1419, opened 2026-07-01
feat: support optional memo on votes

lucasmenendez:
Consumer side of the `VoteEnvelope.memo` feature. Depends on **vocdoni/dvote-protobuf#76** (`memo` as `bytes`), which follows #75 (merged).

## What

Adds an optional free-text **`memo`** on votes (max **256 bytes**), for an open "Other" answer. The memo is persisted in consensus state and hashed into the vote, so it is gated behind a **height-activated soft fork** — every validator enables it at the same block.

### Commits
1. `chore` pin `go.vocdoni.io/proto` to the released tag **v1.15.13** (memo as **bytes**).
2. `feat(vochain)` **consensus core** — fork gate (`genesis.VoteMemoActive`) + a soft-fork **activation runbook** doc-comment; `resolveVoteMemo` (gated UTF-8 + 256-byte validation); persistence + hashing in `state`.
3. `test(vochain)` fork gate, `resolveVoteMemo` (size/UTF-8/gating), hash determinism, StateDBVote persistence + empty-memo-absent.
4. `feat(api)` expose `memo` in `GET /votes/{voteId}` and let clients set it via `apiclient.VoteData`.

## Consensus safety

- **`bytes`, not `string`** (see dvote-protobuf#76): proto3 `string` enforces UTF-8 **at decode**, below the fork gate. During a rolling upgrade a vote with invalid-UTF-8 memo bytes would be accepted by nodes without the field (unknown bytes) and rejected by nodes with it (decode error) → **AppHash divergence / halt**, reachable by any crafted vote. `bytes` decodes identically everywhere; UTF-8/size are validated at the app layer in `resolveVoteMemo`, gated by the fork height (deterministic across upgraded nodes).
- **Safe rollout window** — before the activation height the memo is ignored (not stored, not hashed), so an upgraded node produces the same state as a pre-fork node. `Vote.Hash()` mixes in the memo only when non-empty; an empty memo is stored as absent (`nil`), keeping `StateDBVote` bytes identical to a pre-fork vote.
- **Backward compatible** — existing votes never carried a memo; their signatures (over the whole `Tx`) still verify. The memo rides inside the signed `Tx`.

## Activation runbook

See the header doc-comment of `vochain/genesis/forks.go`. Summary: pick height `H = current + margin` (margin > time for all validators **and** gateways to upgrade); set it in the fork table; release the image (watchtower auto-pulls); confirm ≥2/3 voting power upgraded before H; monitor H-1…H+1 for AppHash agreement. A non-upgraded node halts on apphash mismatch (fail-safe) rather than forking.

## Activation is deferred

This PR ships the feature **dormant on every real network** — `dev`/`stage`/`lts` are all `forkNever` in `vochain/genesis/forks.go`, so merging changes nothing on-chain (only `vocdoni/TEST/1` is active from genesis, for tests). Turning it on is deliberately decoupled from shipping the binary: a concrete activation height `H` will be scheduled in a **separate one-line commit** to `forks.go` after this merges, coordinated with the validator rollout per the runbook above.

## Notes
- `GET /votes/{voteId}` reads the memo from authoritative state (not indexed) — consistent with the existing state read for decryption keys. Not added to the paginated list endpoint.
- Pre-existing swagger CI check fails on `main` too (Go 1.26 vs pinned `swag`) — unrelated to this PR.


vochain/transaction/vote_tx.go
```diff
@@ -16,6 +18,30 @@ import (
 	"go.vocdoni.io/proto/build/go/models"
 )
 
+// MaxVoteMemoSize is the maximum allowed size, in bytes, of the optional
+// VoteEnvelope.memo free-text field.
+const MaxVoteMemoSize = 256
```
p4u: I think we have a better place to define these constants. I don't remember, but its probably in types package.

api/vote.go
```diff
@@ -129,6 +129,11 @@ func (a *API) getVoteHandler(_ *apirest.APIdata, ctx *httprouter.HTTPContext) er
 		Date:                 &voteData.Date,
 	}
 
+	// The memo is not indexed; read it from the authoritative state if present.
+	if sdbVote, err := a.vocapp.State.Vote(voteData.Meta.ProcessId, voteData.Meta.Nullifier, true); err == nil {
+		vote.Memo = string(sdbVote.GetMemo())
```
p4u: if memo is bytes, why casting it to string?

apiclient/vote.go
```diff
@@ -39,6 +39,11 @@ type VoteData struct {
 	Election   *api.Election
 	VoteWeight *big.Int
 
+	// Memo is an optional free-text note (max 256 bytes) attached to the vote,
+	// e.g. an open "Other" answer. Only honored by the chain once the memo
+	// soft-fork is active.
+	Memo string
```
p4u: Same here, memo should be []bytes, what's the point for declaring it string?

vochain/genesis/forks.go
```diff
@@ -0,0 +1,78 @@
+package genesis
```
p4u: What is this new file for? Actually I don't think we need to declare a soft-fork for the memo field, its not breaking. We just need to assure all validators apply the change before it is used in a transaction. I'd remove all this fork thing, unless there is a relevant motivation.

vochain/state/votes.go
```diff
@@ -34,6 +34,10 @@ type Vote struct {
 	Weight               *big.Int
 	VoterID              VoterID
 	Overwrites           uint32
+	// Memo is an optional free-text note attached by the voter (max 256 bytes).
+	// It is only set once the memo soft-fork is active for the chain; see
+	// genesis.VoteMemoActive and VoteTxCheck.
+	Memo string
```
p4u: same type mismatch

vochain/transaction/vote_tx.go
```diff
@@ -16,6 +18,30 @@ import (
...
+
+// resolveVoteMemo applies the memo soft-fork rules to a raw memo and returns the
```
p4u: we shouldn't talk about a softfork on vote_tx.go. This code should be agnostic to any specific chain related stuff

vochain/transaction/vote_tx.go
```diff
@@ -16,6 +18,30 @@ import (
...
+// value to store on the vote. The memo field is bytes (no proto decode-time
+// validation), so all validation happens here, at the application layer, where
+// the height gate makes it deterministic across upgraded nodes:
+//   - before activation (active=false) the memo is ignored entirely (returns ""),
+//     so an upgraded node produces the same state as a pre-fork node;
+//   - after activation it must be at most MaxVoteMemoSize bytes and valid UTF-8.
+func resolveVoteMemo(memo []byte, active bool) (string, error) {
```
p4u: is this function actually necessary? I don't think so. We are adding complexity and technical debt by incorporating this soft-fork related code into our core logic files.

vochain/transaction/vote_tx.go
```diff
@@ -143,6 +169,12 @@ func (t *TransactionHandler) VoteTxCheck(vtx *vochaintx.Tx, forCommit bool) (*vs
 		}
 	}
 
+	// [Soft-fork] Optional memo field. See resolveVoteMemo.
+	memoActive := genesis.VoteMemoActive(t.state.ChainID(), height)
```
p4u: not necessary

p4u (changes requested):

p4u:
Not related to this PR, but I let the note here. The dvote-protobuf CI is broken since last commit, please fix it.

lucasmenendez:
> Not related to this PR, but I let the note here. The dvote-protobuf CI is broken since last commit, please fix it.

Not, it is not broken. In fact, it is now fixed, the CI had a deprecated auth flow to publish new releases on npm and @elboletaire and me fixed it last week. It did not pass because we publish the last version manually, but the CI works.

apiclient/vote.go
lucasmenendez: Already answered

api/vote.go
lucasmenendez: Because originally the memo was a `string`, but we change it to `[]byte` because a possible caveat on protobuf layer, not related with go. Thats why it is a `string` at this level: https://github.com/vocdoni/dvote-protobuf/pull/76

vochain/genesis/forks.go
lucasmenendez: I remember several conversations where you mention that it need a softfork. So how can we assure that?

lucasmenendez:
Done — all five outstanding comments addressed in e3202cc2 (rebased onto the current head).

**Fork removed** (`forks.go:1`, `vote_tx.go:25`, `:32`, `:173`). Deleted `vochain/genesis/forks.go` and `forks_test.go`, dropped the `genesis` import and the height gate from `VoteTxCheck`. `resolveVoteMemo` only took an `active` flag because of the gate, so it becomes `validateVoteMemo([]byte) ([]byte, error)` — plain size and UTF-8 validation, no chain-specific concepts left in core tx code, which was your point on `:25`.

**`Memo` is now `[]byte`** (`votes.go:40`). You were right and my earlier answer about the API layer didn't cover this one: `vstate.Vote` is internal state, not JSON. It was the only `string` sitting between `VoteEnvelope.memo` and `StateDBVote.memo`, both `[]byte`, forcing a conversion at every use. `DeepCopy` now clones it, as it must for a slice.

**`MaxVoteMemoSize` moved to `types/consts.go`**, next to `MaxURLLength` and `KeyKeeperMaxKeyIndex`.

Net −95 lines. Build, vet, staticcheck and the full suite are green.

---

Recording the argument for the gate, since you asked what it was for and I'd rather it be on the record than lost.

`StateDBVote.memo` is field 8 of consensus state and `Vote.Hash()` mixes it in (`vochain/state/votes.go`), so a memo *does* move the AppHash. The difference from the CSP fork in #1425 is what triggers the new rule. There it's an operator creating an election, so "coordinate the rollout first" genuinely closes the window. Here it's **any voter attaching a memo** — the moment the binary is on some validators and not others, one memo vote is enough to diverge them, and no amount of operator coordination prevents a permissionless action.

What keeps this narrow is that an absent, nil or zero-length memo still marshals as an absent proto field and is skipped by `Vote.Hash()`, so every memo-less vote stays byte-identical. Tests cover the nil and zero-length cases explicitly. So the exposure is only a real memo cast during the mixed-version window.

Removed it as you asked — flagging it so the rollout is a deliberate call rather than an assumption: every validator wants the new binary before any memo vote is cast, and that isn't enforceable in-protocol once it ships.

vochain/state/votes.go
```diff
@@ -75,6 +80,11 @@ func (v *Vote) Hash() []byte {
 	h.Write(v.Nullifier)
 	h.Write(v.VotePackage)
 	h.Write(v.WeightBytes())
+	// Only mixed in when the voter attached one, so a vote without a memo hashes
+	// identically to one cast before this field existed.
+	if len(v.Memo) > 0 {
+		h.Write(v.Memo)
+	}
```
lucasmenendez, line 87: This makes sense for me, what do you think @p4u

lucasmenendez:
Human review done!

vochain/state/votes.go
p4u: Looks fine

vochain/state/votes.go
```diff
@@ -162,6 +173,15 @@ func (s *State) AddVote(vote *Vote) error {
 			*sdbVote.OverwriteCount = 1
 		}
 	}
+	// Only store the memo when present. proto3 `optional` marshals a set empty
+	// value as a present field, which would change the StateDBVote bytes (and thus
+	// the state hash) versus a vote cast without a memo; leaving it nil keeps them
+	// equal. The else branch also clears a memo carried by the overwritten vote.
+	if len(vote.Memo) > 0 {
```
p4u: The code can be simplified. Probably we can just assign it (without the if/else)

lucasmenendez:
Thanks @argos — worked through all twelve. Five fixed in 672f7705, one skipped by request, and six are against code that isn't in this PR.

## Fixed

**🟡 Memo unauthenticated on anonymous/Farcaster votes.** Confirmed, and it's broader than reported: the votePackage↔proof binding is *commented out* at `zkproof.go:48-58` ("Disabled for issues with the current circuit implementation"), so for anonymous votes the whole envelope is unbound, not just the memo. Farcaster's ed25519 signature covers `SignedFrameMessageBody`, not the envelope. Standard signed votes are fine — `InitializeSignedVote` receives `vtx.Signature`. Took the reject option: `memoAllowed(process)` returns false for `EnvelopeType.Anonymous` and `CensusOrigin_FARCASTER_FRAME`, checked outside the vote-cache branch so CheckTx and DeliverTx can't disagree.

**🟡 Per-row subtree open in `GET /votes`.** Fixed, though not as suggested — `State.IterateVotes` walks *every* vote in the process, so it would be worse than 100 targeted lookups. Added `State.Votes`, a batch sibling of `State.Vote` that opens the subtree once for a set of nullifiers; `votesList` now groups rows by processId, so an `electionId`-filtered page costs one open instead of 100. Worth noting the API passes `committed=true`, which skips `s.tx.Lock()` (`votes.go:242-247`), so this was wasted CPU rather than contention with block commit. Both call sites now log the read error instead of discarding it.

**🟡 Tautological memo tests.** Correct, and thanks — `Hash` accumulates into a `bytes.Buffer` where `Write(nil)` and `Write([]byte{})` are both no-ops, so neither assertion could fail. Dropped them. `TestAddVoteEmptyMemoNotStored` now asserts what its comment claimed, comparing the marshaled `StateDBVote` of a zero-length memo against one cast with no memo at all, rather than just `got.Memo == nil`. Kept the `len(v.Memo) > 0` guard: redundant for a Buffer, but it states the intent on a consensus hash and costs nothing.

**⚪ Stale `start_test.sh` comment** — dropped. **⚪ apiclient length check** — added, so an oversize memo fails before the round trip. No UTF-8 check there: the memo is deliberately opaque bytes now, length is the only constraint.

## Skipped

**🔴 Fork gate** — @p4u asked for the fork removal explicitly ([here](https://github.com/vocdoni/vocdoni-node/pull/1419#discussion_r3707931862)); e3202cc2 is that request. The AppHash reasoning is right and I recorded the residual risk [in this thread](https://github.com/vocdoni/vocdoni-node/pull/1419#issuecomment-5178335426) — the rollout has to reach every validator before any memo vote is cast, and that isn't enforceable in-protocol.

## Not in this PR

Findings 2-6 and 8 target `migrations/0017_indexed_metadata.sql`, `api/metadata_cache.go`, `vochain/indexer/indexer.go`, `vochain/indexer/transaction.go` and `test/api_test.go`. All five already exist on `main` and none appear in `gh pr diff 1419` — the review diffed against a stale `main`, which is also why the metadata-cache work looks "bundled into this PR". No opinion here on whether they're real; the `index_accounts_name` paging regression in particular looks worth someone's time, but against `main`, not this branch.

api/vote.go
```diff
@@ -264,6 +276,28 @@ func (a *API) votesList(params *VoteParams) (*VotesList, error) {
 			TransactionIndex: &vote.TxIndex,
 			BlockTime:        vote.BlockTime,
 		})
+		byProcess[string(vote.ProcessId)] = append(byProcess[string(vote.ProcessId)], len(list.Votes)-1)
+	}
+
+	// The memo is not indexed; read it from the authoritative state, the same way
+	// GET /votes/{voteId} does. State.Votes opens the votes subtree once per
+	// process, so a page filtered by electionId costs a single open.
+	for pid, rows := range byProcess {
```
p4u: Could we index votes? So we don't need this extra iteration.

api/api_types.go
```diff
@@ -233,6 +233,10 @@ type Vote struct {
 	BlockHeight      uint32          `json:"blockHeight,omitempty" extensions:"x-omitempty"`
 	TransactionIndex *int32          `json:"transactionIndex,omitempty" extensions:"x-omitempty"`
 	OverwriteCount   *uint32         `json:"overwriteCount,omitempty" extensions:"x-omitempty"`
+	// Optional opaque note attached by the voter, hex-encoded, at most
+	// types.MaxVoteMemoSize bytes. The chain stores and returns the bytes
+	// verbatim and does not interpret them; the encoding is up to the client.
+	Memo types.HexBytes `json:"memo,omitempty" extensions:"x-omitempty"`
```
p4u: I think HexBytes is not the correct type here. We use HexBytes for identifiers, not for data payloads. Using a base16 codification to encode data is quite an overhead in size. The correct type would be []bytes (so base64 as default JSON) or we can declare it as MemoString and encode it with UTF-8.

api/vote.go
```diff
@@ -129,6 +130,14 @@ func (a *API) getVoteHandler(_ *apirest.APIdata, ctx *httprouter.HTTPContext) er
 		Date:                 &voteData.Date,
 	}
 
+	// The memo is not indexed; read it from the authoritative state if present.
+	// A read failure must not be reported as "this vote has no memo", so log it.
+	if sdbVote, err := a.vocapp.State.Vote(voteData.Meta.ProcessId, voteData.Meta.Nullifier, true); err == nil {
+		vote.Memo = sdbVote.GetMemo()
```
p4u: Why not just assign it, without the if/else?

apiclient/vote.go
```diff
@@ -71,6 +76,16 @@ func (cl *HTTPclient) Vote(v *VoteData) (types.HexBytes, error) {
 	if err != nil {
 		return nil, err
 	}
+	// Attach the optional memo before the envelope is wrapped, marshaled and
+	// signed, so it is covered by the vote signature. Check the size here too:
+	// the node enforces it, but failing before the round trip gives the caller a
+	// usable error instead of an opaque HTTP one.
+	if len(v.Memo) > 0 {
+		if len(v.Memo) > types.MaxVoteMemoSize {
+			return nil, fmt.Errorf("vote memo exceeds max size of %d bytes", types.MaxVoteMemoSize)
```
p4u: This check is not necessary. This can never happend. If so, there is a huge bug in the code, so we should panic. Instead, I'd just not check the size here. It is responsability of the writing proces to ensure the size, not the reading process.

apiclient/vote.go
```diff
@@ -182,6 +197,23 @@ func (cl *HTTPclient) Vote(v *VoteData) (types.HexBytes, error) {
 	return voteAPI.VoteID, nil
 }
 
+// GetVote fetches a single vote (by nullifier / voteID) from the API,
+// including its optional memo.
```
p4u: No need to say this in the docs. Its implicit that we are returining all vote values (including memo).

vochain/state/votes.go
```diff
@@ -75,6 +80,11 @@ func (v *Vote) Hash() []byte {
 	h.Write(v.Nullifier)
 	h.Write(v.VotePackage)
 	h.Write(v.WeightBytes())
+	// Only mixed in when the voter attached one, so a vote without a memo hashes
+	// identically to one cast before this field existed.
+	if len(v.Memo) > 0 {
```
p4u, line 85: Neat: The correct way to check if a []byte value is empty, is by comparing it to Nil.

vochain/state/votes.go
```diff
@@ -248,6 +268,67 @@ func (s *State) Vote(processID, nullifier []byte, committed bool) (*models.State
...
+//
+// The result is positional and always has len(nullifiers) entries: entry i is nil
+// when that nullifier has no vote, or when its vote id could not be derived. Like
+// Vote, an unknown processID is not an error in itself — the subtree view is
+// created lazily and every lookup simply misses, yielding all-nil entries.
+//
+// When committed is false, the operation is executed also on not yet committed
+// data from the currently open StateDB transaction.
+// When committed is true, the operation is executed on the last committed version.
+func (s *State) Votes(processID []byte, nullifiers [][]byte, committed bool) ([]*models.StateDBVote, error) {
```
p4u: Why this new function? Don't we have already a function to return votes? I'm just wondering, since this looks unrelated to the PR objective.

vochain/transaction/vote_tx.go
```diff
@@ -16,6 +17,40 @@ import (
...
+// encoding a voter uses is a matter between them and whoever reads it back.
+//
+// An absent or empty memo returns nil rather than an empty slice: the stored
+// StateDBVote.memo is an optional field, and leaving it unset keeps the vote
+// byte-identical to one cast without a memo at all.
+func validateVoteMemo(memo []byte) ([]byte, error) {
+	if len(memo) == 0 {
+		return nil, nil
+	}
+	if len(memo) > types.MaxVoteMemoSize {
```
p4u, line 27: You just need this check here, the first is redundant.

vochain/transaction/vote_tx.go
```diff
@@ -143,6 +178,16 @@ func (t *TransactionHandler) VoteTxCheck(vtx *vochaintx.Tx, forCommit bool) (*vs
 		}
 	}
 
+	// Optional memo field. See validateVoteMemo and memoAllowed. Both run outside
+	// the vote-cache branch above, so CheckTx and DeliverTx reach the same verdict
+	// on the same envelope.
+	if len(voteEnvelope.GetMemo()) > 0 && !memoAllowed(process) {
+		return nil, fmt.Errorf("vote memo not supported for anonymous or farcaster votes")
+	}
+	if vote.Memo, err = validateVoteMemo(voteEnvelope.GetMemo()); err != nil {
```
p4u: These two checks are also redundant. Checkmemo() already checks the size.

p4u (changes requested):

vochain/state/votes.go
lucasmenendez: I think that it depends of the use case:
```go
a := []byte{}
cmp_nil := a == nil // false
cmp_len := len(a) == 0 // true
```

Compare with `nil` only works when the var is not initialized, if it is empty it does not work.

p4u: What do you mean by use case? The Zero value of a []byte variable is Nil, its not []byte{}. So if nothing is stored in Memo it should be nil. Unless there is a specific initialization to []byte{} somewhere (shouldn't be AFAIU).

lucasmenendez:
Thanks @p4u — all 9 addressed in 1747a857. Two I did differently than written; reasons below.

**`api/vote.go:285` / `votes.go:284` — index votes, drop `State.Votes`.** Done, and it turned out to be the right thread to pull. Migration 0019 adds `votes.memo`, `OnVote` writes it, and both endpoints are served from SQLite again — so `State.Votes`, both state read-backs and the `byProcess` grouping are all gone. One catch worth flagging: `CreateVote` is `REPLACE INTO`, which in SQLite is DELETE + INSERT, so `memo` had to go in the column list or it would silently revert to its default on every overwrite. `TestVoteMemoSurvivesReindex` covers it.

**`api_types.go:239` — HexBytes.** Agreed, changed to `[]byte` (base64). Went with bytes rather than a UTF-8 `MemoString` because the memo is deliberately opaque now — the chain only bounds its length.

**`api/vote.go:136` — the if/else.** Gone; the memo comes off the indexer row, so it's a plain assign.

**`apiclient/vote.go:85` — the size check.** Removed. Fair point about it being the writer's responsibility. For the record it was added last round at argos's request, so this is a deliberate revert, not an oversight.

**`apiclient/vote.go:201` — doc line.** Dropped.

**`vote_tx.go:32` + `votes.go:180`** — these two interact, so I did them together rather than literally. `AddVote` now plain-assigns as you asked. That is only safe because something upstream normalizes empty → nil: `memo` is a proto3 `optional`, where a *set empty* value marshals as **present**, so `[]byte{}` produces different `StateDBVote` bytes — and a different state hash — than a vote cast without a memo. Applying both suggestions as written removes every normalization and breaks that. So the `len(memo) == 0` early return stays, and it now carries a comment saying it is load-bearing. `TestAddVoteMemoNilContract` pins both halves, including that an empty slice really is distinguishable from nil.

**`vote_tx.go:187` — the two checks.** Consolidated into one `checkVoteMemo(memo, process)`, so the call site is a single check — I think that's what you were after with "Checkmemo() already checks the size". One correction though: `len(voteEnvelope.GetMemo()) > 0` was not a duplicate size check, it was a presence check. Without it, `!memoAllowed(process)` rejects **every** anonymous and Farcaster vote whether or not it carries a memo. In the consolidated function the length-zero return sits ahead of the gate for that reason.

---

Also folded in two memo-specific findings from argos's second review, since they were in this code:

- Neither gate had a test — deleting the rejection left the whole suite green. `transaction_zk_test.go` now asserts an anonymous vote cannot carry a memo.
- `memoAllowed` now also rejects **encrypted** elections. The API withholds the votePackage until the encryption keys are revealed; serving an unencrypted memo next to it in real time would defeat exactly that. Withholding it at the API instead would only be cosmetic — it sits in the clear in consensus state either way.

Verified each fix fails without its test: dropping `memo` from `CreateVote`, deleting the `memoAllowed` branch, and deleting the empty→nil return each turn a different test red. Build, vet, staticcheck, gofumpt and the full suite are green; `go generate` produces no diff.

One deployment note: votes indexed before migration 0019 read an empty memo until their indexer db is rebuilt. A migration can't backfill from state. Only DEV is affected.

p4u:
@argos stop doing reviews on this PR unless explicitly asked.

vochain/transaction/vote_tx.go
lucasmenendez: It is not redundant with the size check, it does a different job: it normalizes an empty memo to `nil` before storing. `VoteEnvelope.memo` is a proto3 `optional`, so a set-but-empty value marshals as *present* and would produce different `StateDBVote` bytes (and a different state hash) than a vote cast without a memo at all. It also has to run **before** the `memoAllowed` gate, so a vote carrying no memo stays valid on anonymous/encrypted/farcaster processes. The rationale is in the function doc comment.

lucasmenendez: Removed in d2687dc3. The emptiness now lives inside the gate condition (`len(memo) > 0 && !memoAllowed(process)`) so memo-less votes still pass on anonymous/encrypted/farcaster processes, and an unset memo just flows through as `nil` — no dedicated check.

p4u:
@lucasmenendez can you explain how the memo works for encrypted elections?

lucasmenendez:
> @lucasmenendez can you explain how the memo works for encrypted elections?

@p4u Encrypted elections does not support memo right now: https://github.com/vocdoni/vocdoni-node/blob/d2687dc3033844689177ad067c30d445e4244871/vochain/transaction/vote_tx.go#L54

p4u:
> > @lucasmenendez can you explain how the memo works for encrypted elections?
> 
> @p4u Encrypted elections does not support memo right now: https://github.com/vocdoni/vocdoni-node/blob/d2687dc3033844689177ad067c30d445e4244871/vochain/transaction/vote_tx.go#L54
> 

I think we should support them. Else the functionallity is very limited. Memo can be encrypted with the election key. Of course, that's up to the voter, since the blockchain cannot verify it. But either way I think its better to allow memo also on encrypted elections.

p4u:
> > > @lucasmenendez can you explain how the memo works for encrypted elections?
> > 
> > 
> > @p4u Encrypted elections does not support memo right now: https://github.com/vocdoni/vocdoni-node/blob/d2687dc3033844689177ad067c30d445e4244871/vochain/transaction/vote_tx.go#L54
> 
> I think we should support them. Else the functionallity is very limited. Memo can be encrypted with the election key. Of course, that's up to the voter, since the blockchain cannot verify it. But either way I think its better to allow memo also on encrypted elections.

Actually I think that we can simplify the logic and remove memoAllowed and the election configuration parameter. So we just allow ANY vote-tx to have a memo field. No need to check if its alowed or not. 

I see no point to restrict this. Is there a reason?

lucasmenendez:
> > > > @lucasmenendez can you explain how the memo works for encrypted elections?
> > > 
> > > 
> > > @p4u Encrypted elections does not support memo right now: https://github.com/vocdoni/vocdoni-node/blob/d2687dc3033844689177ad067c30d445e4244871/vochain/transaction/vote_tx.go#L54
> > 
> > 
> > I think we should support them. Else the functionallity is very limited. Memo can be encrypted with the election key. Of course, that's up to the voter, since the blockchain cannot verify it. But either way I think its better to allow memo also on encrypted elections.
> 
> Actually I think that we can simplify the logic and remove memoAllowed and the election configuration parameter. So we just allow ANY vote-tx to have a memo field. No need to check if its alowed or not.
> 
> I see no point to restrict this. Is there a reason?

I am agree, we need to bring support to memos in other type of elections but we can bring that support in a followup PR. We need this on prod as soon as possible to bring support to memos in other layers (saas, integrators, app).

p4u:
> > > > > @lucasmenendez can you explain how the memo works for encrypted elections?
> > > > 
> > > > 
> > > > @p4u Encrypted elections does not support memo right now: https://github.com/vocdoni/vocdoni-node/blob/d2687dc3033844689177ad067c30d445e4244871/vochain/transaction/vote_tx.go#L54
> > > 
> > > 
> > > I think we should support them. Else the functionallity is very limited. Memo can be encrypted with the election key. Of course, that's up to the voter, since the blockchain cannot verify it. But either way I think its better to allow memo also on encrypted elections.
> > 
> > 
> > Actually I think that we can simplify the logic and remove memoAllowed and the election configuration parameter. So we just allow ANY vote-tx to have a memo field. No need to check if its alowed or not.
> > I see no point to restrict this. Is there a reason?
> 
> I am agree, we need to bring support to memos in other type of elections but we can bring that support in a followup PR. We need this on prod as soon as possible to bring support to memos in other layers (saas, integrators, app).

No need for another PR. Just remove restrictions on memo. Memo should be allowed on every vote transaction. Its a code simplification and pushing consensus changes to production is always a complex and risky task. So let's do it now.

p4u:
I don't know if I'm explaining myself correctly. You don't need to implement anything new, you can even decide not to add Encrypted Memo in the application layer. But there is no point to restrict the usage of the memo field only for non-encrypted elections, on the blockchain layer. Memo is an arbitrary 256 Bytes payload that can be attached to any voteTx transaction and its up to the application layer how to use it (if it is encrypted with the election key, with a organization private key, etc. or just not encrypted). 

So from your side you only need to remove the restrictions. No need to more PR, no need to push new breaking changes in the future on this regard.

apiclient/vote.go
```diff
@@ -71,6 +76,11 @@ func (cl *HTTPclient) Vote(v *VoteData) (types.HexBytes, error) {
 	if err != nil {
 		return nil, err
 	}
+	// Attach the optional memo before the envelope is wrapped, marshaled and
+	// signed, so it is covered by the vote signature.
+	if len(v.Memo) > 0 {
```
p4u: why not just: `vote.Memo = v.Memo` ? If both values are bytes, that should be OK. No need for the if()

vochain/state/votes.go
p4u: this was not resolved not commented. 

`if v.Memo != nil { h.Write(v.Memo) }` is the idiomatic way to do so and it is the style we have been following on this repo. So please, unless there is a good reason, change it.

lucasmenendez:
> I don't know if I'm explaining myself correctly. You don't need to implement anything new, you can even decide not to add Encrypted Memo in the application layer. But there is no point to restrict the usage of the memo field only for non-encrypted elections, on the blockchain layer. Memo is an arbitrary 256 Bytes payload that can be attached to any voteTx transaction and its up to the application layer how to use it (if it is encrypted with the election key, with a organization private key, etc. or just not encrypted).
> 
> So from your side you only need to remove the restrictions. No need to more PR, no need to push new breaking changes in the future on this regard.

Understood, i think that now is done.

p4u (approved):

PR #1425, opened 2026-07-30
vochain: fix CSP salted-proof salt derivation (cross-election reuse + weight binding)

### LINKED ISSUE #1424: CSP salt crops ProcessID to 20 bytes → cross-election signature reuse; and cannot bind vote weight in the blind flow
## Summary

The CSP salted-key derivation salts with the **first 20 bytes of the ProcessID**. Because of the ProcessID byte layout, those 20 bytes exclude the nonce, so **all elections of the same organization on the same chain derive the same salt** — a CSP signature (`ECDSA_PIDSALTED` or `ECDSA_BLIND_PIDSALTED`) issued for one election verifies for every other election of that org. This defeats the salted-key guarantee ("no way to reuse a CSP signature for a different election process").

Separately, in the **blind** flow the vote weight is chosen by the voter inside the CA bundle and the chain trusts it verbatim, with no way for the CSP to bind the weight it authorized. Both are fixable with one change to how the salt is derived.

## Where

- `crypto/saltedkey/saltedkey.go` — `SaltSize = 20`; `SaltBlindPubKey`/`SaltECDSAPubKey` copy `salt[:20]`.
- `vochain/transaction/proofs/cspproof/cspproof.go` — `Verify` passes `envelope.ProcessId` (32B) to the salt fns; only the first 20B are used. Returns `Bundle.VoteWeight` verbatim.
- ProcessID layout (`vochain/processid/process_id.go`): `chainID[6] + orgAddr[20] + censusOrigin[1] + envType[1] + nonce[4]`. `[:20]` = `chainID[6] + orgAddr[:14]` → the nonce (bytes 28–31) and the type bytes are dropped.

```
processID (32B) = chainID[6] + organizationAddr[20] + censusOrigin[1] + envType[1] + nonce[4]
salt = ProcessId[:20] = chainID[6] + orgAddr[:14]   // nonce + type + orgAddr tail cropped off
```

## Impact

- **Cross-election reuse (high):** every process of one org+chain shares a salt. A voter authorized for election A can reuse the CSP signature on election B, C, … of the same org. Applies to both `ECDSA_PIDSALTED` and `ECDSA_BLIND_PIDSALTED`.
- **Unbindable weight (medium):** in the blind flow the CSP signs a message it cannot read, so it cannot constrain the weight; the chain reads `Bundle.VoteWeight` as-is. A relay-side attestation is bypassable (submit straight to the Vochain) and, being address-less, is a shareable bearer token. Weight can only be bound where the chain verifies — i.e. in the salted key.

## Proposed fix

1. **Hash before cropping.** Derive the salt from the full ProcessID: `salt = keccak256(ProcessId)[:20]`. All 32 bytes (incl. nonce) then contribute, so each election gets a distinct salt. Fixes the reuse bug for both salted proof types.
2. **Fold weight into the blind salt.** For the weighted anonymous case, derive `salt = keccak256(DOMAIN || ProcessId || weightBE8)[:20]`, where `weightBE8` is the canonical 8-byte big-endian `Bundle.VoteWeight`. The chain re-derives from the *declared* weight and verifies; a voter who declares a weight the CSP did not authorize signs under a different key and the proof fails. This binds weight on-chain, unbypassably, without ever linking weight to identity.
   - Canonical weight encoding is mandatory (fixed 8-byte BE), or a voter could re-encode the same integer to shift the salt.
   - Domain-separate the weighted derivation from the plain one so a weighted-blind signature can never be confused with a plain PIDSALTED signature.
3. **Collision note.** keccak256→20 bytes is 160-bit; birthday bound ≈ 2⁻⁸⁰, negligible. The risk being fixed is the *structured* crop (deterministic same-org collisions), not hash collisions.

## Compatibility / rollout

This is a **soft-fork**: it changes how the CSP salt is derived inside the existing `ECDSA_PIDSALTED` / `ECDSA_BLIND_PIDSALTED` verification, gated by an activation block height. **No new census origin or proof type is introduced** — the proof enums and census configuration are unchanged. Before the fork height nodes derive the salt as today (`ProcessId[:20]`); at and after it they derive `keccak256(ProcessId)[:20]` (plus the weight term for the blind path). Because the old rule is strictly a subset of validity, honest votes cast under the old derivation before the fork remain valid and new votes verify under the new one — a soft-fork, not a chain split. Every CSP signer must switch derivation at the same height: `vocdoni-node` (verify), `vocdoni/blind-csp` (reference signer), and `vocdoni/saas-backend` (`csp/sign.go` / `csp/sign_blind.go` `saltFromProcessID`).

## Acceptance

- Two elections of the same org+chain derive different salted keys after the fork height.
- A blind vote declaring a weight the CSP did not attest is rejected on-chain.
- The change activates by block height (soft-fork), with no new census origin or proof enum; elections created before the fork still verify.

---

Cross-reference vocdoni/saas-backend#617 (adds the blind flow that this enables).

lucasmenendez:
Fixes #1424. Targets `main`.

## Problem

The CSP salt was the raw processID, but the salting primitives consume only its first 20 bytes: `chainID[6] + orgAddr[:14]`. The nonce (bytes 28–31) never entered the salt, so **every election of the same organization on the same chain derived the same salted keypair**. In the blind flow the CSP signs a message it cannot read — a voter authorized for election A could get a bundle for sibling election B blind-signed, and it verified on B. Separately, `Bundle.VoteWeight` was read verbatim on-chain, with no way for the CSP to bind the weight it actually authorized.

## Fix

One derivation for both `ECDSA_PIDSALTED` and `ECDSA_BLIND_PIDSALTED` (`crypto/saltedkey.Salt`, the reference spec for the sibling repos):

```
w    = uint(bundle.voteWeight) big-endian, or 1 when absent
       rejected if w >= 2^160
salt = keccak256( processID || w-as-32-big-endian-bytes )[:20]
```

- All 32 processID bytes contribute → sibling elections derive distinct salts.
- The CSP-authorized weight is hashed in → declaring a weight the CSP did not authorize derives a different key and the proof fails, unbypassably, without linking weight to identity.
- The weight is rendered to a **fixed 32-byte width** so every byte encoding of an integer maps to one salt — the salt binds exactly the value the chain records.
- `SaltSize` stays 20, so the salting primitives and every deployed CSP are untouched; only the hash preimage changes.

### On the earlier additive derivation (thanks to the Argos review)

An earlier revision of this PR used `salt = keccak256(pid)[:20] + w mod 2^160`. That is **additively malleable** and was a critical break: the verifier re-derives the salt from the voter-chosen `w`, so a voter holding a blind signature valid for `saltA = keccak256(pidA)[:20] + 1` can present a bundle for election B declaring `w' = (saltA − keccak256(pidB)[:20]) mod 2^160`, making election B derive the identical salt — reusing the signature across elections and recording an arbitrary weight, with no crypto. Hashing both inputs removes the algebraic handle (hitting a target salt now needs a keccak preimage collision). Caught before any chain was scheduled and before the sibling repos mirrored it. Regression-tested by `TestCSPAdaptiveWeightForgeryRejected`, which fails against the additive rule and passes against the hashed one.

## Soft fork: **per-election**, not per-block-height

The gate lives in `vochain/genesis/forks.go` (`CSPSaltedProofV2Active`), keyed on `process.StartTime` per chainID rather than block height. The block-height gate the memo fork (#1419) uses **cannot** be applied to CSP proofs: a proof is verified at CheckTx and then **not** re-verified at DeliverTx when the vote is served from the vote cache (`VoteTxCheck`). A vote checked at `H-1` under the old rule and delivered at `H` under the new one would be accepted by the proposer and rejected by every node that never saw it in its mempool → **AppHash divergence / chain halt**.

Keying on immutable election data removes that class of bug: the predicate returns the same value at CheckTx and DeliverTx, on every node, regardless of height or mempool timing. Consequences:

- An election never changes rules mid-flight.
- Since votes with `currentTime < StartTime` are rejected, no vote governed by the new rule can be cast before the activation timestamp — the rollout window is **self-enforcing**.
- An external CSP evaluates the same predicate from the public `startDate` + `chainId`, with no chain polling.

`StartTime` is the anchor because it is always set, immutable after creation (`SetProcessDuration` moves the end, not the start), part of consensus state, and exposed off-chain. `StartBlock` (attacker-chosen for timestamp-based elections) and `ProcessBlockRegistry` (mutable on pause, deleted on end, outside the AppHash) were rejected.

### Activation runbook (full text in `forks.go`)

1. **Merge dormant.** All production chains ship at `forkNever` — merging changes nothing on-chain. Only `vocdoni/TEST/1` is active from genesis, for tests.
2. **Upgrade off-chain signers first.** Ship the mirrored derivation + the same `startDate >= T[chainId]` predicate in vocdoni/blind-csp, vocdoni/saas-backend (`csp/sign.go`, `csp/sign_blind.go`) and vocdoni/vocdoni-sdk (see vocdoni/vocdoni-sdk#435, vocdoni/integrator-sdk#18). Both sides must be deployable before either activates.
3. Pick `T` after the last node **and** CSP upgrade and the longest election lead time; set it for the ChainID in the table (one-line commit); release.
4. **Inverted failure mode vs. a height fork:** node divergence is impossible (the predicate is election-local), but a stale CSP after `T` issues proofs the chain rejects — voter-visible failures on one election, never a chain halt. Watch the signer, not the AppHash.
5. Abort before `T` by moving it later or to `forkNever`; after `T` it is committed consensus state.

## Tests

The salted and blind CSP paths previously had **zero** coverage. This PR adds:

- `test/testcommon/testcsp`: the signer counterpart of the verifier (all four `ProofCA` types, either derivation) — the repo had no CSP signer. Doubles as executable documentation for the sibling repos.
- `cspproof_test.go`: full matrix — every proof type × both sides of the fork; cross-election forgery and weight tampering demonstrated to succeed under the legacy derivation and rejected under the new one; **`TestCSPAdaptiveWeightForgeryRejected`** (the adaptive attack that solves for the weight after seeing the salt); overflow guard; canonical-weight; unsalted types byte-identical on both sides.
- `crypto/saltedkey`: determinism, nonce sensitivity (legacy collision pinned), fixed-width canonicalization, priv/pub round-trips, legacy truncation compatibility.
- Chain-level (`vochain/proof_test.go`): CheckTx→deliverTx→commit for both salted types on both chains, including the adaptive forgery, plus a no-cache delivery agreement test.
- e2e: new `cspelectionblindsalted` op (weighted blind pid-salted votes), wired into the testsuite; verified locally against voconed (5 votes × weight 10 → results `[[50 0]]`).

Drive-by: `TestBaseApplicationWithChainID` now propagates its chainID to the app and state as a real node does (it previously only reached cometbft), and `SaltECDSAPubKey` no longer mutates its input key.

## Note for reviewers

The issue text (#1424) specifies block-height activation, two derivations with a domain tag, and 8-byte BE weight concatenation. All three were superseded during design: per-election activation (halt-safety) and a single hashed derivation. The issue should be updated to match.

> Note on base branch: the soft-fork machinery (`vochain/genesis/forks.go`) also exists on `dev` from the memo fork (#1419). Since this PR targets `main`, it introduces `forks.go` standalone; whichever of the two lands second on a shared branch resolves a trivial add/add on that file (keep both fork tables under one `forkNever`).

### DIFF
--- a/vochain/genesis/forks.go
+++ b/vochain/genesis/forks.go
@@ -1,0 +1,104 @@
+package genesis
+
+import "math"
+
+/*
+Soft-fork activation via per-election gating.
+
+A soft fork here is a state-affecting protocol change activated per ChainID, so
+every validator switches behavior on the same elections. It is lighter than the
+EndOfChain hard fork documented in genesis.go (no new ChainID / AppHash /
+blockstore reset): the chain and its ChainID continue across the fork.
+
+Why the gate is keyed on the election and not on the block height. Before the
+activation point an upgraded node must produce byte-for-byte the same state as a
+not-yet-upgraded node, so mixed binary versions during the rollout window cannot
+diverge. Height gating cannot give that here, because a vote proof is verified at
+CheckTx and then NOT re-verified at DeliverTx when it is served from the vote
+cache (see vochain/transaction.VoteTxCheck). A vote checked at H-1 under the old
+rule and delivered at H under the new one would be accepted by the proposer,
+which skips re-verification, and rejected by every node that never saw it in its
+mempool: an AppHash divergence, i.e. a chain halt.
+
+Keying on immutable election data removes that class of bug entirely. The
+predicate depends only on the ChainID and on fields fixed when the election was
+created, so it returns the same value at CheckTx and at DeliverTx, on every node,
+regardless of height, wall clock or mempool timing. An election never changes
+rules mid-flight, and an off-chain signer (a CSP) can evaluate the same predicate
+from public election data without polling the chain height.
+
+### Runbook: activating a per-election soft fork on a network
+
+  1. Merge with every production chain set to forkNever. Behavior is identical to
+     the previous release, so the binary can be deployed at any time with zero
+     divergence risk. This decouples "deploy the binary" from "activate the rule".
+  2. Ship the mirrored change in every off-chain component that must agree on the
+     rule (for the CSP fork below: vocdoni/blind-csp, vocdoni/saas-backend,
+     vocdoni/vocdoni-sdk), behind the same predicate, with the activation point
+     still unset. Both sides must be deployable before either activates.
+  3. Confirm out-of-band that every validator AND gateway runs the new image;
+     there is no in-protocol version negotiation. Watchtower auto-pulls (~30s).
+  4. Pick the activation point comfortably after the last node upgrade, the last
+     off-chain upgrade, and the longest lead time with which that network's
+     organizers create elections ahead of their start date.
+  5. Set it for the target ChainID in the fork table below (replace forkNever);
+     one-line commit. Release the binary: merge to the network's release branch
+     (e.g. release-lts-1); CI (docker-release.yml) pushes the image.
+  6. Monitor the first election that activates the rule. Note the risk profile is
+     inverted versus a height fork: node divergence is impossible because the
+     predicate is election-local, but a stale off-chain signer issues proofs the
+     chain rejects. That surfaces as voter-visible failures on one election, not
+     as a chain halt. Watch the signer, not the AppHash.
+  7. Abort/rollback: BEFORE the activation point, ship a binary moving it later
+     (or to forkNever) to cancel. AFTER it, elections already running under the
+     new rule are committed consensus state and reverting needs a further
+     coordinated fork.
+  8. Removing the legacy branch: only once every election predating the fork has
+     ended on every chain in the table.
+
+Note there is one older, inline height-gated soft fork at
+vochain/transaction/transaction.go, guarding an ISTC reschedule on
+vocdoni/LTS/1.2. It is deliberately left where it is: it is slated for deletion
+once that chain forgets old blocks, not for migration into this table.
+*/
+
+// forkNever is used as an activation point for a soft-fork that has not been
+// scheduled yet on a given chain: the feature stays inactive for every election.
+const forkNever = uint32(math.MaxUint32)
+
+// cspSaltedProofV2Time holds, per ChainID, the unix timestamp from which an
+// election's StartTime activates the fixed CSP salted-proof derivation
+// (crypto/saltedkey.Salt). Elections that started before it keep the legacy
+// derivation, which cropped the processID to 20 bytes and so shared one salted
+// key across every election of an organization (issue #1424).
+//
+// StartTime is the right anchor because it is always set (NewProcessTxCheck
+// defaults it to the current block timestamp and rejects earlier values), it is
+// immutable after creation (SetProcessDuration moves the end, never the start),
+// it is part of the consensus state, and it is exposed off-chain as the
+// election's startDate so a CSP can evaluate the same predicate.
+//
+// It also makes the rollout window self-enforcing: votes with
+// currentTime < process.StartTime are rejected, so no vote governed by the new
+// rule can be cast before the activation timestamp in wall-clock terms.
+//
+// A value of 0 means active for every election. A ChainID absent from the map
+// (custom chains, vocone, and the "test" ChainID used by
+// vochain.TestBaseApplication) or set to forkNever never activates it.
+var cspSaltedProofV2Time = map[string]uint32{
+	"vocdoni/TEST/1":   0,         // always on for the testsuite chain
+	"vocdoni/DEV/36":   forkNever, // TODO: schedule
+	"vocdoni/STAGE/12": forkNever, // TODO: schedule
+	"vocdoni/LTS/1.2":  forkNever, // TODO: schedule (coordinated with the CSP operators)
+}
+
+// CSPSaltedProofV2Active reports whether the fixed salt derivation for
+// ECDSA_PIDSALTED and ECDSA_BLIND_PIDSALTED proofs is active for an election
+// with the given StartTime on the given chain.
+func CSPSaltedProofV2Active(chainID string, electionStartTime uint32) bool {
+	forkTime, ok := cspSaltedProofV2Time[chainID]
+	if !ok || forkTime == forkNever {
+		return false
+	}
+	return electionStartTime >= forkTime
+}
--- a/api/census_test.go
+++ b/api/census_test.go
@@ -15,10 +15,12 @@ import (
 	"google.golang.org/protobuf/proto"
 
 	"go.vocdoni.io/dvote/crypto/ethereum"
+	"go.vocdoni.io/dvote/crypto/saltedkey"
 	"go.vocdoni.io/dvote/data/ipfs"
 	"go.vocdoni.io/dvote/db"
 	"go.vocdoni.io/dvote/db/metadb"
 	"go.vocdoni.io/dvote/httprouter"
+	"go.vocdoni.io/dvote/test/testcommon/testcsp"
 	"go.vocdoni.io/dvote/test/testcommon/testutil"
 	"go.vocdoni.io/dvote/types"
 	"go.vocdoni.io/dvote/vochain"
@@ -221,6 +223,7 @@ func TestCensusProofVerifierArbo(t *testing.T) {
 	// verify the proof
 	electionID := rnd.RandomBytes(32)
 	valid, weight, err := transaction.VerifyProof(
+		app.ChainID(),
 		&models.Process{
 			ProcessId:    electionID,
 			CensusRoot:   censusData.CensusID,
@@ -332,6 +335,7 @@ func TestCensusZk(t *testing.T) {
 	// verify the proof
 	electionID := util.RandomBytes(32)
 	valid, newWeight, err := transaction.VerifyProof(
+		app.ChainID(),
 		&models.Process{
 			ProcessId:    electionID,
 			CensusRoot:   censusData.CensusID,
@@ -413,6 +417,7 @@ func TestCensusProofVerifierCSP(t *testing.T) {
 
 	// verify the proof
 	valid, weight, err := transaction.VerifyProof(
+		app.ChainID(),
 		process,
 		&models.VoteEnvelope{
 			Nonce:     util.RandomBytes(32),
@@ -488,6 +493,7 @@ func TestCensusProofVerifierCSPLegacy(t *testing.T) {
 
 	// verify the proof
 	valid, weight, err := transaction.VerifyProof(
+		app.ChainID(),
 		process,
 		&models.VoteEnvelope{
 			Nonce:     util.RandomBytes(32),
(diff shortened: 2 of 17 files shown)


p4u:
@argos review

lucasmenendez:
Thanks — both findings were spot on, fixed in 9b70be18.

**Critical (additive malleability).** Confirmed the adaptive forgery: with `salt = keccak256(pid)[:20] + w mod 2^160` the verifier re-derives the salt from the voter-chosen `w`, so a voter can solve `w' = (saltA − keccak256(pidB)[:20]) mod 2^160` and make election B land on the salt the CSP already blind-signed for election A — reusing the signature and recording ~2^159 as the weight, no crypto needed. Replaced the derivation with:

```
salt = keccak256( processID || w-as-32-big-endian-bytes )[:20]
```

Hashing both inputs removes the algebraic handle (hitting a target salt now needs a keccak preimage collision), and rendering the weight to a fixed 32-byte width closes the leading-zero encoding seam you noted, so the salt binds exactly the integer recorded at `cspproof.go`. `saltModulus` is gone; the `MaxVoteWeightBits` guard stays as a sane cap / `FillBytes` safety and is no longer load-bearing (doc updated accordingly). `SaltSize` stays 20, so the salting primitives and eventual sibling-repo mirrors are unaffected — only the preimage changed. Since no chain is scheduled yet, no second coordinated rollout.

**Major (vacuous negative tests).** Added `TestCSPAdaptiveWeightForgeryRejected` (both salted types), which computes the collision weight after seeing the salt and asserts rejection, plus a matching case at the CheckTx level in `TestCSPproofSalted`. Decisive check as you suggested: I temporarily reverted `Salt` to the additive body and confirmed the new test **fails** (the forgery verifies) — restored, and it passes against the hash. So it distinguishes a real binding from a malleable one rather than passing vacuously.

Also fixed the `Salt` doc block (dropped the "why addition" / "why the 2^160 guard is mandatory" rationale, added "why hash both inputs" and "why fixed-width") and the PR body's derivation formula. Full suite + gofumpt/vet/staticcheck green; the `cspelectionblindsalted` e2e still lands `[[50 0]]` against voconed.

p4u:
I fixed some CI stuff in main. Rebase and see if tests pass now @lucasmenendez

vochain/genesis/forks.go
```diff
@@ -0,0 +1,104 @@
+package genesis
```
p4u, line 1: Following the comments on #1419, I think this new forking system is over-engineering. I'd first investigate the current LTS chain, how many ECDSA_BLIND elections have been created and how long since the last one. If its not recent (as I believe), we don't need all these soft-fork logic. Nodes sync using state-sync and fast-sync is not longer available (since long ago), so there is no recheck of the ancient blocks anymore. That said, we can just apply the change and assure all (or the 2/3) of the validators apply it before a new election of this kind is created.

p4u (changes requested):

vochain/genesis/forks.go
lucasmenendez: Investigated as you asked, and the data points the other way, so I'd rather you see it before I remove anything.

**First: the affected type is `ECDSA_PIDSALTED`, not `ECDSA_BLIND`.** The change only touches the two *salted* types — `ECDSA` and `ECDSA_BLIND` derive no salt and are byte-identical before and after.

Method: 2604 LTS elections from `/v2/elections`, filtered to `censusOrigin == OFF_CHAIN_CA` by byte 26 of the processID; for each of the 225 that have votes, one vote → `/v2/chain/transactions/{hash}` → `tx.vote.proof.ca.type`.

| type | elections | first | last |
|---|---|---|---|
| `ECDSA_PIDSALTED` | **175** | 2025-03-14 | **2026-07-30** |
| `ECDSA_BLIND` (unaffected) | 21 | 2024-03-23 | 2026-05-12 |
| `ECDSA` (unaffected) | 6 | 2025-03-06 | 2025-03-06 |
| tx not retrievable (pre-2024) | 23 | — | — |

`ECDSA_BLIND_PIDSALTED`: **zero**. So your question answered literally — 21 blind elections, last one ~3 months ago — but the type this PR actually changes is the dominant CSP proof on LTS.

It is not tailing off: 48 CSP elections created in May, 39 in June, 42 in July. The most recent started 2026-07-30, and `6b342d99…030400000000` **is accepting votes right now**, ending 2026-08-05. On DEV they were created yesterday.

Weights are real too — sampled `voteWeight` on salted proofs: 121×`01`, 41 absent, then `02`, `03`, `04`, `0a`, `16`, `64`, `c8` (200). That is what the weight-binding half of the fix protects.

**You're right about the sync argument.** LTS `blockStoreBase` is 7,517,000 against height 8,166,826 — old blocks are gone and never re-verified, so keeping the legacy branch for *historical* proofs is unnecessary. That reason is dead and I'll drop the framing that leans on it.

**Two reasons survive, and neither applied to the memo:**

1. The memo is opt-in per voter and no client sends it yet, so the mixed-version window is empty in practice. Here *every* CSP vote goes through the changed derivation. During the rollout a legacy proof accepted by a non-upgraded proposer is rejected by an upgraded node at DeliverTx — `VoteTxCheck` returns early only `if fromCache`, so any node that never saw the tx in its mempool re-verifies — and that is an AppHash divergence.
2. The signers are off-chain (blind-csp, saas-backend, the SDK). With no scheduled activation point, "upgrade the nodes" and "upgrade every CSP" have to happen simultaneously, or CSP voting breaks on live elections.

Your condition — *assure the validators upgrade before a new election of this kind is created* — needs a moment with no salted CSP election in flight. There isn't one right now.

**What I'd propose instead of removing it:** you're right that 104 lines of fork framework is over-engineering. Cut `forks.go` to the map plus the predicate (~20 lines), drop the runbook doc-comment and the generic "fork system" framing, keep per-election activation. Every chain still ships `forkNever`, so merging changes nothing on-chain and scheduling stays a one-line commit once the CSPs are ready.

Happy to do that, or to remove it entirely if you'd rather coordinate a window with the CSP operators — I just didn't want to make that call on a premise the chain contradicts. Your call.

lucasmenendez:
Superseded by #1434 after discussing the approach.

The fix here changed the derivation of the deployed salted proof types **in place**, which forced a coordinated activation (the fork gate you flagged as over-engineering) because [`ECDSA_PIDSALTED` is in daily use on LTS with elections always in flight](https://github.com/vocdoni/vocdoni-node/pull/1425#discussion_r3713770285) — there is no safe window to flip the rule under existing elections.

#1434 dissolves the problem instead of gating it: `OFF_CHAIN_CA` stays byte-for-byte as today, permanently, and a new census origin `OFF_CHAIN_CA_V2` (vocdoni/dvote-protobuf#79) applies the fixed derivation to elections that opt in at creation. No fork table, no chainID threading, no activation runbook — the census origin itself, immutable in the processID, is the switch. The rollout rule reduces to the same one as the memo: all validators upgrade before the first V2 election is created, which is naturally satisfied since creating one requires SDK/CSP support that ships afterwards.

The derivation, the security analysis (including the adaptive-weight forgery found by review here), and the test matrix carry over unchanged. Branch `1424` is left for reference.

PR #1426, opened 2026-07-30
Indexer/vote activity

p4u:


vochain/indexer/queries/votes.sql
```diff
@@ -22,6 +22,30 @@ LIMIT 1;
...
+JOIN blocks AS b
+	ON v.block_height = b.height
+WHERE v.process_id = sqlc.arg(process_id)
+GROUP BY period
+ORDER BY period;
+
+-- name: HasVotesMissingBlockTime :one
+-- Reports whether any indexed vote references a block which is not indexed, and
+-- thus cannot be dated. Used once at startup as a completeness check: the scan
+-- stops at the first such vote, so it is cheap when the data is complete.
```
lucasmenendez: This comment is backwards. `EXISTS` short-circuits only when it **finds** a missing vote — on a healthy node (the normal case) it scans the entire `votes` table joined against `blocks`, at every startup, on the single-writer connection, inside `startDB` before the node serves anything.

Cheap equivalent using indexes that already exist (`index_votes_block_height_index`, and `blocks.height` is the rowid PK): compare `MIN(votes.block_height)` with `MIN(blocks.height)`. O(log n), and it catches the realistic case of an indexer db younger than the chain.

vochain/indexer/queries/votes.sql
```diff
@@ -22,6 +22,30 @@ LIMIT 1;
...
 SELECT COUNT(*) FROM votes;
 
+-- name: VoteActivity :many
+-- Aggregates the votes of a process into time buckets, using the block timestamp.
+-- The bucket_format argument is a strftime format string which truncates the
+-- timestamp to the desired granularity (e.g. '%Y-%m-%dT%H:00:00Z' for hourly).
+SELECT CAST(strftime(CAST(sqlc.arg(bucket_format) AS TEXT), b.time) AS TEXT) AS period,
+	COUNT(*) AS count
+FROM votes AS v
+JOIN blocks AS b
```
lucasmenendez: This inner `JOIN` is what makes the two comments above necessary. `GetVote`, twelve lines up in this same file, already uses `LEFT JOIN blocks`.

A `LEFT JOIN` here plus an `undatedVotes` count in the response would report the truth to clients and let `HasVotesMissingBlockTime` and the whole bootstrap-reindex path be deleted rather than fixed. As written, `TotalVotes` can silently disagree with `/elections/{id}/votes/count` and a client has no way to tell.

vochain/indexer/indexer.go
```diff
@@ -430,6 +458,20 @@ func (idx *Indexer) AfterSyncBootstrap(inTest bool) {
 	log.Infof("live results recovery computation finished, took %s", time.Since(startTime))
 }
 
+// bootstrapBlocks reindexes the blocks when startDB found the indexed blocks to be
+// incomplete, either because a migration changed them or because votes were indexed
+// whose block was not. It blocks until the reindex is done, so that once the
+// bootstrap returns, every indexed vote can be dated by its block.
+func (idx *Indexer) bootstrapBlocks(inTest bool) {
+	if !idx.blockReindexNeeded.Swap(false) {
```
lucasmenendez: This can't repair the case it exists for. `ReindexBlocks` only walks `BlockStore().Base()..Height()`, so a vote whose block is below the base (pruned or state-synced node) can never be dated. `HasVotesMissingBlockTime` then returns 1 on **every** restart, and each one kicks off a full block+tx reindex of the whole blockstore that fixes nothing. No cap, no backoff, no marker.

Suggestion: check the missing heights actually fall inside the blockstore range before scheduling, or persist an "already tried" marker.

vochain/indexer/indexer.go
```diff
@@ -326,20 +348,26 @@ func (idx *Indexer) blockTxQueries() *indexerdb.Queries {
...
-// if unecrypted). This method might be called on a goroutine after initializing the Indexer.
+// and then execute a set of recovery actions. It first repairs the indexed blocks if
+// startDB found them incomplete, so that afterwards every indexed vote can be dated.
+// It then checks for those processes which are still open (live) and updates all
+// temporary data (current voting weight and live results if unecrypted).
+// This method might be called on a goroutine after initializing the Indexer.
 // TO-DO: refactor and use blockHeight for reusing existing live results
 func (idx *Indexer) AfterSyncBootstrap(inTest bool) {
+	if !inTest {
+		<-idx.App.WaitUntilSynced()
```
lucasmenendez, line 388: Moving this above the `ignoreLiveResults` return changes behaviour for nodes that never sync. `isSynced` only flips in `beginBlock` (vochain/app.go:281), and seed nodes never run consensus yet still build the indexer when `Indexer.Enabled` (cmd/node/main.go:451-480). Previously an `ignoreLiveResults` indexer returned instantly; now it parks forever, along with the 1 Hz poller goroutine `WaitUntilSynced` spawns. Nothing joins on it, so it's a leak rather than a correctness bug.

Suggestion: `if idx.ignoreLiveResults && !idx.blockReindexNeeded.Load() { return }` before the wait.

Minor: `ReindexBlocks` waits on `WaitUntilSynced` again at :478, so this waits twice.

vochain/indexer/indexer.go
```diff
@@ -96,6 +97,11 @@ type Indexer struct {
 
 	// ignoreLiveResults if true, partial/live results won't be calculated (only final results)
 	ignoreLiveResults bool
+
+	// blockReindexNeeded is set by startDB when the indexed blocks are known to be
```
lucasmenendez: The atomic fixes the data race but not the ordering — if `Swap(false)` ran before startDB's `Store(true)`, the repair is silently skipped. It's safe only incidentally, because the bootstrap goroutine is parked in `WaitUntilSynced` while statesync runs `RestoreBackup` → `startDB`.

Worth saying that in the comment, or gating on `readWriteDB != nil` so it's structural.

vochain/indexer/vote.go
```diff
@@ -98,6 +98,51 @@ func (idx *Indexer) VoteList(limit, offset int, processID string, nullifier stri
...
+}
+
+// ErrInvalidVoteBucket is returned when an unsupported time bucket is requested.
+var ErrInvalidVoteBucket = fmt.Errorf("invalid vote activity bucket")
+
+// VoteActivity returns the votes of a process aggregated into time buckets of the
+// given granularity ("hour" or "day"), ordered chronologically. Buckets with no
+// votes are not present in the returned slice.
+//
+// The timestamp of a vote comes from the block it was included in. The bootstrap
```
lucasmenendez: Two things here:

1. The bootstrap doesn't actually guarantee this — see my comment on `bootstrapBlocks`.
2. `CreateVote` is `REPLACE INTO` keyed on nullifier, so an overwritten vote *moves* buckets and a past hour's count can go down. The endpoint's `@Description` says "the number of votes cast", which reads as immutable history. One sentence would cover it.

api/chain.go
```diff
@@ -1049,8 +1049,11 @@ func (a *API) chainTxListByHeightAndPageHandler(_ *apirest.APIdata, ctx *httprou
 	params, err := parseTransactionParams(
 		ctx.URLParam(ParamPage),
 		"",
-		ctx.URLParam(ParamHeight),
 		"",
+		// parseTransactionParams takes hash before height; passing the height
```
lucasmenendez: Good catch, but the comment understates it. The hash filter is a substring match (db/transactions.sql.go:135-140, `INSTR(LOWER(HEX(hash)), LOWER(?5)) > 0` for inputs under 64 chars) and `Height: 0` disables the height filter entirely — so `/chain/blocks/1/transactions/page/0` returned every tx in the *whole chain* whose hex hash contains "1". Wrong data rather than an empty list.

Same reason the new test isn't quite a regression test: on a single-tx chain it passes with the old code too. Txs in two different blocks would pin it.

api/elections.go
```diff
@@ -443,6 +451,52 @@ func (a *API) electionVotesCountHandler(_ *apirest.APIdata, ctx *httprouter.HTTP
...
+	electionID, err := hex.DecodeString(util.TrimHex(ctx.URLParam(ParamElectionId)))
+	if err != nil || electionID == nil {
+		return ErrCantParseElectionID.Withf("(%s): %v", ctx.URLParam(ParamElectionId), err)
+	}
+	// check the election exists, so an unknown id doesn't silently return an empty activity
+	if _, err := getElection(electionID, a.vocapp.State); err != nil {
+		return err
+	}
+
+	bucket := ctx.QueryParam(ParamBucket)
```
lucasmenendez: Two small ones:

- Param validation lives in `api/helpers.go` elsewhere — `parseStatus` (:299) owns the status enum and returns `ErrParamStatusInvalid` itself, rather than the handler translating a persistence-layer sentinel.
- The result is unbounded: no `from`/`to`, no `LIMIT`, while `MaxItemsPerPage` caps every paginated sibling. A year-long election at `bucket=hour` is ~8.7k buckets per call, recomputed each time.

lucasmenendez (commented):
Nice endpoint, and the tx fix is a real one. Most of my notes are about the supporting machinery (the startup completeness check + the reindex bootstrap) rather than the feature itself.

Worth flagging that CI being green doesn't cover the first three comments below — none of them have a test that would fail.

Things I checked and they're fine: `strftime` over the stored `DATETIME` really does normalize to UTC (go-sqlite3 writes `+00:00`; I verified against real values in `testdata/sqlite-backup-0009.sql.zst`), `WHERE v.process_id = ?` is covered by `index_votes_process_id`, 4061 is free and correctly placed, the handler is nil-safe via the `EnableHandlers` guard, swagger is CI-generated so nothing can go stale, and the sibling `parseTransactionParams` call at chain.go:1010 passes no height so there's no second instance of that bug.

vochain/indexer/queries/votes.sql
p4u: Replaced in 9effb06d with VoteBlockHeightBounds: MIN(votes.block_height) vs MIN(blocks.height), both indexed, and the backwards comment is fixed.

vochain/indexer/queries/votes.sql
p4u: Done in 9effb06d: VoteActivity now LEFT JOINs blocks and reports the remainder as undatedVotes (omitempty), so clients can reconcile against /votes/count. Kept the bounded bootstrap repair for the archive-node case alongside it.

vochain/indexer/indexer.go
p4u: Fixed in 9effb06d: the repair is only scheduled when the undated height range intersects [Base, Height]; votes below the base are counted via CountUndatedVotesBelowHeight and logged once at Info. No repeated futile reindexes.

vochain/indexer/indexer.go
p4u: Applied your exact gate in 9effb06d, and ReindexBlocks was split into a waiting wrapper plus an inner reindexBlocks so WaitUntilSynced runs only once. Note: seed nodes taking the early return also skip bootstrapKeyReveals, which matches their previous behavior (they parked before reaching it).

vochain/indexer/indexer.go
p4u: Documented the ordering invariant on both atomics in 9effb06d. Left the structural gate out: readWriteDB is not reachable from the bootstrap path in its current shape, so the comment plus the statesync ordering note seemed the honest minimum.

vochain/indexer/vote.go
p4u: Both in 9effb06d: doc + swagger now say completeness is not guaranteed (undatedVotes carries the remainder) and that an overwritten vote moves buckets, so a past hour's count can decrease.

api/chain.go
p4u: Comment rewritten in 9effb06d to describe the substring-match behavior (wrong data, not an empty list), and TestAPIBlocks now creates txs in two blocks and asserts each per-block list returns exactly its own txs plus an empty list for a tx-less block.

api/elections.go
p4u: Done in 9effb06d: parseBucket lives in helpers.go returning ErrParamBucketInvalid, and the response is bounded — optional from/to (RFC3339, compared via datetime() to avoid lexical offset comparisons) plus a 1000-bucket cap returning ErrTooManyBucketsRequested (4062) with guidance to use day buckets or narrow the window.

lucasmenendez (approved):

PR #1428, opened 2026-07-30
ci: don't request the pushing actor as portal PR reviewer

p4u:
Follow-up to the VOCDONIBOT_PAT rotation. The swagger job now authors the developer-portal PR with the rotated token's account; `reviewers: ${{ github.actor }}` then asks the PR author to review their own PR, which the API rejects (`Review cannot be requested from pull request author`, see run 30544676389 — note the spec PR itself was updated fine). The `SdkDocsReviewer` team request remains.

PR #1429, opened 2026-07-30
ci: author developer-portal docs PRs with a GitHub App

p4u:
Replaces the `VOCDONIBOT_PAT` personal token with an installation token minted per run from the org-owned **vocdoni-docs-bot** GitHub App (ID in the `DOCS_BOT_APP_ID` repo variable, key in the `DOCS_BOT_PRIVATE_KEY` secret — both already in place, and the credential chain was verified by minting a token against the live installation covering `developer-portal`).

Consequences:
- No expiring credential: tokens are ephemeral (1h), the app key does not expire, and the app cannot leave the org.
- Portal PRs are authored by `vocdoni-docs-bot[bot]`, so `reviewers: ${{ github.actor }}` is valid again for every actor, including the former PAT owner. The line is kept.
- `VOCDONIBOT_PAT` becomes unused and can be deleted after merge; the personal PAT can be revoked.

Supersedes #1428, which worked around the author/reviewer collision by dropping the reviewer request.

### DIFF
--- a/.github/workflows/swagger.yml
+++ b/.github/workflows/swagger.yml
@@ -78,13 +78,23 @@ jobs:
             ${{ steps.diff.outputs.stdout }}
             ```
 
+      - name: Mint app token for developer-portal
+        id: app-token
+        if: ${{ github.event_name == 'push' }}
+        uses: actions/create-github-app-token@v2
+        with:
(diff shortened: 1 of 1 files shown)


p4u:
Obsolete: the developer-portal is deprecated and the publishing pipeline is being removed entirely instead. See the removal PR.

PR #1454, opened 2026-09-10
feat(vochain/ist): rework validator voting-power dynamics

altergui:
Two independent problems in the way `updateValidatorScore` manages
validator power, rolled up into one coordinated change since they share
the activation event and the tests are entangled.

## 1. Grace period between "power → 0" and eviction from the set

Before: `power == 0` triggered `RemoveValidator` on the same block. No
tolerance. Re-adding is a manual, coordinated operation.

After:
- `maxPower` bumped 100 → 1000 (finer granularity). Silent validator
  hits the floor in ~66 periods ≈ **2h 12min at 12s/block**.
- Floor at `minPower = 1` instead of zero. A floored validator still
  signs but weighs `1/1000` ≈ **0.1%** in consensus — a "shadow" seat.
- `inactiveGraceBlocks = 150000` (~**21 days** at 12s/block). Only
  after the validator has been at the floor for a full window is
  `RemoveValidator` called. Marker cleared on any recovery period, so
  a partially healthy validator resets the eviction clock.
- `powerIncrement = 10` per positive-signal period (proportional to
  the maxPower bump). Ramp-up time from `newValidatorPower` stays
  ~3h 20min.
- The "keep the last three validators" guard is preserved.

Per-validator marker (`4-byte big-endian uint32` block height) lives in
the existing `TreeExtra` subtree, keyed `"vldIS/" || signingAddress`.
Same idiom as `MaxProcessSize` / `NetworkCapacity`. Three helpers on
`state.State` wrap set / read / clear. `models.Validator` proto is
untouched.

## 2. Score reflects recent participation, not lifetime average

Before: score was `Votes / (currentHeight - joinHeight) * 100`, with
`Votes` cumulative from the day the validator joined. For a validator
that had been voting reliably for a long time,
`Votes ≈ currentHeight - joinHeight`, so once it went silent the ratio
moved by one integer step every tens of thousands of blocks — and
between those steps the "score stable at >= threshold" branch fired
`+powerIncrement` per period, cancelling almost every decay hit.

This pinned a disconnected validator near maxPower for weeks on
production. Reproduced with Lucas' real dashboard state (`Votes =
7,582,178`, `Score = 89`, `gap ≈ 8.5M`): if he goes silent on today's
LTS binary, his `Power` stays at 100 for **~5 months**, only starts
falling once the score crosses under 80, and takes **>2 years** to
reach 0. It also had a large veteran-vs-recent asymmetry: for the
same 2-day outage a veteran validator loses zero power while a recent
one gets expelled in ~3h.

Fix: reset `Votes` and `Height` at every score computation. The
denominator becomes `updatePowerPeriod` (10 blocks) after the first
window, so the score is the participation rate over the just-closed
window. A silent validator's score goes to 0 the very next period and
its power decays every period after that, regardless of how long it
had been running or how much history it carried. `Proposals` is left
cumulative (metric-only, not consensus-critical). `Validator.Height`
was only read by this function (repo-wide grep), so its semantic
change from "join height" to "window-start height" is safe.
Post-outage, veteran and recent are indistinguishable.

`updatePowerPeriod` stays at 10 blocks. Empirically calibrated against
7d of `comet_consensus_validator_missed_blocks` on LTS: the healthiest
10 validators have miss rates in `[0, 0.32]`% — well within
`positiveScoreThreshold = 80` for the natural window variance. The
`~1.04` baseline in `comet_consensus_missing_validators` comes
entirely from an 11th validator that has been fully offline (and is
not being scraped, which is why it does not appear in the
per-validator query).

## Test harness fix

Tests were calling `updateValidatorScore` once per period, but
production runs it once per block via `Istc.Commit`
(`vochain/app.go:202`), committing between calls. Without the
correction the score branch would see only one attributed vote per
period, hiding the reset. Fixed the harness so `advancePeriod` loops
`updatePowerPeriod` blocks and commits after each.

## Deployment

This is a state-transition rule change. Every validator applies the
new rule at every `updatePowerPeriod`-multiple block; running new and
old code side-by-side on the same chain produces diverging state
hashes and splits the set silently.

Roll out via the `EndOfChain + new chainID` runbook documented in
`vochain/genesis/genesis.go` — that is the standing policy since
`0696d741 all: cleanup code that supported forks` (Jun 2024).

Two follow-up PRs will be needed after this one merges:
1. One-liner adding `EndOfChain: <current_tip + ~24h_of_blocks>` to
   the `vocdoni/LTS/1.2` entry in `vochain/genesis/genesis.go`, so all
   validators deterministically stop at the same height.
2. After all validators reach that height and the last committed
   `AppHash` is captured, a PR that adds a `vocdoni/LTS/1.3` entry to
   the `genesis` map with the captured `AppHash`, sets its
   `InitialHeight` to `EndOfChain + 1`, and bumps
   `networks["lts"]` to `"vocdoni/LTS/1.3"`. Nodes preserve the state
   tree, discard the `LTS/1.2` cometbft blockstore, and start
   producing under the new chainID with these rules from the first
   block.

`TreeExtra` has no schema change (the marker is a new key namespace,
not a new subtree), so state stores from before this change do not
require a schema migration on chain restart.

## Test plan

- [x] `go test ./vochain/state/... ./vochain/ist/...` — all pass
- [x] `go test ./vochain/...` — 148 tests across 21 packages pass
- [x] `TestLongLivedSilentValidatorDecaysFast` reproduces the reported
      "months at high power" bug and pins the fix
- [x] Other tests cover: state helpers roundtrip, decay-to-floor,
      marker set/clear lifecycle, expulsion after (shortened) grace
      period, three-validator lower bound
- [ ] Dev-net dry-run of the ceremony: `EndOfChain` on
      `vocdoni/DEV/36`, freeze, capture `AppHash`, start
      `vocdoni/DEV/37` with these rules, kill a miner, observe
      `vochain_validator_power` drop to `minPower` within ~2h, confirm
      miner stays in set for ~21 days before removal, confirm miner
      recovers cleanly when brought back up before grace elapses


p4u:
@argos review

altergui:
Verified all 5 findings — all valid. Fixes in 4 commits on top of `bf985879` (tip `fb4dc492`, MERGEABLE):

- `328da755` — **finding 4**: `newValidatorPower` 5→50 (rescale with `maxPower`, entry weight back to 5% of cap).
- `b345a1c5` — **findings 1 + 5**: Score-window boundary moved to TreeExtra `vldSW/<addr>` (12B: 4B height + 8B lifetime Votes at start). `Validator.Height`/`Votes` stay canonical → `GET /chain/validators` keeps its `joinHeight`/`votes` semantics and the `vochain_validator_votes` metric stops resetting. Score = `(Votes - votesAtStart) / (height - windowStart) * 100`. `vldIS/` docblock 25B→26B while we're here. Regression test `TestScoringPreservesJoinHeightAndLifetimeVotes`.
- `b5ace6fe` — **finding 3**: Three tests with literal-value asserts: 1000→900 in one silent period, 990→1000 in one positive period, floor reached at exactly period 49. Also caught along the way: the docblock claimed "~66 periods" but with uint64 truncation the real number is 49; fixed.
- `fb4dc492` — **finding 2**: New `vochain/cometbft_test.go` with a table test for `validatorUpdate` (surviving-only / removed-only / mixed / empty). Deleting the `removedPubKeys` loop now fails.

`go build ./...` clean, `go test ./vochain/...` green.

cc @p4u

altergui:
Superseded by #1472 (feat/validator-power-dynamics-v2). The whole cleanup stack (#1458 → #1471) has been retargeted onto #1472's branch.

PR #1470, opened 2026-09-14
vochain: hard-reject legacy OFF_CHAIN_CA (LTS/1.4) [DRAFT — do not merge before LTS/1.4]

altergui:
## Summary

**Draft — do not merge before the LTS/1.4 window.** Follow-up to PR #1463 (LTS/1.3 soft-deprecation). Flips the deprecation warning to a hard rejection at `NewProcessTxCheck`. Depends on this branch being stacked on top of `feat/lts13-reject-legacy-csp` (PR #1463): merge #1463 first, rebase this on `main` after, then unmark for review when LTS/1.4 is being cut.

## Sequence

1. **LTS/1.3 ships (tomorrow)** with #1463 — `OFF_CHAIN_CA` still accepted, warning on validator log + POST `warning` field + GET `censusOriginDeprecated: true`.
2. **In parallel**: vocdoni-sdk companion PR adds `CensusType.CSP_V2 = 'csp2'` and bumps `@vocdoni/proto` to v1.15.14. SDK release goes out. Integrators have 2 weeks with the warning visible to them.
3. **~2 weeks later**: LTS/1.4 ceremony. This PR is marked ready-for-review and merged. Legacy `OFF_CHAIN_CA` starts failing at `CheckTx` with `\"census origin OFF_CHAIN_CA is deprecated (issue #1424); use OFF_CHAIN_CA_V2\"`.

## Why the two-phase approach

The vocdoni-sdk (all branches, main at v0.9.2/0.9.4) only emits `OFF_CHAIN_CA` today. A hard rejection in LTS/1.3 would break every integrator whose SDK falls behind. Soft-deprecation gives them a real migration window with concrete signal (warning in the response body), and moves the failure mode to loud-and-early rather than silent-and-later.

## Diff on top of #1463

- `vochain/transaction/election_tx.go`: `OFF_CHAIN_CA` case returns an error instead of logging a warning. Error string matches the warning verbatim so tooling that already parsed the deprecation string keeps working.
- `vochain/transaction/transaction.go`: drop the `response.Log` warning wiring — `CheckTx` never gets past `NewProcessTxCheck` now.
- `api_types.go` / `api/elections.go`: keep `ElectionCreate.Warning` and `ElectionCensus.CensusOriginDeprecated` unchanged — costs nothing when unused, and elections created under `OFF_CHAIN_CA` on prior chains still surface `censusOriginDeprecated: true` on read.
- `cmd/end2endtest/csp.go` + `dockerfiles/testsuite/start_test.sh`: drop the legacy `cspelection` op — its `setupElectionRaw` path now hits the rejection. `cspelectionv2` remains as the sole CSP E2E.
- `vochain/process_test.go`: `TestNewProcessSoftDeprecatesLegacyCSPOrigin` → `TestNewProcessRejectsLegacyCSPOrigin`. Legacy fails with expected error, V2 passes. Drops the `checkTxLog` helper (no longer needed).

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./vochain/...` (includes `TestNewProcessRejectsLegacyCSPOrigin`)
- [ ] Rebase onto `main` after #1463 lands
- [ ] Confirm vocdoni-sdk V2 release is public and integrator communication has gone out before marking ready-for-review


p4u:
This would break backwards compatibility, isn't it @altergui ?

altergui:
> This would break backwards compatibility, isn't it @altergui ?

definitely, it's just a reference draft for a future bump (LTS/1.4) when we decide to reject insecure OFF_CHAIN_CA

